Issue #15, opened 2021-06-15
Need distinction between what is a prerequisite for a program and an event of a program


Sandesh-L:
**Add programId to `event`:**
In the `programevent` model there is a column for `programId` add that column inside of the `event` model. This way it is clear what event belongs to which program.

**Table called preqForProgram:**
Currently there is no way to tell if an event is required for a program. The preqForProgram modal should contain two fields `programId` and `eventId`.  The `programevent` model should be renamed to `preqForProgram` after transferring the `programId` column. The `preqForProgram` should be a relationship between a program and a prerequisite event and not between a program and its events.

One event can only belong to one program, but the event could be required for multiple programs. Having `programevent` model leads us to think that an event might belong to multiple programs. 

**Create a program called training:**
The training program will contain many events and some of those events will be prereqs for multiple other programs. Create a training program (in test data), add multiple events to it, and assign those events as prereqs for other programs.

PR #17, opened 2021-06-16
Differentiate prerequisite and event issue15

sreynit02:
**Problem:** 
Need distinction between what is a prerequisite for a program and an event of a program
This PR fixes issue #15 

**Solution:** 
Added programId to event:
In the programevent model there is a column for programId add that column inside of the event model. This way it is clear what event belongs to which program.

Made Table called preqForProgram: 
There was no way to tell if an event was required for a program. The preqForProgram contains two fields programId and eventId. The programevent model was renamed preqForProgram after transferring the programId column. 

Created a program called training:
The training program contains many events and some of those events are prereqs for multiple other programs. We created  Training program (in test data), added multiple events to it, and assigned those events as prereqs for other programs.

We also refactored the code in /app/logic/events.py and added preqForProgram model migrate_db.sh .

**Test:** 
Run ./reset_database.sh test and check the database to see if preqForProgram is added and column_id is added to event table. prgramEvent should no longer exist. 
Run tests again and they all should pass.


BrianRamsay:
The whole reason we created ProgramEvent is that there are events that are in more than one program, meaning program_id in Event doesn't work.

You need to remove that part of these changes.

BrianRamsay:
Training is not a separate program. Programs are specific CELTS initiatives. Events within a program can be classified as a training. I thought we said that isRequiredForProgram indicated a training?

If we want to make that clearer we can have other boolean flags. Like, we could do isTraining, or isPrerequisite, or isRequired. But creating new tables doesn't give us additional functionality that we need.

bryantal:
@Sandesh-L @sreynit02 
Any update on this PR?

sreynit02:
> @Sandesh-L @sreynit02
> Any update on this PR?

sreynit02:
Sorry, I accidentally closed the PR by accident. But i've reopened it back

BrianRamsay (approved):
Looks good, pending our final conclusion on prerequisite and required events

PR #19, opened 2021-06-17
User Login

BrianRamsay:
Implement user login from shibboleth and config files. Shibboleth login is not complete yet.

### DIFF
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -1,7 +1,8 @@
-from flask import request, render_template
+from flask import request, render_template, g
 
 from app.controllers.main import main_bp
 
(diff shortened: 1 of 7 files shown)


app/controllers/main/routes.py
```diff
@@ -1,7 +1,8 @@
-from flask import request, render_template
+from flask import request, render_template, g
```
bryantal, line 1: What is "g" here?

bryantal (approved):

app/controllers/main/routes.py
BrianRamsay: g is flask's global request variable

PR #20, opened 2021-06-17
Issues 3.8 and 3.8.1 merged. Checking the user on eligibility to register for an event.

Karina-Agliullova:
**Problem:** The volunteer must be able to register for an event. The user must be able to fill out the required forms. In order to register for an event, the volunteer must be eligible to do so. The volunteer must not be banned and must complete the prerequisite(s) for the event prior to registering.

**Solution:** We created a function for Issue 3.8.1 that checks if the user is not banned and if the user attended any required prerequisites for the event. If the user passes through these requirements, they must be able to register for the event. We created another function for that. It checks if the user is eligible for the event, and lets the user register for the event. We also renamed the file "ProgramEvent" to " preqForEvent" so that we can check what event belongs to what program. An event can only belong to one program, but the program can require multiple events as prerequisites, so it makes more sense to have the file "preqForEvent".

**Tests:** We have 4 files to test the functions described above. One of them is called test_events.py and it checks if the event is valid and if the event is in any program or has any trainings. Another test, test_data.py, checks terms, programs, events, preqForProgram, Users, and EventParticipants and if they exist, it executes them. The other test, test_program_eligibility.py, checks if the user is eligible for the event, if they attended prerequisites, if the username and eventId match, and the test suite also has some tests that are supposed to fail because the code can be broken. And another test suite, test_volunteerRegisterEvent, checks username, event id, rsvp, and it also has some broken code to check if it will break the function. In the database, we will have to go to eventParticipant to check if the row was inserted once we run the test suites with all the data in them. We will need to change the column "attended" from 1 to 0 and vice versa in case we change it in the code to see if it will work and not let the user register for the event if prerequisites are not completed or the user is not eligible.


app/controllers/events/meetsReqsForEvent.py
```diff
@@ -0,0 +1,17 @@
+from app.models.user import User
+from app.models.event import Event
+from app.models.eventParticipant import EventParticipant
+from app.models.programBan import ProgramBan
+from app.models.preqForProgram import PreqForProgram
+
+def isEligibleForProgram(event, user):
```
bryantal: Adding some docstring comments here detailing what this function does is a good idea.

app/controllers/events/volunteerRegisterEvent.py
```diff
@@ -0,0 +1,21 @@
+from app.models import*
+from app.models.event import Event
+from app.models.eventParticipant import EventParticipant
+from app.models.user import User
+from app.controllers.events.meetsReqsForEvent import isEligibleForProgram
+import pytest
+
+def volunteerRegister(userid,  eventid):
```
bryantal: Same thing here, add some docstrings to explain what this function does.  It will help clarify what exactly is happening with the conditionals below.

app/controllers/events/volunteerRegisterEvent.py
```diff
@@ -0,0 +1,21 @@
...
+
+    user = User.get(User.username == userid)
+    event = Event.get(Event.id == eventid)
+    # Assuming the student meets the requirement for the events (function wriiten by Zach and KArina)
+
+    if isEligibleForProgram(event, user):
+        if EventParticipant.select().where(EventParticipant.user == user, EventParticipant.event == event) is None:
```
bryantal: To improve readability, move the items inside the .where() onto their own lines.  We want to keep line length to 80ish characters.

app/controllers/events/volunteerRegisterEvent.py
```diff
@@ -0,0 +1,21 @@
...
+            eventParticipant = EventParticipant.create(user = user, event = event, rsvp = True)
```
bryantal: Same thing here with the line length.  Break out the items inside the .create() onto separate lines similar to how I suggested above.

app/models/preqForProgram.py
```diff
@@ -2,7 +2,7 @@
 from app.models.program import Program
 from app.models.event import Event
 
-class ProgramEvent(baseModel):
+class PreqForProgram(baseModel):
```
bryantal: We're going to ditch the PreqForProgram table for now, and just put a field in the event table for it.  Talk to Sandesh about the changes if they have not been brought into development yet.   He has implemented them in a branch he was working on.

bryantal (changes requested):

bryantal (dismissed):

Karina-Agliullova:
PR updated: Fixed the merge conflict

app/controllers/events/meetsReqsForEvent.py
```diff
@@ -0,0 +1,25 @@
+from app.models.user import User
+from app.models.event import Event
+from app.models.eventParticipant import EventParticipant
+from app.models.programBan import ProgramBan
+
+
+def isEligibleForProgram(event, user):
```
BrianRamsay: It doesn't make sense for this function to take an event as an argument, does it?

The function name indicates that you're checking if a user is eligible for a program, so the arguments should be `user` and `program`. We will need to use this function outside of the context of an event (like if we're showing on a user's profile that they are barred from a program).

tests/code/test_program_eligibility.py
```diff
@@ -0,0 +1,42 @@
+import pytest
+from app.models.eventParticipant import EventParticipant
+from app.controllers.events.volunteerRegisterEvent import volunteerRegister
+from peewee import DoesNotExist
+from app.models.user import User
+from app.models.event import Event
+from app.controllers.events.meetsReqsForEvent import isEligibleForProgram
+
+@pytest.mark.integration
+def test_noUserVolunteerRegister():
```
BrianRamsay: It's good that you are testing all of these cases, however a function called `isEligibleX` should really be returning True or False, and not using exceptions to handle normal False cases.

BrianRamsay (changes requested):

app/controllers/events/meetsReqsForEvent.py
sreynit02: We changed the parameter to `programId` and `user`

tests/code/test_program_eligibility.py
Ebenezer997: This function was to test if the user does not exist and if the program does not exist.

tests/code/test_program_eligibility.py
```diff
@@ -0,0 +1,32 @@
...
+    with pytest.raises(DoesNotExist):
+        eligible = isEligibleForProgram(2, 135156)
+
+    with pytest.raises(DoesNotExist):
+        eligible = isEligibleForProgram(9, "lamichhanes2") #Program doesn't exist
+
+
+
+@pytest.mark.integration
+def test_volunteerEligible():
```
BrianRamsay: You don't test any failure cases. You need to test all of the logic in your function. When would this function return False?

tests/code/test_volunteerRegisterEvent.py
```diff
@@ -0,0 +1,48 @@
...
+
+    with pytest.raises(DoesNotExist):
+        volunteer = volunteerRegister("khatts", -1)
+
+    with pytest.raises(DoesNotExist):
+        volunteer = volunteerRegister("khatts", 0)
+
+
+@pytest.mark.integration
+def test_volunteerRegister():
```
BrianRamsay: You need to test the cases when someone wouldn't be registered.

BrianRamsay (changes requested):

tests/code/test_program_eligibility.py
sreynit02: Added more test cases where the function returns False

tests/code/test_volunteerRegisterEvent.py
sreynit02: Added more test cases where someone wouldn't be registered.

app/controllers/events/programEligibility.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def isEligibleForProgram(program, user):
+    """
+    Checks if the user is banned from an event. Checks if the user has completed
+    specific prerequisite(s) for the event.
+    :param program: the id of the program that is passed in
+    :param user: the user that is logged into the system
+    :return: True if the user is not banned and meets the requirements, and False otherwise
+    """
```
BrianRamsay, line 15: Make docstring clearer and closer to what the function is actually doing. Mention in the description what is happening in the function and how to use it.

Parameter description should indicate acceptable data types and values where relevant (and in this case whether or not the user is logged in is irrelevant)

tests/code/test_programEligibility.py
```diff
@@ -0,0 +1,45 @@
+import pytest
+from app.models.eventParticipant import EventParticipant
+from app.controllers.events.volunteerRegisterEvent import volunteerRegister
+from peewee import DoesNotExist
+from app.models.user import User
+from app.models.event import Event
+from app.controllers.events.programEligibility import isEligibleForProgram
+from app.models.program import Program
+@pytest.mark.integration
+def test_noUserVolunteerRegister():
```
BrianRamsay: this name doesn't match the function being tested

tests/code/test_programEligibility.py
```diff
@@ -0,0 +1,45 @@
...
+    with pytest.raises(DoesNotExist):
+        eligible = isEligibleForProgram(2, 135156)
+
+    with pytest.raises(DoesNotExist):
+        eligible = isEligibleForProgram(9, "lamichhanes2") #Program doesn't exist
+
+
+
+@pytest.mark.integration
+def test_volunteerEligible():
```
BrianRamsay: this name doesn't match the function being tested

tests/code/test_volunteerRegisterEvent.py
```diff
@@ -0,0 +1,55 @@
...
+
+    volunteer = volunteerRegister("lamichhanes2", 10)
+
+    print(volunteer)
```
BrianRamsay: extra print statement

app/controllers/events/volunteerRegisterEvent.py
```diff
@@ -0,0 +1,34 @@
...
+    Lets the user register for an event if they are eligible
+    :param userid: indicates the id that the user has
+    :param eventid: indicates the id that the event has
+    :return: a message that indicates what the user has registered for if eligible;
+    otherwise lets the user know that they are not eligible for registration
+    """
+
+    user = User.get(User.username == userid)
+    event = Event.get(Event.id == eventid)
+    program = Event.select(Event.program).where(Event.id == eventid)
```
BrianRamsay: You are still using the argument eventid, not `event` that you just got in the line before

app/controllers/events/volunteerRegisterEvent.py
```diff
@@ -0,0 +1,34 @@
...
+        if not EventParticipant.select().where(EventParticipant.user == user,
+                                           EventParticipant.event == event).exists():
+            eventParticipant = EventParticipant.create(user = user,
+                                                       event = event,
+                                                       rsvp = True)
+        else:
+            return f"{user} is already registered for {event}"
+        return eventParticipant
+    else:
+        return ("User is not eligible for the program")
```
BrianRamsay: The function shouldn't return different data types depending on the status of the function. That is, success and failure should be clear from running the function or the return values.

In this case it makes sense to raise an exception for the failure case, and ignore if the user is already registered and return the same value as if the record was created.

app/controllers/events/volunteerRegisterEvent.py
```diff
@@ -0,0 +1,34 @@
+from app.models import*
+from app.models.event import Event
+from app.models.eventParticipant import EventParticipant
+from app.models.user import User
+from app.controllers.events.programEligibility import isEligibleForProgram
+import pytest
+
+def volunteerRegister(userid,  eventid):
```
BrianRamsay: This function name should be more explicit, like `rsvpForEvent` or `userRsvpForEvent`

BrianRamsay (changes requested):

app/controllers/events/programEligibility.py
Karina-Agliullova: Made the docstring clearer by describing its purpose and parameters' data types and values

tests/code/test_programEligibility.py
Karina-Agliullova: Renamed the test according to its function name

tests/code/test_programEligibility.py
Karina-Agliullova: Renamed the test according to its function name

tests/code/test_volunteerRegisterEvent.py
Karina-Agliullova: Removed the print statement

app/controllers/events/volunteerRegisterEvent.py
Karina-Agliullova: Replaced the eventid argument with the event argument

app/controllers/events/volunteerRegisterEvent.py
Karina-Agliullova: Replaced the function's old name with userRsvpForEvent. Did the same in the test files and renamed the files that started with volunteerRegister

app/controllers/events/volunteerRegisterEvent.py
Karina-Agliullova: Refactored the function by raising an exception and returning the value for the created record

BrianRamsay (approved):
Looks great! Nice work.

Issue #18, opened 2021-06-16
3.9 The user must be able to view upcoming events


Sandesh-L:
The user should have marked the events they are interested in.

List the upcoming events for programs that they have shown interest in.

PR #26, opened 2021-06-18
Populate Upcoming Events on Webpage

sreynit02:
This PR addresses Issue #18.

Did: We aggregated the events of a user using the programs that they have shown interest in. We compared the start date and start time of the events to the current date and time and determined if they are upcoming events or not. On the web browser (`/events/upcoming_events`), we displayed a table which has Description, Program, Start Date, End Date of the events. 
Side Note: In the event model, We also changed the `timeStart` and `timeEnd` fields from `CharField` to `TimeField` so that we can compare time.

Test:
Inside the `test_data.py` file, add or remove interests for a user. Then run `./reset_database.sh test`. Run `flask run` to see the changes in the table. Change the start dates in the events table to before today, those events shouldn't appear in the table. Change the default user in `app/config/default.yaml` to an existing user in `test_data.py`. Run the website and the table will contain upcoming events for that user.


tests/run_tests.sh
```diff
@@ -22,7 +22,7 @@ function header {
 # ui
 function ui {
 	header UI
-	python -m pytest $FLAGS $UI_URLS tests/ui/smoke_test.py tests/ui/functional_test.py
+	python -m pytest $FLAGS $UI_URLS /tests/ui/smoke_test.py tests/ui/functional_test.py
```
BrianRamsay: Was this change intentional? I doubt it is, especially since the second path didn't have the slash added to it.

app/logic/getUpcomingEvents.py
```diff
@@ -0,0 +1,29 @@
...
+from app.models.user import User
+from datetime import date, datetime, time
+
+def getUpcomingEventsForUser(username):
+    user = User.get(User.username == username)
+    interestedEvent = (Event.select(Event, Program, Interest)
+                            .join(Program)
+                            .switch(Event)
+                            .join(Interest, on=(Event.program == Interest.program))
+                            .where(Event.program_id == Program.id, Interest.user == user)
```
BrianRamsay: The first part of this WHERE clause is unecessary, since you already implicitly have a join condition between those two tables.

Also, I think .switch is also unnecessary, as you are explicitly providing the join condition when you join Interest

BrianRamsay (changes requested):
I think you should pass a User object into your function instead of username. Doesn't seem like we need to do that extra query to get the user information.

Also, should `showUpcomingEvents` still be in controllers? I'm not sure what the point of that function is at all, actually, since it just calls `getUpcomingEventsForUser`

tests/run_tests.sh
Sandesh-L: No, it was not. We have changed it back to what it was.

app/logic/getUpcomingEvents.py
sreynit02: We removed the first part of this WHERE clause and .switch and the code still works.

sreynit02: We also removed the showUpcomingEvent.py file and call getUpcomingEventsForUser directly from routes.py

tests/code/test_showUpcomingEvents.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+from app.models.program import Program
+from app.models.interest import Interest
+from app.models.event import Event
+from app.models.user import User
+from app.logic.getUpcomingEvents import getUpcomingEventsForUser
+from peewee import DoesNotExist
+
+@pytest.mark.integration
+def test_getUpcomingEvents():
```
BrianRamsay: Test isn't filled out enough. What happens on an error? Is it getting the correct events? This needs to be done in parallel with development on the function.

app/logic/getUpcomingEvents.py
```diff
@@ -0,0 +1,27 @@
...
+    interestedEvent = (Event.select(Event, Program, Interest)
+                            .join(Program)
+                            .join(Interest, on=(Event.program == Interest.program))
+                            .where(Interest.user == user)
+                            )
+    upcomingEvents = []
+    for event in interestedEvent.objects():
+        today = date.today()
+        times = datetime.now().time()
+        event_time = event.timeStart
```
BrianRamsay: This variable isn't used.

app/logic/getUpcomingEvents.py
```diff
@@ -0,0 +1,27 @@
+from app.models.program import Program
+from app.models.interest import Interest
+from app.models.event import Event
+from app.models.user import User
+from datetime import date, datetime, time
+
+def getUpcomingEventsForUser(user):
+
+    interestedEvent = (Event.select(Event, Program, Interest)
```
BrianRamsay: This name should be plural, since it's a list (sort of)

BrianRamsay: You also can simplify the query a little. You don't need to specify anything in the select (I'm pretty sure), which means you don't need `.objects()` down below. And since you have a program object you are passing through, your jinja can use that for the name, rather than needing to select programName from the query explicitly.

app/logic/getUpcomingEvents.py
```diff
@@ -0,0 +1,27 @@
...
+def getUpcomingEventsForUser(user):
+
+    interestedEvent = (Event.select(Event, Program, Interest)
+                            .join(Program)
+                            .join(Interest, on=(Event.program == Interest.program))
+                            .where(Interest.user == user)
+                            )
+    upcomingEvents = []
+    for event in interestedEvent.objects():
+        today = date.today()
```
BrianRamsay: I'd either put this line outside of the for loop (for efficiency), or just use date.today() in the conditional below.

app/logic/getUpcomingEvents.py
```diff
@@ -0,0 +1,27 @@
...
+        times = datetime.now().time()
```
BrianRamsay: `times` isn't a very clear variable name. Like the comment above, I'd either move it out of the loop (with a different name, like `currentTime`), or just use it directly in the conditional.

app/logic/getUpcomingEvents.py
```diff
@@ -0,0 +1,27 @@
...
+        event_time = event.timeStart
+        #FIXME: Change event.description to event.eventName (or just add it to the dict). When the eventName column is filled
+        if event.startDate >= today and event.timeStart > times:
```
BrianRamsay: Do we want to show them an event they have missed if the event was today but the current time is after the end timestamp? Right now you are including that case in the list.

app/logic/getUpcomingEvents.py
```diff
@@ -0,0 +1,27 @@
...
+                            .join(Program)
+                            .join(Interest, on=(Event.program == Interest.program))
+                            .where(Interest.user == user)
+                            )
+    upcomingEvents = []
+    for event in interestedEvent.objects():
+        today = date.today()
+        times = datetime.now().time()
+        event_time = event.timeStart
+        #FIXME: Change event.description to event.eventName (or just add it to the dict). When the eventName column is filled
```
BrianRamsay: Go ahead and resolve this FIXME now. I think event should have a `name` field in addition to `description`.

BrianRamsay (changes requested):

app/logic/getUpcomingEvents.py
Sandesh-L: Done!

app/logic/getUpcomingEvents.py
Sandesh-L: Done!

app/logic/getUpcomingEvents.py
Sandesh-L: Done!

app/logic/getUpcomingEvents.py
Sandesh-L: Done!

bryantal (dismissed):

app/logic/getUpcomingEvents.py
```diff
@@ -0,0 +1,20 @@
...
+from app.models.user import User
+from datetime import date, datetime, time
+
+def getUpcomingEventsForUser(user):
+
+    interestedEvents = (Event.select()
+                            .join(Interest, on=(Event.program == Interest.program))
+                            .where(Interest.user == user)
+                            )
+    upcomingEvents = []
```
BrianRamsay: I'd consider using a list comprehension here, instead of maintaining a separate data structure. You could essentially replace from here to the end with something like:
 `return [e for e in interestedEvents if e.startDate >= date.today() and e.timeStart > datetime.now().time()]`

BrianRamsay (changes requested):

app/logic/getUpcomingEvents.py
```diff
@@ -0,0 +1,15 @@
...
```
BrianRamsay: you don't need this variable anymore

app/templates/events/showUpcomingEvents.html
```diff
@@ -0,0 +1,23 @@
+<table class="table">
+
+  {% if upcomingEvents|length > 0 %}
+    <thead class="tableHeader">
+      <th scope="col"><h4><strong>Program</strong></h4>
+      <th scope="col"><h4><strong>Event Name</strong></h4>
+      <th scope="col"><h4><strong>Start Date</strong></h4>
+      <th scope="col"><h4><strong>End Date</strong></h4>
+        </thead>
```
BrianRamsay: indentation of the closing tag should match the opening one

BrianRamsay (changes requested):

BrianRamsay:
Only a couple of minor comments left!

app/logic/getUpcomingEvents.py
sreynit02: removed

app/templates/events/showUpcomingEvents.html
sreynit02: fixed

BrianRamsay (approved):
Looks good!

Issue #14, opened 2021-06-14
3.7


tylerpar99:
Volunteers must be able to indicate program interest.

PR #29, opened 2021-06-21
Created Volunteer Indicate Interest Section

gastonjarju:
This PR fixes issue #14 

**Summary:**

Created the following files:
_volunteerIndicateInterest.html_
_volunteerIndicateInterest.py_
_volunteerIndicateInterest.js_

In the _volunteerIndicateInterest.html_ we created the structure for our section. We added a head tag with the section title. Then, we added div tags looping through the programs' database listing all the programs. We then added checkboxes for every program. We also created a conditional that check the database for previously indicated interest and set the starting status of the checkboxes. We added another div tag to contain the flasher class. Finally, we created a small tag that contains the anchor tag, which leads to the CELTS program web page.

In the _volunteerIndicateInterest.js_, we created a function that onclick will be called to send the program_id and user_id to the controller in an Ajax call. To do this, we created a conditional that will check if a user is checking for interest or unchecking to remove interest. We then have the conditional select the appropriate route for deleting or adding interest. We have second conditional that will display a flasher if the Json is returned as Sucess.  

In the _volunteerIndicateInterest.py_, we created our imports and created three functions. The first function queries the database to pull in the appropriate data to render the template. And then in the second function, we have peewee create a row when the user indicates interest. In the third function, we have peewee delete a row when a user removes interest. 

**To Test:**
You can visit this URL: `http://<YOURIP>/volunteerIndicateInterest` you will see our basic section appear if you click on any options you can then check phpmyadmin and you will see the interest table update and if you refresh the page you will see your check box stays there. When you click you will also get a message that says your interests have been updated. If you uncheck a box you will see in php my admin the interest tabel has removed the row and if you refresh the page again you will see the program will stay unchecked. Below the table you will also see our link to the CELTS berea programs page.


BrianRamsay:
Go ahead and merge development into your branch so you can properly use the authenticated user

BrianRamsay (changes requested):
Need to merge in development so you can use the authenticated user instead of your current_user, and so you don't need your own jquery include

app/controllers/main/volunteerIndicateInterest.py
```diff
@@ -0,0 +1,46 @@
...
+
+@main_bp.route('/addInterest/<program_id>/<userID>', methods = ['POST'])
+def addInterest(program_id, userID):
+    """
+    This function updates the interest table by adding a new row when a user
+    shows interest in a program
+    """
+    try:
+        Interest.get_or_create(program = program_id, user = g.current_user)
+        return jsonify({"Success": True})
```
BrianRamsay: Do not use `jsonify({"Success": True})` or False anywhere. You're not checking the value, it's unnecessary.

app/controllers/main/volunteerIndicateInterest.py
```diff
@@ -0,0 +1,46 @@
...
+    return render_template('volunteerIndicateInterest.html',
+                           title="Volunteer Interest",
+                           user = g.current_user,
+                           programs = programs,
+                           interests = interests,
+                           interests_ids = interests_ids)
+
+@main_bp.route('/addInterest/<program_id>/<userID>', methods = ['POST'])
+def addInterest(program_id, userID):
+    """
```
BrianRamsay, line 13: Even though it is a simple function, you should create a separate function to perform the logic. That way you can test that the Interest gets created properly with our test suite.

app/controllers/main/volunteerIndicateInterest.py
```diff
@@ -0,0 +1,46 @@
...
+    try:
+        Interest.get_or_create(program = program_id, user = g.current_user)
+        return jsonify({"Success": True})
+    except Exception as e:
+        print("Error Updating Interest: ", e)
+        return jsonify({"Success": False}),500
+
+@main_bp.route('/deleteInterest/<program_id>/<userID>', methods = ['POST'])
+def deleteInterest(program_id, userID):
```
BrianRamsay: Like the above comments, create a function to test this separately from the controller. And don't those return json objects if you don't check them.

app/static/js/volunteerIndicateInterest.js
```diff
@@ -0,0 +1,39 @@
+$(document).ready(function() {
+$(".form-check-input").click(function updateInterest(el){
+  var programID = $(this).attr('id');
+  var interest = $(this).is(':checked');
+  var userID = $("#escalerapadronl").attr("id")
```
BrianRamsay: This shouldn't be hard-coded anymore.

app/static/js/volunteerIndicateInterest.js
```diff
@@ -0,0 +1,39 @@
...
+  if (interest) {
+  $.ajax({
+    method: "POST",
+    url: "/addInterest/" + programID + '/' + userID,
+    data: programID, userID,
+    success: function(response) {
+      if (response) {
+        $("#flasher-container").prepend('<div class="alert alert-success"" id="flasher" data-dismiss="alert" role="alert">Your Interest have been Updated</div>');
```
BrianRamsay: flashing a message is such a basic thing, let's make a javascript function in base.js that handles it.

app/static/js/volunteerIndicateInterest.js
```diff
@@ -0,0 +1,39 @@
...
+        $("#flasher").delay(3000).fadeOut();
+      }
+    },
+    error: function(request, status, error) {
+      console.log(request.responseText);
```
BrianRamsay: If this has failed, we need to indicate that it didn't work to the user (and probably uncheck the checkbox or refresh)

app/static/js/volunteerIndicateInterest.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+  });
+  }
+  else{
+  $.ajax({
```
BrianRamsay: These two ajax invocations should be combined. The only difference is the url

database/test_data.py
```diff
@@ -144,3 +145,17 @@
     },
 ]
 ProgramEvent.insert_many(programEvents).on_conflict_replace().execute()
+
```
BrianRamsay: There's already a spot where users are inserted. Don't do it twice.

tests/code/test_events.py
```diff
@@ -29,4 +30,8 @@ def test_getEventsInvalidProgram():
     with pytest.raises(DoesNotExist):
         getEvents(program_id= "asdf")
 
+@pytest.mark.integration
+def test_currentUseris():
+    user = indicateInterestController()
```
BrianRamsay: How does this work? I don't see that function name in your changes. And in any case, there's nothing here that is testing adding or deleting an insert.

BrianRamsay (changes requested):

app/controllers/main/volunteerIndicateInterest.py
BrianRamsay: still not done

app/controllers/main/volunteerIndicateInterest.py
```diff
@@ -0,0 +1,54 @@
...
+        else:
+            deleted_interest = Interest.get(Interest.program == program_id and Interest.user == g.current_user)
+            deleted_interest.delete_instance()
+        return jsonify(success=True)
+    except Exception as e:
+        print(e)
+        return "Error Updating Interest", 500
+
+# @main_bp.route('/deleteInterest/<program_id>', methods = ['POST'])
+# def deleteInterest(program_id):
```
BrianRamsay: don't leave big commented out chunks. If you want to go back to it, that's what git is for.

app/logic/test_updateInterest.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def updateInterest(program_id):
+    """
+    This function updates the interest table by adding a new row when a user
+    shows interest in a program
+    """
+    rule = "addInterest"
+    if 'addInterest' in rule:
+        Interest.get_or_create(program = program_id, user = "ramasayb2")
```
BrianRamsay: there's a hard-coded user id in here

app/static/js/volunteerIndicateInterest.js
```diff
@@ -0,0 +1,26 @@
...
+  if (interest) {
+    var routeUrl = "/addInterest/"
+  }
+  else {
+    var routeUrl = "/deleteInterest/"
+  }
+  $.ajax({
+    method: "POST",
+    url: routeUrl + programID,
+    date: programID,
```
BrianRamsay: I'm assuming this is supposed to be `data`, not `date`. But either way the program ID is in the url, so it isn't necessary.

app/static/js/volunteerIndicateInterest.js
```diff
@@ -0,0 +1,26 @@
...
+    success: function(response) {
+        msgFlash("Your interest have been updated", "success")
+    },
+    error: function(request, status, error) {
+    console.log(status,error);
```
BrianRamsay: Improper indentation.

app/static/js/volunteerIndicateInterest.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+  else {
+    var routeUrl = "/deleteInterest/"
+  }
+  $.ajax({
+    method: "POST",
+    url: routeUrl + programID,
+    date: programID,
+    success: function(response) {
+        msgFlash("Your interest have been updated", "success")
```
BrianRamsay: "Your interest **has** been updated"

app/logic/test_updateInterest.py
BrianRamsay: also a hard-coded variable

tests/code/test_events.py
```diff
@@ -29,4 +31,11 @@ def test_getEventsInvalidProgram():
     with pytest.raises(DoesNotExist):
         getEvents(program_id= "asdf")
 
-
+@pytest.mark.integration
```
BrianRamsay: test is inadequate. Doesn't test error condition,

BrianRamsay (changes requested):

app/controllers/main/volunteerIndicateInterest.py
Sandesh-L: added a return of "Successfully added/deleated interest" so that the test suite has an easier thing to assert.

app/static/js/volunteerIndicateInterest.js
Sandesh-L: it was removed already in a previous commit

app/static/js/volunteerIndicateInterest.js
Sandesh-L: already resolved

app/static/js/volunteerIndicateInterest.js
Sandesh-L: a flasher message was already added. We added the reload functionality after the message is displayed.

database/test_data.py
Sandesh-L: was already resolved

tests/code/test_events.py
Sandesh-L: Don't know. We are working on adding new test cases. This might have already been resolved.

app/logic/addRemoveInterest.py
```diff
@@ -0,0 +1,26 @@
...
+    print(rule)
+    print(type(rule))
+    #print(type(rule.rule))
+    if 'addInterest' in str(rule):
+        Interest.get_or_create(program = program_id, user = currentUser)
+    else:
+        deleted_interest = Interest.get(Interest.program == program_id, Interest.user == currentUser)
+        print(deleted_interest)
+        deleted_interest.delete_instance()
+    return 'True'
```
BrianRamsay: this return statement shouldn't be here. It never changes, and isn't even a boolean. If you don't care about what it returns, don't return something.

app/logic/addRemoveInterest.py
```diff
@@ -0,0 +1,26 @@
...
+    rule: Gets the url from the ajax call, specifies to add or remove interest
+    program_id: id of the program the user is interested in
+    """
+    print(rule)
+    print(type(rule))
+    #print(type(rule.rule))
+    if 'addInterest' in str(rule):
+        Interest.get_or_create(program = program_id, user = currentUser)
+    else:
+        deleted_interest = Interest.get(Interest.program == program_id, Interest.user == currentUser)
```
BrianRamsay: This will throw an exception if the interest doesn't exist, and since we're deleting it we probably just want to ignore that case, rather than bothering someone with an exception.

tests/code/test_addRemoveInterest.py
```diff
@@ -0,0 +1,55 @@
+import pytest
+from app.logic.addRemoveInterest import addRemoveInterest
+from flask import json
+
+@pytest.mark.integration
+def test_addRemoveInterest():
```
BrianRamsay, line 6: This misses the case where addRemoveInterest doesn't do what you want. How could the function fail?

BrianRamsay (changes requested):

app/logic/addRemoveInterest.py
Sandesh-L: We've changed it to something more useful... hopefully. We have left the return statement so that we can test that the function worked properly

tests/code/test_addRemoveInterest.py
Sandesh-L: We've added some more test cases

app/controllers/main/routes.py
```diff
@@ -11,3 +12,15 @@ def home():
         #TODO We have to return some sort of error page
         print('Error in main page:', e)
         return "",500
+
+@main_bp.route('/volunteerIndicateInterest', methods = ['GET'])
+def volunteerIndicateInterest():
+    programs = Program.select()
+    interests = Interest.select().where(Interest.user_id == g.current_user)
```
BrianRamsay: use `Interest.user` instead of `Interest.user_id`. Let's avoid referencing ids directly wherever we can

app/templates/volunteerIndicateInterest.html
```diff
@@ -0,0 +1,32 @@
...
+          <input class="form-check-input" type="checkbox" id="{{ program.id }}" {{ status }}>
+          <label class="form-check-label" style="padding-right:50px;" for="{{ program.id }}">{{program.programName}}</label>
+  {% endfor %}
+  </div>
+  <div>
+    <small>
+      <a href="https://www.berea.edu/celts/community-service-programs/volunteer-opportunities/">For more information click here.</a>
+    </small>
+  </div>
+  <div id="flasher-container">
```
BrianRamsay: isn't there a container for the flash messages in base.html?

tests/code/test_addRemoveInterest.py
```diff
@@ -0,0 +1,60 @@
+import pytest
+from app.logic.addRemoveInterest import addRemoveInterest
+from peewee import IntegrityError, DoesNotExist
+
+@pytest.mark.integration
+def test_addRemoveInterest():
+    user = "ramsayb2"
+    program_id = 1
+
```
BrianRamsay, line 9: you are using a variable named `addInterest` for all of these, which is confusing. I would change it to `result` so it matches how it is used

tests/code/test_addRemoveInterest.py
```diff
@@ -0,0 +1,60 @@
...
+def test_addRemoveInvalidInterest():
+
+    program_id = 3
+
+    #test for user that doesn't exist
+    user = "al;skfjelh"
+    rule = 'addInterest'
+    with pytest.raises(IntegrityError):
+        addInterest = addRemoveInterest(rule, program_id, user)
+        assert addInterest == "Successfully added interest"
```
BrianRamsay: you can get rid of the assert. the test is that it expects an exception

BrianRamsay (changes requested):

app/templates/volunteerIndicateInterest.html
Sandesh-L: There is. It was probably created before flashers were added. We have removed it.

BrianRamsay (approved):
Looks good!

PR #32, opened 2021-06-24
Issue 3.10. Viewing service transcript 21

Karina-Agliullova:
**Problem:** The volunteer must be able to view their service transcript/summary. The service transcript includes programs and service-learning courses. The user or an admin should be able to access the transcript.

**Solution:** We created a file called view_transcript.py in the app/controllers/events folder. The file contains three functions: getSLcoursesTranscript(), getProgramTranscript(), and getUser(). The first function returns a list with courses that the user took, displaying a course name, term, instructor of the course, and the hours earned from each course. We had to create a table for course instructors as there was nothing for it. The second function returns a list of programs that the user participated in. displaying a name of the program, term, and hours accrued. The third function returns the first and the last name of the user when passed the user object.

**Tests:** We created a test suite called test_transcripts.py in the celts/tests/code folder. The file contains one test suite called test_getTranscripts() that should pass when the user object passed. The other three test suites check if the passed user in each function does not exist, so it should also pass. We also created some test suites in test_data.py for coursestatus, courses, courseInstructorRecords, courseHoursEarned, programHoursEarned. test_data.py should work with the data that is in it.


app/controllers/events/routes.py
```diff
@@ -8,7 +8,7 @@
 def events():
     events = getEvents()
 
-    return render_template("events/event_list.html", 
+    return render_template("events/event_list.html",
             events=events,
             user="ramsayb2"
```
BrianRamsay: Don't hard-code a user anymore - there's a global option now.

app/controllers/events/view_transcript.py
```diff
@@ -0,0 +1,70 @@
...
+from app.models.courseParticipant import CourseParticipant
+from app.models.program import Program
+from app.models.courseInstructors import CourseInstructors
+from app.models.user import User
+from app.models.term import Term
+from app.models.eventParticipant import EventParticipant
+from app.models.event import Event
+
+
+def getSLCourseTranscript(user):
```
BrianRamsay: This whole function is almost a textbook candidate for a single SQL query. Maybe two, if you wanted to get instructors separate from hours, but I would look into SQL aggregate functions and GROUP BY, or what the peewee version of that would be. Then you wouldn't have to do any of the loops and conditionals

BrianRamsay: And the nice thing about having a test for it (good job), is you've got an easy way to keep running it and verify that the results are correct.

app/controllers/events/view_transcript.py
```diff
@@ -0,0 +1,70 @@
...
+        instructorList = []
+        for instructor in cInstructor:
+            instructorList.append(instructor.user.firstName+" "+ instructor.user.lastName)
+
+        cHours = courseSumDict[cName]
+        courseList.append([cName, cTerm, instructorList, cHours])
+
+    return courseList
+
+def getProgramTranscript(user):
```
BrianRamsay: Like the above, this could be greatly simplified by doing it all in the database query.

app/controllers/events/view_transcript.py
```diff
@@ -0,0 +1,70 @@
...
+
+    for program in programs:
+        pName = program.programName
+        pTerm = program.term.description
+        pHours = programSumDict[pName]
+        programList.append([pName, pTerm, pHours])
+
+    return programList
+
+def getUser(user):
```
BrianRamsay: You create this user function (which is handy), but you don't use it to do the instructor name in the above function. 

It should be named better as well, because you already have a user - it's what you pass in. You wouldn't need to call a function called `getUser`. `getFullName` would be better, but I think an even better option is to add a `fullName` property to the user object itself. Then, whenever you have a user you can just call `user.fullName`.
I would look at add

app/models/courseInstructors.py
```diff
@@ -0,0 +1,7 @@
+from app.models import*
+from app.models.course import Course
+from app.models.user import User
+
+class CourseInstructors(baseModel):
```
BrianRamsay: Follow the naming convention of the other tables. In general, they shouldn't be plural. The table name describes what each record in the table is: in this case, a single instructor.

app/models/program.py
```diff
@@ -1,6 +1,9 @@
 from app.models import*
+from app.models.term import Term
+from app.models.courseStatus import CourseStatus
 from app.models.partner import Partner
 
 class Program(baseModel):
     programName = CharField()
     partner = ForeignKeyField(Partner, null=True)
+    term = ForeignKeyField(Term)
```
BrianRamsay: I don't think Program should have a term in it. What is the use case for this?

app/templates/transcripts.html
```diff
@@ -0,0 +1,39 @@
+<div class = 'container'>
```
BrianRamsay: Looks like there is a lot of hard-coded stuff here.

tests/code/test_transcripts.py
```diff
@@ -0,0 +1,49 @@
...
+from app.controllers.events.view_transcript import getUser
+from app.models.term import Term
+from app.models.course import Course
+from app.models.courseParticipant import CourseParticipant
+from app.models.user import User
+from app.models.eventParticipant import EventParticipant
+
+@pytest.mark.integration
+def test_noGetSLTranscript():
+
```
BrianRamsay, line 14: You're not actually testing your functions here. `User.get_by_id` is what is throwing the DoesNotExist error in all of these tests, not your functions.

tests/code/test_transcripts.py
```diff
@@ -0,0 +1,49 @@
...
+def test_getTranscripts():
+    user = User.get_by_id("neillz")
+    transcript = getSLCourseTranscript(user)
+    assert transcript[0] == ["Databases", "Spring B 2021", ["Zach Neill"], 2.0]
+    assert transcript[1] == ["Spanish Help", "Spring A 2021", ["Zach Neill","Brian Ramsay"], 3.0]
+    transcript = getProgramTranscript(user)
+    assert transcript[0] == ["Training", "Fall 2021", 3.0]
+    assert transcript[1] == ["Adopt A Grandparent", "Summer 2021", 3.0]
+    assert transcript[2] == ["Berea Buddies", "Spring B 2021", 1.0]
+    name = getUser(user)
```
BrianRamsay: I like that you put a test for this as well! 

But it should be in its own test method. (given the comments above about it, it will be in a separate file as well. But as it is now it should be a separate method).

BrianRamsay (changes requested):

sreynit02:
PR Updated

tests/code/test_transcripts.py
```diff
@@ -0,0 +1,76 @@
...
+        sLcourse = getProgramTranscript(User.get_by_id("adfafa"))
+    with pytest.raises(DoesNotExist):
+        sLcourse = getProgramTranscript(User.get_by_id(56498))
+
+@pytest.mark.integration
+def test_getSLTranscripts():
+    user = User.get_by_id("neillz")
+    transcript = getSLCourseTranscript(user)
+
+    assert transcript[0][0]['fullName'] == ["Zach Neill"]
```
BrianRamsay: I'm a little confused by the double array index here. Isn't it just one list of courses? Why would it return each field in its own dictionary that is in it's own list item?

app/models/program.py
sreynit02: We deleted it now, since we are getting information for term from the query where we join program and event.

app/models/courseInstructors.py
sreynit02: It is changed to single instructor now

app/controllers/events/view_transcript.py
sreynit02: Our new function is using a query to get all the necessary information for the course trancript. We used a loop to append the information into a list of dictionaries.

tests/code/test_transcripts.py
sreynit02: We moved `User.get_by_id` into our functions so that they don't accept users that are not in the database.

tests/code/test_transcripts.py
sreynit02: We used double array because we were testing to see if we were able to get the pieces of information (ex: course_name, hours ...) from our query. But we now got rid of the nested lists and just have one list for the whole transcripts so our code looks cleaner.

app/templates/transcripts.html
sreynit02: We decided to get rid of this file and will branch off this PR to work on the front end of this issue. Our new html file will inherit from base.html

app/controllers/events/view_transcript.py
```diff
@@ -0,0 +1,74 @@
...
+    sLCourseInformation = (CourseParticipant
+    .select(CourseParticipant.course, CourseParticipant.user, fn.SUM(CourseParticipant.hoursEarned).alias("hoursEarned"))
+    .group_by(CourseParticipant.course, CourseParticipant.user)
+    .where(CourseParticipant.user == username))
+
+    courses = [item for item in sLCourseInformation.objects()]
+
+    allCoursesList = []
+    listOfEachSLCourseInfo = []
+    for i in range(0,len(courses)):
```
BrianRamsay: The python way to iterate is `for course in courses:`

app/controllers/events/view_transcript.py
```diff
@@ -0,0 +1,74 @@
...
+    list of instructors who teach the course, and hours earned for each course that the user took.
+    :user: model object
+    """
+    username = User.get_by_id(user)
+    sLCourseInformation = (CourseParticipant
+    .select(CourseParticipant.course, CourseParticipant.user, fn.SUM(CourseParticipant.hoursEarned).alias("hoursEarned"))
+    .group_by(CourseParticipant.course, CourseParticipant.user)
+    .where(CourseParticipant.user == username))
+
+    courses = [item for item in sLCourseInformation.objects()]
```
BrianRamsay: I don't think this is necessary, is it? You can iterate through sLCourseInformation without it

app/controllers/events/view_transcript.py
```diff
@@ -0,0 +1,74 @@
...
+from peewee import *
+
+def getSLCourseTranscript(user):
+    """
+    Returns a service-learning course list that contains a name of the course, term,
+    list of instructors who teach the course, and hours earned for each course that the user took.
+    :user: model object
+    """
+    username = User.get_by_id(user)
+    sLCourseInformation = (CourseParticipant
```
BrianRamsay: leave the L lowercase

app/controllers/events/view_transcript.py
```diff
@@ -0,0 +1,74 @@
...
+        .select(EventParticipant.event , fn.SUM(EventParticipant.hoursEarned).alias("hoursEarned"))
+        .where(EventParticipant.user == username)
+        .join(Event)
+        .join(Program)
+        .group_by(EventParticipant.event.program))
+
+    listOfProgramsTranscript = []
+    programTranscript = []
+
+    for i in range(0, len(programInformation)):
```
BrianRamsay: similar to above: `for ep in programInformation:` or something

tests/code/test_transcripts.py
```diff
@@ -0,0 +1,57 @@
...
+@pytest.mark.integration
+def test_noGetProgramTranscript():
+
+    with pytest.raises(DoesNotExist):
+        sLcourse = getProgramTranscript("adfafa")
+    with pytest.raises(DoesNotExist):
+        sLcourse = getProgramTranscript(56498)
+
+@pytest.mark.integration
+def test_getSLTranscripts():
```
BrianRamsay: It's a good idea to test the empty case and make sure there isn't an error. Like, for a valid user without any courses do you just get an empty array like you expect?

BrianRamsay (changes requested):

app/controllers/events/view_transcript.py
sreynit02: changed L to l

app/controllers/events/view_transcript.py
sreynit02: changed line 29 to for course in courses: and got rid of indexing after in that for loop

app/controllers/events/view_transcript.py
sreynit02: changed line 64 to for program in programInformation: and got rid of indexing in that for loop

app/controllers/events/view_transcript.py
sreynit02: Yes, we don't need it. We removed that loop and it still works and just passed in slCourseInformation in our for loop below it and it works.

tests/code/test_transcripts.py
sreynit02: we added a test for a user who doesn't have any cases and it returns an empty array.

app/controllers/events/view_transcript.py
```diff
@@ -0,0 +1,72 @@
...
+    for course in slCourseInformation:
+        user_full_name = course.user.firstName + " "+ course.user.lastName
+        course_name = course.course.courseName
+        description = course.course.term.description
+        hours = course.hoursEarned
+        cInstructor = CourseInstructor.select().where(CourseInstructor.course== course.course)
+        instructorList = []
+        for instructor in cInstructor.objects():
+            instructorList.append(instructor.user.firstName+" "+ instructor.user.lastName)
+        #creates a list for all information of a course that a courseParticipant is involved in
```
BrianRamsay, line 32: - Because this select is a simple select, you don't need to use `.objects()` to get the values out (I'm pretty sure). 
- `cInstructor` isn't a great name, because it is potentially a list, so it should be plural or, better yet, an indication that it is a query object. 
- When you have a simple transform of a list in python, writing it as a list comprehension is preferred. 
- Another consideration is that you don't need to do that javascript-like ` + " " + ` string concatentation because of f-strings in python. For short strings like this one the readability is a matter of personal preference, so that's up to you if you want to change it. 

Taking all that together, this code might be something like:

`instructorQuery = CourseInstructor.select().where(CourseInstructor.course== course.course)`
`instructorList = [f"{i.user.firstName} {i.user.lastName}" for i in instructorQuery]`

app/controllers/events/view_transcript.py
```diff
@@ -0,0 +1,72 @@
...
+
+def getSlCourseTranscript(user):
+    """
+    Returns a service-learning course list that contains a name of the course, term,
+    list of instructors who teach the course, and hours earned for each course that the user took.
+    :user: model object
+    """
+    username = User.get_by_id(user)
+    slCourseInformation = (CourseParticipant
+    .select(CourseParticipant.course, CourseParticipant.user, fn.SUM(CourseParticipant.hoursEarned).alias("hoursEarned"))
```
BrianRamsay: these should be indented at least 1 level

app/controllers/events/view_transcript.py
```diff
@@ -0,0 +1,72 @@
...
+        for instructor in cInstructor.objects():
+            instructorList.append(instructor.user.firstName+" "+ instructor.user.lastName)
+        #creates a list for all information of a course that a courseParticipant is involved in
+        listOfEachSlCourseInfo = [ user_full_name , course_name , description , hours , instructorList ]
+        allCoursesList.append(listOfEachSlCourseInfo)
+
+    return allCoursesList
+
+
+def getProgramTranscript(user):
```
BrianRamsay: You have user and username flipped for your variable naming. If you are expecting a username passed in to the function and not a user object you should call it `username`, not `user`. down below, if you are getting a User object with peewee, you should call the variable `user`, not `username`.

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #34, opened 2021-06-27
Footer breaks on small screen sizes


romano-w:
The footer text breaks into a new line on smaller screen sizes because it is too wide. This breaks the layout because the footer uses a fixed height and the text is vertically centered by having a high line height.

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #34: Footer breaks on small screen sizes](https://trello.com/c/coM5HmhV/279-celts-34-footer-breaks-on-small-screen-sizes)

PR #35, opened 2021-06-27
Base template responsive update

### LINKED ISSUE #23: Make base templates responsive
Currently, the base template does not break into a new layout for smaller screen sizes. The sidebar should probably become an offcanvas element and the document should flow accordingly.

romano-w:
The layout breaks on smaller screen sizes with the sidebar being replaced with a hamburger button that opens an offcanvas version of the sidebar. The main content is full width in the smaller layout which will give more space for content.
Closes #23


romano-w:
Additionally, font awesome icons were replaced with the new Bootstrap 5 icon pack but if we don't want to do that it is simple to revert.

romano-w:
The footer still does break on smaller screen sizes (when the text wraps the line height becomes a problem) but I created issue #34 to address that.

BrianRamsay (approved):

PR #37, opened 2021-07-06
Created a page where admins can create new events

Sandesh-L:
Summary:
For our issue, we had to create a webpage where admins could add new events.  We first created the UI and the webpage for the admin. Then we added some minor functionality such as datepickers and timepickers. Afterward, we reformated the UI to work with bootstrap. Then we worked on retrieving data from the UI. After that, we added validation on both the front and backend. We also added test cases. (it's a lot)

To test:
run the website the go to <yourIP>:8080/1/create_event
Try to create an event. 
hit save after leaving some fields blank to check for required fields.
Set start time to be after end time to check for an example of backend validation
try to create an event that already exists.
check the training for program toggle without checking required for the program for another example of backed validation.
(run the test cases for most of the other stuff)


app/controllers/admin/createEvents.py
```diff
@@ -0,0 +1,35 @@
...
+from flask import json, jsonify
+from flask import request
+from app.controllers.admin import admin_bp
+from app.logic.adminCreateEvent import getTermDescription
+from app.logic.adminNewEvent import createEvent, manageNewEventData
+from app.logic.validateNewEvent import validateNewEventData
+from app.models.facilitator import Facilitator
+from flask import flash, redirect, url_for, g
+
+@admin_bp.route('/createEvents', methods=['POST'])
```
BrianRamsay: These names should be singular. I thought when you were demoing this change was already made, so I'm not sure what we were seeing.

app/templates/admin/createEvents.html
```diff
@@ -0,0 +1,151 @@
...
+{% block scripts %}
+  {{super()}}
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-select/1.12.1/js/bootstrap-select.js"></script>
+  <script src="/static/js/createEvents.js"></script>
+
+{% endblock %}
+
+{% block styles %}
+  {{super()}}
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css">
```
BrianRamsay: we don't want an old bootstrap included

app/logic/adminCreateEvent.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def getTermDescription():
+
+    termDescriptions = Term.select(Term.description)
+    descriptions = [des.description for des in termDescriptions.objects()]
+
+    return descriptions
+
+def getCurrentTerm():
```
BrianRamsay: I think you should just return a Term object. Whatever calls this function can just use term.description to get the description, and then the function is more useful because the ID goes with it.

app/logic/adminCreateEvent.py
```diff
@@ -0,0 +1,23 @@
+from app.models import*
+from app.models.term import Term
+from app.models.user import User
+
+
+def getTermDescription():
```
BrianRamsay: A list of terms makes more sense to me than a list of descriptions, because the description is accessible from the term. Also, the name should be plural.

app/logic/adminCreateEvent.py
```diff
@@ -0,0 +1,23 @@
...
+
+def getCurrentTerm():
+
+    currentTerm = Term.select(Term.description).where(Term.isCurrentTerm == 1)
+    term = [t.description for t in currentTerm.objects()] #This could be done better. but how??
+    return term[0]
+
+def getFacilitators():
+
+    facilitators = User.select(User).where((User.isFaculty == 1) | (User.isCeltsAdmin == 1) | (User.isCeltsStudentStaff == 1))
```
BrianRamsay: can you just return this variable? What do you gain by making it a list later? It's still iterable both ways.

app/logic/adminNewEvent.py
```diff
@@ -0,0 +1,40 @@
+
+from app.models.term import Term
+from app.models.event import Event
+from app.models.facilitator import Facilitator
+
+def manageNewEventData(eventData):
```
BrianRamsay: 'manage' is too generic a name. I am not sure what this function is supposed to do

app/logic/adminNewEvent.py
```diff
@@ -0,0 +1,40 @@
...
+                              isRecurring = newEventData['recurringEvent'],
+                              isRsvpRequired = newEventData['eventRSVP'], #rsvp
+                              isRequiredForProgram = newEventData['eventRequiredForProgram'],
+                              isTraining = newEventData['eventIsTraining'],
+                              isService = newEventData['eventServiceHours'],
+                              startDate =  newEventData['eventStartDate'],
+                              endDate =  newEventData['eventEndDate'],
+                              program_id = newEventData['programId'])
+
+    eventID = Event.select(Event.id).where((Event.eventName == newEventData['eventName']) &
```
BrianRamsay: you don't need to select the event after you create it. You can just use eventEntry later

app/logic/adminNewEvent.py
```diff
@@ -0,0 +1,40 @@
...
+
+    for checkBox in eventCheckBoxes:
+        if checkBox not in eventData:
+            eventData[checkBox] = False
+
+    return eventData
+
+def createEvent(newEventData):
+
+    term = Term.select(Term.id).where(Term.description == newEventData['eventTerm'])
```
BrianRamsay: I would rework your form so that you're passing term_id around. That way you don't need to select from Term just to insert the proper term below.

app/models/event.py
```diff
@@ -4,12 +4,13 @@
 
 class Event(baseModel):
     eventName = CharField()
-    program = ForeignKeyField(Program, null = True)
+    program = ForeignKeyField(Program.id, null = True)
```
BrianRamsay: Do you know why specifying this was necessary? We don't have to do it on any other ForeignKeyField.

app/static/js/createEvents.js
```diff
@@ -0,0 +1,72 @@
+
+$(document).ready(function(){
+
+  $(".readonly").on('keydown paste', function(e){ //makes the input fields act like readonly (readonly doesn't work with required)
+        if(e.keyCode != 9) // ignore tab
+            e.preventDefault();
+    });
+
+  $.datepicker.setDefaults({
+    dateFormat:'yy-mm-dd'
```
BrianRamsay: needs to be mm-dd-yy

app/static/js/createEvents.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+    $("#endDatePicker").datepicker({minDate: new Date(newYear, newMonth, newDay)});
+    $("#endDatePicker").datepicker( "option", "minDate", new Date(newYear, newMonth, newDay));
+  }
+}
+
+
+
+
+function dropHandler(ev) {
```
BrianRamsay: Seems like this is functionality we should be able to get from jquery ui or something

BrianRamsay (changes requested):

app/static/js/createEvents.js
BrianRamsay: Though since we're not worrying about files yet you could just drop the whole thing

app/models/event.py
gastonjarju: It was neccessary at some point, but it seems it is no more neccessary because when we removed it and tested the web page and it all seemed to work.

app/controllers/admin/createEvents.py
```diff
@@ -0,0 +1,42 @@
...
+
+    if g.current_user.isCeltsAdmin:
+        #add check for admin
+        eventData = request.form.copy() #since request.form returns a immutable dict. we need to copy to change the
+        newEventData= setValueForUncheckedBox(eventData)
+
+
+        startDate = parser.parse(newEventData['eventStartDate'], dayfirst=True) ##THIS IS WHAT WE WANT
+        print(startDate)
+        print(type(startDate))
```
BrianRamsay: make sure all of your debug output is gone

app/controllers/admin/routes.py
```diff
@@ -1,8 +1,29 @@
...
-@admin_bp.route('/testing', methods=['GET'])
+@admin_bp.route('/testing_things', methods=['GET'])
 def testing():
     return "<h1>Hello</h1>"
+
+@admin_bp.route('/<program>/create_event', methods=['GET'])
+def createEventPage(program):
+    listOfTerms = Term.select()
+    facilitators = getAllFacilitators()
+    user = g.current_user
```
BrianRamsay: g.current_user is available to all templates, so you don't need to pass it in separately

app/logic/adminNewEvent.py
```diff
@@ -0,0 +1,44 @@
...
+    return eventData
+#
+# def setDateFormatYMD(eventData):
+#
+#     newEventData['eventStartDate'] = datetime.datetime.strptime(newEventData['eventStartDate'], '%m-%d-%Y').strftime('%Y-%m-%d')
+#     newEventData['eventEndDate'] = datetime.datetime.strptime(newEventData['eventEndDate'], '%m-%d-%Y').strftime('%Y-%m-%d')
+#     newEventData['eventStartDate'] = setDateFormatYMD(newEventData['eventStartDate'])
+#
+#     eventStartDate = datetime.datetime.strptime(newEventData['eventStartDate'], '%m-%d-%Y').strftime('%Y-%m-%d')
+#     eventEndDate = datetime.datetime.strptime(newEventData['eventEndDate'], '%m-%d-%Y').strftime('%Y-%m-%d')
```
BrianRamsay: Remove commented out code (for the most part)

tests/code/test_AdminCreateEvent.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+from app.logic.adminCreateEvent import getTermDescription, getCurrentTerm, getFacilitators
+
+@pytest.mark.integration
+def test_getTermDescription():
```
BrianRamsay: This function doesn't exist anymore

tests/code/test_AdminNewEvent.py
```diff
@@ -0,0 +1,44 @@
...
+    assert newData['eventRSVP'] == False
+    assert newData['eventServiceHours'] == False
+    assert newData['eventIsTraining'] == False
+
+    #test for one missing key
+    eventData = {'eventRequiredForProgram':'on','eventRSVP':'', 'eventServiceHours':True }
+    newData = manageNewEventData(eventData)
+
+    assert newData['eventIsTraining'] == False
+    with pytest.raises(AssertionError):
```
BrianRamsay: I'm not quite sure what the negative assertion is here. If you just want to see if the data is still there, check that it has the right value

tests/code/test_validateNewEvent.py
```diff
@@ -0,0 +1,57 @@
...
+from app.logic.validateNewEvent import validateNewEventData
+
+@pytest.mark.integration
+def test_correctValidateNewEventData():
+
+    validateEventData = {'eventEndDate': '2021-07-03', 'eventStartDate':'2021-07-02',
+                         'eventStartTime': '10:33', 'eventEndTime':'10:35',
+                         'eventIsTraining':'on', 'eventRequiredForProgram':True,
+                         'eventName':'Berea Buddies', 'eventDescription':'Berea Buddies Training'}
+    validateNewEvent = validateNewEventData(validateEventData)
```
BrianRamsay: save the return value into two variables, not one as a list

BrianRamsay (changes requested):

tests/code/test_AdminNewEvent.py
Sandesh-L: The value of newData['eventIsTraining'] is False

app/controllers/admin/createEvents.py
```diff
@@ -0,0 +1,37 @@
...
+
+@admin_bp.route('/createEvent', methods=['POST'])
+def createEvent():
+
+    if g.current_user.isCeltsAdmin:
+
+        eventData = request.form.copy() #since request.form returns a immutable dict. we need to copy to change the
+        newEventData= setValueForUncheckedBox(eventData)
+
+        # reforamt date int Y-m-d for the backend
```
BrianRamsay: A couple typos. Also, you're not reformatting, you're getting a datetime object

app/controllers/admin/createEvents.py
```diff
@@ -0,0 +1,37 @@
...
+        newEventData['eventStartDate'] = parser.parse(newEventData['eventStartDate'], dayfirst=True)
+        newEventData['eventEndDate'] = parser.parse(newEventData['eventEndDate'], dayfirst=True)
+
+
+        # function to validate data
+        validNewEventData, eventErrorMessage = validateNewEventData(newEventData)
```
BrianRamsay: variable names could use some work. Booleans should often use `is` in the name. I was confused if validNewEventData was a dictionary of validated data or True or False. Since you're using it in an `if` statement I'd call it `dataIsValid`.

also, `eventErrorMessage` is a bit generic. In this specific case, the error message is about validation, so you could name it `validationErrorMessage` or something like that.

app/static/js/createEvents.js
```diff
@@ -0,0 +1,52 @@
...
+    $("#startDatePicker").datepicker().datepicker("show"); // Shows the start date datepicker when glyphicon is clicked
+  });
+
+  $("#calendarIconEnd").click(function() {
+      $("#endDatePicker").datepicker().datepicker("show"); // Shows the start date datepicker when glyphicon is clicked
+    });
+
+  $("#checkIsTraining").click(function(){
+    var status = $("input[name='eventIsTraining']:checked").val()
+    console.log(status);
```
BrianRamsay: debug output still in here

app/templates/base.html
```diff
@@ -8,13 +8,21 @@
...
+
     {% endblock %}
 
     {% block scripts %}
         <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
         <script src="https://code.jquery.com/ui/1.12.1/jquery-ui.js"></script>
         <script src="{{url_for('static', filename='js/bootstrap.bundle.min.js') }}?u={{lastStaticUpdate}}" type="text/javascript"></script>
         <script src="{{url_for('static', filename='js/base.js') }}?u={{lastStaticUpdate}}" type="text/javascript"></script>
+        <script src="https://code.jquery.com/ui/1.12.1/jquery-ui.js"></script>
+        <script src="//cdnjs.cloudflare.com/ajax/libs/timepicker/1.3.5/jquery.timepicker.min.js"></script>
```
BrianRamsay: I would include the timepicker css and js just on the pages that need it. there aren't that many.

tests/code/test_AdminNewEvent.py
```diff
@@ -0,0 +1,45 @@
...
+
+    assert newData['eventRequiredForProgram'] == 1
+    assert newData['eventRSVP'] == 2
+    assert newData['eventServiceHours'] == 3
+    assert newData['eventIsTraining'] == 4
+
+
+
+
+def test_createNewEvent():
```
BrianRamsay: test is empty

tests/code/test_GetAllFacilitators.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+from app.logic.getAllFacilitators import getAllFacilitators
+
+
+@pytest.mark.integration
+def test_getAllFacilitators():
+    userFacilitator = getAllFacilitators()
```
BrianRamsay, line 7: since it is a getAll function, you should do a test where there are multiple facilitators, as well as a test where there are none

tests/code/test_GetAllFacilitators.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+@pytest.mark.integration
+def test_getAllFacilitators():
+    userFacilitator = getAllFacilitators()
+
+
+    assert userFacilitator[0].username == "ramsayb2"
+    assert userFacilitator[0].isFaculty == 0
+    assert userFacilitator[0].isFaculty == False
```
BrianRamsay, line 13: Since you want it to be False, just check if it's False. the check against 0 doesn't really add anything

BrianRamsay (changes requested):

app/controllers/admin/createEvents.py
```diff
@@ -0,0 +1,37 @@
...
+
+            flash("Event successfully created!")
+            return redirect(url_for("admin.createEventPage", program=newEventData['programId']))
+
+        else:
+            flash(validationErrorMessage)
+            return redirect(url_for("admin.createEventPage", program=2)) #FIXME: have this redirect to main programs page (or some appropriate non admin page).
+
+    flash("Only celts admins can create an event!")
+    return redirect(url_for("admin.createEventPage", program=2)) #FIXME: have this redirect to main programs page (or some appropriate non admin page).
```
BrianRamsay: I would move these up to the top. That way the admin check and the outcome if they are not an admin are all together, like a gateway to the rest of the controller method.

app/templates/admin/createEvents.html
```diff
@@ -0,0 +1,156 @@
+{% extends "base.html"%}
+{% block scripts %}
+  {{super()}}
+  <script src="/static/js/createEvents.js"></script>
+  <script src="//cdnjs.cloudflare.com/ajax/libs/timepicker/1.3.5/jquery.timepicker.min.js"></script>
+  <script src="https://code.jquery.com/ui/1.12.1/jquery-ui.js"></script>
```
BrianRamsay: jquery UI should be part of base.html already

tests/code/test_AdminNewEvent.py
BrianRamsay: What I mean is, why are you checking to see if an assert raises an exception rather than just asserting the value you expect directly?

BrianRamsay (changes requested):

tests/code/test_AdminNewEvent.py
gastonjarju: We have removed the exception and use assert to test.

BrianRamsay:
We're trying to do too much in one PR. Let's get this PR in and merged, and then you can work on recurring events. That way other people aren't waiting on the finished code.

tests/code/test_GetAllFacilitators.py
gastonjarju: We tested for multiple users by indexing the different facilitator objects in the list. We couldn't really think of a way to test for no facilitators especially after having tested for multiple facilitators. And removing all facilitators from the test_data.py would cause the website functionality to fail. So we just asserted that the length of the facilitators list was >=1 so it makes sure there is at least one facilitator.

tests/code/test_AdminNewEvent.py
gastonjarju: Created a test suite that test the createNewEvent function.

gastonjarju:
> We're trying to do too much in one PR. Let's get this PR in and merged, and then you can work on recurring events. That way other people aren't waiting on the finished code.

We have fixed all the PR comments made in the last 24 hours. However, we did push with the incorrect commit message. The correct message should be "Created a test suite to test the createNewEvent function. Remove the unneccessary recurring events radio buttons and associated js functionalities.Cleaned up some backend code for the recurring events."

app/logic/adminNewEvent.py
```diff
@@ -0,0 +1,63 @@
+from app.models.event import Event
+from datetime import *
+# from datetime import timedelta
+from app.models.facilitator import Facilitator
+# from dateutil import parser
+
+def calculateRecurringEventFrequency(recurringEventInfo):
```
BrianRamsay: This function does not return anything. I'm not sure how you can use it.

app/logic/adminNewEvent.py
```diff
@@ -0,0 +1,63 @@
...
+
+    eventCheckBoxes = ['eventRequiredForProgram','eventRSVP', 'eventServiceHours', 'eventIsTraining', 'eventIsRecurring']
+
+    for checkBox in eventCheckBoxes:
+        if checkBox not in eventData:
+            eventData[checkBox] = False
+
+    return eventData
+
+def createNewEvent(newEventData):
```
BrianRamsay, line 19: A function like this should really return the Event it has created, rather than a message that you want to be passed to the user. Let the controller worry about what message to create on success.

app/logic/adminNewEvent.py
```diff
@@ -0,0 +1,63 @@
...
+                                  endDate =  newEventData['eventEndDate'],
+                                  program = newEventData['programId'])
+
+        facilitatorEntry = Facilitator.create(user = newEventData['eventFacilitator'],
+                                                  event = eventEntry)
+
+        return ("Event successfully created!")
+
+    except:
+        raise
```
BrianRamsay: Make sure you are re-raising the actual exception that was thrown, and not a new generic Exception. Though really, if you are doing nothing in the except block except re-raising the exception, there's no reason for the try/except in the first place.

tests/code/test_AdminNewEvent.py
```diff
@@ -0,0 +1,64 @@
...
+    eventInfo =  {'eventRequiredForProgram':True,'eventRSVP':False, 'eventServiceHours':False,
+                  'eventIsTraining':True, 'eventIsRecurring':False, 'eventStartDate': '2021-12-12',
+                   'eventEndDate':'2022-06-12', 'programId':1, 'eventLocation':"a big room",
+                   'eventEndTime':'21:00', 'eventStartTime':'18:00', 'eventDescription':"Empty Bowls Spring 2021",
+                   'eventName':'Empty Bowls Spring','eventTerm':1,'eventFacilitator':"ramsayb2"}
+    alertMessage = createNewEvent(eventInfo)
+    assert alertMessage == "Event successfully created!"
+
+
+    # FIXME: the test below don't work as expected ... is there a fix or should it be deleated?
```
BrianRamsay: what doesn't work as you expect? It doesn't raise the exceptions? Or it raises different exceptions from what you think? You definitely need some tests for function failure

tests/code/test_AdminNewEvent.py
```diff
@@ -0,0 +1,64 @@
...
+@pytest.mark.integration
+def test_createNewEvent():
+
+    eventInfo =  {'eventRequiredForProgram':True,'eventRSVP':False, 'eventServiceHours':False,
+                  'eventIsTraining':True, 'eventIsRecurring':False, 'eventStartDate': '2021-12-12',
+                   'eventEndDate':'2022-06-12', 'programId':1, 'eventLocation':"a big room",
+                   'eventEndTime':'21:00', 'eventStartTime':'18:00', 'eventDescription':"Empty Bowls Spring 2021",
+                   'eventName':'Empty Bowls Spring','eventTerm':1,'eventFacilitator':"ramsayb2"}
+    alertMessage = createNewEvent(eventInfo)
+    assert alertMessage == "Event successfully created!"
```
BrianRamsay: If you return the event you create, you have more options for testing that it was created successfully. Even without that, you should verify that the event and facilitator are created as expected.

tests/code/test_calculatingRecurringEventFrequency.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+from app.logic.adminNewEvent import calculateRecurringEventFrequency
+
+@pytest.mark.integration
+def test_calculateRecurringEventFrequency():
```
BrianRamsay: This tests that the function runs, but you don't actually test if it works. And empty function would pass these tests.

app/controllers/admin/createEvents.py
```diff
@@ -0,0 +1,43 @@
...
+from app.logic.validateNewEvent import validateNewEventData
+from flask import flash, redirect, url_for, g
+
+
+@admin_bp.route('/createEvent', methods=['POST'])
+def createEvent():
+
+    if not g.current_user.isCeltsAdmin:
+
+        flash("Only celts admins can create an event!")
```
BrianRamsay: You should set the flash type so it's not always the 'success' color

BrianRamsay (changes requested):

app/logic/adminNewEvent.py
Sandesh-L: It is being used in a separate branch/PR (#41). we created it before splitting the issues up. The function has been deleted from this branch to avoid confusion.

tests/code/test_calculatingRecurringEventFrequency.py
Sandesh-L: This test was for a function that is not being used in this PR. The file has been removed to avoid confusion. The other PR for recurring events has the proper tests.

tests/code/test_AdminNewEvent.py
Sandesh-L: Sorry I wasn't very clear on the comment. The issue is that we are not getting any errors at all.  When we put on "hi, how are you?" into a date field, peewee doesn't raise any errors. And it's not just the date field. I think it's all fields that are having this issue.

app/logic/adminNewEvent.py
Sandesh-L: We removed the try/except

BrianRamsay (approved):

PR #44, opened 2021-07-14
Refactor to use new ProgramEvent model and table

BrianRamsay:
Events can be associated with multiple programs now. 

Refactored existing code to use the new table, and added a couple of convenience fields to the Event model.


app/logic/programEligibility.py
```diff
@@ -3,29 +3,34 @@
...
     """
     Verifies if a given user is eligible for a program by checking if they are
     banned from a program and if they have attended all the required events for a program.
+
     :param program: accepts a Program object or a valid programid
     :param user: accepts a User object or userid
     :return: True if the user is not banned and meets the requirements, and False otherwise
     """
     user = User.get_by_id(user)
+    user = User.get_by_id(user)
```
sheggen: Duplicate of line 18?

sheggen (dismissed):

app/logic/programEligibility.py
BrianRamsay: Removed.

Issue #45, opened 2021-07-14
Create warning label for Track Hourse Page.


tylerpar99:
The warning label needs creation to make Admins aware of Volunteers who have not completed training for an event.

PR #47, opened 2021-07-15
PR for Warning label track hours issue 45(This issue is based off of branch: track_volunteer_hours_issue42)

tylerpar99:
This PR fixes issue #45 

Summary: 
 The issue asked that there be a way for an admin to tell that a participant that has checked into or rsvp to an event be flagged if they have not completed all prerequisites for the event. In order to achieve this, we queried the database and pushed on to a list of all the eventIDs that were set as prerequisites to the program the participant is attending. We then queried the database using the eventIDS and pushed onto a list all the users who attended any of the prerequisite events, using this list we then checked how many times a user name appeared and created a conditional that required the user to appear as many times as the length of the eventIDs list. If the user met the amount they would stay on the list, if the user did not they would be removed from the list. After deleting duplicates we then passed the list to the HTML and used jinja to flag any users who appeared on the track hours page who were not on the list of eligible participants. 

To Test:  
There is a test function in the test suite labeled `test_warning.py`. Running this should pass.
You can also visit the URL: `http://<YOURIP>/<PROGRAMID>/<EVENTID>/track_hours`. This URL will display a table with every user set to attend or currently at the event, if any of them have not met the requirements for the event you will see a red warning icon. Hovering over this will display a ToolTip that will tell you that the user has not completed the prerequisites.

Notes: This branch is forked off over the branch listed in the title. If any changes are made to the listed branch this will have to be updated.


BrianRamsay (changes requested):
This needs to have development merged in - there are still references to Event.program

app/controllers/admin/routes.py
```diff
@@ -28,23 +34,27 @@ def studentSearchPage():
 
 @admin_bp.route('/<programID>/<eventID>/track_volunteers', methods=['GET'])
 def trackVolunteersPage(programID, eventID):
+
+    trainingEvents = ProgramEvent.select().where(ProgramEvent.program == programID)
+    trlist = [training.event for training in trainingEvents if training.event.isTraining]
```
BrianRamsay: this should all be done in the query

app/logic/trackAttendees.py
```diff
@@ -0,0 +1,13 @@
...
+from app.models.event import Event
+from peewee import fn
+
+def trainedParticipants(programID, trlist):
+    """
+    This function tracks the users who have attended every Prerequisite
+    event and adds them to a list that will not flag them when tracking hours.
+    """
+    eventTrainingDataList = [participant.user.username for participant in (EventParticipant.select().where(EventParticipant.event.in_(trlist)))]
+    attendedTraining = list(dict.fromkeys(filter(lambda user: eventTrainingDataList.count(user) == len(trlist), eventTrainingDataList)))
```
BrianRamsay: This is fancy, but really seems like it could be simplified into one query with the calling code. `Select all users who do not have a missing event participant record for any training event in the program`

app/static/js/trackVolunteers.js
```diff
@@ -1,36 +1,36 @@
 $(document).ready(function(){
+  $('[data-bs-toggle="tooltip"]').tooltip();
+  $("#trackHoursInput").on("keyup", function() {
 
 // Search functionalities from the volunteer table in the UI
   $("#trackVolunteersInput").on("keyup", function() {
+>>>>>>> track_volunteer_hours_issue42:app/static/js/trackVolunteers.js
```
BrianRamsay: merge conflict artifact

tests/code/test_warning.py
```diff
@@ -0,0 +1,17 @@
...
+from peewee import DoesNotExist
+from app.logic.trackVolunteerHours import prereqParticipants
+from app.controllers.admin.routes import trackVolunteerHoursPage
+from app.models.eventParticipant import EventParticipant
+
+@pytest.mark.integration
+def test_prereqParticipants():
+
+    prlist = [1, 2]
+    prereqParticipant = prereqParticipants(1, prlist)
```
BrianRamsay: this function name changed. This would have thrown an error if you ran the test suite

tests/code/test_warning.py
```diff
@@ -0,0 +1,17 @@
...
+    eventPreqDataList = []
+
+    eventPreqDataList = [participant.user.username for participant in (EventParticipant.select().where(EventParticipant.event.in_(prlist)))]
+    attendedPreq = list(dict.fromkeys(filter(lambda user: eventPreqDataList.count(user) == len(prlist), eventPreqDataList)))
```
BrianRamsay: you are re-implementing the function here, rather than testing it

BrianRamsay (changes requested):

tests/code/test_warning.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+from peewee import DoesNotExist
+from app.logic.trackAttendees import trainedParticipants
+
+@pytest.mark.integration
+def test_prereqParticipants():
+
+    prlist = [1, 2]
+    attendedPreq = prereqParticipant = trainedParticipants(1, prlist)
```
BrianRamsay: this double assignation doesn't seem intentional

app/logic/trackAttendees.py
BrianRamsay: we aren't going to worry about this now

tests/code/test_warning.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+from peewee import DoesNotExist
+from app.logic.trackAttendees import trainedParticipants
+
+@pytest.mark.integration
+def test_prereqParticipants():
```
BrianRamsay: Test an empty case as well

BrianRamsay (changes requested):

BrianRamsay:
There are conflicts to resolve

tests/code/test_warning.py
BrianRamsay: Not sure why it is saying Outdated, this is still relevant

tests/code/test_warning.py
BrianRamsay: Not sure why it is saying Outdated, this is still relevant

BrianRamsay (changes requested):
There are still a couple of outstanding comments in the test

BrianRamsay (approved):
nice!

PR #48, opened 2021-07-15
Added `programCategory` Table 

sreynit02:
Issue: Programs fall into four different categories: Student Led Services,  Training and Education, Bonner Scholars, One-time Events.

Did: Created a new table called `ProgramCategory` with a field called categoryName that is a char field and primary key. Added `programCategory` field in `Program` table as the `foreignKeyField` to `ProgramCategory` table. Added programCategories in `test_data.py` and added more programs and events so that each `programCategory` has an event and program.

Test: 
-reset database 
-run tests and all tests still pass

### DIFF
--- a/app/models/program.py
+++ b/app/models/program.py
@@ -2,7 +2,9 @@ from app.models import*
 from app.models.term import Term
 from app.models.courseStatus import CourseStatus
 from app.models.partner import Partner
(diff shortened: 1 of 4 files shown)


BrianRamsay:
Not relevant anymore

PR #52, opened 2021-07-16
Add recurring event issue46

Sandesh-L:
Summary:

Added functionality to the recurring event checkbox. Where it displays an events table and the end datepicker. After the user selects an end date and start date the table will be populated with all of the events that would be created for that one recurring event.
Created a `calculateRecurringEventFrequency` function that checks the start date and end date and creates a list of dictionaries that contain the dates and event names for the new events that will be created.
Added an if conditional to the `createNewEvent` function that checks if the newly created event is supposed to be recurring or not and then either creates one event or multiple events for however many weeks the events recurs.
Wrote some test cases for  the `calculateRecurringEventFrequency` function and a test case for `createNewEvent`.

To Test:
Run the test on your terminal.
Go to the `<your_ip>:8080/<program_id>/create_event` page and then create an event as you normally would.
If you check the "this event occurs weekly" box then it should show the end date and the table.
Filling out the end date and start date should populate the table will the new events that will be created. 
Hitting save will create those events in the database.
You can double-check the events are created using PHPMyAdmin.

**NOTE:** This branch is forked from PR #37. This should probably not be merged until that one is.


app/controllers/admin/createEvents.py
```diff
@@ -0,0 +1,52 @@
...
+from flask import flash, redirect, url_for, g
+
+@admin_bp.route('/makeRecurringEvents', methods=['POST'])
+@admin_bp.route('/createEvent', methods=['POST'])
+def createEvent():
+
+    if 'makeRecurringEvents' in request.url_rule.rule:
+        recurringEventInfo = request.form.copy()
+        recurringEvents = calculateRecurringEventFrequency(recurringEventInfo)
+        return recurringEvents
```
BrianRamsay: There doesn't seem to be a reason to have this be in the same method.

BrianRamsay:
Development needs to be merged into the branch, and conflicts resolved

app/logic/adminNewEvent.py
```diff
@@ -24,32 +52,58 @@ def createNewEvent(newEventData):
...
+                                          isService = newEventData['eventServiceHours'],
+                                          startDate =  newEventData['eventStartDate'],
+                                          endDate =  newEventData['eventEndDate'],
+                                          program = newEventData['programId'])
+
+                programEvent = ProgramEvent.create(program=program, event=newEvent)
+
+                facilitatorEntry = Facilitator.create(user = newEventData['eventFacilitator'],
+                                          event = newEvent)
+
```
BrianRamsay, line 86: Since the `else` case is the same as the main case, I would combine them. You just need to only loop once and don't add the "Week X" string, and they are the same. Also, program shouldn't be in the event.create() list - it probably was added back in during conflict resolution.

app/static/js/createEvents.js
```diff
@@ -33,6 +32,34 @@ $(document).ready(function(){
...
+          url: "/makeRecurringEvents",
+          data: eventDatesAndName, //get the startDate, endDate and eventName as a dictionary
+          success: function(jsonData){
+            var recurringEvents = JSON.parse(jsonData)
+            var recurringTable = $("#recurringEventsTable")
+            $("#recurringEventsTable tbody tr").remove();
+
+            for (var event of recurringEvents){
+              recurringTable.append("<tr><td>"+event.eventName+"</td><td><input name='week"+event.week+"' type='hidden' value='"+event.Date+"'>"+event.Date+"</td></tr>");
+              console.log(event)
```
BrianRamsay: Let's get rid of this debug output

app/logic/adminNewEvent.py
```diff
@@ -4,6 +4,34 @@
...
+    counter = 0
+    for i in range(0, ((endDate-startDate).days +1), 7):
+        counter += 1
+        recurringEvents.append({'eventName': f"{eventName} Week {counter}",
+                                'Date':startDate.strftime('%m-%d-%Y'),
+                                "week":counter})
+        startDate += timedelta(days=7)
+
+    return json.dumps(recurringEvents)
```
BrianRamsay: I think this function should return recurringEvents straight up, and let the controller do the json conversion

app/controllers/admin/createEvents.py
```diff
@@ -19,15 +24,19 @@ def createEvent():
...
 
-
         # convert date into datetime object (Y-m-d) for the backend
-        newEventData['eventStartDate'] = parser.parse(newEventData['eventStartDate'], dayfirst=True)
-        newEventData['eventEndDate'] = parser.parse(newEventData['eventEndDate'], dayfirst=True)
-
+        newEventData['eventStartDate'] = parser.parse(newEventData['eventStartDate'])
+        newEventData['eventEndDate'] = parser.parse(newEventData['eventEndDate'])
+
+        for entry in range(1,len(newEventData)):
```
BrianRamsay: I don't quite understand this for loop here or in `createNewEvent`. This is looping for each key/value pair in the dictionary, right? Which is not related to the number of new events they want to create.

What you should be doing (it seems to me) is looping through the array from `calculateRecurringEventFrequency`, because that has exactly the number of entries that you are needing to create.

Actually, you could get away with not including `weekX` keys in the event data at all, because you can just regenerate them with `calculateRecurringEventFrequency`

BrianRamsay (changes requested):

BrianRamsay (approved):
Looks good! Thanks.

Issue #51, opened 2021-07-15
CELTS should have ability to use Kiosk Mode.


tylerpar99:
When signing in volunteers for an event the CELTS team should have access to a kiosk mode.

PR #54, opened 2021-07-21
Celts kiosk mode issue#51

tylerpar99:
This PR fixes issue #51 

Summary:
Created a Kiosk page for the CELTS team that will allow them to quickly scan in participants who have shown up at an event. In order to accomplish this, we built a plain page with a header that will display the Current event you signing in for and a Text box that will auto focus on page load and when you toggle full screen. The participant will then scan their badge and the Kiosk page will send their scan code to the back end where we parse the scan and pull out the B number and toggle the user as attended in the database and then upon success flash a message displaying the user first and last name and a successful sign in. We also added a flasher that will display if a user is already signed in to the event. In case a participant forgets their id badge upon arriving at the event the user can manually enter in their B number and press enter and the same steps will be completed. 

To Test:
We have added functions to the test suite that will pass if ran. You can also load up the Kiosk page at this URL: `http://<YOURIP>/1/kiosk` and use a scanner to scan your Id badge or manually type in your Id number. If you are in the system as an eligible participant signed up for this event you will be signed in. If you attempt twice you will be told you are already signed in. If any error occurs you will be flashed a message to see the attendant.

Notes: By default, the browser has priority over the `esc` button. The way we have designed the Kiosk page when clicking the provided toggle the page will enter and exit full screen and hide any elements not necessary for a Participant to use. Clicking `esc` will exit fullscreen but will not bring back the navbar and the footer. Clicking a second time will complete this action.


app/controllers/events/routes.py
```diff
@@ -19,3 +20,33 @@ def showUpcomingEvent():
...
+
+@events_bp.route('/<eventid>/kiosk', methods=['GET'])
+def loadKiosk(eventid):
+    event = Event.get_by_id(eventid)
+    return render_template("/events/eventKiosk.html",
+                            event = event,
+                            eventid = eventid)
+
+@events_bp.route('/signintoKiosk', methods=['POST'])
+def signinKiosk():
```
BrianRamsay: let's rename these functions (esp. the one in app/logic) to remove the kiosk references. Kiosk mode is how things are displayed, but what's happening here is event sign-in.

app/logic/signinKiosk.py
```diff
@@ -0,0 +1,21 @@
+
+from flask import Flask
+from app.models.eventParticipant import EventParticipant
+from app.models.user import User
+
+def sendkioskData(bnumber, eventid):
+    """Accepts scan input and signs in the user. If user exists or is already
+    signed in will return flasher"""
+    bNumberToUsername = User.select().where(User.bnumber == bnumber).objects()[0]
```
BrianRamsay: Use `User.get()` for this - it's designed to get one record from a table.

BrianRamsay: Also, the variable is not holding a `bNumberToUsername`. it's a User.

app/logic/signinKiosk.py
```diff
@@ -0,0 +1,21 @@
+
+from flask import Flask
+from app.models.eventParticipant import EventParticipant
+from app.models.user import User
+
+def sendkioskData(bnumber, eventid):
+    """Accepts scan input and signs in the user. If user exists or is already
+    signed in will return flasher"""
+    bNumberToUsername = User.select().where(User.bnumber == bnumber).objects()[0]
+    attendedParticipants = (EventParticipant.select(EventParticipant.user)
```
BrianRamsay: Rather than do the query, create a list, and then check if the user is in the list, just check for the user directly in the query. If nothing comes back, they aren't in the list.

app/logic/signinKiosk.py
```diff
@@ -0,0 +1,21 @@
...
+    """Accepts scan input and signs in the user. If user exists or is already
+    signed in will return flasher"""
+    bNumberToUsername = User.select().where(User.bnumber == bnumber).objects()[0]
+    attendedParticipants = (EventParticipant.select(EventParticipant.user)
+                                            .where(EventParticipant.attended, EventParticipant.event == eventid))
+
+    attendedList = [participant.user for participant in attendedParticipants ]
+
+    if bNumberToUsername not in attendedList:
+        (EventParticipant.update({EventParticipant.attended: True})
```
BrianRamsay: Does this work if they aren't in the Event Participant table yet? I don't see that case being handled.

app/static/js/eventKiosk.js
```diff
@@ -0,0 +1,104 @@
+var hitEnter = false;
+
+$(document).keydown(function(e){
+  if (e.key === "Escape") {
+    console.log("get off")
```
BrianRamsay: the console logging needs to be removed.

app/static/js/eventKiosk.js
```diff
@@ -0,0 +1,104 @@
+var hitEnter = false;
+
+$(document).keydown(function(e){
+  if (e.key === "Escape") {
+    console.log("get off")
+    $("#fullscreenCheck").prop("checked", false)
+    toggleFullscreen();
+  }
+  else if(e.key === "Enter") {
+      hitEnter = true;
```
BrianRamsay: Do you need a global variable vs passing the value around?

app/static/js/eventKiosk.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+  else if(e.key === "Enter") {
+      hitEnter = true;
+      submitData();
+      hitEnter = false;
+
+  }
+});
+
+function toggleFullscreen() {
```
BrianRamsay, line 11: Did you pull this function from somewhere online? If so, you should credit it in a comment

app/static/js/eventKiosk.js
```diff
@@ -0,0 +1,104 @@
...
+        $("#submitScannerData").val("").blur();
+        $('#submitScannerData').focus();
+      }
+    })
+  }
+}
+
+function hideElements(hide=true) {
+
+  if (hide == true) {
```
BrianRamsay, line 81: you can just use `hide`, no need for the equality check

tests/code/test_signinKiosk.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+from app.logic.signinKiosk import sendkioskData
+from app.models.eventParticipant import EventParticipant
+
+@pytest.mark.integration
+def test_sendKioskDataKiosk():
```
BrianRamsay, line 6: This doesn't cover all of the cases that the function handles, and doesn't clean up the data after it.

BrianRamsay (changes requested):

app/logic/signinKiosk.py
```diff
@@ -0,0 +1,30 @@
...
+from flask import Flask
+from app.models.eventParticipant import EventParticipant
+from app.models.user import User
+from app.models.programBan import ProgramBan
+from app.logic.programEligibility import isEligibleForProgram
+
+def sendUserData(bnumber, eventid, programid):
+    """Accepts scan input and signs in the user. If user exists or is already
+    signed in will return user and login status"""
+    bNumberToUser = User.get(User.bnumber == bnumber)
```
BrianRamsay: `bNumberToUser` is what's happening in this line, but it shouldn't be the name of the variable. The variable is just a User object. A better variable name would be just `user`, or maybe `signInUser`

app/logic/signinKiosk.py
```diff
@@ -0,0 +1,30 @@
...
+    if not isEligibleForProgram(programid, bNumberToUser):
+        userStatus = "banned"
+    elif ((EventParticipant.select(EventParticipant.user)
+                           .join(User)
+                           .where(EventParticipant.attended, User.bnumber == bnumber, EventParticipant.event == eventid))
```
BrianRamsay: Once you've gotten a user object out of the bnumber, you should just use that object. I wouldn't use bnumber again when you can just use your user object. It also lets you avoid joining to User (that is, your where clause could just have `EventParticipant.user == bNumberToUser` like you do below)

app/logic/signinKiosk.py
```diff
@@ -0,0 +1,30 @@
...
+        userStatus = "success"
+        if EventParticipant.get_or_none(EventParticipant.user == bNumberToUser, EventParticipant.event == eventid):
+            (EventParticipant.update({EventParticipant.attended: True})
+                             .where(EventParticipant.user == bNumberToUser, EventParticipant.event == eventid)).execute()
+        else:
+            EventParticipant.insert([{EventParticipant.user: bNumberToUser,
+                                      EventParticipant.event: eventid,
+                                      EventParticipant.rsvp: False,
+                                      EventParticipant.attended: True,
+                                      EventParticipant.hoursEarned: 0}]).execute()
```
BrianRamsay: Do we want to set the proper hoursEarned here, since they are getting marked as attending?

app/static/js/eventKiosk.js
```diff
@@ -0,0 +1,97 @@
...
+  if (e.key === "Escape") {
+    $("#fullscreenCheck").prop("checked", false)
+    toggleFullscreen();
+  }
+  else if(e.key === "Enter") {
+      submitData(true);
+  }
+});
+
+function toggleFullscreen() {                                     // Source: https://stackoverflow.com/questions/1125084/how-to-make-the-window-full-screen-with-javascript-stretching-all-over-the-scre
```
BrianRamsay: while we're editing things, stick this comment on the line before, rather than as a trailing comment on the line

tests/code/test_signinKiosk.py
BrianRamsay: You're cleaning up now, which is good, but you're still only exercising one of the two paths. You either update or insert in this function and the test only checks one.

BrianRamsay (changes requested):

BrianRamsay (approved):
Looks good!

Issue #43, opened 2021-07-12
1.2.5


tylerpar99:
The CELTS Admin must be able to delete event

PR #55, opened 2021-07-21
Delete event issue 43

tylerpar99:
This PR fixes issue #43 

Summary: 
We created the ability for an admin to delete an Event after creation. We merged the Edit events page from a separate branch and simply added a Cancel event button that will hide if you are creating an event and show if you are deleting one.  We then made a `deleteEvent` function that will find the event by the given Id and recursively delete every instance of the event ID we set delete nullable to true so the records will be thoroughly deleted.

To Test: 
You can run the test suite and see our function will pass. You can also visit this URL: `http://<YOURIP>/1/2/edit_event` and you will see a red `Cancel Event` button upon clicking you will get a message notifying you that the event has been successfully canceled. You can then check the database in the events table and see it is no longer there. You can also visit the URL: `http://<YOURIP>/1/create_event` and see that the Cancel Event button does not show.

Notes: This page depends on the Edit Events page. Any changes to that page will require this PR to be updated.


app/controllers/admin/deleteEvent.py
```diff
@@ -0,0 +1,21 @@
...
+from flask import Flask, redirect, flash, g, url_for
+from app.models.event import Event
+from app.controllers.admin import admin_bp
+from app.models.eventParticipant import EventParticipant
+from app.models.outsideParticipant import OutsideParticipant
+from app.models.facilitator import Facilitator
+from app.models.programEvent import ProgramEvent
+
+def deleteEvent(program, eventId):
+    deleteallInstances = [EventParticipant, Facilitator, OutsideParticipant, ProgramEvent]
```
BrianRamsay: is this still used? I don't see any other references to it

app/controllers/admin/deleteEvent.py
```diff
@@ -0,0 +1,21 @@
...
+    try:
+
+        if Event.get_or_none(Event.id == eventId):
```
BrianRamsay: Isn't the get syntax different from the where clause syntax? I'd think you would want `id=eventID`

app/controllers/admin/deleteEvent.py
```diff
@@ -0,0 +1,21 @@
...
+            deleteEvent = Event.get_by_id(eventId)
```
BrianRamsay: You are selecting the event twice. Store the result from get_or_none above and avoid the duplicated database query

app/logic/adminCreateEvent.py
```diff
@@ -0,0 +1,23 @@
...
+    termDescriptions = Term.select(Term.description)
+    descriptions = [des.description for des in termDescriptions.objects()]
+
+    return descriptions
+
+def getCurrentTerm():
+
+    currentTerm = Term.select(Term.description).where(Term.isCurrentTerm)
+    term = [t.description for t in currentTerm.objects()] #This could be done better. but how??
+    return term[0]
```
BrianRamsay: `currentTerm = Term.get(isCurrentTerm=True)`
`return currentTerm.description`

I think you should return the whole term object though, rather than just the description. In which case the function is a one-liner. ` return Term.get(isCurrentTerm = True)`

app/logic/adminCreateEvent.py
```diff
@@ -0,0 +1,23 @@
...
+
+def getFacilitators():
+
+    facilitators = User.select(User).where((User.isFaculty == 1) | (User.isCeltsAdmin == 1) | (User.isCeltsStudentStaff == 1))
+    listOfFacilitators = [fac for fac in facilitators.objects()]
```
BrianRamsay: you don't need .objects() when you are just returning the fields from one table. Likewise, you don't need to specify `User.select(User)`, `User.select()` will suffice.

app/logic/adminNewEvent.py
```diff
@@ -50,7 +50,31 @@ def createNewEvent(newEventData):
 
             facilitatorEntry = Facilitator.create(user = newEventData['eventFacilitator'],
                                                       event = newEvent)
+        return newEvent
     else:
         raise Exception("Invalid Data")
 
-    return newEvent
+def eventEdit(newEventData):
+
```
BrianRamsay, line 91: Use the new `'valid'` key that Sandesh added in his last update to confirm the data is ready to go

BrianRamsay (changes requested):

app/controllers/admin/deleteEvent.py
BrianRamsay: Looks like you can do both. nm

app/logic/validateNewEvent.py
```diff
@@ -43,7 +43,11 @@ def validateNewEventData(newEventData):
     event = Event.select().where((Event.eventName == newEventData['eventName']) &
                              (Event.description == newEventData['eventDescription']) &
                              (Event.startDate == newEventData['eventStartDate']))
+
     if event.exists():
+        if ignoreExist == True:
```
BrianRamsay: This doesn't actually ignore anything. I'd probably flip this variable to be positive rather than negative. e.g., `checkExists` and default it to true. Then, your if statement can be `if checkExists and event.exists():`, which is pretty clear.

tests/code/test_deleteEvent.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+from app.controllers.admin.deleteEvent import deleteEvent
+from app.models.event import Event
+
+@pytest.mark.integration
+def test_deleteEvent():
```
BrianRamsay, line 6: you should test a case where it doesn't already exist as well

tests/code/test_editEvent.py
```diff
@@ -0,0 +1,33 @@
...
+@pytest.mark.integration
+def test_beforeEdit():
+    eventId = 4
+    beforeEdit = Event.get_by_id(eventId)
+
+    assert beforeEdit.eventName == "First Meetup"
+
+@pytest.mark.integration
+def test_afterEdit():
+    eventFunction: editEvent(newEventData)
```
BrianRamsay: you're not actually using your editEvent function are you? I'm not sure what this line is doing, but you are editing the event yourself in the test, not using the function.

BrianRamsay: not sure how this test passes, actually. seems like a syntax error

BrianRamsay (changes requested):

BrianRamsay (dismissed):
Looks good!

BrianRamsay (approved):
Looks good!

Issue #42, opened 2021-07-09
1.2.3 Track attendance and service hours of volunteers


gastonjarju:
1.2.3 The CELTS administrator must be able to track attendance and service hours of volunteers at each event

PR #57, opened 2021-07-22
Track volunteer hours issue42

gastonjarju:
Summary:
This PR resolves issue #42
- Created the UI features for the track hours page with the following functionalities:
    The URL : http://<YourIPAddress>:8080/<programID>/<eventID>/track_hours#
    A table to display the volunteers.
    Checkboxes to mark when a volunteer is present for the event
    Checking the box automatically gives the volunteer certain numbers of hours based on the duration of the event.
    The admin can also change the amount of hours the volunteer received in the Total hours column input fields.
    A search bar to enable the admin to search volunteers info from the table.
    Also created an add button that pops up a modal to search and add volunteers.
    In the modal, we added a search bar that the admin could use to search for users from the user table database and 
    add them to  the event.
   The modal also has a close and add paticipant button for the admin to close the modal or add a volunteer after a search 
   respectively.
   Added an Event Form and Track Hours nav tabs to interchange the pages when click.
   Added a save button at the bottom of the volunteer table to enable the admin to save the changes made each time.
   Also added an Event name  Header and a Program name header that changes each time the eventID or programID changes.
Note: The Event form nav tab has a random link to google.com because the Event form page does not exist in development yet.

- Created test cases in the test file to test our functions.


app/controllers/admin/routes.py
```diff
@@ -1,8 +1,62 @@
...
+
+            eventLengthInHours = getEventLengthInHours(startTime, endTime, eventDate)
+
+            program = Program.get_by_id(programID)
+
+            return render_template("/events/trackVolunteerHours.html",
+                                    eventParticipantsData = list(eventParticipantsData),
+                                    eventLength = eventLengthInHours,
+                                    program = program)
+        except DoesNotExist:
```
BrianRamsay: Is a raw exception message really what you want the user to see?

app/controllers/admin/routes.py
```diff
@@ -1,8 +1,62 @@
...
+            raise DoesNotExist
+
+        except:
+            abort(404)
+
+    else:
+        raise Exception("User must be admin to view this page.")
```
BrianRamsay: Handling this with an exception is probably not what we want. We should either give them a 403 forbidden, or inform them and redirect them somewhere else.

app/controllers/admin/searchTrackHoursVolunteer.py
```diff
@@ -0,0 +1,40 @@
...
+from flask import json, jsonify
+from app.controllers.admin import admin_bp
+from app.models.event import Event
+from app.models.user import User
+from app.models.eventParticipant import EventParticipant
+from app.models.user import User
+from peewee import *
+
+@admin_bp.route('/searchTrackHoursVolunteers/<query>', methods = ['GET'])
+def searchTrackHoursVolunteers(query):
```
BrianRamsay: This search functionality needs to be in a separate function so you can test it without having to go through the ui

app/controllers/admin/searchTrackHoursVolunteer.py
```diff
@@ -0,0 +1,40 @@
...
+                    searchTerm += "%"
+                    results = User.select().where(User.isStudent & User.firstName ** firstName & User.lastName ** lastName)
+                    for participant in results:
+                        if participant not in resultsDict:
+                            resultsDict[f"{participant.firstName} {participant.lastName} ({participant.username})"] = f"{participant.firstName} {participant.lastName} ({participant.username})"
+
+        dictToJSON = json.dumps(resultsDict)
+        return dictToJSON
+
+    except Exception as e:
```
BrianRamsay: You should output the actual exception message so you can debug problems

app/logic/updateTrackHours.py
```diff
@@ -0,0 +1,75 @@
...
+def getEventLengthInHours(startTime, endTime, eventDate):
+    """
+    Converts the event length hours into decimal
+    parameters: startTime- start time event (type: time)
+                endTime- end time event (type: time)
+                eventDate- date of the event (type: datetime)
+    """
+    #can only subtract datetime objects, not time objects. So convert time into datetime
+    eventLength = datetime.combine(eventDate, endTime) - datetime.combine(eventDate, startTime)
+    (h, m, s) = str(eventLength).split(':')
```
BrianRamsay: can't you pull these values out of a datetime directly, rather than cast to a string and split?

app/logic/updateTrackHours.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    updates the events participant table with data from the track hours webpage
+
+    param: participantData- a dictionary that contains data from every row of the page along with the associated username.
+    """
+
+    for user in range(1, len(participantData)):
+        if f'username{user}' in participantData:
+            username = participantData[f'username{user}']
+            if (User.select().where(User.username == username)).exists():
```
BrianRamsay: I would use `User.get_or_none` for this - much shorter

app/logic/updateTrackHours.py
```diff
@@ -0,0 +1,75 @@
...
+
+                try:
+                    if participantData['checkbox_'+ username] == "on": #if the user is marked as present
```
BrianRamsay: What if there is no EventParticipant entry for the user here? Do you want to say they don't exist, or should we add them?

app/logic/updateTrackHours.py
```diff
@@ -0,0 +1,75 @@
...
+                        (EventParticipant.update({EventParticipant.hoursEarned: float(participantData['inputHours_'+ username])}).where(EventParticipant.event == participantData['event'],
```
BrianRamsay: update both values in one update statement.

app/logic/updateTrackHours.py
```diff
@@ -0,0 +1,75 @@
...
+                        EventParticipant.user == participantData[f'username{user}'])).execute()
+
+                    (EventParticipant.update({EventParticipant.hoursEarned: 0}).where(EventParticipant.event == participantData['event'],
+                                             EventParticipant.user == participantData[f'username{user}'])).execute()
+            else:
+                return "Volunteer does not exist."
+        else:
+            break
+
+    return "Volunteer table successfully updated!"
```
BrianRamsay: I'm not a big fan of functions returning text messages about success or failure. Those messages should belong to the controller, not the application functions

app/logic/updateTrackHours.py
```diff
@@ -0,0 +1,75 @@
...
+
+        alreadyVolunteered = (EventParticipant.select().where(EventParticipant.user==userName, EventParticipant.event==volunteerEventID)).exists()
+        if alreadyVolunteered:
+            return("Volunteer already exists.")
+        else:
+            EventParticipant.create(user=userName, event = volunteerEventID, attended = True, hoursEarned = eventLengthInHours)
+            return "Volunteer successfully added!"
+
+    except Exception as e:
+        return "Error when adding volunteer", 500
```
BrianRamsay: logic functions should not concern themselves with the messaging to the user and the http status codes

tests/code/test_updateTrackHours.py
```diff
@@ -0,0 +1,88 @@
+import pytest
+from app.logic.updateTrackHours import getEventLengthInHours, addVolunteerToEvent, updateTrackHours
+from app.models.eventParticipant import EventParticipant
+from datetime import datetime
+
+
+@pytest.mark.integration
+def test_getEventLengthInHours():
```
BrianRamsay: I like this test!

tests/code/test_updateTrackHours.py
```diff
@@ -0,0 +1,88 @@
...
+    eventLengthInHours = 67
+    #test that volunteer is already registered for the event
+    volunteerToEvent = addVolunteerToEvent(user, volunteerEventID, eventLengthInHours)
+    assert volunteerToEvent == "Volunteer already exists."
+
+    #test for adding user as a participant to the event
+    user = "agliullovak"
+    volunteerToEvent = addVolunteerToEvent(user, volunteerEventID, eventLengthInHours)
+    assert volunteerToEvent == "Volunteer successfully added!"
+    (EventParticipant.delete().where(EventParticipant.user == user, EventParticipant.event == volunteerEventID)).execute()
```
BrianRamsay: good cleanup

tests/code/test_updateTrackHours.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # tests if user does not exist in the database
+    participantData = {'inputHours_jarjug':100, 'checkbox_jarjug':"on", 'event':3, 'username1': 'jarjug'}
+    volunteerTableUpdate = updateTrackHours(participantData)
+    assert volunteerTableUpdate ==  "Volunteer does not exist."
+
+    # tests for the case when the checkbox is not checked (user is not present)
+    participantData = {'inputHours_ramsayb2':100, 'event':3, 'username1': 'ramsayb2'}
+    volunteerTableUpdate = updateTrackHours(participantData)
+    assert volunteerTableUpdate == "Volunteer table successfully updated!"
```
BrianRamsay: Make sure you undo changes that the test makes

BrianRamsay (changes requested):

app/controllers/admin/routes.py
gastonjarju: No. We have changed it to a customized exception message (Event ID and Program ID mismatched)

app/logic/updateTrackHours.py
gastonjarju: We will want to say "Volunteer does not exist."

app/controllers/admin/changeTrackVolunteer.py
```diff
@@ -0,0 +1,50 @@
...
+def searchTrackVolunteers(query):
+    '''Accepts user input and queries the database returning results that matches user search'''
+    query = query.strip()
+    search = query.upper()
+    splitSearch = search.split()
+    try:
+        return searchVolunteers(query)
+
+    except Exception as e:
+        return e, 500
```
BrianRamsay: If you're just returning any exception that is thrown there's no reason for a try/catch

app/controllers/admin/changeTrackVolunteer.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@admin_bp.route('/<programID>/<eventID>/track_volunteers', methods=['POST'])
+def updateVolunteerTable(programID, eventID):
+    updateTrackVolunteersMsg = updateTrackVolunteers(request.form)
+    if updateTrackVolunteersMsg == None:
+        flash("Volunteer table succesfully updated")
+    else:
+        flash("Error adding volunteer")
```
BrianRamsay: This function has no return value. It will always be None. If you want to flash that error adding volunteer, you need to catch the exception that is raised.

BrianRamsay: Also, `updateTrackVolunteers` is confusing to me as a function name, because there are two verbs in the name, "update", and "track"

app/logic/updateTrackVolunteers.py
```diff
@@ -0,0 +1,63 @@
...
+           volunteerEventID - id of the event the volunteer is being registered for
+           eventLengthInHours - how long the event lasts (how may hours to give the student) (type: float)
+    '''
+    try:
+        user = user.strip("()")
+        userName=user.split('(')[-1]
+
+        if not EventParticipant.get_or_none(user=userName, event = volunteerEventID):
+            EventParticipant.create(user=userName, event = volunteerEventID, attended = True, hoursEarned = eventLengthInHours)
+        return True
```
BrianRamsay: I'm a little confused by this function. If they already have a record in `EventParticipant`, attended and hoursEarned won't be updated, but they will be there if they aren't in the table yet. Seems inconsistent. I would think we'd want the end database result to be the same, whether or not they had a record already there.

app/controllers/admin/routes.py
```diff
@@ -1,16 +1,50 @@
...
+        if ProgramEvent.get_or_none(ProgramEvent.event == eventID, ProgramEvent.program == programID):
+            eventParticipantsData = EventParticipant.select().where(EventParticipant.event == eventID)
+
+            eventParticipantsData = eventParticipantsData.objects()
+
+            event = Event.get_by_id(eventID)
+            program = Program.get_by_id(programID)
+            startTime = event.timeStart
+            endTime = event.timeEnd
+            eventDate = event.startDate #start date and end date will be the same
```
BrianRamsay: I don't think there's a good reason for 3 separate variables here. They aren't used anywhere else other than the next line, and the original sources aren't that long.

app/logic/updateTrackVolunteers.py
```diff
@@ -0,0 +1,63 @@
...
+        if f'username{user}' in participantData:
+            username = participantData[f'username{user}']
+            if (User.get_or_none(User.username == username)):
+
+                try:
+                    if participantData['checkbox_'+ username] == "on": #if the user is marked as present
+                        (EventParticipant.update({EventParticipant.hoursEarned: float(participantData['inputHours_'+ username]),EventParticipant.attended: True}).where(EventParticipant.event == participantData['event'],
+                                                 EventParticipant.user == participantData[f'username{user}'])).execute()
+
+                except:   #if there is no checkbox for user then they are not present for the event.
```
BrianRamsay: When you are using a try/except as a control flow, catch the exact exception you are expecting. With the current code, you won't know if you make a syntax error in the `try` because you are catching every exception and treating it as normal.

BrianRamsay (changes requested):

app/controllers/admin/changeVolunteer.py
```diff
@@ -0,0 +1,51 @@
...
+from app.logic.updateVolunteers import updateVolunteers, addVolunteerToEvent
+from app.models.user import User
+from peewee import *
+
+@admin_bp.route('/searchVolunteers/<query>', methods = ['GET'])
+def getVolunteers(query):
+    '''Accepts user input and queries the database returning results that matches user search'''
+    query = query.strip()
+    search = query.upper()
+    splitSearch = search.split()
```
BrianRamsay: `search` and `splitSearch` are never used here.

app/controllers/admin/changeVolunteer.py
```diff
@@ -0,0 +1,51 @@
...
+        flash("Error adding volunteer")
+
+    return redirect(url_for("admin.trackVolunteersPage", programID=programID, eventID=eventID))
+
+
+@admin_bp.route('/addVolunteerToEvent/<user>/<volunteerEventID>/<eventLengthInHours>', methods = ['POST'])
+def addVolunteer(user, volunteerEventID, eventLengthInHours):
+    user = user.strip("()")
+    userName=user.split('(')[-1]
+    succesfullyAddedVolunteer = addVolunteerToEvent(userName, volunteerEventID, eventLengthInHours)
```
BrianRamsay: I think you should get a user object either here or at the beginning of addVolunteerToEvent so you can validate that the user is an existing user before you try to do anything with the username.

app/logic/updateVolunteers.py
```diff
@@ -0,0 +1,64 @@
...
+                    (EventParticipant.update({EventParticipant.attended: False, EventParticipant.hoursEarned: 0}).where(EventParticipant.event == participantData['event'],
+                        EventParticipant.user == participantData[f'username{user}'])).execute()
+                except Exception as e:
+                    print(e)
+                    return False
+            else:
+                return False
+        else:
+            break
+    return True
```
BrianRamsay, line 47: One thing I didn't think about with this function. You're getting a list of people, but the whole function will return false if one fails. Do you want to do that or do we want to continue on and try to update the other participants?

I'm not attached to one solution in particular, just wanted to make sure you had thought about it.

BrianRamsay (changes requested):

app/logic/updateVolunteers.py
Sandesh-L: We've decided to leave it as it is.

app/logic/searchVolunteers.py
```diff
@@ -0,0 +1,32 @@
...
+    else:
+        for searchTerm in splitSearch: #searching for specified first and last name
+            if len(searchTerm) > 1:
+                searchTerm += "%"
+                results = User.select().where(User.isStudent & User.firstName ** firstName & User.lastName ** lastName)
+                for participant in results:
+                    if participant not in resultsDict:
+                        resultsDict[f"{participant.firstName} {participant.lastName} ({participant.username})"] = f"{participant.firstName} {participant.lastName} ({participant.username})"
+
+    dictToJSON = json.dumps(resultsDict)
```
BrianRamsay, line 31: do the json dump in the controller, rather than the search function

BrianRamsay (approved):
Nice work!

PR #58, opened 2021-07-22
PR Volunteer event management all event page

zachneill:
This is the PR for the balsamic page https://balsamiq-wireframes.appspot.com/?state={%22action%22:%22open%22,%22ids%22:[%221StA6txFQr7w6c3gJrNHYKxEm6PmXxpZ9%22],%22userId%22:%22114880022803373311847%22}# 

The user must be able to view events grouped by program. The programs are sorted by the currently available categories. Check the balsamic page above for the wireframe that we used to create this page. 
We added a conditional statement in the template to check if the type of user which will show them different buttons in the table of the accordion body. If the user is a student we give them the RSVP button and if the user is a CeltsAdmin or Faculty we give them the Email button.

We used bootstrap 5's nav-pills to display the categories. We implemented a generic accordion with Jinja's macro feature. The macro sources from the queries in events.py that retrieve the correct data from the database. The render template renders all the queries and all variables at once in /events/<term>/ in routes.py. The data is term-sorted. The page uses no JS, per request from Scott and Brian. 

To test: Go to /events/<term> to check that all events for that term exist. There is also a test_eventList.py file. This test suite primarily tests the return statement from groupingEvents(term) in the events.py file. 

Notes: Email button , RSVP and links to the events do not work because they are dependent on a page that hasn't been created yet.


app/controllers/events/routes.py
```diff
@@ -1,18 +1,24 @@
...
 from app.logic.getUpcomingEvents import getUpcomingEventsForUser
 
-@events_bp.route('/events', methods=['GET'])
-def events():
-    events = getEvents()
+@events_bp.route('/events/<term>/', methods=['GET'])
+def events(term):
+
+    eventsDict = groupEventsByCategory(term)
+    print(eventsDict)
```
BrianRamsay: Don't leave temporary debug output in the code

app/templates/events/event_list.html
```diff
@@ -1,8 +1,106 @@
...
+</div>
+{% macro createAccordion(categoryPrograms) %}
+  <div class="accordion" id="categoryAccordion">
+    {% for program,events in categoryPrograms.items() %}
+      <div class="accordion-item">
+        <h2 class="accordion-header" id="accordion__header_{{program}}">
+          <button class="accordion-button collapsed"
+                  type="button"
+                  data-bs-toggle="collapse"
+                  data-bs-target="#accordion__body_{{program}}_{{ loop.index }}"
```
BrianRamsay: loop.index could have a conflict with another tab. For example, if Adopt-A-Grandparent is always first in the list, it will have the same ID each time.

tests/code/test_eventList.py
```diff
@@ -0,0 +1,34 @@
...
+    studentLedEvents = (Event.select(Event, Program.id.alias("program_id"))
+                             .join(ProgramEvent)
+                             .join(Program)
+                             .where(Program.isStudentLed,
+                                    Event.term == 1))
+    groupedEvents = groupEventsByProgram(studentLedEvents)
+    assert groupedEvents == {Program.get_by_id(1): [Event.get_by_id(1), Event.get_by_id(2)] , Program.get_by_id(2): [Event.get_by_id(4)]}
+
+@pytest.mark.integration
+def test_groupEventsByCategory():
```
BrianRamsay, line 49: This function has a lot of other queries that are a little different. You should set up test data so that this test checks them all, not just studentLedEvents. And actually, it's not really testing that one either, is it, just making sure there aren't any errors in execution.

BrianRamsay (changes requested):

app/controllers/events/routes.py
Karina-Agliullova: Removed line 12

app/templates/events/event_list.html
sreynit02: added a parameter to the createAccordion macro mainly so that it could be used in the id. passed in differnt integer for that parameter when we use the createAccordion macro for different pill.

tests/code/test_eventList.py
sreynit02: Made tests for all queries

BrianRamsay (approved):
Nice work!

PR #60, opened 2021-07-26
Event form #8

Ebenezer997:
1.Event form or volunteer view for user to RSVP
2.'fixes issues Event Form #8' 
3.This will show the user a program's event details and allow the user to RSVP for the event. It is a view-only page for students to just read and rsvp for an event. 
we added read-only tags so the students are  unable to edit any inputs, used jinja statements to check if a user is an admin, staff, or student, added a new RSVP button
4.The testing is done by using PHP Myadmin. editing a user as a student then try to RSVP as the user. Add user to programban and try to RSVP for that event. If the next page after the RSVP button is clicked flashes "Cannot RSVP student(name) is banned" or when is a student who is eligible the next page will flashes Successfully registered for the event.
URL : http://IP:8080/<programID>/<eventID>/edit_event
Note: the next page after clicking RSVP is supposed to showUpComingEvents page but the flasher does work so we redirected it to AddParticipant  page.


BrianRamsay:
lots of merge conflicts. You need to get development into your branch

BrianRamsay:
Hit the wrong button. You need to get development into your branch or you are going to start having more and more conflicts. I'm a little worried that there are too many right now.

app/controllers/admin/createEvents.py
```diff
@@ -53,12 +51,9 @@ def createEvent():
...
             return redirect(url_for("admin.createEventPage", program=2)) #FIXME: have this redirect to main programs page (or some appropriate non admin page).
 
-
-
 def selectFutureTerms(currentTermid):
     futureTerms = (Term.select().where(Term.id >= currentTermid)
                                 .where((Term.year <= (Term.get_by_id(currentTermid)).year + 2)))
-
-    listOfTerms = [future.description for future in futureTerms]
+    listOfTerms = [future for future in futureTerms]
```
BrianRamsay: This is redundant. If you really needed a list instead of an iterable query object you could do `listOfTerms = list(futureTerms)`, but I don't think you need even that.

app/controllers/main/routes.py
```diff
@@ -26,6 +27,14 @@ def volunteerIndicateInterest():
                            interests_ids = interests_ids)
 
 
+@main_bp.route('/rsvpForEvent', methods = ['POST'])
+def RSVPCaller():
+    print("Hello")
```
BrianRamsay: no debug prints

app/controllers/main/routes.py
```diff
@@ -26,6 +27,14 @@ def volunteerIndicateInterest():
...
+    eventData = request.form.copy()
```
BrianRamsay: is there a reason to copy the event data first?

app/controllers/main/volunteerRegisterEvents.py
```diff
@@ -0,0 +1,67 @@
...
+    eventId = eventData['eventId']
+    program = eventData['programId']
+    isEligible = userRsvpForEvent(userId, eventId)
+    listOfRequirements = unattendedRequiredEvents(program, userId)
+
+
+    if not isEligible: # if they are banned
+        flash(f"Cannot RSVP: {userId.firstName} is banned")
+
+    elif len(listOfRequirements) > 0:
```
BrianRamsay: We should talk about this kind of warning. Because the trainings aren't required anymore.

app/controllers/main/volunteerRegisterEvents.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+        else:
+            reqListToString = str(listOfRequirements[0])
+            reqListToString = 'Berea Buddies'
+            flash(f"{userId.firstName} {userId.lastName} successfully registered. Warning! You have not attended the following required trainings: {reqListToString}")
+
+    #if they are eligible
+    else:
+        RSVPupdate =(EventParticipant.update({EventParticipant.rsvp: True})
```
BrianRamsay: weren't they already registered above in userRsvpForEvent?

app/controllers/main/volunteerRegisterEvents.py
```diff
@@ -0,0 +1,67 @@
...
+from app.models.eventParticipant import EventParticipant
+from app.controllers.admin import admin_bp
+from app.controllers.events import events_bp
+from app.logic.userRsvpForEvent import userRsvpForEvent, unattendedRequiredEvents
+from flask import flash, request, redirect, url_for, g
+
+
+
+@admin_bp.route('/rsvpForEvent', methods=['POST'])
+def volunteerRegister():
```
BrianRamsay, line 13: There's too much application logic in this controller method

app/logic/programEligibility.py
```diff
@@ -9,14 +9,14 @@
...
+    banned from a program.
 
     :param program: accepts a Program object or a valid programid
     :param user: accepts a User object or userid
     :return: True if the user is not banned and meets the requirements, and False otherwise
     """
-    program = Program.get_by_id(program)
-    if (ProgramBan.select().where(ProgramBan.user == user)) and (ProgramBan.select().where(ProgramBan.program == program)):
+
+    if (ProgramBan.select().where(ProgramBan.user == user, ProgramBan.program == program)):
```
BrianRamsay: I'd add a .exists() to the end of this query to be explicit what we care about, since it is in an if statement.

app/logic/userRsvpForEvent.py
```diff
@@ -16,11 +16,33 @@ def userRsvpForEvent(user,  event):
...
+    isEligible = isEligibleForProgram(program, user)
+    if isEligible:
+        newParticipant = EventParticipant.get_or_create(user = rsvpUser, event = rsvpEvent, rsvp = True)[0]
+        return newParticipant
+    return isEligible
 
-    raise Exception("User is not eligible")
+
+
+def unattendedRequiredEvents(program, user):
```
BrianRamsay, line 31: There is no test for this function. that means this was developed in the wrong order

BrianRamsay (changes requested):

app/controllers/main/routes.py
Sandesh-L: nope. in fact this function is unnecessary

app/controllers/main/volunteerRegisterEvents.py
Sandesh-L: Yup

app/controllers/admin/createEvents.py
Sandesh-L: we fixed this but based on the slack conversations it looks like this will change.

Sandesh-L:
To test for events that have already passed, go to the 2/5/edit_event. Then change admin status to see different options.

BrianRamsay (approved):
Nice!

Issue #4, opened 2021-06-11
Student Search Page


bryantal:
This is where a CELTS admin will search for students that are volunteers or student staff to see their information

SRS Items:
**3.0.1**	The CELTS administrator must be able to search for all volunteers -SL/GJ
**3.0.2**	The CELTS administrator must be able to search for all CELTS student staff

See Admin UI Wireframe: https://balsamiq-wireframes.appspot.com/?state=%7B%22ids%22:%5B%221etrefTLsvWQntesh5_TDdC66TRVzRpHf%22%5D,%22action%22:%22open%22,%22userId%22:%22109665407847548855125%22,%22resourceKeys%22:%7B%7D%7D
The slides titled "3.2 Student Search" and "3.2 Students Searched" are what we want this page to look like.
![image](https://user-images.githubusercontent.com/28715731/121741402-7c26f880-cacc-11eb-9dac-8e0cba83f16d.png)
![image](https://user-images.githubusercontent.com/28715731/121741424-8517ca00-cacc-11eb-90d5-5596ebfebeda.png)

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #4: Student Search Page](https://trello.com/c/wK8jkBEq/268-celts-4-student-search-page)

PR #62, opened 2021-07-28
Create student search page issue#4

gastonjarju:
Summary:
This PR resolves issue #4 
* Created a simple search bar on the student search page that the admin can use to search users by their username/first name/ last name. We also made sure that if the volunteer's name pops up and the admin clicks on their name, it will redirect them to the volunteer's profile page.

To test:
You can visit the web page with  URL: http://:8080/search_student.  To test, try search any user from the database either by their username/first name/ last name. Click their name and it should redirect you to their profile page.

### DIFF
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -1,7 +1,7 @@
 from flask import request, render_template
 from flask import Flask, redirect, flash, url_for
 from app.models.user import User
-from app.controllers.admin import admin_bp, updateSearchStudent
+from app.controllers.admin import admin_bp, getStudent
 
 @admin_bp.route('/testing', methods=['GET'])
 def testing():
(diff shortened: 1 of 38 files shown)


tests/code/test_searchStudents.py
```diff
@@ -0,0 +1,42 @@
...
+
+    query = '123'
+    searchResults = searchVolunteers(query)
+    assert len(searchResults) == 0
+
+    #test with multiple users matching query
+    User.get_or_create(username = 'sawconc', firstName = 'Candace', lastName = 'Sawcon', bnumber = '021556782')
+    query = 'sa'
+    searchResults = searchVolunteers(query)
+    print(searchResults)
```
BrianRamsay: remove this debug output

app/controllers/admin/routes.py
```diff
@@ -1,26 +1,32 @@
...
 from app.controllers.admin.changeVolunteer import getVolunteers
 
 
 @admin_bp.route('/testing_things', methods=['GET'])
 def testing():
     return "<h1>Hello</h1>"
 
+@admin_bp.route('/search_student', methods=['GET'])
+def studentSearchPage():
+    students = User.select()
```
BrianRamsay: you are selecting all users here and passing it to the search page, but I don't see where you use it on the page. the search is set up to use ajax rather than autocomplete, right?

BrianRamsay (changes requested):

app/controllers/admin/routes.py
Sandesh-L: yup

BrianRamsay (approved):
Looks good!

PR #65, opened 2021-07-29
volunteer search on the sidebar now leads to the volunteer search page

Sandesh-L:
This PR fixes issue #63 

volunteer search on the sidebar now leads to the volunteer search page and highlights the Volunteer search button.

We changed the href in the sidebar.html to be `/student_search` the URL for the search page. We also added a jinja tag that sets the Volunteer Search button to be `active` if the current path is `/student_search`.

To test:
Flask run and go to any webpage with a sidebar. 
Click on Volunteer Search on the sidebar.
It should take you to the new page and highlight the button.

NOTE: The URL for the event management page doesn't exist yet so that button doesn't work. This branch is a fork from PR #62 so that should be merged before this one.

### DIFF
--- a/app/controllers/admin/getStudent.py
+++ b/app/controllers/admin/getStudent.py
@@ -1,6 +1,6 @@
 from app.logic.searchStudents import searchVolunteers
 from app.controllers.admin import admin_bp
-from flask import flash,redirect, url_for, request
+from flask import flash,redirect, url_for, request, abort
 from app.controllers.main import main_bp
 from peewee import *
 
@@ -24,6 +24,4 @@ def volunteerProfile():
     volunteerName= request.form.copy()
     username = volunteerName['searchStudentsInput'].strip("()")
     user=username.split('(')[-1]
-    # print(userName)
-    # print(volunteerName)
     return redirect(url_for('main.profilePage', username=user))
--- a/app/controllers/admin/getStudent.py
+++ b/app/controllers/admin/getStudent.py
@@ -3,6 +3,8 @@ from app.controllers.admin import admin_bp
 from flask import flash,redirect, url_for, request, abort
 from app.controllers.main import main_bp
 from peewee import *
+from flask import json
+
 
 
 @admin_bp.route('/searchStudents/<query>', methods = ['GET'])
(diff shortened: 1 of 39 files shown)


app/controllers/admin/getStudent.py
```diff
@@ -0,0 +1,27 @@
...
+def searchStudents(query):
+    '''Accepts user input and queries the database returning results that matches user search'''
+    query = query.strip()
+    search = query.upper()
+    splitSearch = search.split()
+    try:
+        return searchVolunteers(query)
+
+    except Exception as e:
+        return e, 500
```
BrianRamsay: If you're just going to return the error and a 500 code, you don't need to catch the exception. That's the default behavior for exceptions. You catch it if you want to do something like display a friendlier message, or log the output to the console rather than show the user (or do some other logic with it)

app/logic/searchStudents.py
```diff
@@ -0,0 +1,32 @@
...
+    else:
+        for searchTerm in splitSearch: #searching for specified first and last name
+            if len(searchTerm) > 1:
+                searchTerm += "%"
+                results = User.select().where(User.isStudent & User.firstName ** firstName & User.lastName ** lastName)
+                for participant in results:
+                    if participant not in resultsDict:
+                        resultsDict[f"{participant.firstName} {participant.lastName} ({participant.username})"] = f"{participant.firstName} {participant.lastName} ({participant.username})"
+
+    dictToJSON = json.dumps(resultsDict)
```
BrianRamsay: I would do the json dump in the controller, not the function.

app/logic/searchStudents.py
```diff
@@ -0,0 +1,32 @@
+from app.models.user import User
+from app.controllers.main import main_bp
+from flask import json, jsonify
+
+
+
+def searchVolunteers(query):
```
BrianRamsay, line 6: Is there a test for this function? I thought there was, but I'm not seeing it in this PR for some reason

BrianRamsay (changes requested):
Looks pretty good, just a couple of comments and questions

BrianRamsay (approved):

PR #66, opened 2021-07-30
Add term dropdown event management

Karina-Agliullova:
**Note:** Review this PR after PR #58 - Volunteer event management all event page

**Problem:** The user should be able to select a term from a term dropdown on the All events page. The selected term should match the programs that are in that term. Also, by selecting the term, the term name in the heading should be changed to the term selected by the user.

**Solution:** In the event_list.html file, we implemented a dropdown for the term selection. Once the term is selected, it will be passed to the controller to get the correct events that belong to that term. The page will load those events that are included in that term. 

**Testing:** Navigate to /events/<programid> and select any term, and the page should load the events that are in that term and display that term's name in the heading.

### DIFF
--- a/app/templates/events/event_list.html
+++ b/app/templates/events/event_list.html
@@ -12,15 +12,21 @@
 {% endblock %}
 {% block app_content %}
 
-
 <h1 class="text-center">Events List for {{selectedTerm.description}}</h1>
-<div class="d-inline-flex container justify-content-center">
-  <div class="form-group d-inline"><strong>Term *</strong>
-    <select class="form-control" name='eventTerm' id='inputTerm' onchange="location = this.value;">
-      {% for term in listOfTerms %}
-        <option value='/events/{{term}}/' {{ "selected" if selectedTerm.id == term.id }}>{{term.description}} </option>
-      {% endfor %}
-    </select>
+<div class="container-fluid px-1 py-3">
+  <div class="row d-flex align-items-baseline p-3">
+    <div class="col-md-1 ps-1 ms-auto">
+      <h4><strong>Term</strong></h4>
+    </div>
+    <div class="col-md-3 me-auto">
+      <div class="form-group">
+        <select class="form-control" name='eventTerm' id='inputTerm' onchange="location = this.value;">
+          {% for term in listOfTerms %}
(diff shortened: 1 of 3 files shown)


app/templates/events/event_list.html
```diff
@@ -1,8 +1,122 @@
...
-    {{user}}
-    {% for event in events %}
-        <li>{{event}} </li>
-    {% endfor %}
-</ul>
+{% extends "base.html"%}
+{% block scripts %}
+  {{super()}}
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-select/1.12.1/js/bootstrap-select.js"></script>
+  <script src="https://code.jquery.com/ui/1.12.1/jquery-ui.js"></script>
```
BrianRamsay: jquery-ui should go in the base template

app/templates/events/event_list.html
```diff
@@ -1,8 +1,122 @@
...
+  </script>
+{% endblock %}
+{% block styles %}
+  {{super()}}
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-select/1.12.1/css/bootstrap-select.css" />
+  <link rel="stylesheet" href="//code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css">
```
BrianRamsay: jquery-ui should go in the base template

app/templates/events/event_list.html
```diff
@@ -1,8 +1,122 @@
...
+    <button class="nav-link" id="bonnerScholarsEvents"
+    data-bs-toggle="pill" data-bs-target="#pills-bonner-scholars" type="button" role="tab" aria-controls="pills-bonner-scholars" aria-selected="false">Bonner Scholars</button>
+  </li>
+  <li class="nav-item" role="presentation">
+    <button class="nav-link" id="oneTimeEvents"
+    data-bs-toggle="pill" data-bs-target="#pills-one-time" type="button" role="tab" aria-controls="pills-one-time" aria-selected="false">One-time Events</button>
+  </li>
+  </ul>
+</div>
+{% macro createAccordion(categoryPrograms) %}
```
BrianRamsay: this is old, which makes me think you need to update this pr with the latest development changes

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #74, opened 2021-08-05
Fix-mes


gastonjarju:
To complete the fix-me that were left pending.

PR #75, opened 2021-08-06
Fix me issue#74

gastonjarju:
Summary:
This PR fixes issue #74 
- Re-routed the create events page to the `events/<program>` page after an event is created. Also re-routed delete events page to event list page after an event is deleted. Renamed a CELTS program to one of the programs that do not involve student-led or Booner Scholar.
- Added a remove RSVP button on the events/term web page and gave it functionalities such that they could both RSVP or un-rsvp from an event.
- Added the functionality that will disable the RSVP button if the event has already passed in the event/terms web page.

Note: There are a couple of fixmes pending in development.


app/controllers/main/volunteerRegisterEvents.py
```diff
@@ -31,6 +30,11 @@ def volunteerRegister():
     #if they are eligible
     else:
         flash("Successfully registered for event!","success")
+    print(eventData['eventId'], eventData['programId'])
+    print('rsvp??', (EventParticipant.select().where((EventParticipant.user == userId) &(EventParticipant.event == eventData['eventId'])& (EventParticipant.rsvp == True))).exists())
```
BrianRamsay: debug output should be removed

app/static/js/event_list.js
```diff
@@ -0,0 +1,39 @@
...
+  rsvpInfo = {eventId: eventID,
+              programId: programID,
+              from: 'ajax'}
+
+  $.ajax({
+    url: "/rsvpForEvent",
+    type: "POST",
+    data: rsvpInfo,
+    success: function(s){
+        location.reload() // ?????????????????????????????????????
```
BrianRamsay: what do the question marks mean?

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #87, opened 2021-08-23
Volunteer profile combined

### LINKED ISSUE #81: Combine Volunteer Profile
The program interest section of the volunteer profile needs to be combined with the table that shows a volunteer's upcoming events.

bryantal:
This pull merges the volunteer indicate interest section with the volunteer upcoming events table into one full volunteer profile.  I have also added a new directory for volunteer in templates to put the html file for the volunteer profile in.  I also put the controller for the volunteer upcoming events in the same place as the controller for the indicate interest section.


BrianRamsay (changes requested):
The tests don't pass, likely because of errors introduced resolving merge conflicts.

Also, merge the latest development to fix a line ending issue

BrianRamsay (changes requested):
test failures in the build

app/controllers/main/routes.py
```diff
@@ -12,30 +13,26 @@ def home():
     print(f"{g.current_user.username}: {g.current_user.firstName} {g.current_user.lastName}")
     return render_template('main/home.html', title="Welcome to CELTS!")
 
-@main_bp.route('/volunteerIndicateInterest', methods = ['GET'])
-def volunteerIndicateInterest():
+@main_bp.route('/profile/<username>', methods = ['GET'])
```
BrianRamsay, line 16: you don't actually use <username> here. You probably should use username instead of the current user so that admins could view other students' profile pages. (you'll need to check for authorization)

BrianRamsay: it looks like there is some authorization later, but I would move it up to the top. And right now if a celts admin is viewing this page it will be a list of their own events, not the events connected to the profile of `<username>`

app/templates/volunteer/volunteerProfile.html
```diff
@@ -0,0 +1,58 @@
...
+        {% for event in upcomingEvents %}
+        <tr>
+          <td>{{event.singleProgram.programName if event.singleProgram else '--'}}</td>
+          <td>{{event.eventName}}</td>
+          <td>{{event.startDate}}</td>
+          <td>{{event.endDate}}</td>
+        </tr>
+        {% endfor %}
+      {% else %}
+        <h2> Please be more interesting <h2>
```
BrianRamsay: This should be real text now.

BrianRamsay (changes requested):

app/templates/volunteer/volunteerProfile.html
```diff
@@ -27,7 +27,7 @@
         </tr>
         {% endfor %}
       {% else %}
-        <h2> Please be more interesting <h2>
+        <h2> Please select the programs you would like to receive updates on. <h2>
           <!-- FIXME: There should be a link here to where they can select interests to programs -->
```
BrianRamsay: Get rid of the FIXME if it isn't relevant

app/controllers/main/routes.py
```diff
@@ -15,14 +15,16 @@ def home():
...
+        return "", 500 #TODO: Link to an error page
+    upcomingEvents = getUpcomingEventsForUser(username)
     programs = Program.select()
     interests = Interest.select().where(Interest.user == g.current_user)
     interests_ids = [interest.program for interest in interests]
     if username == g.current_user.username or g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff:
         return render_template('/volunteer/volunteerProfile.html',
                                title="Volunteer Interest",
-                               user = g.current_user,
+                               user = username,
```
BrianRamsay: you were passing a User object, and now you are passing a text username. Does it still work?

app/controllers/main/routes.py
```diff
@@ -12,30 +13,28 @@ def home():
...
+        return render_template('/volunteer/volunteerProfile.html',
+                               title="Volunteer Interest",
+                               user = username,
+                               programs = programs,
+                               interests = interests,
+                               interests_ids = interests_ids,
+                               upcomingEvents = upcomingEvents)
+    else:
+        return "", 500
+        abort(403)
```
BrianRamsay: Nothing after a `return` statement will be executed, so which do you want? A 500 error or a 403 error? Hint: 403 is an access forbidden error.

app/controllers/main/routes.py
BrianRamsay: I'm guessing no, since your template calls `user.username`, which doesn't work if user is just a string of text. Gotta test after you make changes.

app/controllers/main/routes.py
```diff
@@ -12,30 +13,28 @@ def home():
     print(f"{g.current_user.username}: {g.current_user.firstName} {g.current_user.lastName}")
     return render_template('main/home.html', title="Welcome to CELTS!")
 
-@main_bp.route('/volunteerIndicateInterest', methods = ['GET'])
-def volunteerIndicateInterest():
+@main_bp.route('/profile/<username>', methods = ['GET'])
+def profilePage(username):
+    if username != g.current_user.username or g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff:
+        return "", 500 #TODO: Link to an error page
```
BrianRamsay: don't leave this as a TODO. Think about the user experience

app/templates/volunteer/volunteerProfile.html
```diff
@@ -0,0 +1,58 @@
...
+        {% for event in upcomingEvents %}
+        <tr>
+          <td>{{event.singleProgram.programName if event.singleProgram else '--'}}</td>
+          <td>{{event.eventName}}</td>
+          <td>{{event.startDate}}</td>
+          <td>{{event.endDate}}</td>
+        </tr>
+        {% endfor %}
+      {% else %}
+        <h2> Please select the programs you would like to receive updates on. <h2>
```
BrianRamsay: I'd say the text should be more like "Choose interests below to be notified about upcoming events."

BrianRamsay (changes requested):

app/controllers/main/routes.py
```diff
@@ -12,30 +13,27 @@ def home():
...
-
 @main_bp.route('/profile/<username>', methods = ['GET'])
 def profilePage(username):
-
-    if username == g.current_user.username or g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff:
-        return f'<h1>Profile page for {username}</h1>'
-
-    abort(403)
+    if not g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff or g.current_user.username == username:
+        print(g.current_user.username)
```
BrianRamsay: Print more than just the username if you are wanting to print something. Otherwise remove it.

app/controllers/main/routes.py
```diff
@@ -12,30 +13,27 @@ def home():
...
+        return "Access Denied", 403
+    else:
+        profileUser = User.select().where(User.username == username)
```
BrianRamsay: use get_by_id or get when you have the primary key

app/controllers/main/routes.py
```diff
@@ -12,30 +13,27 @@ def home():
...
+        upcomingEvents = getUpcomingEventsForUser(username)
+        programs = Program.select()
+        interests = Interest.select().where(Interest.user == g.current_user)
```
BrianRamsay: you're still using current_user here instead of the user you get in line 22

BrianRamsay (changes requested):
There is also a merge conflict with development to resolve.

app/controllers/main/routes.py
```diff
@@ -12,30 +13,31 @@ def home():
...
-                           programs = programs,
-                           interests = interests,
-                           interests_ids = interests_ids)
-
-
 @main_bp.route('/profile/<username>', methods = ['GET'])
 def profilePage(username):
+    if not g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff or g.current_user.username == username:
+        return "Access Denied", 403
+    else:
```
BrianRamsay: Since the `if` is returning, you don't need an else. You can skip it and unindent the rest

app/controllers/main/routes.py
```diff
@@ -12,30 +13,31 @@ def home():
...
-                           title="Volunteer Interest",
-                           user = g.current_user,
-                           programs = programs,
-                           interests = interests,
-                           interests_ids = interests_ids)
-
-
 @main_bp.route('/profile/<username>', methods = ['GET'])
 def profilePage(username):
+    if not g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff or g.current_user.username == username:
```
BrianRamsay: the way this logic is written, a celts student staff member will receive a 403 Forbidden. And so would any student viewing their own profile

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #90, opened 2021-08-23
Make Navbar Accessible 

sreynit02:
Issue: For admins, The all events page and the student search page need to be accessible from the navbar. For students, their volunteer profile page and the all events page need to be accessible from the navbar.

Solution: For Event Management Tab in the navbar, we added the link to the events page which loads the events of the current term by default. 

Test: 
-navigate to /search_student (since we don't have a working homepage yet, we will use this page since it also has the navbar)
-Select Event Management on the navbar. This should lead to the event page of the current term. Change the term and the page should load events for the selected term.
-Select Volunteer Search. This should lead to the volunteer search.


app/controllers/events/routes.py
```diff
@@ -13,6 +13,12 @@
 
 @events_bp.route('/events/<term>/', methods=['GET'])
 def events(term):
+    #set term to current term when events page is accessed from the navbar
```
BrianRamsay, line 16: All of this could be replaced by `Term.get_or_none(isCurrentTerm = True)` or something like that. But... see overall comments.

app/templates/sidebar.html
```diff
@@ -9,7 +9,7 @@ <h1 class="visually-hidden">Admin sidebar</h1>
     <hr>
     <ul class="nav nav-pills flex-column mb-auto">
       <li class="nav-item">
-        <a href="#" class="nav-link text-white {{ 'active' if request.path =='/event_management' }}" aria-current="page">
+        <a href="/events/<term>/" class="nav-link text-white {{ 'active' if request.path =='/events/<term>/' }}" aria-current="page">
```
BrianRamsay: If you need placeholder text, descriptive text would be better, like `/events/current/`

BrianRamsay (changes requested):
A better solution in my opinion would be to add current_term to the global variable g, and then get the current term on each request.

Make sure you aren't selecting every single request.

app/controllers/events/routes.py
sreynit02: replaced line 18-21 with `term = g.current_term`

app/templates/sidebar.html
sreynit02: changed  placeholder text of the link in `sidebar.html` to `/events/current/`

sreynit02:
created g.current current term

app/controllers/events/routes.py
```diff
@@ -25,7 +29,9 @@ def events(term):
         listOfTerms = listOfTerms,
         rsvpedEventsID = rsvpedEventsID,
         currentTime = currentTime,
-        user = g.current_user)
+        user = g.current_user,
+        testing_currentTerm = g.current_term)
```
BrianRamsay: do you need to pass this to the template?

app/logic/loginManager.py
```diff
@@ -53,3 +55,9 @@ def getLoginUser():
         session['username'] = user.username
 
     return user
+
+def getCurrentTerm():
+
+    term = Term.get_or_none(isCurrentTerm = True)
```
BrianRamsay: go ahead and get rid of the variable and just make this function a one-liner

app/__init__.py
```diff
@@ -62,3 +62,9 @@ def inject_environment():
 @app.before_request
 def load_user():
     g.current_user = getLoginUser()
+
+from app.logic.loginManager import getCurrentTerm
+@app.before_request
+def load_currentTerm():
+
```
BrianRamsay: I'd get rid of this newline, since this function is only one line to begin with. Keep the one-liner next to the definition

app/logic/loginManager.py
BrianRamsay: and remove the newline after the `def` line

BrianRamsay (changes requested):
pretty close!

app/controllers/events/routes.py
sreynit02: Yes, because there is an instance where we check if the user is a celtsAdmin or celtsStudentStaff to show the email button.

sreynit02: removed line 33.

BrianRamsay (approved):

PR #94, opened 2021-08-26
Issue #82 - Landing page

### LINKED ISSUE #82: Landing Page
When running the application, a user needs a starting page to land on, so they can use the navbar to get around the application.  The landing page should be the all events page.
image

For volunteers, their profile page should be the landing page.
image

Karina-Agliullova:
**Problem:** The landing page for everyone should be the events list page

**Solution:** In app/controllers/events/routes.py, we removed the route that goes through "/events/<term>" so that we could make an events list page a landing page. We added a selectedTerm instead for the users to be able to navigate on the events list page. We also added another route with "/" so that the landing page is set to the events list page by default. To be able to navigate through the terms, we implemented a currentTerm query that is set to the currentTerm that is selected in a database.


BrianRamsay (changes requested):
Your code overlaps quite a bit with what Sreynit has in https://github.com/BCStudentSoftwareDevTeam/celts/pull/90. Can you get together with her and combine your changes?

hmanalai:
@BrianRamsay In PR #90, they had `if not term.isdigit():` we took it out because it doesn't really serve a purpose in our implementation and it's not clear why we're checking for term not being a digit. Is there something we're missing?

BrianRamsay (changes requested):
The reason they have that conditional is that the route from the sidebar is '/events/current', so they are checking to see if the current term is requested.

I see you've changed the route altogether. Does that still work with their changes? Is there a new way to get to the events, or are we just using the home page for that?

Edit: Also, there's a conflict to resolve

hmanalai:
> 
> 
> The reason they have that conditional is that the route from the sidebar is '/events/current', so they are checking to see if the current term is requested.
> 
> I see you've changed the route altogether. Does that still work with their changes? Is there a new way to get to the events, or are we just using the home page for that?
> 
> Edit: Also, there's a conflict to resolve

With our changes, the homepage will be the events page. It will be the first page all users see.

hmanalai:
@BrianRamsay We fixed the sidebar link to now direct to `/` and also moved the route to be in the main route instead of events route.

BrianRamsay (approved):

PR #98, opened 2021-08-27
Table for issue 78

### LINKED ISSUE #78: Table To View All Service Learning Course Proposals
Faculty UI Wireframe: https://balsamiq-wireframes.appspot.com/?state=%7B%22ids%22:%5B%221ziGyTzz6J7iHjyzmsUzj0yzvuFdITg3n%22%5D,%22action%22:%22open%22,%22userId%22:%22109665407847548855125%22,%22resourceKeys%22:%7B%7D%7D

The faculty need a table that populates with all of their service learning courses/course proposals to view.
SRS Items: **5.2** The faculty must be able to view all of their SL Course Proposals

This slide is what we want this to look like for faculty:
image

zachneill:
This PR is for issue #78 
Issue was to create a page where instructors can see and modify their courses, a.k.a. the following wireframe: https://balsamiq-wireframes.appspot.com/?state={%22action%22:%22open%22,%22ids%22:[%221ziGyTzz6J7iHjyzmsUzj0yzvuFdITg3n%22],%22userId%22:%22114880022803373311847%22} 

In addition to the table, we also implemented a working withdraw function. However, "renew," "edit," and "view" are dependent on pages not yet created. So they are blank. 

Url:
/courseProposals 

To test: 
Go to the url and select "Withdraw" from the select box. The course should be removed. 
Also, go to the file test_proposals.py and have at it.


app/controllers/admin/changeSLAction.py
```diff
@@ -0,0 +1,15 @@
+from flask import request, flash
+from app.controllers.admin import admin_bp
+from app.models.course import Course
+from app.models.courseInstructor import CourseInstructor
+from app.models.courseParticipant import CourseParticipant
+from peewee import *
+
+@admin_bp.route('/withdrawCourse/<course>', methods = ['POST'])
+def withdrawCourse(course):
+    course = Course.get(Course.courseName == course)
```
BrianRamsay: I don't think you want this url to use the course name, because then you have to worry about spaces and punctuation in the url

app/static/js/changeProposalAction.js
```diff
@@ -0,0 +1,31 @@
+
+// withdrawing a course
+course = null
```
BrianRamsay: Figure it out without using a global variable

app/templates/admin/createSLProposalTable.html
```diff
@@ -0,0 +1,65 @@
...
+    <tbody>
+        {% for course in courseDict %}
+          <tr>
+            <td>{{courseDict[course]['name']}}</td>
+            <td>{{courseDict[course]['faculty']}}</td>
+            <td>{{courseDict[course]['term']}}</td>
+            <td>{{courseDict[course]['status']}}</td>
+            <td>
+
+              <select class="form-select" id="{{courseDict[course]['name']}}" onchange='changeAction(this)'>
```
BrianRamsay: You can't use a name as an id. spaces, punctuation, etc don't work in IDs

app/logic/getSLInstructorTableData.py
```diff
@@ -0,0 +1,20 @@
+from app.models.course import Course
+from app.models.user import User
+from app.models.courseInstructor import CourseInstructor
+
+def getProposalData(user):
+    """Returns dictionary with data used to populate SL proposal table"""
+    courses = (Course.select()
+                     .where(CourseInstructor.user==user)
+                     .join(CourseInstructor))
```
BrianRamsay: Add an explicit order_by

app/logic/getSLInstructorTableData.py
```diff
@@ -0,0 +1,20 @@
...
+                     .where(CourseInstructor.user==user)
+                     .join(CourseInstructor))
+    courseDict = {} #any reason why this is a dictionary of dictionaries, wouldn't a list of dictionaries be easier to work with?
+    #I just thought it would be easier to read. But a list is a-ok too.
+    for course in courses:
+        otherInstructors = (CourseInstructor.select().where(CourseInstructor.course==course))
+        faculty = [f"{instructor.user.firstName} {instructor.user.lastName}" for instructor in otherInstructors]
+        courseDict[course.courseName] = {
+        "name":course.courseName,
+        "faculty": ', '.join(faculty),
```
BrianRamsay: I wouldn't comma-separate the list here. Leave the data as intact as possible until you need to use it.

app/logic/getSLInstructorTableData.py
```diff
@@ -0,0 +1,20 @@
...
+        "term":course.term.description,
```
BrianRamsay: likewise, what if you needed the term id later? If you pass the term object here, you can get whatever you need later, and aren't restricted to description.

BrianRamsay (changes requested):

app/static/js/changeProposalAction.js
Sandesh-L: Done!

app/controllers/admin/changeSLAction.py
Sandesh-L: Changed it to use courseID instead

app/templates/admin/createSLProposalTable.html
Sandesh-L: Changed the id to be courseID instead of course name

app/logic/getSLInstructorTableData.py
Sandesh-L: it now uses the term object.

app/logic/getSLInstructorTableData.py
Sandesh-L: Done! The whole list is passed in and jinja will do the join when adding the data to each row.

app/logic/getSLInstructorTableData.py
Sandesh-L: I added a .order_by(Course.id) is that what you meant? or should I be ordering by something else?

BrianRamsay: Depends on what you want to order by. course ID is fine, if you don't have another preference, but think about what it will look like a few years down the road. Which courses do you want showing first?

BrianRamsay (approved):

PR #102, opened 2021-08-31
Consolidate student search methods

### LINKED ISSUE #84: Consolidate student search methods
There are multiple routes and functions that handle searching for students.
`app/logic/searchVolunteers.py`, `app/logic/searchStudents.py`, and ` searchVolunteers() in app/controllers/admin/routes.py` all have a different implementation.

We need to consolidate to a single search method. multiple routes are ok, as long as they are for different purposes and handled by (probably) one controller method.

mupotsal:
Issue #84 : The logic for Searching students and Volunteers was in two different files but the functions within them had little differences. The main difference was the type of the data they were returning. 

Solution: Combined the two functions into one called searchVolunteersAndStudents in searchVolunteersAndStudents.py

Test:

1. Navigate to /1/1/track_volunteers
2. Click the + sign to get the search modal pop up. 
3. Begin Typing "Zach" and a student named Zach should pop up. 
4. Testing for other places where the student is searched is still in progress since the search function have been used in different branch

Screen Shot 2021-08-31 at 10 45 21 AM


app/logic/searchVolunteersAndStudents.py
```diff
@@ -3,8 +3,7 @@
 from flask import json, jsonify
 
 
-
-def searchVolunteers(query):
+def searchVolunteersAndStudents(query):
```
BrianRamsay: searchVolunteersAndStudents() is too wordy. It should reflect what you are actually doing in the function, not how it will be used. So, `searchStudents()`, or even `searchUsers()`.

BrianRamsay: Also, don't mess with JSON here. Let the controller handle changing the representation if it needs to. It's easier to use and easier to test if the function returns a dictionary.

BrianRamsay (changes requested):
Right below the code you edited in routes.py is another search implementation. That one needs to be combined as well.

app/logic/searchVolunteersAndStudents.py
sreynit02: changed `searchVolunteersAndStudents() ` to `searchUsers()`. Moved JSON to controller and changed test accordingly.

BrianRamsay (changes requested):
There are Windows line endings (from development). I've fixed them in development so merge dev in here to resolve them in this branch.

BrianRamsay (changes requested):
There is still a third search that should use searchUsers. See issue description.

BrianRamsay (changes requested):
Getting close! But now events/routes.py has searchUsers and searchVolunteers, two different controller methods that are almost identical. I would probably change the front-ends to be able to use the same back-end search. It wouldn't surprise me if there's duplication in the javascript as well, to handle the results of a search.

mupotsal:
We have updated the pull request. The search  methods are merged.

BrianRamsay (approved):
Nice work!

PR #107, opened 2021-09-03
Refactor volunteer view of event form

### LINKED ISSUE #100: Refactor Volunteer View of Event Form
The event form that volunteers see when they click to go register needs to be changed.  Right now, when a volunteer views an event form, it is just the same as what an admin sees when creating the event, but with read only tags in each dropdown, textbox, etc.  We should change this to be a more readable form without all of the textboxes.  Make sure you are set to be a student in the database and then go to the URL: <YOUR_IP_ADDRESS>:8080/program/eventId/edit_event - Replace program and eventId with the ID's of the program and event ID of the event you want to view.
This:
image
Instead of this:
image

mupotsal:
Issue #100 : The UI view for users who are volunteers still has input boxes even though they can not edit the contents. This is an enhancement of the view.

The following is how it looks like before. 

Screen Shot 2021-09-03 at 9 23 24 AM

The following show the view looks like after. 
Screen Shot 2021-09-03 at 9 24 56 AM

Test:

1. Login in as a student and navigate to /1/1/edit_event
2. The view should be the same as the second uploaded picture.
3. Login as a Celts students staff or Celts admin and navigate to /1/1/edit_event
4. The following view should be displayed.
Screen Shot 2021-09-03 at 9 28 27 AM


app/templates/admin/createEvents.html
```diff
@@ -59,15 +59,19 @@ <h2> New Event Form</h2>
         </div>
         <div class="d-grid gap-1.5">
           <div class="form-group">
-            <label for='eventName'> <strong> Event Name * </strong></label> <br>
-            <input class="form-control"  id='inputEventName' value="{{eventInfo.eventName}}" name="eventName"
-                   {{'required' if user.isCeltsAdmin or user.isCeltsStudentStaff else 'readonly'}}>
```
BrianRamsay, line 64: You're missing the 'required' attribute on all of these form elements now.

BrianRamsay (changes requested):

app/templates/admin/createEvents.html
mupotsal: We have added required back :)

mupotsal: RSVP button functionality is fixed in PR #110

bryantal:
Do we want the facilitators field to be required?  I forgot what we decided on that.  If so, the selectpicker for that field needs a required tag.  If not, then the facilitators field should not show up on the volunteer view when there are no facilitators to display.

BrianRamsay (approved):

PR #110, opened 2021-09-07
Volunteer Remove RSVP Error

### LINKED ISSUE #105: Volunteer Remove RSVP Error
When a student/volunteer goes to remove the RSVP for an event they registered for, it gives an error and does not update the database to remove the RSVP for that student.  This issue stems from the User model not being imported in _app/controllers/main/routes.py_.

sreynit02:
This PR fixes issue #105 : The remove rsvp button was not working: gave an error and did not update in the database.

Fixed: imported the missing models and libraries and added a deletion functionality.

Test:
Log in as a student and navigate to /events/current and RVSP for an event by clicking on the RSVP button. Check the database in the eventParticipant table and it should create an entry for that event. Then, remove rsvp for that same event and get the eventParticipant table again. The previous entry should be removed.


bryantal (dismissed):

app/controllers/main/routes.py
```diff
@@ -82,14 +85,11 @@ def RemoveRSVP():
...
     eventData = request.form
     userId = User.get(User.username == g.current_user)
     eventId = eventData['eventId']
     program = eventData['programId']
-
     currentEventParticipant = EventParticipant.get(EventParticipant.user == userId, EventParticipant.event == eventId)
-
-    currentEventParticipant.delete_instance()
-    flash("Successfully unregistered for event!","success")
+    currentEventParticipant.delete_instance(recursive=True)
```
BrianRamsay: I doubt you want `recursive=True` here.

BrianRamsay (changes requested):

app/controllers/main/routes.py
sreynit02: removed

BrianRamsay (dismissed):

PR #121, opened 2021-09-17
Volunteer rsvp status

### LINKED ISSUE #113: Check for volunteer RSVP status doesn't work correctly
When a volunteer is looking at an event form, the query that checks if they are RSVP'd is flawed.  It gets the the EventParticipant record and if it exists, it will display that they have RSVP'd for the event even if the RSVP status on the EventParticipant record is false.
Related Files:
- _app/controllers/admin/routes.py_
- _app/templates/admin/createEvents.html_

image

sreynit02:
This PR fixes PR #113: RSVP status wasn't working correctly

Solution: Made EventRsvp table mainly to keep track of volunteers who have RSVPed to the event. Removed rsvp status from EventParticipant table. Now the logic is when a volunteer RSVP for an event, we create an entry for them in the EventRsvp table and the EventParticipant Table. When a user is added to an event, we just create an entry for them in the EventParticipant table. When a volunteer wants to remove RSVP from an event, we will remove their entry from both of the tables (EventRSVP and EventParticipant).

Test: 
Log in as a volunteer. RSVP for an event. Check the database and there should be an entry created in the EventRsvp and Eventparticipant.
Remove RSVP for the same event. The entry for that event should be removed from the EventRSVP and EventParticipant table.


BrianRamsay (changes requested):
I think you've made this more confusing with the extra table. Now there's an EventRsvp, but EventParticipant is still doing double duty. So there are three different cases total instead of two. EventParticipant should just be about attendance, and EventRsvp should be about prior registration for an event.

mupotsal:
What extra fields do you think goes to eventRsvp table? If an individual is inside the eventRsvp table that means they have RSVP'ed for the respective program which has their entry. EventParticipant does not have information about RSVP'ng.

bryantal:
It looks like the rsvp status was removed from the eventParticipant table, and that is now only handled in the eventRsvp table which is what we want right?

sreynit02:
This PR is fixed. EventParticipant table now only takes care of attendance and EventRsvp table only takes care of rsvp.

app/controllers/admin/routes.py
```diff
@@ -27,15 +28,18 @@ def trackVolunteersPage(programID, eventID):
     if g.current_user.isCeltsAdmin:
         if ProgramEvent.get_or_none(ProgramEvent.event == eventID, ProgramEvent.program == programID):
             eventParticipantsData = EventParticipant.select().where(EventParticipant.event == eventID)
+            eventRsvpData = EventRsvp.select().where(EventRsvp.event == eventID)
+            eventRsvpUsers = []
             eventParticipantsData = eventParticipantsData.objects()
```
BrianRamsay: this line with .objects is unnecessary. (should be, unless I'm getting mixed up with peewee)

app/controllers/admin/routes.py
```diff
@@ -27,15 +28,18 @@ def trackVolunteersPage(programID, eventID):
...
-
+            for data in eventRsvpData:
+                eventRsvpUsers.append(data.user)
```
BrianRamsay: This extra step of changing to a list of users is unnecessary. You can just get user objects from the original query. This may also be the case with the event participants query.

app/controllers/events/routes.py
```diff
@@ -5,6 +5,28 @@
...
+def events(term):
+    #set term to current term when events page is accessed from the navbar
+    if not term.isdigit():
+        term = g.current_term
+
+    currentTime = datetime.now()
+    eventsDict = groupEventsByCategory(term)
+    listOfTerms = Term.select()
+    participantRSVP = EventRsvp.select().where(EventRsvp.user == g.current_user)
+    rsvpedEventsID = [event.event.id for event in list(participantRSVP)]
```
BrianRamsay: I don't think you need the `list()` call here. Query objects are iterable.

app/controllers/events/routes.py
```diff
@@ -5,6 +5,28 @@
 from app.controllers.events import events_bp
 from app.logic.events import getEvents
 from app.logic.participants import sendUserData
+from datetime import datetime
+from app.models.eventRsvp import EventRsvp
+
+@events_bp.route('/events/<term>/', methods=['GET'])
```
BrianRamsay: Isn't there already a route that does this?

app/templates/events/trackVolunteers.html
```diff
@@ -98,7 +98,7 @@ <h5 class="modal-title" id="exampleModalLabel">Add a volunteer</h5>
         <td>{{participant.user.email}}</td>
         <td>{{participant.user.phoneNumber}}</td>
 
-        <td>{% if participant.rsvp == 1 %} Yes {% else %} No {% endif %}</td>
+        <td>{% if participant.user in eventRsvpUsers  %} Yes {% else %} No {% endif %}</td>
```
BrianRamsay: Doesn't this whole participant bit need to be changed? There won't be any participants here until the event actually happens, which makes it useless for adding volunteers to an event beforehand. It's fine if we're going to make a new ticket to do that work, but we should note it somewhere.

BrianRamsay (changes requested):

app/controllers/admin/routes.py
sreynit02: you're right we don't need .object()

app/controllers/admin/routes.py
sreynit02: Joined User and EventRsvp table and select from that.

app/templates/events/trackVolunteers.html
sreynit02: Note: A possibility of adding the EventRsvp people at the Track Volunteer table.

app/controllers/events/routes.py
sreynit02: Yes. In main/routes.py there's a route that does the same thing and it's the one that we used for our navigation bar. So I deleted this one since we don't use it anywhere else.

app/controllers/events/routes.py
sreynit02: This function is no longer applicable because there's a duplicate in main/routes.py. Made the change in main/routes.py instead.

sreynit02:
PR is fixed and ready to be re-reviewd.

BrianRamsay (approved):

PR #123, opened 2021-09-20
Fixing test data

Ebenezer997:
We updated the events and program names of the test data to the correct names for the Celt's program. Fix the event ids for specific events available. Added new events for new programs. Fix event names to be more clear.
To test reset the database and see if nothing breaks.


database/test_data.py
```diff
@@ -204,6 +204,43 @@
...
+        "programName": "Food Drive",
+        "isStudentLed": False,
+        "isBonnerScholars": False,
+    },
+    {
+        "id": 11,
+        "programName": "Other Events",
+        "isStudentLed": False,
+        "isBonnerScholars": False,
+    }
```
BrianRamsay: We don't have an "Other Events" program. Events can just have no program.

BrianRamsay (changes requested):
You can't change the test data without fixing the tests you break. There are many failures in the build.

BrianRamsay (changes requested):
There are still broken tests.

BrianRamsay:
Tests are still broken

BrianRamsay (changes requested):
There are still a number of failing tests

BrianRamsay (changes requested):
I get two test failures after resetting. Getting closer

BrianRamsay (approved):

PR #125, opened 2021-09-22
Added Contributors page

### LINKED ISSUE #124: Create Contributors Page
We have a link that is supposed to lead to a contributors page in our footer, but it doesn't link to anything because the page doesn't exist.  Let's create it.

Ebenezer997:
Created  HTML file for the contributor page. We follow the format of the LSF contributors page. We added contributors to the default .yml file call the contributors from there.
To test, click on the student software development team link on the footer.


app/config/default.yml
```diff
@@ -23,3 +23,42 @@ mail:
     default_sender: '' #FIXME: Add sender email address
     override_addr: "j5u6j9w6v1h0p3g1@bereacs.slack.com"
     reply_to_address: "j5u6j9w6v1h0p3g1@bereacs.slack.com"
+
+
+contributors:
```
BrianRamsay, line 28: We should add Will Romano, and also any of the folks who worked on the Balsamiq mockups last year.

app/controllers/main/routes.py
```diff
@@ -113,3 +114,9 @@ def RemoveRSVP():
 
     flash("Successfully unregistered for event!", "success")
     return redirect(url_for("admin.editEvent", eventId=eventId, program=program))
+@main_bp.route('/contributors',methods = ['GET'])
+def contributors():
+    contribs = app.config['contributors']
+    return render_template("/Contributors.html",
```
BrianRamsay: the name of this file should be lower-case, like all of our other templates

app/controllers/main/routes.py
```diff
@@ -113,3 +114,9 @@ def RemoveRSVP():
...
+           cfg = contribs
```
BrianRamsay: You don't need to pass this into the page. The application config is available to templates https://stackoverflow.com/questions/7104198/flask-accessing-the-config-variable-in-the-template

app/templates/Contributors.html
```diff
@@ -0,0 +1,30 @@
+{% extends "base.html" %}
+{% set currentPage = "contributors" %}
```
BrianRamsay: Is this variable used? It's declared after base.html is included, and not used in your app_content block.

BrianRamsay (changes requested):

Ebenezer997:
Fixed PR issuess

app/controllers/main/routes.py
```diff
@@ -113,3 +114,9 @@ def RemoveRSVP():
 
     flash("Successfully unregistered for event!", "success")
     return redirect(url_for("admin.editEvent", eventId=eventId, program=program))
+@main_bp.route('/contributors',methods = ['GET'])
+def contributors():
+    # contribs = app.config['contributors']
+    return render_template("/contributors.html"
+           # cfg = contribs
```
BrianRamsay: don't just leave it commented out. It's not useful anymore, and if someone really wants to see what it used to be they can look at the history of the file.

BrianRamsay (changes requested):
Looks good, just one change.

BrianRamsay (approved):
Looks good, thanks.

PR #126, opened 2021-09-22
Vs transcript interface #96

### LINKED ISSUE #96: Volunteer Service Transcript Interface
We have the functionality for generating the volunteer's transcript of their hours, but we still need the tab on the volunteer profile to view all of the hours they volunteer has participated in programs.  The transcript functionality is found in _app/logic/transcript.py_
Volunteer Interface Wireframe: https://balsamiq-wireframes.appspot.com/?state={%22action%22:%22open%22,%22ids%22:[%221NWCXtb1Wvskk2VJNgxxk2lIl937hzC90%22],%22userId%22:%22109665407847548855125%22}
image

MayJue:
This PR fixes #96
- Create the front and back end for the service transcript for volunteers. 
- Create test for the transcript logic 

to test: 
- go to     /serviceTranscript
-  make sure it matches this page :  https://balsamiq-wireframes.appspot.com/?state={%22action%22:%22open%22,%22ids%22:[%221NWCXtb1Wvskk2VJNgxxk2lIl937hzC90%22],%22userId%22:%22113015963196064013745%22}


app/controllers/main/routes.py
```diff
@@ -2,18 +2,33 @@
...
 from app.logic.users import addRemoveInterest
 from app.logic.participants import userRsvpForEvent, unattendedRequiredEvents
+from app.logic.transcript import *
 from app.logic.events import groupEventsByCategory
 from datetime import datetime
 
+@main_bp.route('/')
+def home():
+    print(f"{g.current_user.username}: {g.current_user.firstName} {g.current_user.lastName}")
+    return render_template('main/home.html', title="Welcome to CELTS!")
```
BrianRamsay: Not sure how this old route got added back in. I'm guessing because of merge conflicts, but it makes me worried about other merge conflict resolutions.

app/controllers/main/routes.py
```diff
@@ -113,3 +128,22 @@ def RemoveRSVP():
 
     flash("Successfully unregistered for event!", "success")
     return redirect(url_for("admin.editEvent", eventId=eventId, program=program))
+
+
+@main_bp.route('/serviceTranscript', methods = ['GET'])
+def serviceTranscript():
+    username = "neillz"
```
BrianRamsay: hard-coded username

app/logic/transcript.py
```diff
@@ -9,52 +9,91 @@
...
-        .join(ProgramEvent)
-        .join(Event)
-        .join(EventParticipant)
-        .where(EventParticipant.user == user)
-        .group_by(Program, Event.term)
-        .order_by(Program.programName))
+    eachHours = (EventParticipant.select(fn.SUM(EventParticipant.hoursEarned).alias("hoursEarned")).where(EventParticipant.user == username)
+    + CourseParticipant.select(fn.SUM(CourseParticipant.hoursEarned).alias("hoursEarned")).where(CourseParticipant.user == username))
+
+    totalHours = (eachHours.select_from(fn.SUM(eachHours.c.hoursEarned).alias("hoursEarned")))
```
BrianRamsay: This seems like an odd way to add two values together

app/logic/transcript.py
```diff
@@ -9,52 +9,91 @@
...
+
+    return totalHours
+
+def getStartDate(username):
```
BrianRamsay: I'd change this to getStartYear, since it is returning a year.

app/templates/main/serviceTranscript.html
```diff
@@ -0,0 +1,110 @@
...
+{% block scripts %}
+  {{super()}}
+
+
+{% endblock %}
+
+{% block styles %}
+  {{super()}}
+
+{% endblock %}
```
BrianRamsay: You can get rid of these blocks if you just need the default behavior

app/templates/main/serviceTranscript.html
```diff
@@ -0,0 +1,110 @@
...
+
+
+{% block app_content %}
+
+<h1 class="text-center pb-5">{{g.current_user.firstName}} {{g.current_user.lastName}}</h1>
```
BrianRamsay: Can't use the current-user here, or admins won't be able to print other people's transcripts

app/controllers/main/routes.py
```diff
@@ -2,18 +2,33 @@
...
+    programs = Program.select()
+    interests = Interest.select().where(Interest.user == g.current_user)
+    interests_ids = [interest.program for interest in interests]
+    return render_template('volunteerIndicateInterest.html',
+                           title="Volunteer Interest",
+                           user = g.current_user,
+                           programs = programs,
+                           interests = interests,
+                           interests_ids = interests_ids)
+
```
BrianRamsay, line 19: I suspect this block is a problem, too

BrianRamsay (changes requested):

app/controllers/main/routes.py
```diff
@@ -116,7 +116,29 @@ def RemoveRSVP():
     currentRsvpParticipant.delete_instance()
 
     flash("Successfully unregistered for event!", "success")
-    return redirect(url_for("admin.editEvent", eventId=event.id))
+    return redirect(url_for("admin.editEvent", eventId=eventId, program=program))
```
BrianRamsay: This is wrong. There is no need to pass the program into editEvent. If it's not something you changed, it's likely a merge conflict issue.

BrianRamsay: Notice that it's using a variable `eventId` which doesn't exist in the surrounding code

app/controllers/main/routes.py
```diff
@@ -116,7 +116,29 @@ def RemoveRSVP():
...
+
+@main_bp.route('/<username>/serviceTranscript', methods = ['GET'])
+def serviceTranscript(username):
+    if username != g.current_user and not g.current_user.isAdmin:
```
BrianRamsay: `g.current_user` is a user object and `username` is a text string. You should compare like with like.

app/controllers/main/routes.py
```diff
@@ -116,7 +116,29 @@ def RemoveRSVP():
...
+        abort(403)
+    else:
```
BrianRamsay: Go ahead and remove the `else`. The abort() above exits the function

app/controllers/main/routes.py
```diff
@@ -116,7 +116,29 @@ def RemoveRSVP():
...
+
+        programs = getProgramTranscript(username)
```
BrianRamsay: You should validate `username` before you pass it along to the functions

app/controllers/main/routes.py
```diff
@@ -116,7 +116,29 @@ def RemoveRSVP():
...
+        SLCourses, instructorDict = getSlCourseTranscript(username)
```
BrianRamsay: variables should be camelCase, unlike `SLCourses`

app/logic/transcript.py
```diff
@@ -9,52 +9,108 @@
...
-        .where(EventParticipant.user == user)
-        .group_by(Program, Event.term)
-        .order_by(Program.programName))
+    trainingData = (EventParticipant.select(EventParticipant.event, EventParticipant.hoursEarned)
+                                    .join(Event)
+                                    .where(EventParticipant.user == username, Event.isTraining))
+
+    return trainingData
+
+def getTotalHour(username):
```
BrianRamsay: this method name should be plural

app/logic/transcript.py
```diff
@@ -9,52 +9,108 @@
...
+    eventHours = EventParticipant.select(fn.SUM(EventParticipant.hoursEarned)).where(EventParticipant.user == username).scalar()
+    courseHours =  CourseParticipant.select(fn.SUM(CourseParticipant.hoursEarned)).where(CourseParticipant.user == username).scalar()
+    if eventHours and courseHours:
+        totalHours = eventHours + courseHours
+        return totalHours
+    elif eventHours:
+        return eventHours
+    elif courseHours:
+        return courseHours
+    return 0
```
BrianRamsay: You can avoid this whole awkward if/else with this one statement: `return (eventHours or 0) + (courseHours or 0)`

BrianRamsay (changes requested):

BrianRamsay:
Reminder, this still has some resolution needed

app/controllers/main/routes.py
```diff
@@ -118,6 +118,31 @@ def RemoveRSVP():
...
+@main_bp.route('/<username>/serviceTranscript', methods = ['GET'])
+def serviceTranscript(username):
+    user = User.get_or_none(User.username == username)
+    if user is None:
+        abort(404)
+    if user != g.current_user and not g.current_user.isAdmin:
+        abort(403)
+
+    programs = getProgramTranscript(username)
+    slCourses, instructorDict = getSlCourseTranscript(username)
```
BrianRamsay: We can simplify this method by either making the instructorDict use Course objects for keys, allowing you to loop through the one dictionary to get both courses and instructors, OR add an instructors backreference to CourseInstructor, which will let you access Course.instructors wherever you have a Course object

app/logic/transcript.py
```diff
@@ -9,52 +9,110 @@
...
+                + CourseParticipant.select(Term.year)
+                    .join(Course)
+                    .join(Term).where(CourseParticipant.user == username)).order_by(Event.term.year)
+
+    startDate = startDate.first()
+    if startDate:
+        return startDate.event.term.year
+    return "N/A"
+
+def getUserdata(username):
```
BrianRamsay: unused method

tests/code/test_transcripts.py
```diff
@@ -1,63 +1,197 @@
...
+    addVolunteerToEvent('namet', testingTrainingEvent.id, 2)
+
+    username = "namet"
+    adminName = "ramsayb2"
+
+    testingTrainingsExist = getTrainingTranscript(username)
+    testingTrainingNotExist = getTrainingTranscript(adminName)
+
+
+    assert testingTrainingNotExist.exists() == False
```
BrianRamsay: this one would be clearer with `assert not testingTrainingNotExist.exists()`

tests/code/test_transcripts.py
```diff
@@ -1,63 +1,197 @@
...
+    assert [event.event.name == "Test Training Event" for event in testingTrainingsExist]
```
BrianRamsay: I don't think this is testing what you think. This assert will always be true if `testingTrainingsExist` has anything in it. If you're wanting to test that, just use `assert testingTrainingsExist`.

tests/code/test_transcripts.py
```diff
@@ -1,63 +1,197 @@
...
+    assert testingTrainingsExist.exists()
```
BrianRamsay: Testing if there is anything in an array should usually come before testing the contents of that array. You want to do the broader tests first, so that the specific ones don't fail for reasons other than what you are testing. It's easier to follow the test output, and if the broader ones are later they are useless, because the test will always have failed before you get to them.

tests/code/test_transcripts.py
```diff
@@ -1,63 +1,197 @@
...
+    username = "namet"
+    adminName = "ramsayb2"
+
+    testingBonnerExist = getBonnerScholarEvents(username)
+    testingBonnerNotExist = getBonnerScholarEvents(adminName)
+
+
+    assert testingBonnerNotExist.exists() == False
+    assert [bonner.event.name == "Test Bonner Event" for bonner in testingBonnerExist]
+    assert testingBonnerExist.exists()
```
BrianRamsay: Same comments as `testingTrainings`

tests/code/test_transcripts.py
```diff
@@ -1,63 +1,197 @@
...
 
-    user3 = "agliullovak" #user with no course
-    transcript3 = getSlCourseTranscript(user3)
-    assert transcript3 == []
+    testingSLCExist, instructorDict = getSlCourseTranscript(username)
+    testingSLCNotExist, instructorDict = getSlCourseTranscript(adminName)
+
+    assert testingSLCExist.exists()
+    assert [slc.course.courseName == "Test Course" for slc in testingSLCExist]
+    assert testingSLCNotExist.exists() == False
```
BrianRamsay: Same comments for the others

tests/code/test_transcripts.py
```diff
@@ -1,63 +1,197 @@
...
+    testingProgramEvent = Event.get(Event.name == "Test program Event")
+
+    addVolunteerToEvent('namet', testingProgramEvent.id, 2)
+
+    testingProgramExist = getProgramTranscript(username)
+    testingProgramNotExist = getProgramTranscript(adminName)
+
+    assert testingProgramNotExist.exists() == False
+    assert [program.program.programName == "Test Program Event" for program in testingProgramExist]
+    assert testingProgramExist.exists()
```
BrianRamsay: ditto

tests/code/test_transcripts.py
```diff
@@ -1,63 +1,197 @@
...
+    testingCourse.delete_instance(recursive = True, delete_nullable = True)
+    assert Course.get_or_none(Course.id == testingCourse.id) is None
+    # delete program
+    testingProgramEvent = Event.get(Event.name == "Test program Event")
+    deleteEvent(testingProgramEvent)
+    assert Event.get_or_none(Event.id == testingProgramEvent) is None
+    # delete user
+    user = User.get(User.username == "namet")
+    user.delete_instance(recursive = True, delete_nullable = True)
+    assert User.get_or_none(User.username == "namet") is None
```
BrianRamsay, line 201: Managing your test data this way has created dependencies between your tests. If earlier tests fail or are run independently they will leave bad data in the database. If the tests run in a different order they could fail (this one will, in any case).

If you want to have the same data set for a group of tests, but don't want to use the test data in the database, use `setup` and `teardown` methods (search for how to do that with pytest).

tests/code/test_transcripts.py
```diff
@@ -1,63 +1,197 @@
...
-    assert transcript[2] == ['Empty Bowls', 'Spring A 2021', 5.0]
-
-    user = "ramsayb2" #user who's not involved in any program
-    transcript = getProgramTranscript(user)
-    assert transcript == []
+def testingProgram():
+
+
+    username = "namet"
+    adminName = "ramsayb2"
```
BrianRamsay: these variables are assigned too far away from where they are being used.

BrianRamsay (changes requested):

BrianRamsay (approved):
Looks good!

PR #131, opened 2021-09-28
Note model change

### LINKED ISSUE #77: Change to Note Model to Support Notes on Courses
Admins will be leaving feedback on different questions of service learning course proposals.  We don't have anything in our database to support this, so this will require more discussion (Possibly only need to add a foreign key to the model that leads to the course, but notes could be for each question answered, so unsure of the best way to implement this.  Should discuss with Brian, Alex, and Hila)
CELTS Admin Wireframe: https://balsamiq-wireframes.appspot.com/?state={%22action%22:%22open%22,%22ids%22:[%221etrefTLsvWQntesh5_TDdC66TRVzRpHf%22],%22userId%22:%22109665407847548855125%22}
Look at the service learning slides titled: 4. Admin review to get an idea of how the admin comments are supposed to function

mupotsal:
This PR fixes issue #77. The problem was that in the database, there was no way to store Notes and relate them to their respective questions. 

Solution: 
 We have changed the model for notes and added other models to meet the Notes Requirements as needed by the application. Two new models namely courseQuestion and questionNote have been added.

Test: 
Check is the added models meet the specified needs.


database/test_data.py
```diff
@@ -479,12 +507,13 @@
...
-        "sectionBQuestion4": "",
-        "sectionBQuestion5": "",
-        "sectionBQuestion6": ""
+        # "sectionBQuestion1": "",
+        # "sectionBQuestion2": "",
+        # "sectionBQuestion3": "",
+        # "sectionBQuestion4": "",
+        # "sectionBQuestion5": "",
+        # "sectionBQuestion6": "",
+
```
BrianRamsay, line 510: Don't comment things out that should just be removed.

database/test_data.py
```diff
@@ -570,9 +599,53 @@
...
+    "course":1,
+    "questionContent":" This is another random question",
+    "questionNumber":4,
+    },
+    {
+    "course":1,
+    "questionContent":" Why are you interested in teaching this course?",
+    "questionNumber":5,
+    }
+]
```
BrianRamsay, line 614: Make sure you have more than one course in there, or else you can't properly test that your code handles all cases.

BrianRamsay: Though I just noticed that there's not functionality with this change, so I'm ok with leaving it for now, and adding more data when there's something to test.

BrianRamsay (changes requested):

BrianRamsay:
Just one small change needed

database/test_data.py
sreynit02: removed comments

sreynit02:
PR is fixed

BrianRamsay (approved):

PR #132, opened 2021-09-28
Creates a Global Search User Js File

hmanalai:
**Summary**: This PR places the search user javascript in its own javascript file that can be imported by any other js file that wants to use the functionality. It also moves the backend logic for searching users from the admin route to the main route.


app/static/js/searchUser.js
```diff
@@ -0,0 +1,31 @@
...
+        },
+        error: function(request, status, error) {
+          console.log(status, error);
+        }
+      })
+    },
+     select: function(event , ui) {
+       var user = ui.item.value
+       $(`#${inputId}`).val(ui.item.value)
+       $(`#${formId}`).submit()
```
BrianRamsay: we should check if the element exists before trying to submit it. Do you get a console error right now? Because your searchUser invocation in searchStudents.js doesn't pass a formId.

BrianRamsay (changes requested):
Looks very nice! Just one question.

app/static/js/searchUser.js
hmanalai: I forgot to remove the ajaxURL parameter. Since all requests will be routed to the same URL, that's no longer needed. I will get rid of it.

hmanalai: Since now the only parameters one can pass are the input id and form id, I don't think we need to check for their existence because whoever is using this function would have to pass ids of ones that exist.

BrianRamsay (changes requested):
The volunteer search on the event page doesn't work. "error NOT FOUND" for every query

hmanalai:
@BrianRamsay we updated the PR.

BrianRamsay (changes requested):
There is a huge gap on the track volunteers page that is not there in development. Happens for me at all screen sizes.
Screen Shot 2021-09-30 at 5 01 25 PM

hmanalai:
@BrianRamsay Fixed the gap issue and reformatted the HTML file a bit.

BrianRamsay (approved):

Issue #99, opened 2021-08-27
Admin Email Volunteer After Event Creation


bryantal:
CELTS Admin UI Wireframe: https://balsamiq-wireframes.appspot.com/?state=%7B%22ids%22:%5B%221etrefTLsvWQntesh5_TDdC66TRVzRpHf%22%5D,%22action%22:%22open%22,%22userId%22:%22109665407847548855125%22,%22resourceKeys%22:%7B%7D%7D

We want to give admins the ability to email necessary volunteers about an event they have created (i.e. volunteers who are interested in a program would receive an email for the event page for a training for that program.)
![image](https://user-images.githubusercontent.com/28715731/131168315-ae02aa32-02a6-4b51-bad7-4792e573a166.png)
![image](https://user-images.githubusercontent.com/28715731/131168373-a1967da2-bd15-4b85-86a7-15cadd300bcb.png)

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #99: Admin Email Volunteer After Event Creation](https://trello.com/c/U0RxgKFu/303-celts-99-admin-email-volunteer-after-event-creation)

PR #133, opened 2021-09-29
Allow admins to send email through the event list page

Sandesh-L:
**Summary:** 
Added functionality to the email buttons in the event list page (now the home page). Created a modal that pops up when the email button is clicked. The user can fill out the subject and body of the email as well as who is sending the email and who receives it. The senders can be the CELTS Admin or CELTS Student Staff. The recipients can be people who are interested in the program or only people who are RSVP'ed. 

**To Test:** 
Go to the website's main page and open one of the accordions (make sure the user is an Admin or CeltsStudentStaff). Click on the email button next to the email. Fill out the different fields and hit send.

**Files Changed**: 
default.yml
production.yml
staging.yml
routes.py
event.py
event_list.js
event_list.html
test_events.py
**Files Added:**
test_emailHandler.py
emailModal.html
emailHandler.py
email.py


zachneill:
This branch was created to fix issue #99. Adding this comment to link them since it's not stated in the PR

app/logic/emailHandler.py
```diff
@@ -0,0 +1,89 @@
...
+from app.models.emailTemplate import EmailTemplate
+from app import app
+import sys
+from pathlib import Path
+
+def load_config(file):
+    """ Loads the 'default' file in the config directory. """
+    with open(file, 'r') as ymlfile:
+        cfg = yaml.load(ymlfile, Loader=yaml.FullLoader)
+    return cfg
```
BrianRamsay: The configuration is already loaded in celts. We don't want to copy in LSF's worse configuration handling

app/logic/emailHandler.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def getVolunteerEmails(programID = None, eventID = None, emailRecipients = "interested"):
```
BrianRamsay: I'd split this into two methods, `getParticipantEmails` and `getInterestedEmails`

app/config/default.yml
```diff
@@ -4,7 +4,7 @@ use_tracy: False
 use_banner: False
 override_file: "local-override.yml"
 default_user: "ramsayb2"
-celts_admin_contact: "celts@berea.edu"
+celts_admin_contact: "bramsayr@gmail.com"
```
BrianRamsay: should still be `celts@berea.edu` (per Sandesh)

app/logic/emailHandler.py
```diff
@@ -0,0 +1,89 @@
...
+            MAIL_OVERRIDE_ALL=self.default['mail']['override_addr']
+        )
+
+        self.emailInfo = emailInfo
+        self.mail = Mail(app)
+        self.mail.connect()
+
+    def updateSenderEmail(self):
+        """ Updates who is sending the emails based on the event_list form. """
+        if '@' in self.emailInfo['emailSender']: #if the current user is sending the email
```
BrianRamsay: Note: we can use a mailto link so they send the email from their own mail client

app/templates/events/event_list.html
```diff
@@ -88,8 +89,11 @@ <h2 class="accordion-header" id="accordion__header_{{program}}">
...
                       {% if (user.isCeltsAdmin) or (user.isCeltsStudentStaff) %}
-                        <td><i class="bi bi-envelope-fill"> Email</i></td>
-                        {# FIXME: the Email icon should be clickable and lead to email template #}
+                        <td>
+                            <button class="btn btn-primary" data-bs-toggle="modal" data-bs-target="#emailModal" onclick="passEventInfo({{event.id}}, {{program.id}}, {{selectedTerm}})">
+                              <span class="bi bi-envelope-fill"> Email</span>
+                            </button>
+                        </td>
                       {% else %}
                         {% if (event.startDate < currentTime.date()) or (event.timeStart < currentTime.time() and event.startDate == currentTime.date())%}
```
BrianRamsay: Let's make this logic a property of the Event model, so we can use it like so: `if event.isPast:`

app/templates/events/event_list.html
```diff
@@ -106,10 +110,79 @@ <h2 class="accordion-header" id="accordion__header_{{program}}">
           </div>
         </div>
       </div>
+      {{emailModal}}
     {% endfor %}
   </div>
 {% endmacro %}
 
+{% set emailModal %}
```
BrianRamsay: separate this modal into its own file and include it. This way others can also include the email modal.

tests/code/test_emailHandler.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+from flask_mail import Message
+from app.logic.emailHandler import *
+
```
BrianRamsay, line 9: We will want to figure out how to test the email sending itself, probably with a combination of Mocking and different configs for testing.

BrianRamsay (changes requested):

app/logic/emailHandler.py
```diff
@@ -0,0 +1,84 @@
...
+            MAIL_PORT=app.config['mail']['port'],
+            MAIL_USERNAME= app.config['mail']['admin_username'],
+            MAIL_PASSWORD= app.config['mail']['admin_password'],
+            REPLY_TO_ADDRESS= app.config['mail']['reply_to_address'],
+            MAIL_USE_TLS=app.config['mail']['tls'],
+            MAIL_USE_SSL=app.config['mail']['ssl'],
+            MAIL_DEFAULT_SENDER=app.config['mail']['default_sender'],
+            #ALWAYS_SEND_MAIL=default['ALWAYS_SEND_MAIL']
+            MAIL_OVERRIDE_ALL=app.config['mail']['override_addr']
+        )
```
BrianRamsay: there's no reason to update all of these config values. You aren't even using them anymore.

app/logic/emailHandler.py
```diff
@@ -0,0 +1,84 @@
...
+    def updateSenderEmail(self, message):
+        """ Updates who is sending the emails based on the event_list form. """
+
+        if self.emailInfo['emailSender'] == 'CELTS Admins':
+            message.sender = app.config['mail']['admin_username']
+            # app.config.update(
+            #     mail_username= app.config['mail']['admin_username'],
+            #     mail_password = app.config['mail']['admin_password']
+            # )
+            # print("\n\n"+app.config["mail_username"])
```
BrianRamsay: Don't leave commented out code in, especially if it is out of date

app/logic/emailHandler.py
```diff
@@ -0,0 +1,84 @@
...
+
+        elif self.emailInfo['emailSender'] == 'CELTS Student Staff':
+            app.config.update(
+                mail_username = app.config['mail']['staff_username'],
+                mail_password = app.config['mail']['staff_password']
+            )
```
BrianRamsay, line 52: I don't like swapping application-wide config values for a decision that is local to one email.

app/controllers/events/email.py
```diff
@@ -0,0 +1,36 @@
...
+
+    if emailInfo['emailRecipients'] == "interested":
+        emails = getInterestedEmails(emailInfo['programID'])
+    elif emailInfo['emailRecipients'] == "eventParticipant":
+        emails = getParticipantEmails(emailInfo['eventID'])
+    else:
+        flash("Unable to determine email recipients", "danger")
+
+    if emails == None:
+        flash("Error getting email recipients", "danger")
```
BrianRamsay: You're going to double flash here if there are no emails. I'd remove the `else` condition above, since you are handling the error case here.

app/logic/emailHandler.py
```diff
@@ -0,0 +1,84 @@
...
+                for email in emails:
+                    message.recipients = [email]
+                    conn.send(message)
+        else:
+            if app.config['mail']['override_addr']:
+                message.recipients = [app.config['mail']['override_addr']]
+
+            message.reply_to = app.config['mail']["reply_to_address"]
+            self.mail.send(message)
+        return 1
```
BrianRamsay: We don't need to be returning integers. Let's use True

BrianRamsay: Should you check the return value of the actual message sending to make sure it succeeded? I'm not sure what the possible return values are.

BrianRamsay: This is also a good method to have a try/catch, so that you can return False if there was an exception in email sending. That way the calling code in the controller only has to concern itself with a true or false outcome, not true, false, and exception

app/static/js/event_list.js
```diff
@@ -1,5 +1,34 @@
...
+  //   // console.log($("#subject").val())
+  //   // subject = self.emailInfo['subject'].replace(" ",'%20')
+  //
+  //   // window.location.href = "mainto:"+$("#emailSender").val()+"?Subject="+$("#subject").val()
+  //   window.location.href = "mailto:"+$("#emailSender").val()
+  //   // window.location.href = `mailto:${encodeURIComponent(("#emailSender").val())}?subject=+${("#subject").val()}`
+  //   // `{encodeURIComponent($("#subject").val())}&body=${encodeURIComponent($("#message").val())}`;
+  // })
+
+});
```
BrianRamsay: Let's get rid of this code if it isn't needed. If it might be needed, add some comments to indicate the use case

BrianRamsay (changes requested):

app/config/default.yml
```diff
@@ -12,17 +12,19 @@ db:
...
-    port: 465                # use port 465 for gmail, use 25 for Berea
-    username: ''             # add the email of the sender (for gmail)
-    password: ''             # Add the password of the sender (for gmail)
-    tls: False
-    ssl: True
-    default_sender: '' #FIXME: Add sender email address
-    override_addr: "j5u6j9w6v1h0p3g1@bereacs.slack.com"
-    reply_to_address: "j5u6j9w6v1h0p3g1@bereacs.slack.com"
+
+enabled: True
```
BrianRamsay: this is now too generic a name. It should be `MAIL_ENABLED`

app/config/default.yml
```diff
@@ -12,17 +12,19 @@ db:
...
+MAIL_PORT: 465                # use port 465 for gmail, use 25 for Berea
+MAIL_USERNAME: 'bramsayr@gmail.com'      # add the email of the sender (Admin) (for gmail)
+MAIL_PASSWORD: 'celtsTest'             # Add the password of the sender (for gmail)
+# staff_username: 'sceggenh@gmail.com'   # the email of the sender (celtsStudentStaff)
+# staff_password: 'celtsTest2'
+MAIL_USE_TLS: False
+MAIL_USE_SSL: True
+MAIL_DEFAULT_SENDER: 'bramsayr@gmail.com' #FIXME: Add sender email address
+MAIL_OVERRIDE_ALL: "j5u6j9w6v1h0p3g1@bereacs.slack.com"
+REPLY_TO_ADDRESS: "j5u6j9w6v1h0p3g1@bereacs.slack.com"
```
BrianRamsay: This should also start with `MAIL_` to be consistent

app/config/staging.yml
```diff
@@ -3,7 +3,7 @@ use_tracy: True
 use_banner: True
 
 mail:
-    enabled: False
+    enabled: True
```
BrianRamsay: don't forget to change the other config files with the new arrangement

BrianRamsay (changes requested):

BrianRamsay:
Is this ready for more review or still in progress?

app/config/default.yml
BrianRamsay: still relevant

app/config/default.yml
BrianRamsay: also still relevant

app/config/production.yml
```diff
@@ -3,11 +3,16 @@ use_tracy: True
...
+admin_username: 'bramsayr@gmail.com'      # add the email of the sender (Admin) (for gmail)
+admin_password: 'celtsTest'             # Add the password of the sender (for gmail)
+staff_username: 'sceggenh@gmail.com'   # the email of the sender (celtsStudentStaff)
+staff_password: 'celtsTest2'
+MAIL_SUPRESS_SEND: False
+MAIL_USE_TLS: False
+MAIL_USE_SSL: True
+# MAIL_DEFAULT_SENDER: 'bramsayr@gmail.com' #FIXME: Add sender email address
+MAIL_OVERRIDE_ALL: "j5u6j9w6v1h0p3g1@bereacs.slack.com"
+REPLY_TO_ADDRESS: "j5u6j9w6v1h0p3g1@bereacs.slack.com"
```
BrianRamsay: Don't repeat all of the values in default.yml.

app/config/staging.yml
```diff
@@ -2,11 +2,16 @@ use_shibboleth: False
...
+admin_username: 'bramsayr@gmail.com'      # add the email of the sender (Admin) (for gmail)
+admin_password: 'celtsTest'             # Add the password of the sender (for gmail)
+staff_username: 'sceggenh@gmail.com'   # the email of the sender (celtsStudentStaff)
+staff_password: 'celtsTest2'
+MAIL_SUPRESS_SEND: False
+MAIL_USE_TLS: False
+MAIL_USE_SSL: True
+# MAIL_DEFAULT_SENDER: 'bramsayr@gmail.com' #FIXME: Add sender email address
+MAIL_OVERRIDE_ALL: "j5u6j9w6v1h0p3g1@bereacs.slack.com"
+REPLY_TO_ADDRESS: "j5u6j9w6v1h0p3g1@bereacs.slack.com"
```
BrianRamsay: Only put the things in here that are different from default.yml

app/controllers/events/email.py
```diff
@@ -0,0 +1,59 @@
...
+def emailVolunteers():
+    """ Uses emailHandler to send an email with the form in event_list. """
+
+    emailInfo = request.form.copy()
+    if '@' in emailInfo['emailSender']: # if they are using mailto instead
+        return redirect(url_for("main.events", selectedTerm = emailInfo['selectedTerm']))
+    else:
+        if emailInfo['programID'] == 'Unknown':
+            program = ProgramEvent.get(ProgramEvent.event == emailInfo['eventID'])
+            emailInfo['programID'] = program.id
```
BrianRamsay: `ProgramEvent` doesn't have an `id` field. I'm guessing this wasn't tested.

Also, events aren't guaranteed to only have one program.

app/logic/emailHandler.py
```diff
@@ -0,0 +1,61 @@
...
+
+def getInterestedEmails(programID = None):
+    """ Gets emails of students interested in a program. """
+    volunteersToEmail = User.select().join(Interest).where(Interest.program == programID)
+    return [user.email for user in volunteersToEmail]
+
+def getParticipantEmails(eventID = None):
+    """ Gets emails of students participating in an event. """
+    volunteersToEmail = User.select().join(EventParticipant).where(EventParticipant.event == eventID)
+    return [user.email for user in volunteersToEmail]
```
BrianRamsay, line 29: This should probably be getting RSVPs rather than participants, right? Unless we're sending the email after the event (so we probably should have a third recipient option)

app/logic/emailHandler.py
BrianRamsay: still relevant. we need more defined return values vs exceptions.

app/controllers/events/email.py
```diff
@@ -0,0 +1,59 @@
...
+
+        email = emailHandler(app,emailInfo)
+        emailSent = email.sendEmail(Message(emailInfo['subject'],
+                                           emails, # recipients
+                                           emailInfo['message']),
+                                           emails) # passed for sending individually
+        if emailSent == 1:
+            flash("Email successfully sent!", "success")
+        else:
+            flash("Error sending email", "danger")
```
BrianRamsay, line 43: It never returns anything other than 1. We need a try/except to handle errors, not return values (based on the current return values)

app/static/js/event_list.js
BrianRamsay: still relevant

app/static/js/event_list.js
```diff
@@ -1,5 +1,42 @@
 
 
+$(document).ready(function(){
+
+  $("#sendEmail").click(function(){
+    console.log($("#emailSender").val())
+    console.log($("#message").val())
```
BrianRamsay: don't need debug output

app/static/js/event_list.js
```diff
@@ -1,5 +1,42 @@
...
+
+function passEventInfo(eventID, programID, selectedTerm){
+  // when the email button for an event is clicked, this passes the eventID and programID to the modal
+  $('#emailModal').modal('show')
+  $(".modal-body #eventID").val(eventID)
+  console.log($(".modal-body #eventID").val())
+  $(".modal-body #programID").val(programID)
+  console.log($(".modal-body #programID").val())
+  $(".modal-body #selectedTerm").val(selectedTerm)
+  console.log($(".modal-body #selectedTerm").val())
```
BrianRamsay: don't need debug output

BrianRamsay (changes requested):

app/logic/emailHandler.py
Sandesh-L: ``mail.send`` doesn't seem to have a return value. I added a try except block. The function will return false if it does run into an error or just return true if there are no errors.

app/logic/emailHandler.py
Sandesh-L: This will be put in as a separate issue.

tests/code/test_emailHandler.py
Sandesh-L: testing for the mail has been added. Used ``Flask_Mail``'s testing functionalities.

app/logic/emailHandler.py
Sandesh-L: only updates the necessary config values now.

app/logic/emailHandler.py
Sandesh-L: This still exists, I couldn't figure out a way to get around it.

app/logic/emailHandler.py
Sandesh-L: removed it

app/config/staging.yml
Sandesh-L: I couldn't think of anything that would be different in staging vs default so I removed the mail info from staging.

app/config/production.yml
Sandesh-L: removed it. Couldn't think of what would be different in here from default so I just removed the whole mail config from ``production.yml`` it will just use what is in default.

app/controllers/events/email.py
Sandesh-L: I added try and except to the email handler so it now returns either True or False.

app/controllers/events/email.py
```diff
@@ -0,0 +1,59 @@
...
+    emailInfo = request.form.copy()
+    if '@' in emailInfo['emailSender']: # if they are using mailto instead
+        return redirect(url_for("main.events", selectedTerm = emailInfo['selectedTerm']))
+    else:
+        if emailInfo['programID'] == 'Unknown':
+            emailInfo['eventID'] = 2
+            programs = ProgramEvent.select(ProgramEvent.program).where(ProgramEvent.event == emailInfo['eventID'])
+            listOfPrograms = [program.program for program in programs.objects()]
+            emailInfo['programID'] = listOfPrograms
+            print("\n\n\n\n\n list of programs:", listOfPrograms, "\n\n\n\n\n\n\n")
```
BrianRamsay, line 25: Debug output

app/controllers/events/email.py
```diff
@@ -0,0 +1,59 @@
...
+
+    emailInfo = request.form.copy()
+    if '@' in emailInfo['emailSender']: # if they are using mailto instead
+        return redirect(url_for("main.events", selectedTerm = emailInfo['selectedTerm']))
+    else:
+        if emailInfo['programID'] == 'Unknown':
+            emailInfo['eventID'] = 2
+            programs = ProgramEvent.select(ProgramEvent.program).where(ProgramEvent.event == emailInfo['eventID'])
+            listOfPrograms = [program.program for program in programs.objects()]
+            emailInfo['programID'] = listOfPrograms
```
BrianRamsay, line 24: The intermediate variable `listOfPrograms` is unnecessary (though doesn't harm anything). I don't think you need .objects() either.

app/controllers/events/email.py
```diff
@@ -0,0 +1,59 @@
...
+
+@events_bp.route('/mailto', methods=['POST'])
+def getEmails(emailGroup, programID=None, eventID=None):
+    """This is where the ajax call should get the mailto info"""
+
+    if emailGroup=='interested' and programID:
+        emails = getInterestedEmails(programID)
+
+    elif emailGroup == 'eventParticipant' and eventID:
+        email = getInterestedEmails(eventID)
```
BrianRamsay, line 54: mismatch between `email` and `emails`. As it is now, `emails` won't be returned

app/logic/emailHandler.py
```diff
@@ -0,0 +1,70 @@
...
+from pathlib import Path
+
+def getInterestedEmails(listOfPrograms):
+    """ Gets emails of students interested in a program. """
+    emails = []
+    print("\n\n\n\n\n",listOfPrograms,"\n\n\n\n\n")
+
+    volunteersToEmail = User.select(User.email).join(Interest).join(Program, on=(Program.id == Interest.program)).where(Program.id.in_([p.id for p in listOfPrograms]))
+    print("\n\n\n\n\n",volunteersToEmail,"\n\n\n\n\n")
+    print("\n\n\n\n\n", volunteersToEmail.objects(),"\n\n\n\n\n")
```
BrianRamsay, line 22: more debug output to remove

app/logic/emailHandler.py
```diff
@@ -0,0 +1,70 @@
...
+import webbrowser
+import time
+from pathlib import Path
+
+def getInterestedEmails(listOfPrograms):
+    """ Gets emails of students interested in a program. """
+    emails = []
+    print("\n\n\n\n\n",listOfPrograms,"\n\n\n\n\n")
+
+    volunteersToEmail = User.select(User.email).join(Interest).join(Program, on=(Program.id == Interest.program)).where(Program.id.in_([p.id for p in listOfPrograms]))
```
BrianRamsay, line 20: This line needs to be split up

app/templates/emailModal.html
```diff
@@ -0,0 +1,66 @@
+<script src="/static/js/event_list.js"></script>
+<div class="modal fade" id="emailModal" tabindex="-1" aria-labelledby="emailModalLabel" aria-hidden="true">
+  <div class="modal-dialog modal-dialog-scrollable modal-lg">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h4 class="modal-title m-auto" id="emailModalLabel">'Event/Program' Email Template</h4>
+        <button type="button" class="btn-close position-absolute end-0 me-3" data-bs-dismiss="modal" aria-label="Close"></button>
+      </div>
+      <div class="modal-body">
+        <h6 class="text-center">Last Email was sent to ____ on 10/20/20</h6>
```
BrianRamsay, line 10: We should either remove this, implement it, or indicate that it is a nonfunctioning placeholder

BrianRamsay (changes requested):

app/logic/emailHandler.py
```diff
@@ -0,0 +1,70 @@
...
+from app import app
+import sys
+import webbrowser
+import time
+from pathlib import Path
+
+def getInterestedEmails(listOfPrograms):
+    """ Gets emails of students interested in a program. """
+    emails = []
+    print("\n\n\n\n\n",listOfPrograms,"\n\n\n\n\n")
```
BrianRamsay, line 18: more debug

Issue #117, opened 2021-09-13
CELTS Admin Manage Service Learning Courses


bryantal:
SRS Item: 5.6 - The CELTS administrator must be able to manage SL Course Proposals

This will be the table where CELTS admins can view and manage all service learning course proposals for the upcoming term.
![image](https://user-images.githubusercontent.com/28715731/133150309-7879fac4-5a3d-4a05-9826-3f99deaa49bf.png)

CELTS SRS: https://docs.google.com/document/d/17URsAbRVdhJ9lBABzlM3Zk7KKbmuD2m9aYMT3tG2lkw/edit
CELTS Admin Wireframe: https://balsamiq-wireframes.appspot.com/?state=%7B%22ids%22:%5B%221etrefTLsvWQntesh5_TDdC66TRVzRpHf%22%5D,%22action%22:%22open%22,%22userId%22:%22109665407847548855125%22,%22resourceKeys%22:%7B%7D%7D

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #117: CELTS Admin Manage Service Learning Courses](https://trello.com/c/RyLfxymg/313-celts-117-celts-admin-manage-service-learning-courses)

mupotsal:
@Sandesh-L How is this issue different from what you have been working on?

Sandesh-L:
@mupotsal We only implemented the remove course option. None of the other ones. This looks to be the admin side of the sl courses. We only built the view for the professors. The PR was merged so you can check app/templates/admin/createSLProposalTable.html for the html that we built

PR #136, opened 2021-09-30
Slc admin view issue117

tylerpar99:
# This PR fixes issue #117 

## Note: Left sent section of approved course blank as we are waiting on a decision to be made about how we are going to store the information regarding what emails have been sent.


app/controllers/admin/routes.py
```diff
@@ -174,3 +175,28 @@ def addParticipants():
...
+    '''
+    Renders the page for admins to manage Course Proposals
+    '''
+    currentTerm = Term.get(Term.isCurrentTerm)
+    if termId == None:
+        termId = currentTerm
+    else:
+        termId = (int(termId))
+
+    print(type(termId))
```
BrianRamsay: debug print

app/controllers/admin/routes.py
```diff
@@ -174,3 +175,28 @@ def addParticipants():
...
+    pending = pendingCourses(termId)
+    approve = approveCourses(termId)
+    print(termId, "================")
```
BrianRamsay: debug print

app/controllers/admin/routes.py
```diff
@@ -174,3 +175,28 @@ def addParticipants():
...
     return render_template('addParticipants.html',
                             title="Add Participants")
+
+@admin_bp.route('/courseManagement', methods = ['GET', 'POST'])
+@admin_bp.route('/<termId>/courseManagement', methods = ['GET', 'POST'])
+def courseManagement(termId = None):
+    '''
+    Renders the page for admins to manage Course Proposals
+    '''
+    currentTerm = Term.get(Term.isCurrentTerm)
```
BrianRamsay: there is already a global to get the current term `g.current_term`

app/controllers/admin/routes.py
```diff
@@ -174,3 +175,28 @@ def addParticipants():
...
+    if termId == None:
+        termId = currentTerm
+    else:
+        termId = (int(termId))
```
BrianRamsay: you should validate termId by getting a Term object at the beginning of the controller. Then you can use that Term object later

app/static/js/courseManagement.js
```diff
@@ -0,0 +1,19 @@
+function changeTerm() {
+  var termId = $('option:selected', selectTerm).attr('value');
+  console.log("lalalalalalal", termId);
```
BrianRamsay: debug output

app/static/js/courseManagement.js
```diff
@@ -0,0 +1,19 @@
...
+    type: "POST",
+    url: "/courseManagement" + "/" ,
+    data: JSON.stringify(termId),
+    contentType: "application/json",
+    success: function(response){
+
+      }
+    })
+    $('form').submit();
+};
```
BrianRamsay, line 4: I'm confused about this ajax call that does nothing on success, and this form submission that doesn't wait for the ajax to finish

tests/code/test_course_management.py
```diff
@@ -0,0 +1,49 @@
...
+                                        isAllSectionsServiceLearning = 0,
+                                        isPermanentlyDesignated = 0)
+
+        testingCourse = Course.create(courseName = "Testing Pending",
+                                        term = 3,
+                                        status = 2,
+                                        courseCredit = "12",
+                                        createdBy = "Tyler Parton",
+                                        isAllSectionsServiceLearning = 0,
+                                        isPermanentlyDesignated = 0)
```
BrianRamsay, line 25: Why not add these to the test data, instead of specifying it all here?

BrianRamsay: Also, the test fails for me. `/Users/ramsayb2/dev/celts/tests/code/test_course_management.py:46: AssertionError: assert 'Testing Approved' in []`

BrianRamsay (changes requested):

tests/code/test_course_management.py
tylerpar99: I ran the test and It passes for me, not sure if there is a difference in test data causing it, if so I don't know what that would be. Regarding the reason we added the creations there, the idea was regardless of how someone manipulated the test data out test would pass. If we added it to test data the issue is someone can go in and change it which would fail our test. We figured the best way to rectify this was to have it create and delete the event within the test and regardless of what someone had in their own branch it would pass. @BrianRamsay

app/controllers/admin/routes.py
```diff
@@ -174,3 +175,26 @@ def addParticipants():
...
+@admin_bp.route('/<termId>/courseManagement', methods = ['GET', 'POST'])
+def courseManagement(termId = None):
+    '''
+    Renders the page for admins to manage Course Proposals
+    '''
+    currentTerm = g.current_term
+    if termId == None:
+        termId = currentTerm
+    else:
+        termId = Term.get(Term.id == termId)
```
BrianRamsay: These lines can be made a bit simpler, in addition to handling the case where an invalid termId came in.

```
term = Term.get_or_none(Term.id == termId)
if not term:
    term = g.current_term
```

tests/code/test_course_management.py
BrianRamsay: A good plan. I think it does make sense to have pure test data connected to the test

app/logic/courseManagement.py
```diff
@@ -0,0 +1,34 @@
...
+    '''
+    Queries the database to get all the neccessary information for
+    pending courses.
+    '''
+
+    approveCourses = (CourseInstructor.select(Course.courseName, Term.description, CourseInstructor.user, Course.status)
+                    .join(Course)
+                    .join(CourseStatus)
+                    .switch(Course)
+                    .join(Term)).where(CourseInstructor.course.term.id == termId, Course.status.status == "Approve")
```
BrianRamsay: Shouldn't this status be "Approved" instead of "Approve"? Or does it mean something other than that this course is approved?

BrianRamsay: And if it should be approved, that language should change everywhere (including variable and method names)

app/controllers/admin/routes.py
BrianRamsay: Notice that instead of calling a variable `termId`, it's called `term`, because that's what it is. A `Term` object, not a numeric ID.

app/templates/admin/courseManagement.html
```diff
@@ -0,0 +1,87 @@
...
+
+{% block app_content %}
+
+<div id="content">
+  <h1 class="text-center pt-1 pb-4">Service Learning Course Proposal Management</h1>
+  <div class="text-center">
+    <form id="termSelector" action="" method="post">
+      <select class="w-25" id="selectTerm" onchange="formSubmit(this.value)">
+          {% for term in terms %}
+            <option value="{{term.id}}" {{ "selected" if term.id == termId }}> {{ term.description }}</option>
```
BrianRamsay: Does this actually work? Because `termId` is an object here, not a number. If it does work, you're just getting lucky that the object is smart enough to convert itself to a number for the operation (or rather, that it has implemented the equality operator with an integer by comparing to its ID)

BrianRamsay (changes requested):

app/controllers/admin/routes.py
```diff
@@ -160,3 +158,27 @@ def addParticipants():
 
     return render_template('addParticipants.html',
                             title="Add Participants")
+
+@admin_bp.route('/courseManagement', methods = ['GET', 'POST'])
+@admin_bp.route('/<term>/courseManagement', methods = ['GET', 'POST'])
```
BrianRamsay: I'd flip that around so the url is `/courseManagement/<term>`

app/controllers/admin/routes.py
```diff
@@ -160,3 +158,27 @@ def addParticipants():
...
+    Renders the page for admins to manage Course Proposals
+    '''
+
+    term = Term.get_or_none(Term.id == term)
+    if not term:
+        term = g.current_term
+
+
+    pending, pendingCourseInstructor = pendingCourses(term)
+    approved, approvedCourseInstructor = approvedCourses(term)
```
BrianRamsay: These return values should be combined. One is a list of course ids, and the other is a dictionary with keys that are course ids. That's the same data. If you return just the dictionary you have the data you need

app/controllers/admin/routes.py
```diff
@@ -160,3 +158,27 @@ def addParticipants():
...
+    terms = selectFutureTerms(g.current_term)
```
BrianRamsay: This function is gone now

app/logic/courseManagement.py
```diff
@@ -0,0 +1,56 @@
...
+from app.models.course import Course
+from app.models.term import Term
+
+
+def pendingCourses(termId):
+    '''
+    Queries the database to get all the neccessary information for
+    non approved and non completed courses.
+    '''
+
```
BrianRamsay, line 12: You should be able to get course and instructor info in one query. An alternate option would be to use the courseInstructors backreference that was added to Course

BrianRamsay (changes requested):
There are some changes that can be made to this PR in light of other changes that have gone in since (like the courseInstructors backreference)

BrianRamsay (approved):

PR #145, opened 2021-10-10
Faculty service learning proposal issue#76

### LINKED ISSUE #76: Faculty Create Service Learning Course Proposals
This issue covers SRS item 5.1 - The faculty must be able to create SL Course Proposals (From the [CELTS SRS](https://docs.google.com/document/d/17URsAbRVdhJ9lBABzlM3Zk7KKbmuD2m9aYMT3tG2lkw/edit))

The faculty will need to have functionality to create service learning course proposals through a page like this one:
image

See Faculty UI Wireframe: https://balsamiq-wireframes.appspot.com/?state=%7B%22ids%22:%5B%221ziGyTzz6J7iHjyzmsUzj0yzvuFdITg3n%22%5D,%22action%22:%22open%22,%22userId%22:%22109665407847548855125%22,%22resourceKeys%22:%7B%7D%7D

hmanalai:
* Fixes issue #76 
* This PR creates new pages for faculty service-learning proposals using [this wireframe](https://balsamiq-wireframes.appspot.com/?state=%7B%22ids%22:%5B%221ziGyTzz6J7iHjyzmsUzj0yzvuFdITg3n%22%5D,%22action%22:%22open%22,%22userId%22:%22109665407847548855125%22,%22resourceKeys%22:%7B%7D%7D)


app/templates/serviceLearning/slcQuestionnaire.html
```diff
@@ -0,0 +1,46 @@
+{% extends "base.html" %}
+{% block scripts %}
+
+  {{super()}}
+  <style media="screen"></style>
```
BrianRamsay: You're including a style comment in the scripts block. There's another block for that, right?

app/controllers/serviceLearning/routes.py
```diff
@@ -0,0 +1,99 @@
...
+    if request.method == "POST":
+        courseData["courseName"] = request.form.get("courseName")
+        courseData["courseAbbreviation"] = request.form.get("courseAbbreviation")
+        courseData["courseCredit"] = request.form.get("credit")
+        courseData["courseInstructorPhone"] = request.form.get("courseInstructorPhone")
+        courseData["regularOccurenceToggle"] = request.form.get("regularOccurenceToggle")
+        courseData["termId"] = request.form.get("term")
+        courseData["slSectionsToggle"] = request.form.get("slSectionsToggle")
+        courseData["slDesignation"] = request.form.get("slDesignation")
+        courseData["permanentDesignation"] = request.form.get("permanentDesignation")
```
BrianRamsay: Nothing is done with this data structure

app/controllers/serviceLearning/routes.py
```diff
@@ -0,0 +1,99 @@
...
+        return redirect("/slcQuestionnaire")
+
+    terms = Term.select()
+    return render_template('serviceLearning/slcProposal.html', terms=terms)
+
+@serviceLearning_bp.route('/slcQuestionnaire', methods=['GET', 'POST'])
+def slcQuestionnaire():
+    """ This page renders slc questionnare """
+    if request.method == "POST":
+        term = Term.get(Term.id == courseData["termId"])
```
BrianRamsay: courseData can't come from that other controller method, right?

BrianRamsay (changes requested):
Starting to look good, but I'm thinking there's more to work on here, right?

BrianRamsay:
A comment I forgot to mention - let's structure the URLs a bit with a prefix, like `/slc/` or `/servicelearning/`

app/controllers/serviceLearning/routes.py
```diff
@@ -0,0 +1,76 @@
...
+
+@serviceLearning_bp.route('/serviceCourseManagement', methods = ['GET'])
+def serviceCourseManagement():
+    """This is a Temporary Page for the Service Course Managment Screen."""
+    courseDict = getServiceLearningCoursesData(g.current_user)
+    return render_template('serviceLearning/slcManagment.html',
+        instructor=g.current_user,
+        courseDict=courseDict)
+
+@serviceLearning_bp.route('/slcNewProposal', methods=['GET', 'POST'])
```
BrianRamsay: I think you saw my comment before and thought I was referring to the templates, but we should be consistent about all of these URLs within Service Learning. They should all have the same format, e.g., `/slc/ROUTE` or `/servicelearning/ROUTE`, so users know where they are in the application. For the templates actually it isn't so necessary, since they are all in the servicelearning directory already.

app/static/js/slcNewProposal.js
```diff
@@ -0,0 +1,113 @@
...
+  $("#instructorTable tr").each(function(a, b) {
+    courseInstructors.push($('.instructorName', b).text());
+  });
+  return $.ajax({
+    url: "/courseInstructors",
+    data: JSON.stringify(courseInstructors),
+    type: "POST",
+    contentType: "application/json",
+    success: function () {
+      console.log("success");
```
BrianRamsay: don't need success debug output

app/controllers/serviceLearning/routes.py
```diff
@@ -0,0 +1,76 @@
...
+
+    terms = Term.select()
+    return render_template('serviceLearning/slcNewProposal.html', terms=terms)
+
+instructorsDict = {}
+@serviceLearning_bp.route('/courseInstructors', methods=['POST'])
+def getInstructors():
+    instructorObjectList = []
+    instructorsList = request.get_json()
+    for instructor in instructorsList:
```
BrianRamsay: I would avoid using `instructor` twice in this block. Sometimes it's not a big deal, but I think indicating that the first one is unprocessed would be good, like `rawInstructor`.

BrianRamsay (changes requested):
UI Notes on the new proposal form:

The Instructor search should only search faculty (and you could fix the currently broken logic while you are in there, or put that in as a separate ticket).
There is still default placeholder text.
The term dropdown should exclude past terms.
I think the 3-dot step indicator should be in a line with the buttons.
What do you think about putting the buttons and step indicator at the top, so they don't move? Or at top and bottom?

BrianRamsay:
Oops, I just noticed it was marked as a Draft. If the review comments are not relevant, go ahead and ignore them.

BrianRamsay (approved):

PR #146, opened 2021-10-12
Admin Management Page

sreynit02:
Issue: There needs to be an Admin Management page that allows the admin to add users as Celts Admin or Celts Student Staff. Admin should be able to remove a user from being a Celts Admin or Celts Student Staff. Only Admin has access to this page.
Admin Management Page
Test: 
1. Add a valid user who's not a Celts Admin as a Celts Admin. After clicking **Add**, a success flash message should pop up saying they are a Celts Admin now. In the database, user.isCeltsAdmin = 1. Then, remove admin privilege from them; there should also be a green flash message and user.isCeltsAdmin = 0.
2.  Add a valid user who's not a Celts Student Staff as a Celts Student Staff. After clicking **Add**, a success flash message should pop up saying they are a Celts Student Satff now. n the database, user.isCeltsStudentStaff = 1. Then, remove Celts Student Satff privilege from that same user. A green flash message should appear and user.isCeltsStudentStaff = 0. 
3. reset database and run tests; all the tests should pass.


app/controllers/admin/userManagement.py
```diff
@@ -0,0 +1,34 @@
+from flask import Flask, render_template,request, flash, g, abort
+from app.models.user import User
+from app.controllers.admin import admin_bp
+from app.logic.userManagement import addCeltsAdmin,addCeltsStudentStaff,removeCeltsAdmin,removeCeltsStudentStaff
+import re
```
BrianRamsay: I like to organize the import statements. python standard library imports should go at the top

app/controllers/admin/userManagement.py
```diff
@@ -0,0 +1,34 @@
...
+        addCeltsStudentStaff(user)
+        flash(username+ " has been added as a Celts Student Staff", 'success')
+    elif method == "removeCeltsAdmin":
+        removeCeltsAdmin(user)
+        flash(username+ " is no longer a Celts Admin ", 'success')
+    elif method == "removeCeltsStudentStaff":
+        removeCeltsStudentStaff(user)
+        flash(username+ " is no longer a Celts Student Staff", 'success')
+
+    return ("sucess")
```
BrianRamsay: success is misspelled

app/logic/userManagement.py
```diff
@@ -0,0 +1,25 @@
+from app.models.user import User
+
+def addCeltsAdmin(user):
+    user = User.get_by_id(user)
+    userSelect = User.get(User.username == user.username)
```
BrianRamsay: I'm curious what you are trying to do here. What is wrong with the user object that you get in line 4, that you need to get the same one again in line 5?

app/static/js/searchStudent.js
```diff
@@ -5,5 +5,6 @@ function callback() {
 }
 
 $("#searchStudentsInput").on("input", function() {
+  console.log("The function is called");
```
BrianRamsay: remove debug output

app/static/js/userManagement.js
```diff
@@ -0,0 +1,133 @@
...
+    data: data,
+    success: function(s){
+        location.reload()
+    },
+    error: function(error, status){
+        console.log(error, status)
+    }
+
+  })
+}
```
BrianRamsay, line 67: Way too much mostly duplicated code. All of these dropdowns are essentially the same, you should use the same code to manage them.

app/templates/admin/userManagement.html
```diff
@@ -0,0 +1,63 @@
...
+      <div class="input-group">
+        <input type="input" id="removeCeltsStudentStaffInput" name="removeCeltsStudentStaffInput" class="form-control" placeholder="Search" oninput="" autocomplete="off"/>
+        <button type="button" class="btn btn-secondary btn-sm" disabled>
+          <span class="bi bi-search"></span>
+        </button>
+      </div>
+      <div class="col-sd-3"><button type="button" class="btn btn-danger btn-sm" id="removeCeltsStudentStaff">Remove</button></div>
+  </div>
+</div>
+</div>
```
BrianRamsay, line 43: I would even suggest generating the html in a loop (possibly), or at least using a macro to generate the inputs and buttons.

BrianRamsay (changes requested):

BrianRamsay:
Forgot to mention also, the page doesn't scale well when the screen-size shrinks. The space in the middle of the page doesn't shrink while the inputs do.

app/controllers/admin/userManagement.py
mupotsal: fixed

app/controllers/admin/userManagement.py
mupotsal: fixed

app/logic/userManagement.py
mupotsal: We removed it. It was redundant indeed!

app/static/js/userManagement.js
mupotsal: Refactored the code

app/templates/admin/userManagement.html
mupotsal: Created Macro that generates buttons and inputs.

mupotsal: > Forgot to mention also, the page doesn't scale well when the screen-size shrinks. The space in the middle of the page doesn't shrink while the inputs do.

We have resolved the screen shrink issue.

BrianRamsay (approved):
Good changes! I'm merging the branch so we have it ready to go, and we'll put in a new ticket to fix the whitespace issue when the page resizes.

PR #147, opened 2021-10-14
Fixes Remove Volunteer Functionality in Track Volunteer Page

### LINKED ISSUE #144: Remove Volunteer from event functionality not working.
**Issue**: 
There is an issue with using modules in javascript files. Functions from javascript files imported to HTML with type = "module" can not be accessed.
Screen Shot 2021-10-07 at 5 34 56 PM
 For instance, in trackVolunteers.js functions called _removeVolunteerFromEvent()_ and  _toggleVolunteersInputBox()_ can not be accessed. 

**Reproduction:** 
Navigate to _/event/7/track_volunteers_ try to delete one of the participants there. You will not be able to do so. If you open the console, you get an error called 
Screen Shot 2021-10-07 at 5 34 06 PM

**Notes**

You will notice that if you remove _module_ as the type, the functions can be executed but the search functionality breaks.

hmanalai:
* Fixes issue #144 
* Summary: The javascript was changed to a module that caused some function calls to break because they were no longer accessible through HTML. This PR fixes that issue by creating event listeners in the javascript file. It also fixes some querying issues in the python file that failed to retrieve user and event data.

### DIFF
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -8,7 +8,7 @@ from app.models.user import User
 from app.models.eventParticipant import EventParticipant
 from app.logic.searchUsers import searchUsers
 from app.logic.volunteers import updateVolunteers, addVolunteerToEvent, getEventLengthInHours
-from app.logic.participants import trainedParticipants
+from app.logic.participants import trainedParticipants, getPresentParticipants
 from app.models.user import User
 from app.models.eventRsvp import EventRsvp
 
@@ -21,7 +21,6 @@ def getVolunteers(query):
 
 @admin_bp.route('/event/<eventID>/track_volunteers', methods=['GET'])
 def trackVolunteersPage(eventID):
-
     try:
         event = Event.get_by_id(eventID)
     except DoesNotExist as e:
@@ -34,30 +33,31 @@ def trackVolunteersPage(eventID):
     if not program:
         return "TODO: What do we do for no programs or multiple programs?"
 
-    attendedTraining = trainedParticipants(program)
+    trainedParticipantsList = trainedParticipants(program)
+    presentParticipants = getPresentParticipants(event)
+
     if not g.current_user.isCeltsAdmin:
         abort(403)
 
-    eventParticipantsData = (EventParticipant
+    eventRsvpData = (EventRsvp
         .select()
-        .where(EventParticipant.event==event))
-
-    eventRsvpUsers = (User
-        .select(User, EventRsvp)
-        .join(EventRsvp)
         .where(EventRsvp.event==event))
 
-    eventLengthInHours = getEventLengthInHours(event.timeStart, event.timeEnd,  event.startDate)
+    eventLengthInHours = getEventLengthInHours(
+        event.timeStart,
+        event.timeEnd,
+        event.startDate)
+
     isPastEvent = (datetime.now() >= datetime.combine(event.startDate, event.timeStart))
 
     return render_template("/events/trackVolunteers.html",
-                            eventParticipantsData=list(eventParticipantsData),
-                            eventRsvpUsers=eventRsvpUsers,
-                            eventLength=eventLengthInHours,
-                            program=program,
-                            event=event,
-                            isPastEvent=isPastEvent,
-                            attendedTraining=attendedTraining)
+        eventRsvpData=list(eventRsvpData),
+        presentParticipants=presentParticipants,
+        eventLength=eventLengthInHours,
+        program=program,
+        event=event,
+        isPastEvent=isPastEvent,
+        trainedParticipantsList=trainedParticipantsList)
 
 @admin_bp.route('/event/<eventID>/track_volunteers', methods=['POST'])
 def updateVolunteerTable(eventID):
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -8,7 +8,7 @@ from app.models.user import User
 from app.models.eventParticipant import EventParticipant
 from app.logic.searchUsers import searchUsers
 from app.logic.volunteers import updateVolunteers, addVolunteerToEvent, getEventLengthInHours
-from app.logic.participants import trainedParticipants, getPresentParticipants
+from app.logic.participants import trainedParticipants, getEventParticipants
 from app.models.user import User
 from app.models.eventRsvp import EventRsvp
 
@@ -34,7 +34,7 @@ def trackVolunteersPage(eventID):
         return "TODO: What do we do for no programs or multiple programs?"
 
     trainedParticipantsList = trainedParticipants(program)
-    presentParticipants = getPresentParticipants(event)
+    eventParticipants = getEventParticipants(event)
 
     if not g.current_user.isCeltsAdmin:
         abort(403)
@@ -52,7 +52,7 @@ def trackVolunteersPage(eventID):
 
     return render_template("/events/trackVolunteers.html",
         eventRsvpData=list(eventRsvpData),
-        presentParticipants=presentParticipants,
+        eventParticipants=eventParticipants,
         eventLength=eventLengthInHours,
         program=program,
         event=event,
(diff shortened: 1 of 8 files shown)


app/controllers/admin/volunteers.py
```diff
@@ -87,12 +89,10 @@ def addVolunteer(volunteer, volunteerEventID, eventLengthInHours):
...
-
+    return jsonify({"Success": True}), 200
 
 @admin_bp.route('/removeVolunteerFromEvent/<user>/<eventID>', methods = ['POST'])
 def removeVolunteerFromEvent(user, eventID):
     (EventParticipant.delete().where(EventParticipant.user == user, EventParticipant.event == eventID)).execute()
-    flash("Volunteer successfully removed")
-    return "Volunteer successfully removed"
+    flash("Volunteer successfully removed", "success")
+    return jsonify({"Success": True}), 200
```
BrianRamsay: let's not do the `"Success" : True` thing if we're not actually checking the returned data for success or failure

app/logic/participants.py
```diff
@@ -82,3 +82,16 @@ def unattendedRequiredEvents(program, user):
...
+
+def getPresentParticipants(event):
+        presentParticipantsDict = {}
+        presentParticipants = (EventParticipant
+            .select()
+            .where(EventParticipant.event==event)
+            .filter(EventParticipant.attended==1))
+
+        for participant in presentParticipants:
+            presentParticipantsDict[participant.user.username] = participant.hoursEarned
```
BrianRamsay: You don't need to change it if you don't want to since the current code is clear, but you could replace everything but the query in this function with a dictionary comprehension: `return {p.user.username: p.hoursEarned for p in presentParticipants}`

BrianRamsay (changes requested):
I don't see a test for getPresentParticipants

app/controllers/admin/volunteers.py
```diff
@@ -67,32 +72,27 @@ def updateVolunteerTable(eventID):
...
-def addVolunteer(volunteer, volunteerEventID, eventLengthInHours):
-    volunteer = volunteer.strip("()")
-    username = volunteer.split('(')[-1]
-    user = User.get(User.username == username)
-    succesfullyAddedVolunteer = addVolunteerToEvent(user, volunteerEventID, eventLengthInHours)
+@admin_bp.route('/addVolunteerToEvent/<volunteer>/<eventId>', methods = ['POST'])
+def addVolunteer(volunteer, eventId):
+    username = volunteer.strip("()").split('(')[-1]
+    user = User.get(User.username==username)
+    succesfullyAddedVolunteer = addVolunteerToEvent(user, eventId)
```
BrianRamsay: `succesfullyAddedVolunteer` is spelled incorrectly

app/logic/volunteers.py
```diff
@@ -19,45 +21,67 @@ def getEventLengthInHours(startTime, endTime, eventDate):
...
-                    if participantData['checkbox_'+ username] == "on": #if the user is marked as present
-                        (EventParticipant.update({EventParticipant.hoursEarned: float(participantData['inputHours_'+ username]),EventParticipant.attended: True}).where(EventParticipant.event == participantData['event'],
-                                                 EventParticipant.user == participantData[f'username{user}'])).execute()
-
+                    if participantData[f'checkbox_{username}']: #if the user is marked as present
+                        hoursEarned = float(participantData['inputHours_'+ username])
+                        if eventParticipant:
+                            ((EventParticipant
+                                .update({
+                                    EventParticipant.hoursEarned: hoursEarned})
```
BrianRamsay: Where to line-break is often a personal preference thing, but I think this one makes it harder to read. I'd keep the dictionary all on one line, since there is only one key:value pair

app/logic/volunteers.py
```diff
@@ -19,45 +21,67 @@ def getEventLengthInHours(startTime, endTime, eventDate):
...
+                                hoursEarned=0))
             else:
                 return False
         else:
             break
     return True
 
 
-def addVolunteerToEvent(user, volunteerEventID, eventLengthInHours):
+def addVolunteerToEvent(user, volunteerEventID):
```
BrianRamsay: We should clarify this method name now that the semantics have changed. It's not clear from the name whether it is messing with participant or rsvp

BrianRamsay: Also, it seems like this change should have broken a test, but there isn't a corresponding test change. If it's missing a test, you should add it.

app/logic/volunteers.py
```diff
@@ -19,45 +21,67 @@ def getEventLengthInHours(startTime, endTime, eventDate):
 
 def updateVolunteers(participantData):
```
BrianRamsay: This method should definitely have a test.

BrianRamsay (changes requested):

tests/code/test_volunteers.py
```diff
@@ -45,47 +46,49 @@ def test_getEventLengthInHours():
 
 
 @pytest.mark.integration
-def test_addVolunteerToEvent():
+def test_addVolunteerToEventRsvp():
```
BrianRamsay, line 49: This test and `test_updateEventParticipants` both test the return value of the function, but don't actually check that the function did what it was supposed to do. Each one should have a check to confirm that the record is properly in the database when the function returns true.

BrianRamsay (changes requested):

BrianRamsay:
Also, there is a conflicted file to resolve.

tests/code/test_volunteers.py
hmanalai: I wanted to ask about this and whether we should change the asserts. Thanks! we will make the change

hmanalai:
@BrianRamsay We updated the PR with the new changes. It is ready to merge. 
I also wanted to note that, in this PR we have left the `attended` column in the `EventParticipant` table deleting it required modifying other files and the PR would become bigger than we wanted. 

We are creating a new issue for that and will put in a separate PR.

app/logic/volunteers.py
```diff
@@ -17,47 +19,61 @@ def getEventLengthInHours(startTime, endTime, eventDate):
...
+def updateEventParticipants(participantData):
     """
-    updates the events participant table with data from the track Volunteer webpage
+    Create new entry in event participant table if user does not exist. Otherwise, updates the record.
 
     param: participantData- a dictionary that contains data from every row of the page along with the associated username.
     """
-
+    event = Event.get_or_none(Event.id==participantData['event'])
+    eventId = participantData['event']
```
BrianRamsay: I would clean up your usage of the passed in event. 

You get the event object on line 28, but don't check that it is valid. You should validate that the event exists (aborting with the appropriate error), and then once it is valid use it everywhere you need an id, which would mean using `event.id` everywhere in this function in place of `eventId` and `participantData['event']`.

The general principle here is that it is better to use your own cleaned or validated or processed data than direct user input, so once you've gone through the trouble of getting the event object you might as well use it everywhere. It isolates the form field name as well, so you only reference the 'event' in the form once.

BrianRamsay (changes requested):
One other thing that I noticed.

app/logic/volunteers.py
```diff
@@ -17,47 +19,64 @@ def getEventLengthInHours(startTime, endTime, eventDate):
...
+def updateEventParticipants(participantData):
     """
-    updates the events participant table with data from the track Volunteer webpage
+    Create new entry in event participant table if user does not exist. Otherwise, updates the record.
 
     param: participantData- a dictionary that contains data from every row of the page along with the associated username.
     """
+    event = Event.get_or_none(Event.id==participantData['event'])
+    if not event:
+        raise Exception("Event does not exist.") # ???
```
hmanalai, line 30: I wasn't sure what error I should raise, so I did this. Do u have another suggestion @BrianRamsay ?

app/logic/volunteers.py
hmanalai: Made these changes.

app/logic/volunteers.py
BrianRamsay: I think that's fine. There are a few different ways to get to a 500 error, and this is one of them

BrianRamsay (approved):

BrianRamsay:
Looks good!

Issue #129, opened 2021-09-24
Term Management Page


bryantal:
CELTS Admin Wireframe: https://balsamiq-wireframes.appspot.com/?state={%22action%22:%22open%22,%22ids%22:[%221etrefTLsvWQntesh5_TDdC66TRVzRpHf%22],%22userId%22:%22109665407847548855125%22}

We do not have a way to set the current term in CELTS.  We would like to have an "Admin Management" or "Admin Roles" tab in the sidebar that will lead to this page.  This page populate with all the terms for the year listed out, and admins will be able to choose to set one as the current term
![image](https://user-images.githubusercontent.com/28715731/134743399-0c737052-3c52-4b2c-b321-c7aeeb84b887.png)
The bottom half of this page is what we want for this, and it will be combined with the user management page later on.

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #129: Term Management Page](https://trello.com/c/KYSzgxKr/318-celts-129-term-management-page)

PR #148, opened 2021-10-15
Changes Current Term in Admin Management

zachneill:
This is a PR for Issue #129 . I forked off branch admin_management which is on PR #146 .

This issue is regarding adding a way to set the current term. It is supposed to be on the admin management page. 

I created a jinja for loop that adds a button for each term in the database. When clicked on, the JS in userManagement.js makes the clicked-on button active and the rest inactive. I used an Ajax call to send data from a hidden tag to the userManagement.py controller file. It calls changeCurrentTerm(term) in the userManagement.py logic file. changeCurrentTerm(term) updates the g.current_term's isCurrentTerm key to false and get_by_id(term)'s isCurrentTerm to true. If successful, the page reloads and a flasher appears.

TO TEST:
Go to YOUR_IP_ADDRESS/userManagement
Select a different term. 
Hit "Make Current Term"
The current term should change to the term you chose. 

Additional tests are in test_userManagement.py

NOTE: I asked Alex, and he okayed the implementation of buttons as opposed to a dropdown, but said to make note of it for the future.


app/logic/changeCurrentTerm.py
```diff
@@ -0,0 +1,2 @@
+def changeTerm():
+    pass
```
BrianRamsay: This file isn't needed, apparently

app/controllers/admin/userManagement.py
```diff
@@ -0,0 +1,52 @@
...
+        flash(username+ " is no longer a Celts Admin ", 'success')
+    elif method == "removeCeltsStudentStaff":
+        removeCeltsStudentStaff(user)
+        flash(username+ " is no longer a Celts Student Staff", 'success')
+
+    return ("success")
+
+@admin_bp.route('/userManagement', methods = ['GET'])
+def userManagement():
+    terms = Term.select()
```
BrianRamsay: We don't want to be sending every single term for choosing the current term. That list will be huge eventually. I would suggest using the existing getFutureTerms method, or something similar, where you get one or two of the previous terms and then the next few.

app/controllers/admin/userManagement.py
```diff
@@ -0,0 +1,52 @@
...
+    totalTerms = len([term.id for term in terms])
+    if g.current_user.isAdmin:
+        return render_template('admin/userManagement.html',
+                                terms=terms,
+                                selectedTerm = Term.get_by_id(g.current_term),
```
BrianRamsay: g.current_term is already available to the template

app/controllers/admin/userManagement.py
```diff
@@ -0,0 +1,52 @@
...
+    elif method == "removeCeltsStudentStaff":
+        removeCeltsStudentStaff(user)
+        flash(username+ " is no longer a Celts Student Staff", 'success')
+
+    return ("success")
+
+@admin_bp.route('/userManagement', methods = ['GET'])
+def userManagement():
+    terms = Term.select()
+    totalTerms = len([term.id for term in terms])
```
BrianRamsay: you can just do `len(terms)` here. But... it's also not necessary to pass `totalTerms` at all. jinja can do a count filter to get the number of terms, like {{terms|count}}.

app/templates/admin/userManagement.html
```diff
@@ -0,0 +1,65 @@
...
+          value='{{ term.id }}'>{{term.description}}</button>
+        </div>
+        {% endfor %}
+        <button class="btn btn-primary my-2 float-end"
+        id="submitButton" role="button">Make Current Term</button>
+      </div>
+    </div>
+  </div>
+  <input type="hidden" value="{{selectedTerm}}" id="termInput">
+  <input type="hidden" value="{{totalTerms}}" id="totalTerms">
```
BrianRamsay: Rather than have to store the total terms to use later, you could just count the number of buttons in #currentTermList. That's pretty easy to do in js, and you don't have to manage a hidden field

tests/code/test_userManagement.py
```diff
@@ -0,0 +1,51 @@
...
+    with pytest.raises(DoesNotExist):
+        removeCeltsStudentStaff("1234")
+
+def test_changeCurrentTerm():
+    # test via g.current_term
+    changeCurrentTerm(2)
+    assert g.current_term == 2
+    # test via isCurrentTerm
+    newTerm = changeCurrentTerm(1)
+    assert newTerm.isCurrentTerm
```
BrianRamsay, line 49: I'd also test that g.current_term is different after the second call, and that the old current term doesn't have isCurrentTerm = True anymore.

BrianRamsay (changes requested):

app/logic/changeCurrentTerm.py
zachneill: File has been deleted

app/controllers/admin/userManagement.py
zachneill: Updated code to reflect this

app/controllers/admin/userManagement.py
zachneill: uses selectFutureTerms now

app/controllers/admin/userManagement.py
zachneill: fixed. uses loop.index also

app/templates/admin/userManagement.html
zachneill: removed hidden field(s), works

tests/code/test_userManagement.py
zachneill: tests changed/added

app/logic/utils.py
```diff
@@ -6,6 +6,14 @@ def selectFutureTerms(currentTermid):
                                 .where((Term.year <= (Term.get_by_id(currentTermid)).year + 2)))
 
     return futureTerms
+
+def selectSurroundingTerms(currentTermid, prevTerms=2):
```
BrianRamsay, line 4: Now that you have this function you should get rid of `selectFutureTerms`

app/controllers/admin/userManagement.py
```diff
@@ -29,6 +32,19 @@ def manageUsers():
...
 @admin_bp.route('/userManagement', methods = ['GET'])
 def userManagement():
+    terms = selectSurroundingTerms(g.current_term.id)
     if g.current_user.isAdmin:
-        return render_template('admin/userManagement.html')
+        return render_template('admin/userManagement.html',
+                                terms=terms)
     abort(403)
+
+@admin_bp.route('/changeCurrentTerm', methods=['POST'])
```
BrianRamsay: Let's make the pure admin routes have 'admin' in them. So, `/admin/changeTerm`, `/admin/manageUsers`, etc. And `/userManagement` should be less specific now, maybe just `/admin`

app/controllers/admin/userManagement.py
```diff
@@ -29,6 +32,19 @@ def manageUsers():
...
+def changeTerm():
+    try:
+        termData = request.form
+        term = int(termData["id"])
+        changeCurrentTerm(term)
+        flash('Current term changed successfully', 'success')
+    except:
+        flash('Error. Current term request unsuccessful', 'warning')
+    return ('success')
```
BrianRamsay: You return 'success' even if the function has an exception. If you just need a return value, return something blank. If you actually want to return a success or failure indicator, make sure it is correct.

app/static/js/userManagement.js
```diff
@@ -41,7 +41,15 @@ $(document).ready(function() {
   $("#removeCeltsStudentStaff").on("click", function() {
     submitRequest("removeCeltsStudentStaff", "#removeCeltsStudentStaffInput")
   });
-
+  for (var i=1; i<=$('#currentTermList .term-btn').length; i++){
+    console.log("clicked")
```
BrianRamsay: remove debug output

BrianRamsay (changes requested):
Getting close! Changes look good.

app/logic/utils.py
zachneill: selectFutureTerms removed

zachneill: actually, selectFutureTerms is being used in the admin routes file.

app/static/js/userManagement.js
zachneill: removed

app/controllers/admin/userManagement.py
zachneill: returns `" "`

app/controllers/admin/userManagement.py
zachneill: uses /admin. sidebar also directs to /admin

app/logic/utils.py
BrianRamsay: I meant you should remove it and change where it was being used to use your new function.

BrianRamsay (changes requested):

BrianRamsay (commented):
Nice work. I will note that you changed the behavior when you replaced `selectFutureTerms` with `selectSurroundingTerms`, because you didn't call it with `prev_term=0` to match how `selectFutureTerms` worked. But I think that's ok. They can tell us if they'd rather have the dropdown start with the current term

BrianRamsay (approved):

PR #149, opened 2021-10-19
Demo branch

mupotsal:
This PR documets issues we found in preparation for the upcoming demo. The PR also fixed some issues that were noted.  

Task | Notes
-- | --
Volunteer logs in and indicates interest for a program | * The volunteer indicated the interest button was not updating but we have worked on the issue before and the function got moved in another PR. We modified the following line to get it to work. interests_ids = [interest.program.id for interest in interests]
CELTS admin creates training event for that program | there isn’t a template for creating student led training events. jinja2.exceptions.TemplateNotFound: /admin/createStudentLedTrainingEvents.htmlWill there be training events that are specifically for an event
CELTS admin sends email to interested volunteers | Got this error: ConnectionRefusedError: [Errno 61] Connection refusedThe email console shows up
Volunteer RSVPs for the training event/attends and gets hours | There is a small glitch. The links under the event name do not work. The end point should be /event/<event id>/edit not /event/<event id>/edit_event.
CELTS admin creates event for the program | This works fine (both creation and cancellation)
CELTS admin sends email to eligible volunteers | ConnectionRefusedError: [Errno 61] Connection refused
Volunteer RSVPs for the event | This works fine after we added a try statement to accept both eventId and Id.
CELTS admin views the volunteer in the track volunteers table and sees if training is completed | There should be an rsvp table to be able to track those who RSVP’ed to the event. Prior to the event, Celts admin can’t track those who didn’t RSVP to the event if they completed the training
Using the Kiosk page to swipe in | There were multiple flashes after =scanning the id, so we fixed that one on the demo-branch. Typing in the Bnumber manually doesn’t let one sign into the event. We fixed this by adding an else to the conditional statement that extract the number (demo-branch)
Volunteer viewing their service transcript |  
CELTS admin searching for a student and viewing their service transcript | Admin searching for a student is not leading them to the profile page. This error: The method is not allowed for the requested URL. (Gaston & Luis working on profile page).
CELTS admin creating all volunteer training event | Admin is able to create all volunteer training events. However, it doesn’t show up in training and education as it should.
Adding a volunteer who isn't registered to the event in the track volunteers table | Working
  |  
Volunteer viewing their service transcript | Working (not sure of how to navigate to with without typing the url)

</div></b>

### DIFF
--- a/app/controllers/events/routes.py
+++ b/app/controllers/events/routes.py
@@ -33,11 +33,15 @@ def signinEvent():
     eventid = request.form["eventid"]
     bnumber = request.form["bNumber"]
     programid = ProgramEvent.select(ProgramEvent.program).where(ProgramEvent.event == eventid)
-    if len(bnumber) >= 9:
+
+    if bnumber[0]==";" and bnumber[-1]=="?": # scanned bNumber starts with ";" and ends with "?"
         bnumber = "B"+ bnumber[1:9]
     else:
-        bnumber = "B"+ bnumber[0:]
-
+        if not bnumber[0].isalpha():
+            bnumber = "B"+ bnumber[0:]
+        else:
+            if not bnumber[0].upper() == "B":
+                return "", 500
     try:
         kioskUser, userStatus = sendUserData(bnumber, eventid, programid)
         if userStatus == "banned":
--- a/app/controllers/events/routes.py
+++ b/app/controllers/events/routes.py
@@ -37,11 +37,10 @@ def signinEvent():
     if bnumber[0]==";" and bnumber[-1]=="?": # scanned bNumber starts with ";" and ends with "?"
         bnumber = "B"+ bnumber[1:9]
     else:
-        if not bnumber[0].isalpha():
-            bnumber = "B"+ bnumber[0:]
-        else:
-            if not bnumber[0].upper() == "B":
-                return "", 500
+        if bnumber[0].isdigit():
+            bnumber = "B"+ bnumber[0:8]
+        elif bnumber[0].upper() != "B":
+            return "", 500
     try:
         kioskUser, userStatus = sendUserData(bnumber, eventid, programid)
         if userStatus == "banned":
--- a/app/controllers/events/routes.py
+++ b/app/controllers/events/routes.py
@@ -33,8 +33,11 @@ def signinEvent():
     eventid = request.form["eventid"]
     bnumber = request.form["bNumber"]
     programid = ProgramEvent.select(ProgramEvent.program).where(ProgramEvent.event == eventid)
-    if len(bnumber) == 23:
+    if len(bnumber) >= 9:
         bnumber = "B"+ bnumber[1:9]
+    else:
+        bnumber = "B"+ bnumber[0:]
+
     try:
         kioskUser, userStatus = sendUserData(bnumber, eventid, programid)
         if userStatus == "banned":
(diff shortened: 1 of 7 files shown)


app/controllers/main/routes.py
```diff
@@ -81,7 +81,12 @@ def volunteerRegister():
     for the event they have clicked register for.
     """
     eventData = request.form
-    event = Event.get_by_id(eventData['id'])
+    try:
+        event = Event.get_by_id(eventData['eventId'])
+    except:
+        event = Event.get_by_id(eventData['id'])
```
BrianRamsay: I'd just change this to whatever it needs to be, rather than accept both. If there are multiple places that hit this controller, let's just change them both to `id` or to `eventId`.

BrianRamsay (changes requested):

BrianRamsay:
Thanks for that list of features. I'm working on the student-led trainings page, and the PR for email sending is close to done.

app/controllers/main/routes.py
sreynit02: we changed the name to eventId.

app/static/js/eventKiosk.js
```diff
@@ -52,7 +52,8 @@ function eventFlasher(flash_message, status){
 }
 
 function submitData(hitEnter = false){
-  if($("#submitScannerData").val().length > 20 || hitEnter == true){
+  if($("#submitScannerData").val().length == 23 || hitEnter == true){
+    console.log("Submitting the data");
```
BrianRamsay: No debug code please

app/templates/admin/createSingleEvent.html
```diff
@@ -65,9 +65,9 @@ <h1 id="title">{{page_title}}</h1>
     <div class="row p-2">
       <div class="col-md-6 col-sm-12 ">
         {% if eventData.id %}
-          <input type="hidden" name="id" value="{{eventData.id}}" >
+          <input type="hidden" name="eventId" value="{{eventData.id}}" >
```
BrianRamsay: You can't make this change without changing everywhere that it is used. In this case, creating new events and editing events won't work, because it expects `id` in the event data. If you want to avoid fixing that functionality you should use `id` everywhere instead of `eventId`. If you really want `eventId`, you need to find all of the event-related logic that uses `id` and convert those as well.

BrianRamsay (changes requested):

app/static/js/eventKiosk.js
```diff
@@ -52,7 +52,7 @@ function eventFlasher(flash_message, status){
 }
 
 function submitData(hitEnter = false){
-  if($("#submitScannerData").val().length > 20 || hitEnter == true){
+  if(hitEnter == true){
```
BrianRamsay: don't compare equality between booleans and booleans. Just use the variable that is true or false. `if(hitEnter) {`

app/controllers/events/routes.py
```diff
@@ -33,8 +33,11 @@ def signinEvent():
     eventid = request.form["eventid"]
     bnumber = request.form["bNumber"]
     programid = ProgramEvent.select(ProgramEvent.program).where(ProgramEvent.event == eventid)
-    if len(bnumber) > 20:
+    if len(bnumber) >= 9:
         bnumber = "B"+ bnumber[1:9]
+    else:
+        bnumber = "B"+ bnumber[0:]
```
BrianRamsay: The logic here needs to be more robust. You know what characters begin and end the scanned string, so I would check for those rather than length for that case. And then if it's possible it was typed, you should check if there's a 'B' in front or not, and then you should make sure you have 8 numbers.

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #151, opened 2021-10-22
Remove RSVP Button from Create Event Template


bryantal:
When creating an event, the RSVP button should not be there.  We'll need to remove it since it causes errors when clicked.

![image](https://user-images.githubusercontent.com/28715731/138475250-1261724d-5f81-491e-8db0-e466a5c2d5bb.png)

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #151: Remove RSVP Button from Create Event Template](https://trello.com/c/XJCD7ORX/327-celts-151-remove-rsvp-button-from-create-event-template)

Boorclark:
@BrianRamsay I believe this issue can be resolved. The issue has already been fixed.

PR #161, opened 2021-10-28
Remove RSVP button from createEvent Page

sreynit02:
Issue #151 : When creating an event, the RSVP button should not be there since RSVP is not possible if the event hasn't been created.

Fix: Remove the button when the 'create' event action is done. The button is still there if the admin is editing or student is just RSVPing for the event.

Test: 
Create an event using a template. There should only be a 'Save' button at the bottom of that page. 
Template


bryantal (dismissed):

app/templates/admin/createSingleEvent.html
```diff
@@ -247,8 +247,12 @@ <h1 id="title">{{page_title}}</h1>
                 {% if userHasRSVPed %}
                   <button type="submit" class="btn btn-danger"  formaction="/rsvpRemove">Remove RSVP</button>
                 {% else %}
-                  <button type="submit" id="rsvpEvent" class="btn btn-primary" formaction="/rsvpForEvent" {{'disabled' if isPastEvent}}>RSVP</button>
+                 {% if request.path[-6:] != "create"%}
```
BrianRamsay: There is already a variable `isNewEvent` set up in this template for this case. It's easier to use that one than include the logic again.

app/templates/admin/createSingleEvent.html
```diff
@@ -247,8 +247,12 @@ <h1 id="title">{{page_title}}</h1>
...
+                 <button type="submit" id="rsvpEvent" class="btn btn-primary" formaction="/rsvpForEvent" {{'disabled' if isPastEvent}}>RSVP</button>
+                 {% endif %}
                 {% endif %}
+                {% else %}
+                <button type="submit" id="rsvpEvent" class="btn btn-primary" formaction="/rsvpForEvent" {{'disabled' if isPastEvent}}>RSVP</button>
```
BrianRamsay: Not sure how this extra `else` and rsvp button snuck in here. Does this page load? I would think it would complain about the extra `{% else %}` and `{% endif %}` without an `{% if %}`

BrianRamsay (changes requested):

app/templates/admin/createSingleEvent.html
mupotsal: Screen Shot 2021-11-01 at 9 40 03 AM
It is for this if, I have made it clear now by indentation. I hope it helps.

app/templates/admin/createSingleEvent.html
mupotsal: Screen Shot 2021-11-01 at 9 41 01 AM
It is for this if, I have made it look better by indentation.

app/templates/admin/createSingleEvent.html
```diff
@@ -244,18 +244,22 @@ <h1 id="title">{{page_title}}</h1>
...
-                  <button type="submit" class="btn btn-danger"  formaction="/rsvpRemove">Remove RSVP</button>
-                {% else %}
-                  <button type="submit" id="rsvpEvent" class="btn btn-primary" formaction="/rsvpForEvent" {{'disabled' if isPastEvent}}>RSVP</button>
-                {% endif %}
-              {% endif %}
-
+                  {% if userHasRSVPed %}
+                    <button type="submit" class="btn btn-danger"  formaction="/rsvpRemove">Remove RSVP</button>
+                  {% else %}
+                     {% if request.path[-6:] != "create"%}
```
BrianRamsay: This is still wrong. There is already a variable in the template for this logic.

app/templates/admin/createSingleEvent.html
```diff
@@ -244,18 +244,22 @@ <h1 id="title">{{page_title}}</h1>
...
+                     <button type="submit" id="rsvpEvent" class="btn btn-primary" formaction="/rsvpForEvent" {{'disabled' if isPastEvent}}>RSVP</button>
+                     {% endif %}
+                 {% endif %}
+            {% else %}
+                <button type="submit" id="rsvpEvent" class="btn btn-primary disabled" formaction="/rsvpForEvent" >RSVP</button>
+            {% endif %}
```
BrianRamsay: There does not need to be two different rsvp buttons. figure out how to only have the html once. It should only be there if the event is in the future, and if we are not creating a new event

BrianRamsay (changes requested):

app/templates/admin/createSingleEvent.html
mupotsal: I have resolved that and. it should be good now.

bryantal (dismissed):

BrianRamsay (approved):

PR #162, opened 2021-10-29
Add RSVP status to Interest Table

sreynit02:
Issue: We need to add RSVP status to show if a user has rsvp for events or not in the upcoming events table. 

Solution: Queried the eventRsvp table for events that the user has RSVP'ed for. Use if statement to check if the upcomingEvents is in eventRsvp.
SCR


bryantal (approved):
So much easier with the RSVP table you all created.  Looks good!

BrianRamsay (approved):

Issue #160, opened 2021-10-28
Add courseProposals Route


zachneill:
The PR #98 was merged into the development. However, its route, /courseProposals doesn't exist in the admin routes file anymore. All the frontend is in development. I think it's a matter of re-adding the route and possibly making minor changes to the withdraw button.

zachneill:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #160: Add courseProposals Route](https://trello.com/c/oDYw5mHK/334-celts-160-add-courseproposals-route)

PR #165, opened 2021-10-29
Fixes Course Proposals Page

zachneill:
This PR is for #160 . The issue was likely that PR #98 conflicted with later merges and likely got overwritten. 

Most of the front end is the same. The biggest change is updating the delete query since there were additional foreign key field dependencies. 

Files renamed: 
- logic/getSLInstructorTableData.py --> courseProposals.py
- controllers/admin/changeSLAction.py --> REMOVED and moved to courseProposals.py logic file
- static/js/changeProposalAction.js --> courseProposals.js
- templates/admin/createSLProposalTable.html --> courseProposals.html 

To test:
Go to /courseProposals 
Delete any proposal. 
The database updates all traces of the course. 

NOTE: Only withdraw works. Other issues focus on view, edit, etc. functionality.


app/controllers/admin/routes.py
```diff
@@ -143,6 +142,26 @@ def addRecurringEvents():
     recurringEvents = calculateRecurringEventFrequency(preprocessEventData(request.form.copy()))
     return json.dumps(recurringEvents, default=str)
 
+@admin_bp.route('/courseProposals', methods=['GET'])
```
bryantal: Should do a check to make sure that the user is a faculty.  I think a specific route with the user's username would also be ideal.  Similar to the volunteer profile.

app/controllers/admin/routes.py
```diff
@@ -143,6 +142,26 @@ def addRecurringEvents():
...
+    courseDict = getProposalData(g.current_user)
+    try:
+        return render_template("/admin/courseProposals.html",
+                                instructor = g.current_user,
+                                courseDict = courseDict)
+    except Exception as e:
+        print('Error while creating table: ', e)
+        return "", 500
+
+@admin_bp.route('/courseProposals/<courseID>/withdraw/', methods = ['POST'])
```
bryantal: Another case where a check of the user is needed.  Make sure a user can't mess around with urls and withdraw a course that isn't theirs.

bryantal (changes requested):

app/controllers/admin/routes.py
bryantal: Oops, misunderstood this, but it still needs a check to see if the user is an admin.  I was able to get to this page and withdraw a course as a student through putting in the URL

app/controllers/admin/routes.py
bryantal: *Check to see if the user is a CELTS admin

bryantal:
Need to merge with development

bryantal (commented):
Need to merge with development

zachneill:
I accidentally used the wrong branch (transcript-link-167) so I've moved it over. Should be good now

app/controllers/admin/routes.py
```diff
@@ -155,6 +154,26 @@ def addRecurringEvents():
...
+@admin_bp.route('/courseProposals', methods=['GET'])
+def createTable():
+    courseDict = getProposalData(g.current_user)
+    try:
+        return render_template("/admin/courseProposals.html",
+                                instructor = g.current_user,
+                                courseDict = courseDict)
+    except Exception as e:
+        print('Error while creating table: ', e)
+        return "", 500
```
BrianRamsay: There's no reason for this try/except. Any exception will already trigger a 500 error and print the message by default. You should only use a blanket try/except like this if you are needing to catch a specific exception, if you want to suppress or sanitize an error message, or if you want to continue through an error

app/controllers/admin/routes.py
```diff
@@ -155,6 +154,26 @@ def addRecurringEvents():
...
+
+@admin_bp.route('/courseProposals/<courseID>/withdraw/', methods = ['POST'])
```
BrianRamsay: I'd drop the last `/` on the URL as well

app/controllers/admin/routes.py
```diff
@@ -174,3 +193,33 @@ def addParticipants():
 
     return render_template('addParticipants.html',
                             title="Add Participants")
+
+@admin_bp.route('/<username>/courseProposals', methods=['GET'])
+def createTable(username, unauthorized=False):
```
BrianRamsay: Where is unauthorized coming from here? Are you using some sort of decorator somewhere?

app/controllers/admin/routes.py
```diff
@@ -174,3 +193,33 @@ def addParticipants():
...
+    if unauthorized:
+        flash("Unauthorized to perform this action", 'warning')
+    try:
+        if authorized(username):
+            courseDict = getProposalData(g.current_user)
+            return render_template("/admin/courseProposals.html",
+                                    instructor = g.current_user,
+                                    courseDict = courseDict,
+                                    username = username)
```
BrianRamsay: I'd be more explicit here. ie, "whose username?" It's hard to know in the template

app/controllers/admin/routes.py
```diff
@@ -174,3 +193,33 @@ def addParticipants():
...
+        else:
+            flash("Unauthorized to view page", 'warning')
+            return redirect(url_for('main.events'))
+    except Exception as e:
+        print('Error while creating table: ', e)
+        return "", 500
+
+@admin_bp.route('/<username>/courseProposals/<courseID>/withdraw/', methods = ['POST'])
```
BrianRamsay: Why are there two different course withdrawal URLs and two different controller methods for doing it?

app/logic/courseProposals.py
```diff
@@ -19,3 +22,19 @@ def getProposalData(user):
         "term":course.term,
         "status":course.status.status}
     return courseDict
+
+def deleteProposal(courseID):
+    """Deletes proposal of ID passed in. Removes foreign keys first.
+    Key Dependencies: QuestionNote, CourseQuestion, CourseParticipant,
+    CourseInstructor"""
+    course = Course.get(Course.id == courseID)
+    notes = (QuestionNote.select()
```
BrianRamsay: Just change this query to a delete and execute it. No need to loop through each note.

BrianRamsay (changes requested):

app/controllers/admin/routes.py
zachneill: done

app/controllers/admin/routes.py
zachneill: try/except removed!

app/controllers/admin/routes.py
zachneill: removed the arbitrary ones. whoops, mis-pulling from branch went under the radar

app/logic/courseProposals.py
zachneill: done!

app/controllers/admin/routes.py
zachneill: refactored code so unauthorized is unneeded.

app/templates/admin/courseProposals.html
```diff
@@ -23,13 +24,12 @@ <h1 class="text-center mt-5 mb-2">{{g.current_user.firstName}} {{g.current_user.
...
         {% for course in courseDict %}
           <tr>
-            <td>{{courseDict[course]['name']}}</td>
-            <td>{{courseDict[course]['faculty']|join(", ")}}</td>
-            <td>{{courseDict[course]['term'].description}}</td>
-            <td>{{courseDict[course]['status']}}</td>
+            <td>{{ courseDict[course]['name'] }}</td>
+            <td>{{ courseDict[course]['faculty']|join(", ") }}</td>
+            <td>{{ courseDict[course]['term'].description }}</td>
+            <td>{{ courseDict[course]['status'] }}</td>
```
bryantal: Some strange spacing going on here.  Was this intentional?

tests/code/test_proposals.py
```diff
@@ -1,17 +1,54 @@
...
+from peewee import DoesNotExist
 from app.models.program import Program
+from app.models.course import Course
+from app.models.courseParticipant import CourseParticipant
+from app.models.courseInstructor import CourseInstructor
+from app.models.courseQuestion import CourseQuestion
 from app.models.user import User
-from app.logic.getSLInstructorTableData import getProposalData
-from app.controllers.admin.changeSLAction import withdrawCourse
+from app.logic.courseProposals import deleteProposal, getProposalData, authorized
```
bryantal: Not sure if this is just on my end, but I'm getting an error from this line about importing authorized from courseProposals.py.  Doesn't look like you use it in this test suite either.  Is it needed?

bryantal (changes requested):

app/logic/courseProposals.py
```diff
@@ -19,3 +24,13 @@ def getProposalData(user):
...
         "status":course.status.status}
     return courseDict
+
+def deleteProposal(courseID):
+    """Deletes proposal of ID passed in. Removes foreign keys first.
+    Key Dependencies: QuestionNote, CourseQuestion, CourseParticipant,
+    CourseInstructor, Note"""
+    course = Course.get(Course.id == courseID)
+    questions = CourseQuestion.select().where(CourseQuestion.course == course)
+    notes = [note for note in Note.select(Note.id).where(QuestionNote.question.in_(questions)).distinct().join(QuestionNote)]
```
BrianRamsay: Rather than this list comprehension, you can just directly convert it with `list(your query)`. I'm surprised that just using the query doesn't work in the delete on line 36 though.

BrianRamsay (changes requested):

BrianRamsay (changes requested):
Do you see an error in test_proposals.py? If so, fix pls. If not, I'll take a look at it and see why it doesn't like my environment.

zachneill:
> Do you see an error in test_proposals.py? If so, fix pls. If not, I'll take a look at it and see why it doesn't like my environment.

@BrianRamsay I fixed it!

zachneill:
[CELTS #160: Add courseProposals Route](https://trello.com/c/oDYw5mHK/334-celts-160-add-courseproposals-route)

zachneill:
#145 integrates all the code from this pr so I will close it

PR #169, opened 2021-11-01
Volunteer profile admin view issue#5

gastonjarju:
**Summary:**
Created the Volunteer Profile page (both the Admin and Volunteer View). 
On both pages, we created two tables: one for the list of programs a user showed interest in and the other for displaying all the CELTS programs on a table alongside the  Eligibility, Training Status, and Checkboxes columns. 
Added a modal for the Admin to ban or unban a volunteer from a specific program. The Admin can also edit a volunteer's interest using the checkboxes. When a checkbox is checked off, the page automatically loads and updates the Interest table. 

**To Test:** 
Go to the (IP_Address/profile/username) and make sure that the user is an Admin or a student to see the Admin view or Volunteer View page respectively. Click on the show interest checkboxes and you will see how the page automatically reloads and the Interest Table updated.
For admins to ban or unban a user, click on the edit link and a modal will pop up. Fill in the end date for the ban and indicate the reason for the ban or unban and click the ban/unban button. This will update the checkmark on the web page and also store the notes for the ban/unban in the note table database. 

**Notes:**
We are writing test suites.


app/controllers/admin/routes.py
```diff
@@ -17,12 +17,16 @@
...
+from app.models.programBan import ProgramBan
+from app.logic.getSLInstructorTableData import getProposalData
 from app.logic.participants import trainedParticipants
 from app.logic.volunteers import getEventLengthInHours
 from app.logic.utils import selectFutureTerms
-from app.logic.events import deleteEvent, getAllFacilitators, attemptSaveEvent, preprocessEventData, calculateRecurringEventFrequency
+from app.logic.events import deleteEvent, getAllFacilitators, getUpcomingEventsForUser , attemptSaveEvent, preprocessEventData, calculateRecurringEventFrequency
 from app.controllers.admin import admin_bp
+from app.logic.users import isEligibleForProgram
+from app.logic.searchUsers import searchUsers
```
BrianRamsay: You added a lot of imports in this file, but there are no other changes. Why did you need the imports?

app/controllers/main/routes.py
```diff
@@ -37,38 +40,72 @@ def events(selectedTerm=None):
...
+    """
+    This function displays the information of a volunteer to the user
+    """
+    if (g.current_user.username == username) or g.current_user.isAdmin:
+         upcomingEvents = getUpcomingEventsForUser(username)
+         programs = Program.select()
+         interests = Interest.select().where(Interest.user == username)
+         programBan = ProgramBan.select().where(ProgramBan.user == username)
+         interests_ids = [interest.program for interest in interests]
+         eventParticipant = EventParticipant.select().where(EventParticipant.user == username)
```
BrianRamsay: This should be plural (and I'd probably change the name to something like `attendedEvents`). Actually, do we use this? Or are you supposed to be getting from EventRsvp?

app/controllers/main/routes.py
```diff
@@ -37,38 +40,72 @@ def events(selectedTerm=None):
         currentTime = currentTime,
         user = g.current_user)
 
-@main_bp.route('/profile/<username>', methods = ['GET'])
-def profilePage(username):
-    if not g.current_user.isCeltsAdmin and not g.current_user.isCeltsStudentStaff and g.current_user.username != username:
-        return "Access Denied", 403
+@main_bp.route('/profile/<username>', methods=['GET'])
+def viewVolunteersProfile(username):
+    """
```
BrianRamsay, line 57: You should 404 if the user doesn't exist

app/controllers/main/routes.py
```diff
@@ -37,38 +40,72 @@ def events(selectedTerm=None):
...
+    This function displays the information of a volunteer to the user
+    """
+    if (g.current_user.username == username) or g.current_user.isAdmin:
+         upcomingEvents = getUpcomingEventsForUser(username)
+         programs = Program.select()
+         interests = Interest.select().where(Interest.user == username)
+         programBan = ProgramBan.select().where(ProgramBan.user == username)
+         interests_ids = [interest.program for interest in interests]
+         eventParticipant = EventParticipant.select().where(EventParticipant.user == username)
+         currentTerm = Term.select().where(Term.isCurrentTerm == 1)
```
BrianRamsay: Use `g.current_term`

app/controllers/main/routes.py
```diff
@@ -37,38 +40,72 @@ def events(selectedTerm=None):
...
+         trainingEvents = (ProgramEvent.select().join(Event).where((Event.term == currentTerm) & (Event.isTraining == 1)))
+         trainingChecklist = {}
+         for program in programs:
+             trainingChecklist[program.id] = trainedParticipants(program.id)
```
BrianRamsay: This is a lot of work (a query for each program). It seems like it's coming at the problem a little backwards. You're getting a list of everyone who has completed the training for each program, rather than just looking to see if the volunteer has completed the trainings in each program.

app/controllers/main/routes.py
```diff
@@ -37,38 +40,72 @@ def events(selectedTerm=None):
...
+         eligibilityTable = []
+         for i in programs:
+             eligibilityTable.append({"program" : i,
+                                      "completedTraining" : (username in trainedParticipants(i)),
```
BrianRamsay: You went through all that work to make `trainingChecklist` and you aren't using it here. You're doing a new query again for each of these.

app/controllers/main/routes.py
```diff
@@ -37,38 +40,72 @@ def events(selectedTerm=None):
...
+def viewVolunteersProfile(username):
+    """
+    This function displays the information of a volunteer to the user
+    """
+    if (g.current_user.username == username) or g.current_user.isAdmin:
+         upcomingEvents = getUpcomingEventsForUser(username)
+         programs = Program.select()
+         interests = Interest.select().where(Interest.user == username)
+         programBan = ProgramBan.select().where(ProgramBan.user == username)
+         interests_ids = [interest.program for interest in interests]
```
BrianRamsay: our variables should be camelCase. And you shouldn't name it `interests_ids` if it doesn't contain IDs. It contains programs they are interested in.

app/controllers/main/routes.py
```diff
@@ -37,38 +40,72 @@ def events(selectedTerm=None):
...
+            interests_ids = interests_ids,
+            trainingChecklist = trainingChecklist,
+            upcomingEvents = upcomingEvents,
+            eligibilityTable = eligibilityTable,
+            volunteer = User.get(User.username == username),
+            user = g.current_user)
+    abort(403)
+
+
+@main_bp.route('/banUnbanUser/<program_id>/<username>', methods=['POST'])
```
BrianRamsay: Instead of having banUnbanUser, make two routes. /banUser and /unbanUser. Each route is a specific action, though they can use the same controller method, then you don't have to pass banOrUnban either.

app/controllers/main/routes.py
```diff
@@ -37,38 +40,72 @@ def events(selectedTerm=None):
...
+@main_bp.route('/deleteInterest/<program_id>/<username>', methods = ['POST'])
+@main_bp.route('/addInterest/<program_id>/<username>', methods = ['POST'])
+def updateInterest(program_id, username):
     """
     This function updates the interest table by adding a new row when a user
     shows interest in a program
     """
     rule = request.url_rule
-    user = g.current_user
+    username = username
```
BrianRamsay: This is a meaningless assignment

app/controllers/main/routes.py
```diff
@@ -37,38 +40,72 @@ def events(selectedTerm=None):
...
-@main_bp.route('/addInterest/<program_id>', methods = ['POST'])
-def updateInterest(program_id):
+
+@main_bp.route('/deleteInterest/<program_id>/<username>', methods = ['POST'])
+@main_bp.route('/addInterest/<program_id>/<username>', methods = ['POST'])
+def updateInterest(program_id, username):
     """
     This function updates the interest table by adding a new row when a user
     shows interest in a program
     """
```
BrianRamsay, line 164: You should check if the user and program exist, and 404 or 500 if they don't.

app/logic/events.py
```diff
@@ -140,10 +141,44 @@ def groupEventsByCategory(term):
                          "One Time Events" : groupEventsByProgram(oneTimeEvents)}
     return categorizedEvents
 
+def eventEdit(newEventData):
```
BrianRamsay: It concerns me that this is in here. Is this from a merge conflict? It is really old, and wrong.

app/logic/users.py
```diff
@@ -10,28 +12,58 @@ def isEligibleForProgram(program, user):
...
-
-    if (ProgramBan.select().where(ProgramBan.user == user, ProgramBan.program == program).exists()):
+    now = datetime.datetime.now()
+    if (ProgramBan.select().where(ProgramBan.user == user, ProgramBan.program == program, ProgramBan.endDate > now).exists()):
         return False
 
     return True
 
-def addRemoveInterest(rule, program_id, currentUser):
+def addRemoveInterest(rule, program_id, username):
```
BrianRamsay: Make this two functions. The controller should worry about whether or not "add" or "delete" is in the rule, and then call the appropriate function.

app/logic/users.py
```diff
@@ -10,28 +12,58 @@ def isEligibleForProgram(program, user):
...
         try:
-            deleted_interest = Interest.get(Interest.program == program_id, Interest.user == currentUser)
+            deleted_interest = Interest.get(Interest.program == program_id, Interest.user == username)
             deleted_interest.delete_instance()
             return "Successfully removed interest"
         except:
             return "This interest does not exist"
+
+
+def banUnbanUser(banOrUnban, program_id, username, note, banEndDate, creator):
```
BrianRamsay: I would create a banUser method and an unbanUser method. And don't forget your tests!

BrianRamsay (changes requested):
First pass at a review.

You also have Windows line endings on 3 files. You should fix those.
http://ssdt-documentation.berea.edu/en/line-endings

BrianRamsay:
Don't forget to fix conflicts. 

In future, don't go so long on one PR without getting it ready to merge.

app/controllers/main/routes.py
```diff
@@ -37,38 +40,72 @@ def events(selectedTerm=None):
...
+            programs = programs,
+            eventParticipant = eventParticipant,
+            interests = interests,
+            trainingEvents = trainingEvents,
+            programBan = programBan,
+            interests_ids = interests_ids,
+            trainingChecklist = trainingChecklist,
+            upcomingEvents = upcomingEvents,
+            eligibilityTable = eligibilityTable,
+            volunteer = User.get(User.username == username),
```
BrianRamsay: You should probably get this up at the top and use it instead of username everywhere. That way if it is invalid you fail at the beginning, instead of at the end. It also makes the code clearer

app/controllers/main/routes.py
```diff
@@ -37,38 +40,72 @@ def events(selectedTerm=None):
...
+            user = g.current_user)
```
BrianRamsay: `g` is available to the templates, you don't have to pass in the current user.

app/static/js/volunteerProfile.js
```diff
@@ -0,0 +1,80 @@
...
+    var programID = $(this).attr('id');
+    var interest = $(this).is(':checked');
+    var username = $(this).attr('name');
+
+    if (interest) {
+      var routeUrl = "addInterest";
+    }
+    else {
+      var routeUrl = "deleteInterest";
+    }
```
BrianRamsay: This if/else could be a lot simpler with a ternary operator: `var routeUrl = interest ? "addInterest" : "deleteInterest;`

app/static/js/volunteerProfile.js
```diff
@@ -0,0 +1,80 @@
...
+      else {
+        $("#banVolunteerButton").prop("disabled", true);
+      }
+
+  });
+  // Reloading page after user clicks on the show interest checkbox
+  $(document).ready(function() {
+    $(".form-check-input").click(function(){
+      location.reload(true);
+      console.log("Help me God")
```
BrianRamsay: don't need debug output

app/static/js/volunteerProfile.js
```diff
@@ -0,0 +1,80 @@
...
+
+  });
+
+  $("#banVolunteerNote, #banVolunteerEndDate").change(function () {
+    if( ($("#banVolunteerNote").val()) && ($("#banVolunteerEndDate")).val()) {
+            $("#banVolunteerButton").prop("disabled", false);
+      }
+      else {
+        $("#banVolunteerButton").prop("disabled", true);
+      }
```
BrianRamsay: Instead of switching on that boolean expression, you could just set the button to it:
```
$("#banVolunteerButton").prop("disabled", ! ($("#banVolunteerNote").val() && $("#banVolunteerEndDate").val()));
```
or to be clearer in two lines:
```
var enableButton = ($("#banVolunteerNote").val() && $("#banVolunteerEndDate").val());
$("#banVolunteerButton").prop("disabled", !enableButton);
```
Also, in this whole file you'd save yourself a lot of typing and line space if you just used 'ban' instead of 'banVolunteer' everywhere. The extra word doesn't really add to the clarity, and you use 'unban' instead of 'unbanVolunteer'

app/templates/admin/serviceTranscript.html
```diff
@@ -0,0 +1,150 @@
+{% extends "base.html"%}
```
BrianRamsay: I'm not sure why you guys have service transcript changes in here. You weren't using this file, right?

app/templates/main/volunteerProfile.html
```diff
@@ -0,0 +1,117 @@
+{% extends "base.html"%}
+{% block scripts %}
+  {{super()}}
+  <script src="//cdnjs.cloudflare.com/ajax/libs/timepicker/1.3.5/jquery.timepicker.min.js"></script>
+  <script src="/static/js/volunteerProfile.js"></script>
+  <script type="text/javascript" src="{{url_for('static', filename='js/ .js') }}?u={{lastStaticUpdate}}"></script>
```
BrianRamsay: This is not including anything. Where did this come from?

app/templates/main/volunteerProfile.html
```diff
@@ -0,0 +1,117 @@
...
+ <div class="row">
+   <table class="table">
+            <tbody id="volunteerTable">
+              <thead>
+                <tr>
+                  <th scope="col">Program</th>
+                  <th scope="col">Interested?</th>
+                  <th scope="col">Completed Training</th>
+                  {% if user.isCeltsAdmin %}
+                  <th scope="col">Elegibility</th>
```
BrianRamsay: "Eligibility" is all i's

app/templates/main/volunteerProfile.html
```diff
@@ -0,0 +1,117 @@
...
+              </thead>
+              {% for program in eligibilityTable %}
+              <tr>
+                <td id="programNameTable">{{program.program.programName}}</td>
+                <td><input class="form-check-input" type="checkbox" id="{{program.program.id}}" name="{{volunteer.username}}" {% if program.program in interests_ids %} checked {% endif %}></td>
+                <td>{% if volunteer.username in trainingChecklist[program.program.id] %}<i class="bi bi-check-lg"></i> {% else %} <i class="bi bi-x-lg"></i>{% endif %}</td>
+                {% if user.isCeltsAdmin %}
+                <td class="programBan">
+                  {% if program.isNotBanned %}<i class="bi bi-check-lg"></i><button type="button" class="ban btn btn-link" id="{{program.program}}" name="{{program.program.programName}}" value="Ban">Edit</button>
+                  {% else %} <i class="bi bi-x-lg"></i><button type="button" class="ban btn btn-link" id="{{program.program}}" name="{{program.program.programName}}" value="Unban">Edit</button>
```
BrianRamsay: even though `program.program` works here, you're relying on the object being automatically cast to a string that is just the id. I would be explicit and use `program.program.id`

BrianRamsay (changes requested):
There has been a lot of code written, so don't be concerned about how many suggestions there are for change.

Also, a couple of tests are failing. Are they failing in your environment?

app/controllers/admin/routes.py
gastonjarju: Removed the unused imports

app/controllers/main/routes.py
gastonjarju: We have removed the unused query, "eventParticipant" and "programBan."

app/controllers/main/routes.py
gastonjarju: We removed the "currentTerm" query because we did not use it in our code.

app/controllers/main/routes.py
gastonjarju: Added a try and except to take care of the 404 error message if the user does not exist. Leads an unknown user to a 404 error page that says, "user does not exist."

app/controllers/main/routes.py
Lyisus: Changed this variable name to "programsInterested" and changed the HTML file accordingly

app/controllers/main/routes.py
Lyisus: Deleted this part as it is being done later in the file in the eligibilityTable, will try to fix the query there, instead.

app/controllers/main/routes.py
gastonjarju: Deleted the trainingChecklist because we are not using it.

app/controllers/main/routes.py
Lyisus: Deleted the meaningless assignment

app/logic/events.py
Lyisus: Deleted this as it was old and outdated

app/controllers/main/routes.py
Lyisus: Refactored according to this suggestion

app/controllers/main/routes.py
Lyisus: Refactored the controller and HTML according to this comment's useful suggestion.

app/static/js/volunteerProfile.js
Lyisus: Debug output has been eliminated

app/templates/main/volunteerProfile.html
gastonjarju: Resolved the spellings.

app/templates/admin/serviceTranscript.html
Lyisus: We did not work with this file. After a little investigating, it seems to be the result of a previous merge conflict where we kept the file not knowing it was no longer being used. There is a another serviceTranscript.html in the templates/main folder that is in development while this one, in templates/admin, is not.

app/templates/main/volunteerProfile.html
Lyisus: Maybe from Ebeneezer's branch when we merged with it at the beginning. We trusted that what he had was necessary for his code to work and didn't think of refactoring it at the time. These scripts will be moving away from our files soon enough.

gastonjarju:
> Don't forget to fix conflicts.
> 
> In future, don't go so long on one PR without getting it ready to merge.

app/templates/main/volunteerProfile.html
gastonjarju: They have been removed.

app/controllers/main/routes.py
gastonjarju: Added the check.

app/templates/main/volunteerProfile.html
gastonjarju: Fixed!

app/static/js/volunteerProfile.js
gastonjarju: Fixed.

app/controllers/main/routes.py
```diff
@@ -45,42 +46,76 @@ def events(selectedTerm=None):
...
-        return "Access Denied", 403
+@main_bp.route('/profile/<username>', methods=['GET'])
+def viewVolunteersProfile(username):
+    """
+    This function displays the information of a volunteer to the user
+    """
+    try:
+        User.get(User.username == username)
+    except Exception as e:
+        print(e)
```
BrianRamsay: You don't need to print this one. There's no extra information we would get from that output

app/controllers/main/routes.py
```diff
@@ -45,42 +46,76 @@ def events(selectedTerm=None):
...
+         # for program in programs:
+         #     trainingChecklist[program.id] = trainedParticipants(program.id)
+         eligibilityTable = []
+         for program in programs:
+              notes = ProgramBan.select().where(ProgramBan.user == username,
+                                                ProgramBan.program == program,
+                                                ProgramBan.endDate > datetime.datetime.now())
+              noteForDict = "None"
+              for j in notes:
+                  noteForDict = j.banNote.noteContent
```
BrianRamsay: This is a long way around to get only the most recent note, and it relies on the query being sorted implicitly. I'd just get the most recent straight out of the query, or even use a slice of the list of all notes (like `list(notes)[-1].banNote.noteContent`)

app/controllers/main/routes.py
```diff
@@ -45,42 +46,76 @@ def events(selectedTerm=None):
...
-        return "Error retrieving user profile", 500
+        return "Error Updating Ban", 500
+
 
-@main_bp.route('/deleteInterest/<program_id>', methods = ['POST'])
-@main_bp.route('/addInterest/<program_id>', methods = ['POST'])
-def updateInterest(program_id):
+@main_bp.route('/deleteInterest/<program_id>/<username>', methods = ['POST'])
+@main_bp.route('/addInterest/<program_id>/<username>', methods = ['POST'])
+def updateInterest(program_id, username):
```
BrianRamsay: I would reorganize this route to put username first, since that's how our other user-related routes are. like `<username>/delete_interest/<program_id>`

BrianRamsay: The ban/unban routes should follow the same pattern

app/logic/events.py
```diff
@@ -147,6 +147,8 @@ def getUpcomingEventsForUser(user,asOf=datetime.datetime.now()):
     events = (Event.select(Event)
                             .join(ProgramEvent)
                             .join(Interest, on=(ProgramEvent.program == Interest.program))
+                            # .join(ProgramBan, on=(ProgramEvent.program == ProgramBan.program))
+                            # .where(ProgramBan.program != ProgramEvent.program)
```
BrianRamsay: remove commented out code

app/templates/main/volunteerProfile.html
BrianRamsay: Doesn't look like it

app/static/js/volunteerProfile.js
```diff
@@ -0,0 +1,71 @@
+$(document).ready(function(){
+
+  $(".form-check-input").click(function updateInterest(){
+    var programID = $(this).attr('id');
+    var interest = $(this).is(':checked');
+    var username = $(this).attr('name');
+
+    var routeUrl = interest ? "addInterest" : "deleteInterest;
```
BrianRamsay: syntax error

BrianRamsay (changes requested):
The message when the events list is empty doesn't look very good.

app/static/js/volunteerProfile.js
gastonjarju: Fixed!

app/logic/users.py
gastonjarju: Fixed!

app/static/js/volunteerProfile.js
gastonjarju: Fixed!

Lyisus: Noticed and fixed with a previous commit

app/logic/events.py
Lyisus: Removed

app/controllers/main/routes.py
Lyisus: Deleted

app/controllers/main/routes.py
Lyisus: Two routes have been made.

app/logic/users.py
Lyisus: Different methods have been implemented.

app/controllers/main/routes.py
BrianRamsay: Now you are doing this query twice. Once at the beginning and once here. And you still are using `username` everywhere instead of the user object.

app/controllers/main/routes.py
BrianRamsay: you are still passing in the current user. You don't need to pass anything.

app/controllers/main/routes.py
```diff
@@ -45,42 +46,99 @@ def events(selectedTerm=None):
...
+            interests = interests,
+            programsInterested = programsInterested,
+            upcomingEvents = upcomingEvents,
+            eligibilityTable = eligibilityTable,
+            volunteer = User.get(User.username == username),
+            user = g.current_user)
+    abort(403)
+
+
+@main_bp.route('/banUser/<program_id>/<username>', methods=['POST'])
```
BrianRamsay: I think these urls should match the interest url format, like `/<username>/ban/<program_id>` and `/<username>/unban/<program_id>`

app/logic/users.py
```diff
@@ -10,28 +12,78 @@ def isEligibleForProgram(program, user):
...
     program_id: id of the program the user is interested in
+    username: username of the user showing disinterest
+
     """
+    try:
+        interestToDelete = Interest.get(Interest.program == program_id, Interest.user == username)
+        interestToDelete.delete_instance()
+        return "Successfully removed interest"
+    except:
+        return "This interest does not exist"
```
BrianRamsay: If they want to remove an interest and one doesn't exist, who cares? The end result is the same. 

If this text doesn't ever show to the end user, I wouldn't worry about changing it, but a better plan would be to check for a `DoesNotExist` exception and ignore it. That way any other exception will still be raised for the controller to capture. The way you have it right now the try/except in the controller is unnecessary, because every exception is caught inside this function.

app/logic/users.py
```diff
@@ -10,28 +12,78 @@ def isEligibleForProgram(program, user):
...
+    creator: the admin or person with authority who created the ban
+    """
+    noteForDb = Note.create(createdBy = creator,
+                             createdOn = datetime.datetime.now(),
+                             noteContent = note,
+                             isPrivate = 0)
+    ProgramBan.create(program = program_id,
+                      user = username,
+                      endDate = banEndDate,
+                      banNote = noteForDb.id)
```
BrianRamsay: don't reference `.id` here. Just use `noteForDb` and let peewee handle it

app/static/js/volunteerProfile.js
```diff
@@ -0,0 +1,65 @@
...
+    $("#banEndDate").val("")
+    $("#banVolunteerNote").val("")
+
+    if( $(this).val()=="Unban"){
+      $("#ubanEndDate").hide()
+      $("#banEndDate").val("0001-01-01") //This is a placeholder value for the if statement in line 49 to work properly
+      $("#banNoteDiv").show()
+      $("#banNote").text($(this).attr("note"))
+    }
+
```
BrianRamsay, line 50: Rather than call `$()` multiple times for each element, you should store them each in a variable. That way there aren't as many places you need to change the id should the id change, and you aren't having to make jquery do as much work.

app/static/js/volunteerProfile.js
```diff
@@ -0,0 +1,65 @@
...
+    $("#banButton").attr("username", $(".form-check-input").attr("name"))
+    $("#banButton").attr("banOrUnban", $(this).val());
+    $("#ubanEndDate").show()
+    $("#banNoteDiv").hide()
+    $("#banEndDate").val("")
+    $("#banVolunteerNote").val("")
+
+    if( $(this).val()=="Unban"){
+      $("#ubanEndDate").hide()
+      $("#banEndDate").val("0001-01-01") //This is a placeholder value for the if statement in line 49 to work properly
```
BrianRamsay: Good use of a comment!

BrianRamsay (changes requested):

bryantal:
A few things:
1. The message left when the user has not marked interest in any programs has not changed.  It needs to since its contrast is too high against the white background which is an accessibility issue.  Maybe just make it black text and center it a little better.
2. The datepicker for the ban modal allows me to pick a date that is before the current date and the current date.  It shouldn't allow for either (Our other datepickers have this functionality, so take a look at those).
3. There is a glitch with the note text box where I am able to type in it, but the ban button is still disabled until I click out of the text box.
4. Merge Conflicts.

app/logic/users.py
BrianRamsay: Your change didn't actually address this, as far as I can tell.

app/logic/users.py
```diff
@@ -10,28 +12,79 @@ def isEligibleForProgram(program, user):
...
+    username: username of the user showing disinterest
+
     """
+    interestToDelete = Interest.get(Interest.program == program_id, Interest.user == username)
+    if interestToDelete:
+        interestToDelete.delete_instance()
+        return "Successfully removed interest"
+    else:
+        raise Exception("Interest does not exist.")
+        return False
```
BrianRamsay: You didn't address my earlier comment about this. This case shouldn't cause a failure. What they want the end result to be is already the case - they aren't interested in this particular program. Also, your `return False` will never be reached because you are raising the exception, but there's not a good reason to raise the exception in the first place.

tests/code/test_users.py
```diff
@@ -34,63 +33,78 @@ def test_isEligibleForProgram():
...
+
+    username = "khatts"
+    program_id = 2
+    result = removeUserInterest(program_id, username)
+    assert result == "Successfully removed interest"
+
+    #test removing interest with different program id
+    program_id = 3
+    result = removeUserInterest(program_id, username)
+    assert result == "Successfully removed interest"
```
BrianRamsay: You are only checking successes. No failure cases.

tests/code/test_users.py
```diff
@@ -34,63 +33,78 @@ def test_isEligibleForProgram():
...
+    result = addUserInterest(program_id, username)
     assert result == 'Successfully added interest'
 
-    # test add and removing interest with different program id
-    rule = 'addInterest'
+    # test adding interest with different program id
     program_id = 3
-    result = addRemoveInterest(rule, program_id, user)
+    result = addUserInterest(program_id, username)
     assert result == 'Successfully added interest'
```
BrianRamsay: Are you checking cases where the interest already exists?

app/logic/users.py
```diff
@@ -10,28 +12,79 @@ def isEligibleForProgram(program, user):
...
+    Parameters:
+    program_id: id of the program the user is interested in
+    username: username of the user showing interest
+    """
+    try:
+        Interest.get_or_create(program = program_id, user = username)
+        return "Successfully added interest"
+    except Exception as e:
+        print("Error: ", e)
+        return "An error ocurred while adding the interest"
```
BrianRamsay: As I mentioned to you in person, blanket exceptions can be risky. If you expect a certain exception, catch it specifically. If your function returns errors, the calling code should handle that. If your function raises exceptions, the calling code should handle that. Mixing exceptions and return values can get tricky.

app/controllers/main/routes.py
```diff
@@ -45,42 +46,97 @@ def events(selectedTerm=None):
...
+    program_id: the primary id of the program the student is being unbanned from
+    username: unique value of a user to correctly identify them
+    """
+    postData = request.form
+    unbanNote = postData["note"] # This contains the note left about the change
+    try:
+        return unbanUser(program_id, username, unbanNote, g.current_user)
+    except Exception as e:
+        print(e)
+        return "Error Updating Unban", 500
```
BrianRamsay: "Error Updating Unban" isn't a great error message for the user.

BrianRamsay: Not sure if they will see it or not

BrianRamsay (changes requested):

tests/code/test_users.py
gastonjarju: Fixed!

tests/code/test_users.py
gastonjarju: No. I was checking cases where user adds an interest with a different program_id

app/controllers/main/routes.py
gastonjarju: I have tweaked the exception a little bit and also added a flasher when a user is unbanned.

app/logic/users.py
gastonjarju: We have removed the unnecessary exception raised because if program_id and username does not except exceptions have already been taken care of in the routes.py file.

app/logic/users.py
gastonjarju: Fixed.

app/controllers/main/routes.py
```diff
@@ -47,50 +49,134 @@ def events(selectedTerm=None):
...
+    try:
+        User.get(User.username == username)
+    except Exception as e:
+        return "User does not exist", 404
+    if (g.current_user.username == username) or g.current_user.isAdmin:
+         upcomingEvents = getUpcomingEventsForUser(username)
+         programs = Program.select()
+         interests = Interest.select().where(Interest.user == username)
+         programsInterested = [interest.program for interest in interests]
+         print("===============", username, type(username))
```
BrianRamsay: no debug output

app/controllers/main/routes.py
```diff
@@ -47,50 +49,134 @@ def events(selectedTerm=None):
...
+@main_bp.route('/profile/<username>', methods=['GET'])
+def viewVolunteersProfile(username):
+    """
+    This function displays the information of a volunteer to the user
+    """
+    try:
+        User.get(User.username == username)
+    except Exception as e:
+        return "User does not exist", 404
+    if (g.current_user.username == username) or g.current_user.isAdmin:
```
BrianRamsay: since you have already gotten the user from the database in line 58 (you've done the query anyway), I would just use it here instead of username, like `g.current_user == user`

app/controllers/main/routes.py
BrianRamsay: This comment was not resolved.

app/controllers/main/routes.py
```diff
@@ -47,50 +49,134 @@ def events(selectedTerm=None):
...
-                               user = profileUser,
-                               programs = programs,
-                               interests = interests,
-                               interests_ids = interests_ids,
-                               upcomingEvents = upcomingEvents,
-                               rsvpedEvents = rsvpedEvents,
-                               backgroundTypes = backgroundTypes,
-                               completedBackgroundCheck = completedBackgroundCheck
-                               )
+        flash("Successfully banned user", "success")
```
BrianRamsay: Don't flash before you actually try it. If you do this and there is an exception in `banUser`, the successful flash has already happened.

app/controllers/main/routes.py
```diff
@@ -47,50 +49,134 @@ def events(selectedTerm=None):
...
+#                                user = profileUser,
+#                                programs = programs,
+#                                interests = interests,
+#                                interests_ids = interests_ids,
+#                                upcomingEvents = upcomingEvents,
+#                                rsvpedEvents = rsvpedEvents,
+#                                backgroundTypes = backgroundTypes,
+#                                completedBackgroundCheck = completedBackgroundCheck
+#                                )
+# >>>>>>> development
```
BrianRamsay: Get rid of the merge artifacts once you have resolved them, don't just comment them out

app/templates/main/volunteerProfile.html
```diff
@@ -0,0 +1,159 @@
+{% extends "base.html"%}
+{% block scripts %}
+  {{super()}}
+  <script src="/static/js/volunteerProfile.js"></script>
+{% endblock %}
+{% block styles %}
+  {{super()}}
+    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/timepicker/1.3.5/jquery.timepicker.min.css">
```
BrianRamsay: You don't use a timepicker, do you? If not, this whole block is unnecessary.

tests/code/test_users.py
```diff
@@ -34,63 +34,109 @@ def test_isEligibleForProgram():
...
-    # test for incorrect rule
-    rule = "lkejfiv"
-    result = addRemoveInterest(rule, program_id, user)
-    assert result == None
+        result = addUserInterest(program_id, username)
+        assert result == True
+
+
+@pytest.mark.integration
+def test_banUser():
```
BrianRamsay, line 110: There's no failure in here. What happens if we failed to ban the user? If the answer is there's an exception, that's fine, but verify that in the test.

tests/code/test_users.py
```diff
@@ -34,63 +34,109 @@ def test_isEligibleForProgram():
...
+
+    #test for banning a user from a program with different program is
+    program_id = 3
+    status = banUser (program_id, username, note, banEndDate, creator)
+    assert status == "Successfully banned the user"
+
+
+
+@pytest.mark.integration
+def test_unbanUser():
```
BrianRamsay, line 141: There's no failure in here. What happens if we failed to unban the user? If the answer is there's an exception, that's fine, but verify that in the test.

BrianRamsay (changes requested):
Pretty close, guys! Just a few comments left in addition to fixing failing tests.

app/controllers/main/routes.py
gastonjarju: Fixed the commented part of the code.

app/templates/main/volunteerProfile.html
gastonjarju: Yes, it is unnecessary because we are not using it. Thus, it has been removed.

app/controllers/main/routes.py
Lyisus: Turned line 61 into `if (g.current_user == username) or g.current_user.isAdmin:` and it works fine. I am not sure if that's what you meant by this comment, though.

app/controllers/main/routes.py
```diff
@@ -49,50 +49,107 @@ def events(selectedTerm=None):
...
-                               interests_ids = interests_ids,
-                               upcomingEvents = upcomingEvents,
-                               rsvpedEvents = rsvpedEvents,
-                               backgroundTypes = backgroundTypes,
-                               completedBackgroundCheck = completedBackgroundCheck
-                               )
+        volunteer = User.get(User.username == username)
+    except Exception as e:
+        return "User does not exist", 404
+    if (g.current_user == username) or g.current_user.isAdmin:
```
BrianRamsay: You should use `volunteer` everywhere you can instead of `username`. Once you validate your inputs, use the validated data instead of the inputs. For this code it doesn't make much of a difference because you are ensuring `username` is valid above with the try/except, but it's a good principle to follow.

BrianRamsay (changes requested):
So close! Only one more comment. Sorry it took me so long to make it.

app/controllers/main/routes.py
Karina-Agliullova: We fixed the issue. Added **volunteer** everywhere we could.

BrianRamsay (changes requested):
There are some failing tests and a merge conflict to resolve.

BrianRamsay:
Closed because of a mis-merge and bad test fixes

Issue #167, opened 2021-11-01
Link to Volunteer Service Transcript


bryantal:
We need a link to volunteer service transcript from the user's profile page.  Let's use Bootstrap's nav tabs to complete this similar to how it is implemented on the edit event page and track volunteers page.

![image](https://user-images.githubusercontent.com/28715731/139686919-99a8faa5-dc00-4c5c-9a64-3857a29e59d7.png)
![image](https://user-images.githubusercontent.com/28715731/139686566-dd15d889-05eb-4992-b64c-ba54e049cc1e.png)

route: [IP Address]:8080/profile/[username]

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #167: Link to Volunteer Service Transcript](https://trello.com/c/9DuQan6w/336-celts-167-link-to-volunteer-service-transcript)

PR #188, opened 2021-11-05
Adds Tabs to Link Service Transcript and Upcoming Events

zachneill:
This PR fixes issue #167 where there was no link to the volunteer service transcript page. 

There is now a nav-tab in both serviceTranscript.html and volunteerProfile.html. I also added and reformatted their names to better reflect the wireframe. 

TO TEST:
Go to /profile/<USERNAME> 
There should be a nav-tab that takes the user to the username's transcript. 
Click on "Service Transcript" 
It sends the user to /<USERNAME>/serviceTranscript


zachneill:
This PR accidentally pulled from a different branch. I'm closing it and instead using PR #189

PR #189, opened 2021-11-05
Adds Tabs to Link Service Transcript and Upcoming Events

zachneill:
This PR fixes issue #167 where there was no link to the volunteer service transcript page.

There is now a nav-tab in both serviceTranscript.html and volunteerProfile.html. I also added and reformatted their names to better reflect the wireframe.

TO TEST:
Go to /profile/
There should be a nav-tab that takes the user to the username's transcript.
Click on "Service Transcript"
It sends the user to //serviceTranscript

### DIFF
--- a/app/templates/main/serviceTranscript.html
+++ b/app/templates/main/serviceTranscript.html
@@ -2,8 +2,18 @@
 
 {% block app_content %}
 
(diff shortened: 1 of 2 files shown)


BrianRamsay (changes requested):
If adding tabs, you need to fix the Print display so they don't show up on the printed page

BrianRamsay (approved):
Looks good, thanks

PR #191, opened 2021-11-09
Time selector fix for Firefox and Safari

sreynit02:
Issue: Select picker for time input on create_event page does not allow a user to select a time on Firefox and Safari. They can only type in the time. 

Solution: We added a function to the javascript file of this page to check if the browser is Safari or Firefox and then use the Modernizr library to add the time picker that supports these two browsers. However, the time icon is not showing . 

Test: 
- navigate to create an event page in Firefox and Safari.
- result: 
timepicker


app/static/js/createEvents.js
```diff
@@ -17,11 +16,20 @@ function updateDate(obj) {
...
 }
 
+function updateTime(obj) {
+  if (obj.id == "startTime") {
+    $("#startTime").timepicker({timeFormat: 'h:mm p',});
+  }
+
+  if (obj.id == "startTime") {
+    $("#startTime").timepicker({timeFormat: 'h:mm p',});
+  }
```
BrianRamsay: These are the exact same conditionals.

BrianRamsay: Aren't you already initializing these when the document loads?

app/templates/admin/createSingleEvent.html
```diff
@@ -4,6 +4,8 @@
   {{super()}}
   <script src="/static/js/createEvents.js"></script>
   <script src="//cdnjs.cloudflare.com/ajax/libs/timepicker/1.3.5/jquery.timepicker.min.js"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/modernizr/2.8.3/modernizr.min.js"></script>
+
```
BrianRamsay, line 7: We don't need modernizr anymore

app/templates/admin/createSingleEvent.html
```diff
@@ -116,17 +118,17 @@ <h1 id="title">{{page_title}}</h1>
           <div class="form-inline d-flex">
             <div  class="form-group" >
               <div class="d-inline-flex">
-                <label for="startTime"><strong>Start Time:</strong></label>
+                <label for="startTimePicker"><strong>Start Time:</strong></label>
                 {% if isEditable %}
-                  <input type='time' class="form-control timePicker" id='pickStartTime'value="{{eventData.timeStart}}"  name="timeStart" required>
+                  <input type='time' class="form-control" id='startTime'value="{{eventData.timeStart}}"  name="timeStart" onchange=updateTime(this) required>
```
BrianRamsay: The label doesn't match the id anymore

app/templates/admin/createSingleEvent.html
```diff
@@ -116,17 +118,17 @@ <h1 id="title">{{page_title}}</h1>
...
                 {% else %}
                   <p>{{ eventData.timeStart.strftime("%I: %M %p") if eventData.timeStart }}</p>
                 {% endif %}
               </div>
               <div class='form-group time d-inline-flex' style="padding: 0 0 0 2rem">
-                <label for="endTime"><strong>End Time:</strong></label>
+                <label for="endTimePicker"><strong>End Time:</strong></label>
                 {% if isEditable %}
-                  <input type='time' class="form-control timePicker" id='pickEndTime' value="{{eventData.timeEnd}}" name="timeEnd" required>
+                  <input type='time' class="form-control" id='endTime' value="{{eventData.timeEnd}}" name="timeEnd" onchange=updateTime(this) required>
```
BrianRamsay: The label doesn't match the id anymore

app/templates/admin/createSingleEvent.html
BrianRamsay: Also, since we're already doing stuff to these elements in javascript, including adding an event listener, the onchange event should also be added there.

BrianRamsay (changes requested):

app/static/js/createEvents.js
sreynit02: removed this function since we already initialized when document laods.

app/templates/admin/createSingleEvent.html
sreynit02: removed.

app/templates/admin/createSingleEvent.html
sreynit02: made label match the id. removed onchange event

app/templates/admin/createSingleEvent.html
sreynit02: made label match id

BrianRamsay (changes requested):
I would play with the options a little more on the JQuery UI selector. We don't usually use 24 hour time, and we don't need to see every time option through the night. I'd also see about changing the end time options based on what they selected in start time.

mupotsal:
> I would play with the options a little more on the JQuery UI selector. We don't usually use 24 hour time, and we don't need to see every time option through the night. I'd also see about changing the end time options based on what they selected in start time.

What timeframes should be included in the time picker?

BrianRamsay:
I would say the start options should be from 8am to 10pm, with the end options maybe from 8 to midnight. But they should be able to type any time. The validation can make sure that the end time is after the start time.

hmanalai:
The PR is updated to make the timepicker look the same in all browsers. The timepicker library we are using has some accessibility issues. I looked into other libraries and couldn't find one that did not have accessibility issues or was compatible with Bootstrap 5. 
Also, this PR now fixes issue #219  as well. 

I tried to fix the flash message accessibility (it's not being announced in Firefox) but couldn't find any solutions. So that issue is still open.

hmanalai:
**UPDATE**: After our labor meeting discussion, we decided to use the `timepicker `library only when the browser is Firefox or Safari. The PR is updated to reflect this change.

BrianRamsay (approved):

Issue #171, opened 2021-11-01
Sidebar Doesn't Extend to the Length of the Page


bryantal:
Our sidebar isn't extending the whole length of the page.  I'm pretty sure this affects all pages.

![image](https://user-images.githubusercontent.com/28715731/139723119-ca0612be-3731-4b74-b4b8-7700fe6a677c.png)

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #171: Sidebar Doesn't Extend to the Length of the Page](https://trello.com/c/6Me82YLs/338-celts-171-sidebar-doesnt-extend-to-the-length-of-the-page)

sreynit02:
We were not able to get the sidebar to extend. But we removed the dark "space" to the right of the sidebar. This change is committed to branch 'fix-sidebar' .

PR #197, opened 2021-11-12
Extend and Restrict Sidebar to Bottom of Page

zachneill:
This PR is a solution to issue #171. The sidebar did not extend to the end of the page. I couldn't find a viable solution where the sidebar stays put, so it is now a fixed sidebar. 

I added position-fixed to the nav in base.html. I also added a margin-right: 280px, which is the width of the sidebar in sidebar.html. This pushes everything back to the center. Otherwise, the sidebar's position is ignored, and the content goes under the bar. Printing is unaffected, and because of col-md-auto, mobile view is also unaffected. 

TO TEST:
Go to the CELTS website, and the sidebar should be fixed on all pages.

### DIFF
--- a/app/templates/base.html
+++ b/app/templates/base.html
@@ -34,64 +34,64 @@
 {% block content %}
 
 <div>
-    <div class="row g-0">
-      <div class="col-md-auto d-none d-lg-block">
-        <div style="margin-right: 280px">
-          <nav class="position-fixed float-start">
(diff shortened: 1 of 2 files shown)


BrianRamsay (approved):
Looks good. The original reported problem is really caused by the main container taking up 100% of vertical space, and then the footer trying to squeeze its way back in, but this is a good change.

Issue #190, opened 2021-11-08
Add Background Check Status to Volunteer Profile


bryantal:
We'll need a section, possibly on the bottom side in the table, for admins to keep track of the status of background checks they put in for the volunteers.  We'll probably need a new table for background checks that have a foreign key that points to a user and also contains a passed or failed field.

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #190: Add Background Check Status to Volunteer Profile](https://trello.com/c/QZW2YIRJ/342-celts-190-add-background-check-status-to-volunteer-profile)

PR #198, opened 2021-11-15
Background check

sreynit02:
This PR addresses issue #190. 

We implemented the logic that records if the student has passed the background check. We created a model calls backgroundCheck that has the user as the ForeignKey Field and a boolean field that indicates if that user passed the background check. 

We also have a simple select UI that is located on the User Profile page that indicates if the student has completed the background check. Selecting  Yes or No just records if the user has passed the background check. There is no further logic that relates to banning yet. 

background check

Test: 
1. run tests. All tests in test_volunteer should pass
2. navigate to a student profile and the page that looks like above should show up. select Yes/No and check the backgroundcheck table to see if the passbackgroundCheck column corresponds to what you selected.

### DIFF
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -11,6 +11,7 @@ from app.logic.volunteers import updateEventParticipants, addVolunteerToEventRsv
 from app.logic.participants import trainedParticipants, getEventParticipants
 from app.models.user import User
 from app.models.eventRsvp import EventRsvp
+from app.models.backgroundCheck import BackgroundCheck
 
 
 @admin_bp.route('/searchVolunteers/<query>', methods = ['GET'])
@@ -97,3 +98,10 @@ def removeVolunteerFromEvent(user, eventID):
     (EventRsvp.delete().where(EventRsvp.user==user)).execute()
     flash("Volunteer successfully removed", "success")
     return ""
+
+@admin_bp.route('/', methods = ['POST'])
+def updateBackgroundCheck(user, setTo):
+
+    updateBackgroundStatus = BackgroundCheck.select().where(user == user)
+    updateBackgroundStatus.passBackgroundCheck = setTo
+    updateBackgroundCheck.save()
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -103,7 +103,7 @@ def removeVolunteerFromEvent(user, eventID):
 def updateBackgroundCheck():
     eventData = request.form
     user = eventData['user']
-    setTo = int(eventData['setTo'])
+    checkPassed = int(eventData['checkPassed'])
     type = eventData['bgType']
-    setUserBackgroundCheck(user,type, setTo)
+    setUserBackgroundCheck(user,type, checkPassed)
     return ""
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -7,7 +7,7 @@ from app.models.event import Event
 from app.models.user import User
 from app.models.eventParticipant import EventParticipant
 from app.logic.searchUsers import searchUsers
-from app.logic.volunteers import updateEventParticipants, addVolunteerToEventRsvp, getEventLengthInHours
+from app.logic.volunteers import updateEventParticipants, addVolunteerToEventRsvp, getEventLengthInHours,updateOrCreateVolunteerBackground
 from app.logic.participants import trainedParticipants, getEventParticipants
 from app.models.user import User
 from app.models.eventRsvp import EventRsvp
@@ -99,9 +99,14 @@ def removeVolunteerFromEvent(user, eventID):
     flash("Volunteer successfully removed", "success")
     return ""
 
-@admin_bp.route('/', methods = ['POST'])
-def updateBackgroundCheck(user, setTo):
-
-    updateBackgroundStatus = BackgroundCheck.select().where(user == user)
-    updateBackgroundStatus.passBackgroundCheck = setTo
-    updateBackgroundCheck.save()
+@admin_bp.route('/updateBackgroundCheck', methods = ['POST'])
+def updateBackgroundCheck():
+    print(".............................................................")
+    eventData = request.form
+    user = eventData['user']
+    setTo = int(eventData['setTo'])
+    print(type(setTo))
+    print("....................................................",eventData)
+    print("This is the data we got",user,setTo)
+    updateOrCreateVolunteerBackground(user,setTo)
+    return ""
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -7,7 +7,7 @@ from app.models.event import Event
 from app.models.user import User
 from app.models.eventParticipant import EventParticipant
 from app.logic.searchUsers import searchUsers
-from app.logic.volunteers import updateEventParticipants, addVolunteerToEventRsvp, getEventLengthInHours,updateOrCreateVolunteerBackground
+from app.logic.volunteers import updateEventParticipants, addVolunteerToEventRsvp, getEventLengthInHours,setUserBackgroundCheck
 from app.logic.participants import trainedParticipants, getEventParticipants
 from app.models.user import User
 from app.models.eventRsvp import EventRsvp
@@ -104,5 +104,6 @@ def updateBackgroundCheck():
     eventData = request.form
     user = eventData['user']
     setTo = int(eventData['setTo'])
-    updateOrCreateVolunteerBackground(user,setTo)
+    type = eventData['bgType']
+    setUserBackgroundCheck(user,type, setTo)
     return ""
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -101,12 +101,8 @@ def removeVolunteerFromEvent(user, eventID):
 
 @admin_bp.route('/updateBackgroundCheck', methods = ['POST'])
 def updateBackgroundCheck():
-    print(".............................................................")
     eventData = request.form
     user = eventData['user']
     setTo = int(eventData['setTo'])
-    print(type(setTo))
-    print("....................................................",eventData)
-    print("This is the data we got",user,setTo)
     updateOrCreateVolunteerBackground(user,setTo)
     return ""
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -101,9 +101,13 @@ def removeVolunteerFromEvent(user, eventID):
 
 @admin_bp.route('/updateBackgroundCheck', methods = ['POST'])
 def updateBackgroundCheck():
-    eventData = request.form
-    user = eventData['user']
-    checkPassed = int(eventData['checkPassed'])
-    type = eventData['bgType']
-    setUserBackgroundCheck(user,type, checkPassed)
-    return ""
+    if g.current_user.isCeltsAdmin:
+        eventData = request.form
+        user = eventData['user']
+        checkPassed = int(eventData['checkPassed'])
+        type = eventData['bgType']
+        setUserBackgroundCheck(user,type, checkPassed)
+        return ""
+    else:
+        abort(404)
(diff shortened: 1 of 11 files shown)


app/templates/volunteer/volunteerProfile.html
```diff
@@ -54,7 +54,20 @@ <h2>Which programs do you want to receive information about?</h2>
...
         </small>
       </div>
+
+  </div>
+  <div class="row">
+    <div>
+      {% if g.current_user.isCeltsAdmin %}
+      <label for="check"><strong>Completed their background check:</strong></label>
+      {% set username = user.username%}
+      <select id="check" onchange="updateBackgroundCheck({{username}})">
```
BrianRamsay: Why not just use user.username on line 64?

app/static/js/volunteerIndicateInterest.js
```diff
@@ -24,3 +25,27 @@ $(document).ready(function() {
...
+      setTo : setTo,
+      user: user.id
+  }
+
+  $.ajax({
+    url: "/updateBackgroundCheck",
+    type: "POST",
+    data: data,
+    success: function(s){
+      console.log("success")
```
BrianRamsay: We don't need to log success. We should also flash some sort of visual that the database was updated successfully, like a checkmark next to it that disappears or something

app/static/js/volunteerIndicateInterest.js
```diff
@@ -24,3 +25,27 @@ $(document).ready(function() {
...
   });
 });
+
+
+// onchange="updateBackgroundCheck(1)"
+function updateBackgroundCheck(user){
+  let setTo = $( "#check").val();
+  let data = {
+      setTo : setTo,
+      user: user.id
```
BrianRamsay: What is user.id here? It looks like the template is passing this function a username.

app/static/js/volunteerIndicateInterest.js
```diff
@@ -1,5 +1,6 @@
 $(document).ready(function() {
   $(".form-check-input").click(function updateInterest(el){
+    console.log("passed");
```
BrianRamsay: remove debug output

app/logic/volunteers.py
```diff
@@ -81,3 +81,13 @@ def addVolunteerToEventRsvp(user, volunteerEventID):
 
     except Exception as e:
         return False
+
+def updateOrCreateVolunteerBackground(user, setTo):
```
BrianRamsay: I would call this `setUserBackgroundCheck` or `updateUserBackgroundCheck`

app/controllers/main/routes.py
```diff
@@ -65,7 +67,8 @@ def profilePage(username):
                                interests = interests,
                                interests_ids = interests_ids,
                                upcomingEvents = upcomingEvents,
-                               rsvpedEvents = rsvpedEvents)
+                               rsvpedEvents = rsvpedEvents,
+                               completedBackGroundCheck = completedBackGroundCheck.passBackgroundCheck)
```
BrianRamsay: be consistent with `Background` vs `BackGround`

app/controllers/admin/volunteers.py
```diff
@@ -97,3 +98,11 @@ def removeVolunteerFromEvent(user, eventID):
     (EventRsvp.delete().where(EventRsvp.user==user)).execute()
     flash("Volunteer successfully removed", "success")
     return ""
+
+@admin_bp.route('/updateBackgroundCheck', methods = ['POST'])
+def updateBackgroundCheck():
+    eventData = request.form
+    user = eventData['user']
+    setTo = int(eventData['setTo'])
+    updateOrCreateVolunteerBackground(user,setTo)
```
BrianRamsay: instead of `setTo` I'd call it what it means, not what you are going to do with it. Like `checkPassed`. Also, instead of casting it to an `int`, cast it to a `bool`

BrianRamsay (changes requested):
I've written this comment 3 times now, because information keeps changing. :-) I emailed Sheila and got some more detail on background checks. It's a little more complicated now. Though in some ways less complicated, because before I was going to say we should be tracking the checks by academic year, but that's not something that they would want right now.

There are 3 types of background checks! We need to set and display CAN, SHS, and FBI background check details. This means we will need a `type` field in the table. Probably the best plan is a separate table called `BackgroundCheckType` and `type` is a FK to this table. That table should have `id` (CAN, SHS, FBI) and `description` ("Child Abuse and Neglect", "Safe Hiring Solutions", "Federal Criminal Background"). `BackgroundCheck` should have a primary key of `user` + `type`. This data should live in base_data.py rather than test_data.py.

For the UI, You can decide if there should be one dropdown with the different checks and buttons to pass or reset, or if there should be multiple dropdowns.

app/logic/volunteers.py
sreynit02: changed function to `setUserBackgroundCheck`

app/static/js/volunteerIndicateInterest.js
sreynit02: removed

app/controllers/main/routes.py
sreynit02: changed to Background

app/controllers/admin/volunteers.py
```diff
@@ -97,3 +98,12 @@ def removeVolunteerFromEvent(user, eventID):
     (EventRsvp.delete().where(EventRsvp.user==user)).execute()
     flash("Volunteer successfully removed", "success")
     return ""
+
+@admin_bp.route('/updateBackgroundCheck', methods = ['POST'])
+def updateBackgroundCheck():
+    eventData = request.form
```
BrianRamsay: Only CELTS admins (and not student staff) should be able to change the background check status

app/controllers/main/routes.py
```diff
@@ -58,14 +60,24 @@ def profilePage(username):
         rsvpedEventsList = EventRsvp.select().where(EventRsvp.user == profileUser)
         rsvpedEvents = [event.event.id for event in rsvpedEventsList]
 
+        allUserEntries = list(BackgroundCheck.select().where(BackgroundCheck.user == profileUser))
+        completedBackgroundCheck = {}
+        for entry in allUserEntries:
+            completedBackgroundCheck[entry.type.id] = entry.passBackgroundCheck
```
BrianRamsay: I'd collapse these 3 lines into a dictionary comprehension

app/controllers/main/routes.py
```diff
@@ -58,14 +60,24 @@ def profilePage(username):
...
                                user = profileUser,
                                programs = programs,
                                interests = interests,
                                interests_ids = interests_ids,
                                upcomingEvents = upcomingEvents,
-                               rsvpedEvents = rsvpedEvents)
+                               rsvpedEvents = rsvpedEvents,
+                               backgroundTypes = backgroundTypes,
+                               completedBackgroundCheck = completedBackgroundCheck,
+                               currentUser = g.current_user)
```
BrianRamsay: you don't need to pass a current user. the template can access g.current_user directly

app/logic/volunteers.py
```diff
@@ -81,3 +81,13 @@ def addVolunteerToEventRsvp(user, volunteerEventID):
...
+
+def setUserBackgroundCheck(user,bgType, checkPassed):
+    user = User.get_by_id(user)
+    requiredQuery = BackgroundCheck.select().where(BackgroundCheck.user == user, BackgroundCheck.type == bgType)
+
+    if not requiredQuery.exists():
+        BackgroundCheck.create(user = user, type = bgType, passBackgroundCheck = checkPassed) # Check if this is needed.
+    else:
+        update = BackgroundCheck.update(passBackgroundCheck = checkPassed).where(BackgroundCheck.user == user, BackgroundCheck.type == bgType)
+        update.execute()
```
BrianRamsay: A shorter way with no conditionals since we don't need to maintain any history is to delete any entry with that type, and then insert a new one. Or you could use get_or_create instead of the initial select and edit the field and save.

app/models/backgroundCheck.py
```diff
@@ -0,0 +1,11 @@
+from app.models import *
+from app.models.user import User
+from app.models.backgroundCheckType import BackgroundCheckType
+
+class BackgroundCheck(baseModel):
+    user = ForeignKeyField(User)
+    type = ForeignKeyField(BackgroundCheckType)
+    passBackgroundCheck = BooleanField(default=False)
```
BrianRamsay, line 8: We need a `datePassed` field

app/static/js/volunteerIndicateInterest.js
BrianRamsay: still shows up in the PR...

app/static/js/volunteerIndicateInterest.js
```diff
@@ -24,3 +25,26 @@ $(document).ready(function() {
...
+
+function updateBackgroundCheck(user,bgType){
+
+  let checkPassed = $( "#"+bgType).val();
+  let data = {
+      checkPassed : checkPassed,
+      user: user.id,
+      bgType: bgType
+  }
+  console.log(data)
```
BrianRamsay: no debug output

tests/code/test_volunteers.py
```diff
@@ -109,3 +110,26 @@ def test_updateEventParticipants():
...
+    updatedModel = list(BackgroundCheck.select().where(BackgroundCheck.user == "khatts",BackgroundCheck.type == "SHS"))
+    assert updatedModel[0].passBackgroundCheck == False
+
+    updatebackground = setUserBackgroundCheck("neillz", "FBI",False)
+    updatedModel = list(BackgroundCheck.select().where(BackgroundCheck.user == "neillz",BackgroundCheck.type == "FBI"))
+    assert updatedModel[0].passBackgroundCheck == False
+
+    updatebackground = setUserBackgroundCheck("mupotsal","SHS",True)
+    updatedModel = list(BackgroundCheck.select().where(BackgroundCheck.user == "mupotsal",BackgroundCheck.type == "SHS"))
+    assert updatedModel[0].passBackgroundCheck == True
```
BrianRamsay: since you are interested in one result every time, I would use `.get` instead of `.select` and `list` and `[0]`

BrianRamsay (commented):
Rather than a table with type, description, and the dropdown, I would have just the description and the dropdown in a Background Check section. Avoiding tables helps with mobile responsiveness

app/controllers/admin/volunteers.py
sreynit02: whenever, I cast the data into a boolean, it data is set to always true. Left it as an int instead for now.

app/controllers/admin/volunteers.py
mupotsal: We have made changes to this but the UI should do not allow them to make changes if they are not an admin.

app/controllers/main/routes.py
sreynit02: fixed

app/controllers/main/routes.py
mupotsal: Fixed

app/logic/volunteers.py
mupotsal: Still working on this.

tests/code/test_volunteers.py
mupotsal: fixed

mupotsal:
> Rather than a table with type, description, and the dropdown, I would have just the description and the dropdown in a Background Check section. Avoiding tables helps with mobile responsiveness

Since we are going to have the description, dropdown, and date wouldn't it make sense to have a table?

app/logic/volunteers.py
mupotsal: Worked on this

app/templates/volunteer/volunteerProfile.html
sreynit02: fixed

app/models/backgroundCheck.py
sreynit02: added

BrianRamsay (approved):

Issue #193, opened 2021-11-09
CELTS Student Staff Should Only Have Access to Their Own Program's Events


bryantal:
CELTS Student Staff will mostly be "program managers" and need to have certain permissions so that they are only able to create, edit, and delete events for the program that they work with.  We'll need a permissions table that ties to the user table, and it would have a foreign key that points to a user and tell which program the CELTS Student Staff has permissions for.

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #193: CELTS Student Staff Should Only Have Access to Their Own Program's Events](https://trello.com/c/CHATRanD/347-celts-193-celts-student-staff-should-only-have-access-to-their-own-programs-events)

PR #209, opened 2021-11-26
Student admin permissions

mupotsal:
This PR fixes issue #193 

Summary: 
This PR adds a feature where Celts Student Admins have restrictions on the programs they can create or edit. A Celt's student manager can only create or edit events that belong to the programs where they are project managers. Celts Admins can set what programs the CeltsStudentAdmins can be managers for. 

Test: 
Login the app as neillz. Try to create an event and you see the page below
Screen Shot 2021-11-26 at 3 29 58 PM. This is because right now there are no restrictions on the programs the user can create events from. Now make neillz a Celts Student Admin in the database. Try to create an event again. You should see the image below

 Screen Shot 2021-11-26 at 3 32 21 PM.
This is because neillz is a program manager for the program where Berea Tutoring and Empty bowls events belong to.
Screen Shot 2021-11-26 at 3 33 36 PM

Additionally, if you log in as a celts Admin and navigate to a profile page, you can assign what programs a student admin can be a manager for. 
Screen Shot 2021-11-27 at 12 53 37 AM
If you login as a student  you should see profile like the following. 
Screen Shot 2021-11-27 at 12 54 16 AM


app/controllers/admin/routes.py
```diff
@@ -40,7 +43,15 @@ def switchUser():
 
 @admin_bp.route('/template_select')
 def templateSelect():
-    allprograms = Program.select().order_by(Program.programName)
+    allprograms = []
+    if g.current_user.isCeltsStudentStaff:
+        studentManagerPrograms = list(StudentManager.select().where(StudentManager.user==g.current_user))
+        permissionPrograms = [entry.program.id for entry in studentManagerPrograms]
+        editablePrograms = Program.select().where(Program.id.in_(permissionPrograms)).order_by(Program.programName)
+        allprograms = editablePrograms
```
BrianRamsay: Getting the programs someone has permissions to should be a separate function. I think we will use this several different places.

Also, use one query to get the information, not two with a list comprehension in between

app/controllers/admin/routes.py
```diff
@@ -94,12 +103,16 @@ def createEvent(templateid, programid=None):
     # make sure our data is the same regardless of GET or POST
     preprocessEventData(eventData)
     futureTerms = selectSurroundingTerms(g.current_term)
-
+    queryManager = StudentManager.select().where(StudentManager.user == g.current_user,StudentManager.program==programid)
+    isProgramManager = False
+    if queryManager.exists():
+        isProgramManager = True
```
BrianRamsay: Isn't this what hasPrivilege does?

app/controllers/admin/routes.py
```diff
@@ -129,16 +142,21 @@ def editEvent(eventId):
     userHasRSVPed = EventRsvp.get_or_none(EventRsvp.user == g.current_user, EventRsvp.event == event)
     isPastEvent = (datetime.now() >= datetime.combine(event.startDate, event.timeStart))
 
+    programSelect = ProgramEvent.get(event=eventId)
```
BrianRamsay: You can use `event.singleProgram` for this, since you are assuming one program. Make sure you handle the case where the program doesn't have an event (currently not handled)

app/controllers/admin/userManagement.py
```diff
@@ -39,9 +39,20 @@ def manageUsers():
...
             flash(username+ " is no longer a Celts Student Staff", 'success')
-
     return ("success")
 
+@admin_bp.route('/updateManagers', methods=['POST','GET'])
+def updateProgramManagers():
+    eventData = request.form
+    user = eventData['user']
+    programID = eventData['programID']
+    status = eventData['status']
```
BrianRamsay: I'm not sure there's a reason to use temporary variables here, since you aren't doing anything with them, just passing them along

app/controllers/admin/userManagement.py
```diff
@@ -39,9 +39,20 @@ def manageUsers():
...
+    if status == "true":
+        addProgramManager(user,int(programID))
+    else:
+        removeProgramManager(user,int(programID))
+    return ""
+
```
BrianRamsay, line 66: You should do some error checking here, since there isn't any in the functions.

app/static/js/volunteerIndicateInterest.js
```diff
@@ -24,3 +22,30 @@ $(document).ready(function() {
...
+
+function updateManagers(el,user,status){
+  var programID = el.id;
+  var user = (user[0].id);
+  if (status == undefined){
+    status = true;
+  }
+  let data = {
+      programID : programID,
+      user : user,
```
BrianRamsay: this should be userID, not user

app/templates/admin/createSingleEvent.html
```diff
@@ -70,7 +71,10 @@ <h1 id="title">{{page_title}}</h1>
 
         <div class="d-grid gap-1.5">
           <div class="form-group">
-            <label for='name'> <strong>Event Name</strong></label> <br>
+            <label for='name'> <strong>Event Name {{isProgramManager}}
+
+
+</strong></label> <br>
```
BrianRamsay: Not sure what's going on here. Why do we need a boolean and some extra spaces in the label?

app/templates/volunteer/volunteerProfile.html
```diff
@@ -66,6 +66,45 @@ <h2>Which programs do you want to receive information about?</h2>
...
+            {% set status = 'checked' %}
+          {% else %}
+          {% endif %}
+              <input  type="checkbox" id="{{ program.id }}" {{ status }}  onclick="updateManagers(this,{{user.username}},{{status}})">
+              <label  style="padding-right:50px;" for="{{ program.id }}">{{program.programName}}</label>
+      {% endfor %}
+    </div>
+  </div>
+  {% endif %}
+{% elif g.current_user.isCeltsStudentStaff %}
```
BrianRamsay: I don't think we need to restrict showing the manager information only to student staff

app/templates/volunteer/volunteerProfile.html
```diff
@@ -66,6 +66,45 @@ <h2>Which programs do you want to receive information about?</h2>
...
+        {% if program.id in permissionPrograms %}
+          <p  style="padding-right:50px;">{{program.programName}}</p>
+        {% else %}
+        {% endif %}
+
+    {% endfor %}
+  </div>
+</div>
+{% endif %}
+{% endif %}
```
BrianRamsay, line 100: The nested ifs should be indented

tests/code/test_userManagement.py
```diff
@@ -36,6 +36,21 @@ def test_modifyCeltsStudentStaff():
     with pytest.raises(DoesNotExist):
         removeCeltsStudentStaff("1234")
 
+@pytest.mark.integration
+def test_modifyStudentManager():
+    user = "mupotsal"
+    current_user = User.get(User.username==user)
```
BrianRamsay: You don't use this variable

tests/code/test_userManagement.py
```diff
@@ -36,6 +36,21 @@ def test_modifyCeltsStudentStaff():
...
+    currentManagerStatus = StudentManager.get(user=user,program=2)
+    assert currentManagerStatus.program.id ==2
+    removeProgramManager(user,2)
+    assert hasPrivilege(user, 2) == False
+    currentManagerStatus = StudentManager.select().where(StudentManager.user==user,StudentManager.program ==2)
+    assert currentManagerStatus.exists() == False
+    addProgramManager(user,2)
+    assert hasPrivilege(user, 2) == True
+    currentManagerStatus = StudentManager.get(user=user,program=2)
+    assert currentManagerStatus.program.id ==2
```
BrianRamsay, line 68: I'd put some spaces in here between related chunks of code, with maybe a comment or two indicating the case being tested

app/static/js/volunteerIndicateInterest.js
```diff
@@ -24,3 +22,30 @@ $(document).ready(function() {
     });
   });
 });
+
+function updateManagers(el,user,status){
+  var programID = el.id;
+  var user = (user[0].id);
+  if (status == undefined){
+    status = true;
```
BrianRamsay: status is either `''` (undefined), `'checked'`, or `true` at different times, and you pass `'checked'` when you are removing, as far as I can tell. This is confusing to me. I would separate what you pass to the controller from the string values `''` and `'checked'` from the template. to me, what you are passing is the action you want to take, 'add' or 'remove'.

BrianRamsay: Also, you don't really need to pass the status here. you can use the element you pass in to see if the checkbox is checked or not

BrianRamsay (changes requested):

app/controllers/admin/routes.py
mupotsal: Made a separate function for getting all programs and refactored the queries.

app/controllers/admin/routes.py
mupotsal: Yes, and fixed it.

app/controllers/admin/routes.py
mupotsal: I have used the event.single.program. However, I am not sure I understood the handling that you stated above. The only time we get into this route **'/event/eventId/edit'** is when we want to edit an existing event so it is guaranteed that the event always exists.

app/controllers/admin/userManagement.py
mupotsal: refactored, I did it like that at first for clarity.

app/controllers/admin/userManagement.py
mupotsal: Fixed

app/static/js/volunteerIndicateInterest.js
mupotsal: Fixed.

app/templates/admin/createSingleEvent.html
mupotsal: Fixed

app/templates/volunteer/volunteerProfile.html
mupotsal: Fixed.

tests/code/test_userManagement.py
mupotsal: Done

app/static/js/volunteerIndicateInterest.js
mupotsal: fixed

app/controllers/admin/userManagement.py
BrianRamsay: When you run into an error you should be returning a 500 status code

tests/code/test_userManagement.py
BrianRamsay: I don't see any newlines added to separate the sections.

app/templates/volunteer/volunteerProfile.html
```diff
@@ -67,8 +67,50 @@ <h2>Which programs do you want to receive information about?</h2>
...
+          {% else %}
+          {% endif %}
+              <input  type="checkbox" id="{{ program.id }}" {{ 'checked' if status==1 else 'unchecked' }}  onclick="updateManagers(this,{{user.username}},{{status}})">
+              <label  style="padding-right:50px;" for="{{ program.id }}">{{program.programName}}</label>
+      {% endfor %}
+    </div>
+  </div>
+  {% endif %}
+{% else %}
+<div class="row">
```
BrianRamsay: There's a missing case here. We probably don't want to show this section at all if the user is a regular student (not staff)

BrianRamsay (changes requested):
Just a couple of comments in here. Also, merge development in, resolve conflicts, and check that the test suites pass.

tests/code/test_userManagement.py
sreynit02: added newlines to separate cases.

app/templates/volunteer/volunteerProfile.html
mupotsal: Fixed by adding another case.

app/controllers/admin/routes.py
```diff
@@ -41,7 +44,12 @@ def switchUser():
 
 @admin_bp.route('/template_select')
 def templateSelect():
-    allprograms = Program.select().order_by(Program.programName)
+    allprograms = []
```
BrianRamsay: We need to do a check at the beginning that only admins can see this page, otherwise 403. Right now regular students have the admin view

app/controllers/admin/userManagement.py
```diff
@@ -40,9 +40,26 @@ def manageUsers():
...
+        except:
+            flash('Error while trying to add a manager.')
+    elif int(eventData['status']) == 1:
+        try:
+            removeProgramManager(eventData['userID'],int(eventData['programID']))
+        except:
+            flash('Error while trying to remove a manager.')
+    else:
+        flash('Error while removing a manager.')
+    abort(500)
```
BrianRamsay: You are aborting every time from this controller, so it will always show as a failed request from the user's side.

app/logic/userManagement.py
```diff
@@ -53,8 +72,11 @@ def addNextTerm():
...
             year=newYear,
-            academicYear=newAY, 
+            academicYear=newAY,
             isSummer="Summer" in newDescription.split())
     newTerm.save()
 
     return newTerm
+
+def getPrograms():
+    return Program.select().join(StudentManager).where(StudentManager.user==g.current_user).order_by(Program.programName)
```
BrianRamsay: This should take a passed-in user, rather than use g.current_user

BrianRamsay (changes requested):
Getting close! Thanks

app/logic/userManagement.py
mupotsal: Fixed!

app/controllers/admin/routes.py
mupotsal: Fixed

app/controllers/admin/userManagement.py
BrianRamsay: Also, this should only be a POST route. There's no GET requests coming to this route.

app/controllers/admin/userManagement.py
```diff
@@ -40,9 +40,31 @@ def manageUsers():
...
+    elif int(eventData['status']) == 1:
+        try:
+            removeProgramManager(eventData['userID'],int(eventData['programID']))
+        except:
+            flash('Error while trying to remove a manager.')
+            return "Error while removing manager."
+
+    else:
+        flash('Error while removing a manager.')
+        return "Error while removing manager."
```
BrianRamsay: All of these errors should print the actual error to the console so we can debug the problem. Right now you are catching the exceptions and not telling anyone except the user.

Also, errors should return an appropriate http status code. Before you were returning 500 for everything. Now you are returning 200 for everything (by default). Neither is correct.

BrianRamsay (changes requested):
Still some issues with one controller method, and then we are good to go.

app/controllers/admin/userManagement.py
```diff
@@ -40,9 +40,31 @@ def manageUsers():
...
+    elif int(eventData['status']) == 1:
+        try:
+            return(removeProgramManager(eventData['userID'],int(eventData['programID'])))
+        except:
+            flash('Error while removing a manager.','warning')
+            abort(500,"'Error while trying to remove a manager.'")
+
+    else:
+        flash('Error while removing a manager.','warning')
+        abort(500,"'Error while trying to remove a manager.'")
```
BrianRamsay: We're still missing a piece here. What we need in error handling is:
- A message to the user that doesn't contain potential system information
- Status and/or data to the front-end to know how to respond.
- Debug information to the admins

(Sometimes the first two are combined)

We are missing useful information for the admins here. "Error while trying to remove a manager" doesn't tell us what went wrong, which means someone will have to be able to replicate the error and edit the file to add debug output in order to get more information. That's not nice. You need to give the exception information to the admins (and not to the web user).

Also, that last case is being handled the same but it's not the same. You know exactly what is wrong there - 'status' was not 1 or 0. You could get rid of it entirely by having two routes, `addProgramManager` and `removeProgramManager` instead of a `status` index and and if/else

BrianRamsay (changes requested):
If you aren't sure what I'm looking for, let's meet up.

BrianRamsay (approved):
Woohoo!

PR #210, opened 2021-12-15
Creates an Email handler Class 

hmanalai:
* Creates a new email handler class in a file name `emailHandler2.py`. 
* The class handles processing raw data, fetching new data, creating the email, sending and logging the email. 
* This PR also adds more functionality to the email modal, such as populating the modal with data from the database. 

Note: I added a new model (EmailLog), and new data in `test_data.py`.


app/controllers/events/email.py
```diff
@@ -0,0 +1,53 @@
...
+from app.models.programEvent import ProgramEvent
+from app.models.emailTemplate import EmailTemplate
+from app.models.emailLog import EmailLog
+from app.controllers.main import main_bp
+from app.logic.emailHandler import getInterestedEmails, getParticipantEmails, emailHandler
+from app import app
+
+@main_bp.route('/retrieveEmailTemplate', methods=['GET'])
+def retrieveEmailTemplate():
+    # Couldn't find a way to do this with render_template
```
hmanalai: I was not able to populate the email modal with data using `render_template`, so I did it this way. I'm not sure how else I can fix this.

app/controllers/events/email.py
```diff
@@ -0,0 +1,53 @@
...
+
+@main_bp.route('/fetchEmailLogData/<eventId>', methods=['GET'])
+def fetchEmailLogData(eventId):
+    dateSent = recipients = None
+    # TODO: Bruteforce implementation. Need to figure out a way to get the most recent email log for an event
+    emailLogs = EmailLog.select().where(EmailLog.event==eventId).order_by(EmailLog.dateSent.asc())
+    for emailLog in emailLogs:
+        recipients = emailLog.recipientsCategory
+        dateSent = emailLog.dateSent.strftime('%m/%d/%Y')
+    return {'recipients':recipients, 'dateSent':dateSent}
```
hmanalai: This is a very brute force way to get the most recent email log. I searched about how to get the most recent using peewee but wasn't successful. Any ideas on how to do this?

In the email log, each email is tied to an event. Since users can send multiple emails for a single event, in the modal where it says "last email was sent ...." we need to show the latest one.

BrianRamsay: One option is to do it within the query, either with LIMIT or with a subselect within the query if you are using a database that doesn't support LIMIT. Since we can use some peewee fanciness, that's not necessary, though. You can just use .get() to get the first row (changing the order by). 

`EmailLog.select().where(EmailLog.event==eventId).order_by(EmailLog.dateSent.desc().get())`

app/controllers/events/email.py
BrianRamsay: If you're not sure how to set it up on the javascript side so the ajax populates the modal from the response, you could use `render_template` here and store the resulting html under the `'body'` key, rather than storing the template.

app/controllers/events/email.py
```diff
@@ -0,0 +1,53 @@
...
+    if emailGroup=='interested' and programID:
+        emails = getInterestedEmails(programID)
+
+    elif emailGroup == 'eventParticipant' and eventID:
+        email = getInterestedEmails(eventID)
+
+    else:
+        flash("Unable to determine email recipients", "danger")
+
+    return email
```
BrianRamsay: Still a problem here between `email` and `emails`.  But since you mention `#TODO: Do we still need this?`, I would figure out the answer to that before you see if you need to test this better and fix it.

app/logic/emailHandler.py
```diff
@@ -0,0 +1,69 @@
...
+import webbrowser
+import time
+from pathlib import Path
+
+def getInterestedEmails(listOfPrograms):
+
+    """ Gets emails of students interested in a program. """
+    emails = []
+
+    volunteersToEmail = User.select(User.email).join(Interest).join(Program, on=(Program.id == Interest.program)).where(Program.id.in_([p.id for p in listOfPrograms]))
```
BrianRamsay: This line needs to be split up

app/logic/emailHandler.py
```diff
@@ -0,0 +1,69 @@
...
+        if 'sendIndividually' in self.emailInfo:
+            if app.config.mail['MAIL_OVERRIDE_ALL']:
+                message.recipients = [app.config['MAIL_OVERRIDE_ALL']]
+            with self.mail.connect() as conn:
+                for email in emails:
+                    message.recipients = [email]
+                    conn.send(message)
+        else:
+            try:
+                if app.config['ENV'] == 'testing':
```
BrianRamsay: We probably want to use the override for 'development' as well

app/logic/emailHandler2.py
```diff
@@ -0,0 +1,189 @@
...
+        if 'slCourseId' in self.raw_form_data:
+            self.sl_course_id = self.raw_form_data['slCourseId']
+
+    def fetch_event_programs(self, programId):
+        """ Fetches all the programs of a particular event """
+        # Non-student-led programs have "Unknown" as their id
+        # ---Q: maybe this id should be changed to something more specific?
+        if programId == 'Unknown':
+            programs = ProgramEvent.select(ProgramEvent.program).where(ProgramEvent.event==self.event.id)
+            return [program.program for program in programs.objects()]
```
BrianRamsay: I would rename the variable in 68 as `programEvents`, since that is the object it is fetching.

BrianRamsay (changes requested):
I'm only about halfway through, but submitting this so you can see the suggestions while I'm looking at the rest.

app/logic/emailHandler2.py
```diff
@@ -0,0 +1,189 @@
...
+        if recipients_category == "Interested":
+            recipients = (User.select()
+                .join(Interest)
+                .join(Program, on=(Program.id==Interest.program))
+                .where(Program.id.in_([p.id for p in self.program_ids])))
+
+        if recipients_category == "RSVP'd":
+            recipients = (User.select()
+                .join(EventParticipant)
+                .where(EventParticipant.event==self.event.id))
```
BrianRamsay: If this category is RSVP, shouldn't it use EventRsvp instead of EventParticipant?

app/logic/emailHandler2.py
```diff
@@ -0,0 +1,189 @@
...
+
+        return [recipient for recipient in recipients]
+
+    def replace_general_template_placholders(self, email_body=None):
+        """ Replaces all template placeholders except name """
+        domain = urlparse(request.base_url) # TODO: how to avoid using request?
```
BrianRamsay: I would pass the request in when you create the EmailHandler object. Dependency injection reduces the coupling between the object and the application and makes it easier to test, too

BrianRamsay: or maybe not pass the request, but pass the domain object from urlparse

BrianRamsay (changes requested):
Love the unit tests.

app/logic/emailHandler2.py
hmanalai: Yes, it should be EventRsvp. We will change it.

app/logic/emailHandler2.py
hmanalai: Do you mean something like this?

``` 
url_domain = urlparse(request.base_url).netloc 
email = EmailHanlder(raw_form_data, url_domain)
```

And modify `EmailHandler` class to take in the additional `url_domain` parameter?

app/logic/emailHandler.py
hmanalai: `emailHandler.py` is from the original PR. We didn't change this one because we didn't know if we could delete it. 
If the class defined in `emailHanlder2.py` is doing everything we need it to do, then we can delete this one. 

I believe it does, so we might as well delete this file altogether.

app/logic/emailHandler2.py
BrianRamsay: yes

app/logic/emailHandler.py
```diff
@@ -0,0 +1,188 @@
...
+        # Recipients
+        if 'recipientsCategory' in self.raw_form_data:
+            self.recipients_category = self.raw_form_data['recipientsCategory']
+            self.recipients = self.retrieve_recipients(self.recipients_category)
+
+        # Service Learning Course
+        if 'slCourseId' in self.raw_form_data:
+            self.sl_course_id = self.raw_form_data['slCourseId']
+
+    def fetch_event_programs(self, programId):
```
BrianRamsay: be consistent with under_scores vs camelCase

app/logic/emailHandler.py
```diff
@@ -0,0 +1,188 @@
...
+    def update_email_template(self):
+        try:
+            self.process_data()
+            (EmailTemplate.update({
+                EmailTemplate.subject: self.subject,
+                EmailTemplate.body: self.body,
+                EmailTemplate.replyToAddress: self.reply_to
+            }).where(EmailTemplate.purpose==self.template_identifier)).execute()
+        except Exception as e:
+            print("Error updating email template record: ", e)
```
BrianRamsay, line 188: This method should indicate failure (exception or return value) so that the calling code can handle success vs failure.

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #211, opened 2021-12-20
Admin logs

Ebenezer997:
Created a page to keep track of  Admin activities in the celts application.
Some of the admin activities that are tracked are creating events, editing events, deleting events, and banning a user

Test:
Log in to the app as admin. perform activities such as creating events, editing events, changing background checks, and deleting events. Click on admin logs on the sidebar and you will see all your activities have been recorded.

sssss

### DIFF
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -21,6 +21,7 @@ from app.models.adminLogs import AdminLogs
 from app.logic.participants import trainedParticipants
 from app.logic.volunteers import getEventLengthInHours
 from app.logic.utils import selectSurroundingTerms
+from app.logic.adminLogs import createLog
 from app.logic.events import deleteEvent, getAllFacilitators, attemptSaveEvent, preprocessEventData, calculateRecurringEventFrequency
 from app.logic.courseManagement import pendingCourses, approvedCourses
 from app.controllers.admin import admin_bp
@@ -87,6 +88,8 @@ def createEvent(templateid, programid=None):
             validationErrorMessage = "Unknown Error Saving Event. Please try again"
 
         if saveSuccess:
+            event = Event.get(name=eventData['name'],term=eventData['term'],timeStart=eventData['timeStart'],startDate=eventData['startDate'])
+            createLog("Created event with id:"+str(event.id)+" :name="+event.name+" startDate:"+str(event.startDate) + " found at:"+"/event/"+str(event.id)+ "/edit")
             noun = (eventData['isRecurring'] == 'on' and "Events" or "Event") # pluralize
             flash(f"{noun} successfully created!", 'success')
             return redirect(url_for("main.events", term = eventData['term']))
@@ -121,6 +124,8 @@ def editEvent(eventId):
         eventData = request.form.copy()
         saveSuccess, validationErrorMessage = attemptSaveEvent(eventData)
         if saveSuccess:
+            event = Event.get(name=eventData['name'],term=eventData['term'],timeStart=eventData['timeStart'],startDate=eventData['startDate'])
+            createLog("Edited event with id:"+str(event.id)+" :name="+event.name+" startDate:"+str(event.startDate) + " found at:"+"/event/"+str(event.id)+ "/edit")
             flash("Event successfully updated!", "success")
             return redirect(url_for("admin.editEvent", eventId = eventId))
         else:
@@ -143,7 +148,9 @@ def deleteRoute(eventId):
 
     try:
         term = Event.get(Event.id == eventId).term
+        event = Event.get(Event.id==eventId)
         deleteEvent(eventId)
+        createLog("Deleted event with id"+" "+eventId, +" name: "+event.name)
         flash("Event removed", "success")
         return redirect(url_for("main.events"))
 
@@ -198,7 +205,7 @@ def courseManagement(term = None):
                             term = term)
 @admin_bp.route('/adminLogs', methods = ['GET', 'POST'])
 def adminlogs():
-    allLogs = AdminLogs.select()
+    allLogs = AdminLogs.select().order_by(AdminLogs.createdOn.desc())
 
     return render_template("/admin/adminLogs.html",
                             allLogs = allLogs)
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -18,6 +18,7 @@ from app.models.outsideParticipant import OutsideParticipant
 from app.models.eventParticipant import EventParticipant
 from app.models.programEvent import ProgramEvent
 from app.models.adminLogs import AdminLogs
+from app.models.studentManager import StudentManager
 from app.logic.participants import trainedParticipants
 from app.logic.volunteers import getEventLengthInHours
 from app.logic.utils import selectSurroundingTerms
@@ -26,6 +27,8 @@ from app.logic.courseManagement import pendingCourses, approvedCourses
 from app.controllers.admin import admin_bp
 from app.controllers.admin.volunteers import getVolunteers
 from app.controllers.admin.userManagement import manageUsers
+from app.logic.userManagement import hasPrivilege, getPrograms
+
 
 
 @admin_bp.route('/switch_user', methods=['POST'])
@@ -42,13 +45,22 @@ def switchUser():
 
 @admin_bp.route('/eventTemplates')
 def templateSelect():
-    allprograms = Program.select().order_by(Program.programName)
-    visibleTemplates = EventTemplate.select().where(EventTemplate.isVisible==True).order_by(EventTemplate.name)
+    if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff:
+        allprograms = []
+        if g.current_user.isCeltsStudentStaff:
+            allprograms = getPrograms(g.current_user)
+        else:
+            allprograms = Program.select().order_by(Program.programName)
 
-    return render_template("/events/template_selector.html",
-                programs=allprograms,
-                templates=visibleTemplates
-            )
+        visibleTemplates = EventTemplate.select().where(EventTemplate.isVisible==True).order_by(EventTemplate.name)
+
+        return render_template("/events/template_selector.html",
+                    programs=allprograms,
+                    templates=visibleTemplates
+                )
+    else:
+        abort(403)
+
 
 @admin_bp.route('/eventTemplates/<templateid>/create', methods=['GET','POST'])
 @admin_bp.route('/eventTemplates/<templateid>/<programid>/create', methods=['GET','POST'])
@@ -94,13 +106,15 @@ def createEvent(templateid, programid=None):
 
     # make sure our data is the same regardless of GET or POST
     preprocessEventData(eventData)
+    isProgramManager = hasPrivilege(g.current_user,programid)
     futureTerms = selectSurroundingTerms(g.current_term, prevTerms=0)
 
     return render_template(f"/admin/{template.templateFile}",
             template = template,
             eventData = eventData,
             futureTerms = futureTerms,
-            allFacilitators = getAllFacilitators())
+            allFacilitators = getAllFacilitators(),
+            isProgramManager = isProgramManager)
 
 
 @admin_bp.route('/eventsList/<eventId>/edit', methods=['GET','POST'])
@@ -129,17 +143,19 @@ def editEvent(eventId):
     futureTerms = selectSurroundingTerms(g.current_term)
     userHasRSVPed = EventRsvp.get_or_none(EventRsvp.user == g.current_user, EventRsvp.event == event)
     isPastEvent = (datetime.now() >= datetime.combine(event.startDate, event.timeStart))
+    program = event.singleProgram
 
+    isProgramManager = hasPrivilege(g.current_user,program)
     return render_template("admin/createSingleEvent.html",
                             eventData = eventData,
                             allFacilitators = getAllFacilitators(),
                             futureTerms = futureTerms,
                             isPastEvent = isPastEvent,
-                            userHasRSVPed = userHasRSVPed)
+                            userHasRSVPed = userHasRSVPed,
+                            isProgramManager = isProgramManager)
 
 @admin_bp.route('/event/<eventId>/delete', methods=['POST'])
 def deleteRoute(eventId):
-
     try:
         deleteEvent(eventId)
         flash("Event successfully deleted.", "success")
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -89,7 +89,7 @@ def createEvent(templateid, programid=None):
 
         if saveSuccess:
             event = Event.get(name=eventData['name'],term=eventData['term'],timeStart=eventData['timeStart'],startDate=eventData['startDate'])
-            createLog("Created event with id:"+str(event.id)+" :name="+event.name+" startDate:"+str(event.startDate) + " found at:"+"/event/"+str(event.id)+ "/edit")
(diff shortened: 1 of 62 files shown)


app/controllers/admin/volunteers.py
```diff
@@ -107,7 +109,9 @@ def updateBackgroundCheck():
         checkPassed = int(eventData['checkPassed'])
         type = eventData['bgType']
         setUserBackgroundCheck(user,type, checkPassed)
-        return ""
+        user = User.get_by_id(user)
+        createLog(user.firstName + " "+ user.lastName +"'s backgroundCheck for " +type+ " has been updated to "+ str(bool(checkPassed)))
```
hmanalai: This sentence should be an action sentence like the ones before. It would be modified to "Updated user's background check to ..."

app/controllers/admin/routes.py
```diff
@@ -85,6 +88,8 @@ def createEvent(templateid, programid=None):
             validationErrorMessage = "Unknown Error Saving Event. Please try again"
 
         if saveSuccess:
+            event = Event.get(name=eventData['name'],term=eventData['term'],timeStart=eventData['timeStart'],startDate=eventData['startDate'])
+            createLog("Created event with id:"+str(event.id)+" name: "+event.name+" startDate: "+str(event.startDate) + " found at: "+"/event/"+str(event.id)+ "/edit")
```
hmanalai: Since we are showing these sentences in the log, they should be full sentences and the date should be formatted. Also, if we are going to add a link, it should be a hyperlink so the user can click on it.

Right now it looks like this 

"Created event with id:16 name: All Volunteer Training startDate: 2022-01-20 found at: /event/16/edit"

It would be better if we change it to something like: 

"Created All Volunteer Training event, with start date 01/20/2022." and maybe add a link. 

The same comment applies to other logs too.

app/controllers/admin/routes.py
```diff
@@ -119,6 +124,8 @@ def editEvent(eventId):
         eventData = request.form.copy()
         saveSuccess, validationErrorMessage = attemptSaveEvent(eventData)
         if saveSuccess:
+            event = Event.get(name=eventData['name'],term=eventData['term'],timeStart=eventData['timeStart'],startDate=eventData['startDate'])
```
hmanalai: Can't we get the event using just the name and term? Or is there a possibility of having multiple events with the same name?

Also, would it be possible to have attemptSaveEvent to return the event id (or even better, the event object) instead of boolean? Of course, if that messes up the overall logic then please ignore.

app/controllers/admin/routes.py
hmanalai: Also, I don't think it's necessary to include the event id.

hmanalai: I would also suggest using f string instead of + to concatenate sentences. For example, instead of what's here, we could do 

`f'Created {event.name}, with start date of {event.startDate}' `

app/controllers/admin/routes.py
mupotsal: I think it is possible to have events with the same name on the same semester. However, I can change the logic and we see what Brian thinks.

app/logic/adminLogs.py
```diff
@@ -0,0 +1,9 @@
+from flask import g
+from datetime import datetime
+from app.models.user import User
+from app.models.adminLogs import AdminLogs
+
+def createLog(content):
+    date = datetime.strptime(datetime.now().strftime('%Y %m %d %H:%M:%S'),'%Y %m %d %H:%M:%S')
```
BrianRamsay: you don't need to do all of these formatting to create a datetime object. `datetime.now()` is already one.

BrianRamsay (changes requested):
There's no test for this! The very first thing you do when you are creating a backend function should be to create the test. Even before you start the function.

app/logic/adminLogs.py
mupotsal: Fixed this by just using date.now()

app/controllers/admin/routes.py
```diff
@@ -86,6 +88,8 @@ def createEvent(templateid, programid=None):
             validationErrorMessage = "Unknown Error Saving Event. Please try again"
 
         if saveSuccess:
+            event = Event.get(name=eventData['name'],term=eventData['term'],timeStart=eventData['timeStart'],startDate=eventData['startDate'])
+            createLog(f'Created {event.name}, with start date {str(datetime.strftime(event.startDate, "%m/%d/%Y"))}.')
```
BrianRamsay: I would add these log calls where the action actually happens. So in this case, in the function that actually saves the event. That way if events are created from any source, not just from this controller, it will be logged.

Also, you'll have the information you need to generate the log, without having to query for the data again.

You will have to update the other tests to use a g.current_user, or else they will all break like your current one is.

tests/code/test_logsCreation.py
```diff
@@ -0,0 +1,13 @@
...
+from app.logic.adminLogs import createLog
+from app.models import mainDB
+
+@pytest.mark.integration
+def test_createLogs():
+    with mainDB.atomic() as transaction:
+        currentLogsCount = len(list(AdminLogs.select()))
+        createLog("This is a test log")
+        assert len(list(AdminLogs.select())) == currentLogsCount +1
+        transaction.rollback()
```
BrianRamsay: I like that you are basing this test on the current count of logs, and not on a number that will change with changes to the test data! You should test and make sure that the current user and date are being added properly, though. 

Your test isn't actually passing, because you don't initialize g.current_user

BrianRamsay (changes requested):
I have an architectural change suggestion.

app/controllers/admin/routes.py
BrianRamsay: I had this same thought, Hila. Since we can return multiple events from `saveEventToDb`, this wouldn't quite be accurate even if we made `attemptSaveEvent` return them. I could see an argument to having the first parameter `attemptSaveEvent` returns be an event array though. Ultimately my suggestion is that we should move the logging next to the actual creation of the event. That way you don't have to re-query the database, and the logging happens anytime an event is created without having to remember to add logging every time you call the function.

BrianRamsay:
I think you should put the function and test in with the utility functions, rather than having their own files.

tests/code/test_logsCreation.py
mupotsal: This is resolved. I also added more tests for the required entities.

app/controllers/admin/routes.py
mupotsal: Solved

app/controllers/admin/routes.py
```diff
@@ -120,6 +120,7 @@ def editEvent(eventId):
         eventData = request.form.copy()
         saveSuccess, validationErrorMessage = attemptSaveEvent(eventData)
         if saveSuccess:
+            event = Event.get(name=eventData['name'],term=eventData['term'])
```
BrianRamsay: You don't need to get the event here anymore

app/controllers/admin/routes.py
```diff
@@ -142,6 +143,7 @@ def deleteRoute(eventId):
 
     try:
         term = Event.get(Event.id == eventId).term
+        event = Event.get(Event.id==eventId)
```
BrianRamsay: Both term and event are unnecessary now

app/controllers/admin/volunteers.py
```diff
@@ -107,7 +108,8 @@ def updateBackgroundCheck():
         checkPassed = int(eventData['checkPassed'])
         type = eventData['bgType']
         setUserBackgroundCheck(user,type, checkPassed)
-        return ""
+        user = User.get_by_id(user)
```
BrianRamsay: What is the point of querying the database here?

app/controllers/admin/volunteers.py
```diff
@@ -107,7 +108,8 @@ def updateBackgroundCheck():
...
+        return " "
     else:
         abort(404)
         return ""
```
BrianRamsay: `return` after an `abort` will never be called

app/logic/adminLogs.py
```diff
@@ -0,0 +1,9 @@
+from flask import g
+from datetime import datetime
+from app.models.user import User
```
BrianRamsay: you don't need to import User

app/logic/events.py
```diff
@@ -28,6 +30,18 @@ def deleteEvent(eventId):
     event = Event.get_or_none(Event.id == eventId)
     if event:
         event.delete_instance(recursive = True, delete_nullable = True)
+        if event.startDate:
+            try:
+                createLog(f"Deleted event: {event.name}, which had a startdate of {str(datetime.datetime.strftime(event.startDate, '%m/%d/%Y'))}")
```
BrianRamsay: No need for casting a string to a string

BrianRamsay: Actually, why do you need two different cases here? Either event.startDate is a date coming from the database or it is a string. It can't be both.

app/logic/userManagement.py
```diff
@@ -1,28 +1,37 @@
...
     user.isCeltsAdmin = True
     user.save()
+    createLog(f'Made {user.firstName} {user.lastName} a Celts Admin.')
+
 
 def addCeltsStudentStaff(user):
     user = User.get_by_id(user)
     user.isCeltsStudentStaff = True
     user.save()
+    createLog(f'Made {user.firstName} {user.lastName} a Celts Student Staff.')
```
BrianRamsay: I would change this to read more fluently. "Made ...  a CELTS student staff member"

app/logic/userManagement.py
```diff
@@ -1,28 +1,37 @@
...
 from flask import g, session
+from app.logic.adminLogs import createLog
 
 from playhouse.shortcuts import model_to_dict
 
 def addCeltsAdmin(user):
     user = User.get_by_id(user)
     user.isCeltsAdmin = True
     user.save()
+    createLog(f'Made {user.firstName} {user.lastName} a Celts Admin.')
```
BrianRamsay: "Celts" should be "CELTS" everywhere we have user-accessible text

app/logic/userManagement.py
```diff
@@ -1,28 +1,37 @@
...
     user.isCeltsAdmin = False
     user.save()
+    createLog(f'Removed {user.firstName} {user.lastName} from Celts Admins.')
+
 
 def removeCeltsStudentStaff(user):
     user = User.get_by_id(user)
     user.isCeltsStudentStaff = False
     user.save()
+    createLog(f'Removed {user.firstName} {user.lastName} from a Celts Student Staff.')
```
BrianRamsay: "Removed ... as a CELTS student staff member."

app/logic/volunteers.py
```diff
@@ -89,3 +91,4 @@ def setUserBackgroundCheck(user,bgType, checkPassed):
     deleteInstance.execute()
     update = BackgroundCheck.create(user=user, type=bgType, passBackgroundCheck=checkPassed, datePassed=today)
     update.save()
+    createLog(f"Updated {user.firstName} {user.lastName}'s background check for {bgType} to {str(bool(checkPassed))}.")
```
BrianRamsay: isn't checkPassed already boolean? One of these casts seems unnecessary.

tests/code/test_logsCreation.py
```diff
@@ -0,0 +1,35 @@
...
+            g.current_user = User.get_by_id('mupotsal')
+            currentLogsCount = len(list(AdminLogs.select()))
+            createLog("This is a test log 2")
+            allLogs = list(AdminLogs.select().order_by(AdminLogs.createdOn.desc()))
+            mostRecentLog = allLogs[0]
+            assert mostRecentLog.createdBy == g.current_user
+            assert mostRecentLog.logContent == "This is a test log 2"
+            assert datetime.strftime(mostRecentLog.createdOn,'%Y-%m-%d') == datetime.strftime(datetime.now(),'%Y-%m-%d')
+            assert len(allLogs) == currentLogsCount +1
+            transaction.rollback()
```
BrianRamsay: This second block is the exact same test as the first one, right? I'd say it is unnecessary unless it is testing something different that I'm not aware of

tests/code/test_transcripts.py
```diff
@@ -11,111 +13,113 @@
...
-                                            event = testingBonnerEvent.id,
-                                            attended = True,
-                                            hoursEarned = 2)
-
-    testingProgramEvent = Event.get(Event.name == "Test Program Event")
-    trainingpart = EventParticipant.create(user = "namet",
-                                            event = testingProgramEvent.id,
-                                            attended = True,
-                                            hoursEarned = 2)
+    with app.app_context():
```
BrianRamsay, line 16: You only need `with app.app_context()` around specific blocks. None of the other stuff in this method needs it.

tests/code/test_transcripts.py
```diff
@@ -190,23 +194,24 @@ def testingTotalHours():
...
-    assert Course.get_or_none(Course.id == testingCourse.id) is None
-    # delete program
-    testingProgramEvent = Event.get(Event.name == "Test program Event")
-    deleteEvent(testingProgramEvent)
-    assert Event.get_or_none(Event.id == testingProgramEvent) is None
-    # delete user
-    user = User.get(User.username == "namet")
-    user.delete_instance(recursive = True, delete_nullable = True)
-    assert User.get_or_none(User.username == "namet") is None
+    with app.app_context():
```
BrianRamsay, line 197: same comment as above. Only use this where it is needed.

tests/code/test_userManagement.py
```diff
@@ -1,39 +1,45 @@
...
-    addCeltsAdmin(current_user)
-    current_user = User.get(User.username == user)
-    assert current_user.isCeltsAdmin == True
-    removeCeltsAdmin(current_user)
-    current_user = User.get(User.username == user)
-    assert current_user.isCeltsAdmin == False
+    with app.app_context():
+        g.current_user = "ramsayb2"
+        user = "agliullovak"
+        current_user = User.get(User.username == user)
```
BrianRamsay: it's confusing to have `current_user` and `g.current_user` in the same code.

BrianRamsay (changes requested):
A number of small comments. Looking better!

There is a bunch of debug output in the tests. Tests shouldn't be printing.

tests/code/test_transcripts.py
mupotsal: The test fails when I run it without `app_context`

app/logic/events.py
mupotsal: A date was not formatted well in the tests (tests_events). I have fixed it.

tests/code/test_transcripts.py
mupotsal: Seems everyplace where we use `g.current_user` in tests need `app.context.` Without it it fails.

tests/code/test_transcripts.py
mupotsal: Can you please check again? If I remove it, my tests fail.

tests/code/test_transcripts.py
BrianRamsay: The only line in this entire method that needs to be in the block `with app.app_context()` is line 17.

BrianRamsay: The other tests are similar, with the function calls being tested also needing the app context

BrianRamsay (changes requested):

BrianRamsay (dismissed):

BrianRamsay:
There are merge conflicts to resolve.

BrianRamsay (approved):

Issue #153, opened 2021-10-22
Create new Terms


Sandesh-L:
Implement a way to create terms. Don't let users create a new term, just create a few new terms for them when it is necessary. Perhaps have a button on the term management page that says "Add Term". When clicked, it will automatically create one term in the future. If the current term is Fall 2021 and we have terms listed until Fall 2022, clicking "Add Term" would automatically create a new term, Spring 2022.

Sandesh-L:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #153: Creating New Terms](https://trello.com/c/a2iFGG7I/329-celts-153-creating-new-terms)

PR #212, opened 2021-12-20
Add New Term 

mupotsal:
This PR fixes issue #153. 

Summary. 
Users should be able to add a new term with just a click. 

Test:
Login as a  manager and go to _/admin_. Either check the latest term in the table there or in the database. Click _Add New Term_  and a new term is added which follows the latest term in the table. 
Screen Shot 2021-12-20 at 11 42 12 AM


app/logic/userManagement.py
```diff
@@ -33,3 +33,27 @@ def changeCurrentTerm(term):
...
 
     session["current_term"] = model_to_dict(newCurrentTerm)
+
+def addTerm():
+    terms_table = {"Spring":"Summer",
+                    "Summer":"Fall",
+                    "Fall":"Spring"}
+    terms = list(Term.select().order_by(Term.id))
+    lastCreatedTerm = terms[-1]
+    term = lastCreatedTerm.description
```
hmanalai: maybe rename this variable to termDescription

mupotsal: Fixed

app/logic/userManagement.py
```diff
@@ -33,3 +33,27 @@ def changeCurrentTerm(term):
     newCurrentTerm.save()
 
     session["current_term"] = model_to_dict(newCurrentTerm)
+
+def addTerm():
```
BrianRamsay: I would call this `addNextTerm`, so it is clearer how it works

app/logic/userManagement.py
```diff
@@ -33,3 +33,27 @@ def changeCurrentTerm(term):
...
+    terms_table = {"Spring":"Summer",
+                    "Summer":"Fall",
+                    "Fall":"Spring"}
+    terms = list(Term.select().order_by(Term.id))
+    lastCreatedTerm = terms[-1]
+    termDescription = lastCreatedTerm.description
+    term = termDescription.split()
+    year = termDescription[-1]
+    if termDescription[0]=="Fall": # We only change the year when it is Fall
+        year = int(year) + 1
```
BrianRamsay: This logic won't work if we're going from, say, 2029 to 2030, will it? You're only getting a single digit off of termDescription, not the whole year. The whole year is `term[1]`, not `termDescription[-1]`, if I'm reading it correctly

app/logic/userManagement.py
```diff
@@ -33,3 +33,27 @@ def changeCurrentTerm(term):
...
+
+    createdTermDescription = terms_table[term[0]]+" "+str(year)
+    academicYear = lastCreatedTerm.academicYear
+    if termDescription[0] == "Summer": #we only change academic year when the latest term in the table is Summer
+        previousAY = academicYear.split("-")
+        academicYear = str(int(previousAY[0])+1)+"-"+str(int(previousAY[-1])+1)
+
+    isSummer = False
+    if "Summer" in createdTermDescription.split():
+        isSummer = True
```
BrianRamsay: you can just assign the boolean expression to the `isSummer` variable, without having to use an `if` at all

app/controllers/main/routes.py
```diff
@@ -180,3 +182,20 @@ def searchUser(query):
 @main_bp.route('/contributors',methods = ['GET'])
 def contributors():
     return render_template("/contributors.html")
+
```
BrianRamsay, line 185: Just making sure, is this section in the right place? Don't want it added here if it already exists somewhere else as a result of a merge moving code or something.

app/logic/userManagement.py
BrianRamsay: There's no test for this! This is a very clear case where TDD makes sense. I suspect it would have saved you a lot of time.

BrianRamsay (changes requested):
UI wise, I'd make sure the button text isn't wrapped to the next line, and I would change 'Add New Term' to 'Add Next Term'. It's a little more descriptive.

app/logic/userManagement.py
mupotsal: The problem was with the variable term. It was not used correctly. We have fixed and the logic stays the same and should work.

app/logic/userManagement.py
mupotsal: We did not get this well. Do you mean shortening it to: 

`isSummer = True if "Summer" in createdTermDescription.split() else False`

app/logic/userManagement.py
Ebenezer997: function  addTerm has been changed to addNextTerm

app/controllers/main/routes.py
Ebenezer997: I think is in the right place. I checked it and it does not exist anywhere apart from there

app/logic/userManagement.py
mupotsal: I added test for the AddTerm function

app/logic/userManagement.py
BrianRamsay: No. You are giving a boolean expression to the if statement, just use the same expression for the variable.
`isSummer = "Summer" in createdTermDescription.split()`

tests/code/test_userManagement.py
```diff
@@ -58,3 +60,50 @@ def test_invalidTermInputs():
...
     with pytest.raises(DoesNotExist):
         changeCurrentTerm("womp")
+
+@pytest.mark.integration
+def test_addNextTerm():
+    newTerm = Term.create(description="Fall 2022",year=2022, academicYear= "2022-2023",isBreak=False, isSummer=False,isCurrentTerm=False)
+    newTerm.save()
+
+    terms = list(Term.select().order_by(Term.id))
+    lastCreatedTerm = terms[-1]
```
BrianRamsay: If you are trying to get the id of the most recent term, just use the object for it you already have, `newTerm`

tests/code/test_userManagement.py
```diff
@@ -58,3 +60,50 @@ def test_invalidTermInputs():
...
+    addNextTerm()
```
BrianRamsay: It would make sense to have addNextTerm return the term it creates. Makes it easier to check and easier to use

tests/code/test_userManagement.py
```diff
@@ -58,3 +60,50 @@ def test_invalidTermInputs():
         changeCurrentTerm(100)
     with pytest.raises(DoesNotExist):
         changeCurrentTerm("womp")
+
+@pytest.mark.integration
+def test_addNextTerm():
```
BrianRamsay: Adding a term doesn't have anything to do with user management. Changing the term doesn't either, really, even though it was already here in this file. I would put both of these tests in test_utils or create a new file test_terms

BrianRamsay (changes requested):
a few comments to resolve

tests/code/test_userManagement.py
Ebenezer997: changed the recent term to object newterm

tests/code/test_userManagement.py
Ebenezer997: move both of the tests to test_utils

Ebenezer997: moved both tests to test_utils

app/logic/userManagement.py
Ebenezer997: fixed

BrianRamsay (approved):
I went ahead and made the test changes we talked about last week in the review, and refactored addNextTerm.

I would encourage you to look at the diff here, so you can see what changes I made. Particularly the changes to the test, to make them simpler and also to fully test the function. You weren't testing anything other than description before.
https://github.com/BCStudentSoftwareDevTeam/celts/pull/212/files/ee09b254d495235623f0f97397a4a22164e2d06a..40b6106a5864dc9ccd2476e20031953b7ce01309

Issue #192, opened 2021-11-09
Matching Volunteers to Outside Participants


bryantal:
One missing feature needed on some program's events (Berea Buddies, Tutoring, etc.) is the ability to match volunteers with outside participants.

![image](https://user-images.githubusercontent.com/28715731/140958072-cafb8ce9-1847-4937-b4a2-c713fe5e59dc.png)

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS administrator must be able to match volunteers to outside participants](https://trello.com/c/pf9iSaer/286-celts-administrator-must-be-able-to-match-volunteers-to-outside-participants)

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #192: Matching Volunteers to Outside Participants](https://trello.com/c/ljnvyNzx/343-celts-192-matching-volunteers-to-outside-participants)

Boorclark:
This feature is not going to be developed right now. Closing issue.

PR #214, opened 2021-12-30
Match participants

mupotsal:
This Pull Request solves issue #192. 

**Description**:
Volunteers should be able to be matched with outside participants. This PR has both the backend and frontend implemented for the feature. 

**Test**
Navigate to an already existing event and go to the track volunteers page. 
Screen Shot 2021-12-30 at 10 14 44 AM
Add outside participants first and then match them. Currently, the relationship between volunteers and outside participants is one too many. However, it is important to note that an outside participant can only have one volunteer for a given event. 

**Important Points to note**

- Deleting outside participants also removes them from the volunteers. 
- Outside Participants need to be added separately for each event.


app/controllers/admin/routes.py
```diff
@@ -194,3 +194,9 @@ def courseManagement(term = None):
                             approvedCourses = approved,
                             terms = terms,
                             term = term)
+
+
+@admin_bp.route('/matchVolunteerParticipant',methods = ['GET','POST'])
+def matchParticipants():
+    pass
```
hmanalai: Were you intending to delete this `pass`?

app/templates/events/trackVolunteers.html
```diff
@@ -139,4 +238,31 @@ <h5 class="modal-title" id="exampleModalLabel">Add a volunteer</h5>
...
+        </div>
+      </div>
+      <div class="modal-footer">
+        <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Close</button>
+        <button id="selectParticipantButton" type="button" class="btn btn-primary" disabled>Add participant</button>
+      </div>
+    </div>
+  </div>
+</div>
+
```
hmanalai: @BrianRamsay I'm just curious here. Do you think it's better to have multiple modals or one and we change the content dynamically?

Of course, this would be in cases when modal content is very similar. For example, the above two modals are the same except for the title and an `id` (that I assume is later used in JavaScript).

app/templates/events/trackVolunteers.html
```diff
@@ -114,12 +164,61 @@ <h1 class="text-center">{{event.name}}</h1>
...
+    </thead>
+    <tbody id="volunteerTable">
+      {% for participant in outsideParticipants%}
+      <tr>
+        <td>
+          <input
+            class="form-control"
+            type='hidden'
+            name="username{{loop.index|string}}"
+            id="{{participant.username}}"
```
hmanalai: The outside participant table doesn't have a username column. Were you intending to create the column? Or should the `id` be changed to something else?

app/controllers/admin/volunteers.py
```diff
@@ -92,13 +105,61 @@ def addVolunteer(volunteer, eventId):
         flash("Error when adding volunteer", "danger")
     return ""
 
+@admin_bp.route('/addParticipantToEvent/<volunteer>/<eventId>', methods = ['POST'])
+def addParticipant(volunteer, eventId):
+    email = volunteer.strip("()").split('(')[-1]
```
hmanalai: I'm kind of confused by what this function is supposed to do. From the name, it looks like it's supposed to add a participant to the event participant table. OR maybe add an outside participant to the outside participant table. 

However, looks like it's creating/getting an entry from the Match Participants table, and it's specifically creating it using the outside participant. 

I think the name of the function should be changed to describe what it is doing.

app/controllers/admin/volunteers.py
```diff
@@ -92,13 +105,61 @@ def addVolunteer(volunteer, eventId):
...
+
 @admin_bp.route('/removeVolunteerFromEvent/<user>/<eventID>', methods = ['POST'])
 def removeVolunteerFromEvent(user, eventID):
     (EventParticipant.delete().where(EventParticipant.user==user, EventParticipant.event==eventID)).execute()
     (EventRsvp.delete().where(EventRsvp.user==user)).execute()
+    update = MatchParticipants.get_or_none(MatchParticipants.volunteer==user,MatchParticipants.event==eventID)
+    while update:
+        update.volunteer=None
+        update.save()
+        update = MatchParticipants.get_or_none(MatchParticipants.volunteer==user,MatchParticipants.event==eventID)
```
hmanalai: I am assuming that in lines 139-143 the while loop is there because one volunteer can be matched to multiple outside participants. 
 
Instead of doing a get query for each entry inside the loop, you can do `select().where()` that would get all the entries in one query, and then you can loop through the query result to update the volunteer column to `None`.

hmanalai: The point is the avoid multiple queries.

app/controllers/admin/volunteers.py
```diff
@@ -92,13 +105,61 @@ def addVolunteer(volunteer, eventId):
...
+
     flash("Volunteer successfully removed", "success")
     return ""
 
+@admin_bp.route('/removeParticipantFromEvent/<participant>/<eventID>', methods = ['POST'])
+def removeParticipantFromEvent(participant, eventID):
+    (MatchParticipants.delete().where(MatchParticipants.outsideParticipant==participant, MatchParticipants.event==eventID)).execute()
+    flash("Particpant successfully removed", "success")
+    return ""
+
```
hmanalai: I think we should be consistent with our usage of participants and outside participants. Sometimes it can get confusing.
For me, participant (alone) means it's a student.

hmanalai: If you do decide to change participant to outsideParticipant for consistency, then it'd have to be changed everywhere else in the code as well.

app/logic/searchUsers.py
```diff
@@ -24,3 +25,28 @@ def searchUsers(query):
...
+    query = query.strip()
+    search = query.upper()
+    splitSearch = search.split()
+    resultsDict = {}
+
+    firstName = splitSearch[0] + "%"
+    lastName = " ".join(splitSearch[1:]) +"%"
+
+    if len(splitSearch) == 1: #search for first or last name
+        results = OutsideParticipant.select().where(OutsideParticipant.firstName ** firstName | OutsideParticipant.lastName ** firstName)
```
hmanalai: This function is the same as the one above, except for the two lines of the query (lines 40 and 48). Instead of writing the same code all over again, it would be better to modify the above function to use a different query based on whether the user is an outside participant or not.

app/static/js/searchOutsideParticipant.js
```diff
@@ -0,0 +1,31 @@
+export default function searchOutsideParticipant(inputId, callback, parentElementId=null){
+  var query = $(`#${inputId}`).val()
+  console.log("searchOutsideParticipant is invoked here");
```
hmanalai: We have to remove the debug statements.

hmanalai: Also, this function is very similar to the searchUser one. Can't you use the same javascript file (I believe it's called searchUser)? 
Why the need to define it again?

app/templates/events/trackVolunteers.html
mupotsal: I changed the ID to the participant.email. For outside participants, email is the primary key since they do not have a bNumber. What do you think about that?

app/controllers/admin/volunteers.py
mupotsal: I have renamed the function and the endpoints are now more descriptive of what they are doing.

app/controllers/admin/volunteers.py
mupotsal: I have fixed it. I used the update method instead. I think it is fast too.

app/controllers/admin/volunteers.py
mupotsal: Resolved.

app/controllers/admin/volunteers.py
```diff
@@ -18,7 +19,7 @@
 def getVolunteers(query):
     '''Accepts user input and queries the database returning results that matches user search'''
 
-    return json.dumps(searchUsers(query))
+    return json.dumps(searchUsers(query,"volunteers"))
```
hmanalai, line 23: Instead of "volunteer," you should have "insideParticipant." From the changes I see in the searchUser.js, this code would run the code for outside participant.

app/controllers/admin/volunteers.py
```diff
@@ -92,13 +105,58 @@ def addVolunteer(volunteer, eventId):
         flash("Error when adding volunteer", "danger")
     return ""
 
+@admin_bp.route('/addOutsideParticipantToEvent/<outsideParticipant>/<eventId>', methods = ['POST'])
+def addOutsideParticipant(outsideParticipant, eventId):
```
hmanalai: Please rename this to describe what the function is doing, because we are not really adding outside participants in this function.

app/controllers/admin/routes.py
```diff
@@ -194,3 +194,8 @@ def courseManagement(term = None):
                             approvedCourses = approved,
                             terms = terms,
                             term = term)
+
+
+@admin_bp.route('/matchVolunteerParticipant',methods = ['GET','POST'])
+def matchParticipants():
+    matchVolunteers()
```
BrianRamsay: How are you passing nothing to this function?

app/controllers/admin/volunteers.py
```diff
@@ -92,13 +105,58 @@ def addVolunteer(volunteer, eventId):
...
+    email = outsideParticipant.strip("()").split('(')[-1]
```
BrianRamsay: Passing this info in the url is a bit weird, when you are already posting.

app/logic/searchUsers.py
```diff
@@ -9,18 +10,36 @@ def searchUsers(query):
 
     firstName = splitSearch[0] + "%"
     lastName = " ".join(splitSearch[1:]) +"%"
+    results = None
+    searchId = None
 
     if len(splitSearch) == 1: #search for first or last name
-        results = User.select().where(User.isStudent & (User.firstName ** firstName | User.lastName ** firstName))
+        if searchGroup == "insideParticipant":
```
BrianRamsay: Doesn't this break everything? it's pretty weird that in order to search for students (in any context) you would have to pass 'insideParticipant' instead of 'student'

app/models/matchParticipants.py
```diff
@@ -0,0 +1,13 @@
...
+
+from app.models.outsideParticipant import OutsideParticipant
+
+class MatchParticipants(baseModel):
+    volunteer = ForeignKeyField(User,null=True,default=None)
+    outsideParticipant = ForeignKeyField(OutsideParticipant)
+    event = ForeignKeyField(Event)
+
+    class Meta:
+        primary_key=CompositeKey('outsideParticipant','event')
```
BrianRamsay: Are we ok with making it so we can never have two volunteers matched to the same group of outside participants?

app/static/js/slcNewProposal.js
```diff
@@ -96,7 +96,7 @@ function callback() {
 }
 
 $("#courseInstructor").on('input', function() {
-  searchUser("courseInstructor", callback);
+  searchUser("courseInstructor", callback,"insideParticipant");
```
BrianRamsay: definitely not correct

app/templates/events/trackVolunteers.html
BrianRamsay: When modals are very similar I would tend to say they should use the same html and have pieces swapped out by javascript

BrianRamsay (changes requested):

app/controllers/admin/routes.py
mupotsal: I did not need this here. We already have it in controllers for volunteers.py.

app/controllers/admin/volunteers.py
mupotsal: I have fixed that.

app/models/matchParticipants.py
mupotsal: @BrianRamsay What do you mean here? Currently a volunteer can have many outsideParticipants for a given event. However, an outsideParticipant can only be matched to one volunteer.

app/static/js/slcNewProposal.js
mupotsal: I am not sure why this is incorrect?

app/logic/searchUsers.py
mupotsal: We typically had two search groups, insideParticipant and outsideParticipant. However, I have changed insideParticipant to student. If we agree that insideParticipant are people with B# then it becomes easy in this scenario.

app/controllers/admin/volunteers.py
BrianRamsay, line 23: is this searching students or is this searching people who have volunteered before or who are valid volunteers?

app/controllers/admin/volunteers.py
BrianRamsay: I don't see what is fixed. You are still passing that in the url instead of the post request

app/controllers/admin/volunteers.py
BrianRamsay: I think this is still relevant. addOutsideParticipantToEvent would be more accurate (see: the url)

app/controllers/admin/volunteers.py
```diff
@@ -92,13 +105,68 @@ def addVolunteer(volunteer, eventId):
...
+def matchParticipant():
+    matchData = request.form
+
+    volunteer = matchData['volunteer']
+    outsideParticipant = matchData['outsideParticipant']
+    eventId = matchData['eventId'][0]
+
+    vol = User.get_by_id(volunteer)
+    update = MatchParticipants.get_or_none(MatchParticipants.outsideParticipant==outsideParticipant,MatchParticipants.event==int(eventId),MatchParticipants.volunteer==None)
+    if update != None:
```
BrianRamsay: `if update != None` is the same as `if update` (unless update could be False)

app/controllers/admin/volunteers.py
```diff
@@ -92,13 +105,68 @@ def addVolunteer(volunteer, eventId):
...
+    event = eventId.split(':')
+    newEntry = MatchParticipants.get_or_create(outsideParticipant=email,event=int(event[0]))
+    if newEntry[-1]==False:
+        flash("Participant already added to this event!", "danger")
+    else:
+        flash("Participant succesfully added to the event!", "success")
+    return ""
+
+@admin_bp.route('/matchParticipants', methods = ['POST'])
+def matchParticipant():
```
BrianRamsay: I'm a little confused by this function. You validate `volunteer` by getting `vol`, but then never use `vol`. Also, I think we should move away from the assumption that there is only one volunteer matched to one outside participant. I thought we changed the data structure to let that be many to many.

app/controllers/main/routes.py
```diff
@@ -166,19 +166,21 @@ def serviceTranscript(username):
                             startDate = startDate,
                             userData = user)
 
-@main_bp.route('/searchUser/<query>', methods = ['GET'])
-def searchUser(query):
+@main_bp.route('/searchUser/<query>/<group>', methods = ['GET'])
+def searchUser(query,group):
     '''Accepts user input and queries the database returning results that matches user search'''
+    print( "query ...............", query)
```
BrianRamsay, line 245: remove debug output

app/logic/searchUsers.py
```diff
@@ -9,18 +10,36 @@ def searchUsers(query):
 
     firstName = splitSearch[0] + "%"
     lastName = " ".join(splitSearch[1:]) +"%"
-
+    results = None
+    searchId = None
+    print ("This.................",searchGroup)
```
BrianRamsay: debug output

app/models/matchParticipants.py
BrianRamsay: As we talked before, we want a many-many relationship between volunteers and outside participants for events

app/static/js/searchUser.js
```diff
@@ -1,12 +1,16 @@
-export default function searchUser(inputId, callback, parentElementId=null){
+export default function searchUser(inputId, callback,group,parentElementId=null){
   var query = $(`#${inputId}`).val()
+  console.log("search ...............", group)
```
BrianRamsay: debug output

app/static/js/userManagement.js
```diff
@@ -38,7 +38,7 @@ $(document).ready(function() {
 
   // remove celts student staff
   $("#removeCeltsStudentStaffInput").on("input", function() {
-    searchUser("removeCeltsStudentStaffInput", callback);
+    searchUser("removeCeltsStudentStaffInput", callback, "");
```
BrianRamsay: Add a default to the new parameter so you don't have to change all the locations it is called

app/templates/events/trackVolunteers.html
```diff
@@ -90,24 +99,68 @@ <h1 class="text-center">{{event.name}}</h1>
...
+        <ul>
+          <li> {{entry.firstName}} {{entry.lastName}} </li>
+        </ul>
+        {% endfor%}
+        {% endif %}
         </td>
+
+        <td>
+              {% set ac = accordionCount.value| string() %}
+              {% set accodionFlushNumber = "accordionFlushNumber"+ac %}
```
BrianRamsay: should be 'accordionFlushNumber' both places

app/templates/events/trackVolunteers.html
```diff
@@ -90,24 +99,68 @@ <h1 class="text-center">{{event.name}}</h1>
...
+                <div id="{{flushCollapseNumber}}" class="accordion-collapse collapse" aria-labelledby="flush-headingOne" data-bs-parent="#{{accodionFlushNumber}}">
+                  <div class="accordion-body">
+                    <ul>
+                      {% for outsideParticipant in outsideParticipants%}
+                      {% set status = '' %}
+                      {% if participant.user in matches.keys() %}
+                      {% if outsideParticipant in matches[participant.user] %}
+                      {% set status = 'checked' %}
+                      {% endif %}
+                      {% endif %}
```
BrianRamsay: Indent properly

tests/code/test_participants.py
```diff
@@ -251,3 +252,17 @@ def test_getEventParticipantsWithWrongParticipant():
...
+    matchVolunteers("mupotsal","maryjones@example.gmail.com",8)
+    newTotal = len(list(MatchParticipants.select()))
+    assert newTotal == currentMatches+1
+    MatchParticipants.delete(MatchParticipants.volunteer=="mupotsal",MatchParticipants.outsideParticipant=="maryjones@example.gmail.com",MatchParticipants.event==8).execute()
+
+    currentMatches = len(list(MatchParticipants.select()))
+    matchVolunteers("ramsayb2","maryjones@example.gmail.com",7)
+    newTotal = len(list(MatchParticipants.select()))
+    assert newTotal == currentMatches+1
+    MatchParticipants.delete(MatchParticipants.volunteer=="ramsayb2",MatchParticipants.outsideParticipant=="maryjones@example.gmail.com",MatchParticipants.event==7).execute()
```
BrianRamsay: There is a lot more manipulation of outside participants and matches that happens. We need to get all of that logic into functions that we can test, rather than have it all happening inside of controller methods

BrianRamsay (changes requested):

BrianRamsay (dismissed):

app/controllers/admin/volunteers.py
Ebenezer997: resolved

app/templates/events/trackVolunteers.html
Ebenezer997: resolved

app/controllers/admin/volunteers.py
mupotsal: who are valid volunteers

BrianRamsay:
@mupotsal , Check out the conflicts with Tyler's recent fix.

app/controllers/admin/volunteers.py
BrianRamsay: not resolved

app/controllers/admin/volunteers.py
```diff
@@ -81,16 +84,32 @@ def updateVolunteerTable(eventID):
...
+    username = volunteerData["username"]
     user = User.get(User.username==username)
+    eventId = volunteerData['eventId'][0]
     successfullyAddedVolunteer = addVolunteerToEventRsvp(user, eventId)
     EventParticipant.create(user=user, event=eventId) # user is present
     if successfullyAddedVolunteer:
         flash("Volunteer successfully added!", "success")
     else:
-        flash("Error when adding volunteer", "danger")
+        flash("Error when adding vol    unteer", "danger")
```
BrianRamsay, line 98: There's an extra space in the message.

app/controllers/admin/volunteers.py
```diff
@@ -81,16 +84,32 @@ def updateVolunteerTable(eventID):
...
 
-@admin_bp.route('/addVolunteerToEvent/<volunteer>/<eventId>', methods = ['POST'])
-def addVolunteer(volunteer, eventId):
-    username = volunteer.strip("()").split('(')[-1]
+@admin_bp.route('/addVolunteerToEvent', methods = ['POST'])
+def addVolunteer():
+    volunteerData = request.form
+    username = volunteerData["username"]
     user = User.get(User.username==username)
+    eventId = volunteerData['eventId'][0]
```
BrianRamsay, line 92: Having to add `[0]` is a clue that you are duplicating a form field.

app/templates/events/trackVolunteers.html
```diff
@@ -31,30 +32,35 @@ <h1 class="text-center">{{event.name}}</h1>
...
-<input class="form-control" id="eventID" name="event" value="{{event}}" type='hidden' />
-<div class="flex-container" style="padding: 5px;">
-  <div class='form-group d-inline-flex'>
-    <a class="btn btn-success" role="button" href="#" data-bs-toggle="modal" data-bs-target="#addVolunteerModal" style="margin:5px;">Add volunteer</a>
-    <a class="btn btn-warning" role="button" href= "/eventsList/{{event.id}}/kiosk" style="margin:5px;">Kiosk Entry</a>
+
+<h2 class ="text-center"> Volunteers </h2>
+<form action="#" method="post"><br>
+  <label class='d-none' id="eventLength">{{eventLength}}</label>
+  <input class="form-control" id="eventID" name="event" value="{{event}}" type='hidden' />
```
BrianRamsay, line 39: Don't we want `event.id`? You're just outputting a whole object's string representation

app/controllers/main/routes.py
BrianRamsay: still not done

app/controllers/admin/volunteers.py
```diff
@@ -81,16 +84,32 @@ def updateVolunteerTable(eventID):
...
+        flash("Error when adding vol    unteer", "danger")
+    return ""
+
+@admin_bp.route('/addOutsideParticipantToEvent', methods = ['POST'])
+def addOutsideParticipant():
+
+    outsideParticipantData = request.form
+    email = outsideParticipantData['email']
+    eventId = outsideParticipantData['eventId']
+    event = eventId.split(':')
```
BrianRamsay, line 107: I would do the whole calculation here. Because what is being stored in `event` ISN'T an event, it is an array. Also, why are you passing something that isn't an event ID already in a field called `eventId`?

app/logic/searchUsers.py
```diff
@@ -9,17 +11,35 @@ def searchUsers(query):
...
-            if participant not in resultsDict:
-                resultsDict[participant.username]=model_to_dict(participant)
+            if searchGroup != "outsideParticipant":
+                searchId = participant.username
+                if participant not in resultsDict:
+                    resultsDict[participant.username]=model_to_dict(participant)
+            else:
+                searchId = participant.email
+                if participant not in resultsDict:
+                    resultsDict[participant.email]=model_to_dict(participant)
```
BrianRamsay, line 30: The only thing actually different in these two blocks is the value you are using for a search id. Refactor so the if/else is around the smallest code, and the common code is outside

app/logic/searchUsers.py
```diff
@@ -9,17 +11,35 @@ def searchUsers(query):
...
+
+
     else:
         for searchTerm in splitSearch: #searching for specified first and last name
             if len(searchTerm) > 1:
                 searchTerm += "%"
-                results = User.select().where(User.isStudent & User.firstName ** firstName & User.lastName ** lastName)
+                if searchGroup != "outsideParticipant":
+                    searchId = participant.username
```
BrianRamsay, line 38: You don't have a participant here, do you? Does this code work?

app/static/js/searchUser.js
```diff
@@ -1,25 +1,39 @@
-export default function searchUser(inputId, callback, clear=false, parentElementId=null, columnRequested=null){
-  var query = $(`#${inputId}`).val()
+export default function searchUser(searchOptions){
```
BrianRamsay, line 1: Don't replace the required arguments. Now there is no way to tell by looking at or calling this function what is required to make it work.

app/static/js/trackVolunteers.js
```diff
@@ -11,6 +11,18 @@ $(document).ready(function() {
     }
   });
 
+  var table =  $('#trackOutsideParticipants').DataTable({
+  "fnDrawCallback": function(oSettings) {
+    if ($('#trackOutsideParticipants tr').length < 11) {
+        $('.dataTables_paginate').hide(); //disable search and page numbers when the length of the table is less 11
```
BrianRamsay, line 17: This comment should be above the whole block, so I don't have to read it to know what it does.

app/static/js/trackVolunteers.js
```diff
@@ -36,20 +72,68 @@ $(document).ready(function() {
...
+    error: function(request, status, error) {
+    }
+  });
+});
+
+$(".removeVolunteer").on("click", function() {
+  let username =  $(this)[0].id;
+  let eventId = $('#eventID').val();
+  console.log("This is the username",username);
+  console.log("This is the eventID",eventId);
```
BrianRamsay: Remove debug output.

app/templates/events/trackVolunteers.html
```diff
@@ -31,30 +32,35 @@ <h1 class="text-center">{{event.name}}</h1>
...
+  <div class="flex-container" style="padding: 5px;">
+    <div class='form-group d-inline-flex'>
+      <a class="btn btn-success" id="addVolunteerButton" role="button" href="#" data-bs-toggle="modal" data-bs-target="#addVolunteerModal" style="margin:5px;" >Add volunteer</a>
+      <a class="btn btn-warning" role="button" href= "/event/{{event.id}}/kiosk" style="margin:5px;">Kiosk Entry</a>
   </div>
 </div>
 <form method="post">
   <label class='d-none' id="eventLength">{{eventLength}}</label>
   <input class="form-control" id="eventID" name="event" value="{{event}}" type='hidden'>
```
BrianRamsay, line 48: There should only be one element in an entire page with a given id attribute.

app/models/matchParticipants.py
```diff
@@ -0,0 +1,12 @@
+from app.models import*
+from app.models.user import User
+from app.models.event import Event
+
+from app.models.outsideParticipant import OutsideParticipant
+
+class EventOutsideParticipants(baseModel):
```
BrianRamsay: This file doesn't match the name of the class

BrianRamsay (changes requested):
Why are there no tests? When you write new functionality, start with the test!

app/models/matchParticipants.py
Ebenezer997: resolved

BrianRamsay:
Closing PR, as this feature is not being developed anymore

PR #215, opened 2021-12-31
Fix contrast issue on the Volunteer Profile page

gastonjarju:
Issue:  The message left when the user has not marked interest in any programs has not changed. It needs to since its contrast is too high against the white background which is an accessibility issue. Maybe just make it black text and center it a little better.

Solution:  Added or changed the Bootstrap text class to make the text dark instead of blue so that it would match the white background. Also, added an align attribute to center the text. 

To Test: Navigate to the IP Address/profile/ramsayb2 and you should see the text is centered and color in black. 

image


BrianRamsay:
Merge development in (or volunteer-profile-fix and then development), and let's see what conflicts are left.

PR #217, opened 2022-01-06
Ban button fix me

gastonjarju:
Issue:
There is a glitch with the note text box where I am able to type in it, but the ban button is still disabled until I click out of the text box.

Solution:
Added a .on() Event handler that replaces the .change(). Now, the user does not need to click outside the textarea before the ban/unban button is enabled.

To Test:

Navigate to the IP Address/profile/ramsayb2
Click on "Edit"
Click on the datepicker input field and select an End Date for the ban
Click on the Textarea to insert the reason for the ban, and as soon as you insert a character, the ban button enables. If there is no character, it remains disabled.


BrianRamsay:
Merge development in (or volunteer-profile-fix and then development), and let's see what conflicts are left.

BrianRamsay (approved):

Issue #208, opened 2021-11-24
Screen reader compatibility 


Ebenezer997:
check all CELTS  HTML pages have screen reading compatibility

PR #218, opened 2022-02-09
Sidebar Accessibility

hmanalai:
The sidebar had two accessibility issues:
1. It was not obvious which item in the list was currently selected, which we fixed by adding conditionals to highlight the currently selected link. 
2. There was a hidden text `<h1 class="visually-hidden">Admin sidebar</h1>`  that the NVDA read but was hidden. Since it didn't serve any purpose, we removed it. 

We also added a `skip to main content` link that only shows when it is tabbed into. That link will take the user to the main content of a given page so long as `<main id="maincontent"></main>` is in that page's html file directly after `{% block app_content %}`. For testing purposes, only the Events List page and the Create Event ('/template_select') page  have this main tag. We will add it to the other pages as we get to them.

Note: This PR is related to issue #208. More PRs will be coming soon for other parts of the same issue.

### DIFF
--- a/app/templates/events/event_list.html
+++ b/app/templates/events/event_list.html
@@ -13,6 +13,7 @@
   <link rel="stylesheet" href="//code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css">
 {% endblock %}
 {% block app_content %}
+<main id="maincontent"></main>
 <h1 class="text-center mt-5">Events List for {{selectedTerm.description}}</h1>
 <div class="container-fluid px-1 py-3">
   <div class="row d-flex align-items-baseline p-3">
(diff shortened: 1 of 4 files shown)


Sandesh-L:
Should we add the "skip to main page" button to this PR? Since I think it will be at the beginning of the sidebar.

Sandesh-L:
> Should we add the "skip to main page" button to this PR? Since I think it will be at the beginning of the sidebar.

done!

app/templates/events/event_list.html
```diff
@@ -13,6 +13,7 @@
   <link rel="stylesheet" href="//code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css">
 {% endblock %}
 {% block app_content %}
+<main id="maincontent"></main>
```
BrianRamsay: Let's add this id to a more global location, like an already existing element that surrounds our app content. (as we discussed, potentially line 54 in base.html)

BrianRamsay (changes requested):

app/templates/events/event_list.html
hmanalai: Done.

BrianRamsay (approved):

PR #220, opened 2022-02-16
Fixes edit event and track volunteer accessibility

hmanalai:
This PR fixes the accessibility issues of the edit/create event and track volunteer pages. In the process of fixing accessibility issues, we have also re-designed the track volunteers page. 

Note: This PR is related to issue #208. More PRs will be coming soon for the same issue.

### DIFF
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -214,7 +214,7 @@ def calculateRecurringEventFrequency(event):
         raise Exception("This event is not a recurring event")
 
     return [ {'name': f"{event['name']} Week {counter+1}",
-              'date': event['startDate'] + datetime.timedelta(days=7*counter),
+              'date': (event['startDate'] + datetime.timedelta(days=7*counter)).strftime("%m/%d/%Y"),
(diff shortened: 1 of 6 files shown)


app/logic/events.py
```diff
@@ -214,7 +214,7 @@ def calculateRecurringEventFrequency(event):
         raise Exception("This event is not a recurring event")
 
     return [ {'name': f"{event['name']} Week {counter+1}",
-              'date': event['startDate'] + datetime.timedelta(days=7*counter),
+              'date': (event['startDate'] + datetime.timedelta(days=7*counter)).strftime("%m/%d/%Y"),
```
BrianRamsay, line 217: This change is test-breaking. Make sure it doesn't break the existing application.

BrianRamsay (changes requested):

app/logic/events.py
hmanalai: Done

BrianRamsay (approved):

PR #222, opened 2022-02-18
Admin Management Accessibility

hmanalai:
This PR fixes the accessibility issues on the admin management page. Specifically, 
- we have changed the search icon at the end of the search bar to be an icon, instead of a button. 
- we fixed the labels for the search fields, and adjusted the look of the search bars. 
- we changed the current term to be a card instead of buttons. Now, the screen reader says "selected term" for the currently selected item in the card. 
- we also added the name of the selected term to the flasher. 

Note: This PR is related to issue #208. More PRs will be coming soon for the same issue.

### DIFF
--- a/app/controllers/admin/userManagement.py
+++ b/app/controllers/admin/userManagement.py
@@ -56,8 +56,9 @@ def changeTerm():
         termData = request.form
         term = int(termData["id"])
         changeCurrentTerm(term)
-        flash('Current term changed successfully', 'success')
+        newCurrentTerm = Term.get_by_id(term)
+        flash(f'Current term successfully changed to {newCurrentTerm.description}.', 'success')
     except:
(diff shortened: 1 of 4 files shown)


app/controllers/admin/userManagement.py
```diff
@@ -56,8 +56,9 @@ def changeTerm():
         termData = request.form
         term = int(termData["id"])
         changeCurrentTerm(term)
-        flash('Current term changed successfully', 'success')
+        newCurrentTerm = Term.get_by_id(term)
+        flash(f'Current term successfully changed to {newCurrentTerm.description}.', 'success')
```
BrianRamsay: you should be able to use `g.current_term`, without needing to do another get

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #176, opened 2021-11-03
Time Selectpickers do not Work on Firefox Browser


bryantal:
When creating and editing an event, the time selectpickers are not supported on Firefox browsers.  We'll need to replace them with a time selectpicker that will work on Firefox.

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #176: Time Selectpickers do not Work on Firefox Browser](https://trello.com/c/rP3JyTOd/340-celts-176-time-selectpickers-do-not-work-on-firefox-browser)

mupotsal:
This is also an issue for Safari.

sreynit02:
Liberty and I worked on it but there needs to be some more exploration with the time picker. We left off with this PR#191: https://github.com/BCStudentSoftwareDevTeam/celts/pull/191

hmanalai:
The same issue is happening with the flash messages. The screen reader reads them correctly on Chrome and Edge, but not on firefox.

hmanalai:
Fixed in PR #191

PR #226, opened 2022-03-02
Sidebar Accessibility

### LINKED ISSUE #208: Screen reader compatibility

hmanalai:
This is the last PR for issue #208 about accessibility. 

The PR fixes the sidebar so that it is highlighted as the user navigates through the web pages. It also fixes the current user dropdown at the bottom of the sidebar to show the currently selected user.  

Note: We realized that flash messages are not read in Firefox, however, they are read correctly by the screen reader in Chrome and Edge. We have added a note about this to the timepicker issue  #176 (which is also not working in Firefox).

### DIFF
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -39,7 +39,7 @@ def switchUser():
     return redirect(request.referrer)
 
 
-@admin_bp.route('/template_select')
+@admin_bp.route('/eventTemplates')
 def templateSelect():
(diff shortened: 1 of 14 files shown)


BrianRamsay (changes requested):
One small conflict to fix and then it's g2g

hmanalai:
> One small conflict to fix and then it's g2g

Fixed the merge conflict.

BrianRamsay (approved):

Issue #79, opened 2021-08-17
Faculty Edit Service Learning Course Proposals


bryantal:
Faculty UI Wireframe: https://balsamiq-wireframes.appspot.com/?state=%7B%22ids%22:%5B%221ziGyTzz6J7iHjyzmsUzj0yzvuFdITg3n%22%5D,%22action%22:%22open%22,%22userId%22:%22109665407847548855125%22,%22resourceKeys%22:%7B%7D%7D

The faculty must be able to edit service learning course proposals from the table where they view all of them.
SRS Item: **5.3** The faculty must be able to edit SL  Course Proposals
![image](https://user-images.githubusercontent.com/28715731/129770052-606a2c9b-0705-4706-ae0c-0d33d930e0e9.png)

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #79: Faculty Edit Service Learning Course Proposals](https://trello.com/c/9b572uA4/292-celts-79-faculty-edit-service-learning-course-proposals)

mupotsal:
@zachneill  and @Sandesh-L what issue where you working on that is almost similar to this isssue?

zachneill:
we made pr #98 which is the same page and has withdraw functionality. it was merged and the url is supposed to be /courseProposals. However, I can't seem to get it working while on the development branch.

PR #234, opened 2022-03-18
Create Edit SLC Proposals Page

zachneill:
This is for issue #79 . Existing proposals as well as newly created proposals can be edited. The database updates accordingly. Setting object to true in searchUser returns the actual object instead of a string. 

TO TEST 
Go to /serviceLearning/courseManagement
Click "Edit" in the dropdown on any proposal
Modify any values
Hit submit

We also have test suite: test_edit_slc_course.py

### DIFF
--- a/app/controllers/serviceLearning/routes.py
+++ b/app/controllers/serviceLearning/routes.py
@@ -11,7 +11,7 @@ from app.models.courseParticipant import CourseParticipant
 from app.controllers.serviceLearning import serviceLearning_bp
 from app.logic.searchUsers import searchUsers
 from app.logic.serviceLearningCoursesData import getServiceLearningCoursesData, withdrawProposal
-from app.logic.courseManagement import updatecourse
+from app.logic.courseManagement import updateCourse
 
 @serviceLearning_bp.route('/serviceLearning/courseManagement', methods = ['GET'])
 @serviceLearning_bp.route('/serviceLearning/courseManagement/<username>', methods = ['GET'])
@@ -65,7 +65,6 @@ def slcNewProposal():
         term = Term.get(Term.id==request.form.get("term"))
         status = CourseStatus.get(CourseStatus.status == "Pending")
         coursecheck = Course.get_or_none(Course.courseName == request.form.get("courseName"))
-        print("OVERHEREEEEEE", coursecheck)
         if coursecheck == None:
             course = Course.create(
                 courseName=request.form.get("courseName"),
@@ -89,7 +88,7 @@ def slcNewProposal():
                 CourseInstructor.create(course=course, user=instructor.username)
             return redirect('/serviceLearning/courseManagement')
         else:
-            updatecourse(request.form.copy())
+            updateCourse(request.form.copy())
 
     terms = Term.select().where(Term.year >= g.current_term.year)
     courseData = None
@@ -118,4 +117,4 @@ def withdrawCourse(courseID):
             flash("Unauthorized to perform this action", 'warning')
     except Exception as e:
         flash("Withdrawal Unsuccessful", 'warning')
-    return "" 
+    return ""
--- a/app/controllers/serviceLearning/routes.py
+++ b/app/controllers/serviceLearning/routes.py
@@ -89,7 +89,7 @@ def slcNewProposal():
                 CourseInstructor.create(course=course, user=instructor.username)
             return redirect('/serviceLearning/courseManagement')
         else:
-            updateCourse(request.form.copy())
+            updateCourse(request.form.copy(), instructorsDict)
             return redirect('/serviceLearning/courseManagement')
     terms = Term.select().where(Term.year >= g.current_term.year)
     courseData = None
@@ -112,14 +112,12 @@ def getInstructors():
 def updateInstructorPhone():
     try:
         instructorData = request.get_json()
-        print(instructorData)
         updateInstructorPhone = User.update(phoneNumber=instructorData[1]).where(User.username == instructorData[0]).execute()
-        flash("Instructor's phone number updated", 'success')
-        return ""
+        return "success"
     except Exception as e:
         print(e)
-        flash("Failed to update phone number", 'warning')
         return e
+        
 @serviceLearning_bp.route('/serviceLearning/withdraw/<courseID>', methods = ['POST'])
 def withdrawCourse(courseID):
     try:
--- a/app/controllers/serviceLearning/routes.py
+++ b/app/controllers/serviceLearning/routes.py
@@ -30,10 +30,13 @@ def serviceCourseManagement(username=None):
 @serviceLearning_bp.route('/serviceLearning/editProposal/<courseID>', methods=['GET', 'POST'])
 def slcEditProposal(courseID):
     courseData = CourseQuestion.get(CourseQuestion.course == courseID)
-    print(courseData.course.id)
-    print(courseData.course.courseName)
-    print(courseData.course.isRegularlyOccuring)
-    print(courseData.course.serviceLearningDesignatedSections)
+    courseInstructor = CourseInstructor.select().where(CourseInstructor.course == courseID)
+
+    for inst in courseInstructor:
+        print(inst.user.firstName)
+    isRegularlyOccuring = ""
+    isAllSectionsServiceLearning = ""
+    isPermanentlyDesignated = ""
 
     if courseData.course.isRegularlyOccuring:
         isRegularlyOccuring = "checked"
@@ -43,9 +46,17 @@ def slcEditProposal(courseID):
         isPermanentlyDesignated = "checked"
 
     terms = Term.select().where(Term.year >= g.current_term.year)
+
+    print(type(courseData.course.term))
+
+    for term in terms:
+        print(type(term))
+        if courseData.course.term == term.id:
+            print("Hello")
     return render_template('serviceLearning/slcNewProposal.html',
                                 courseData = courseData,
                                 terms = terms,
+                                courseInstructor = courseInstructor,
                                 isRegularlyOccuring = isRegularlyOccuring,
                                 isAllSectionsServiceLearning = isAllSectionsServiceLearning,
                                 isPermanentlyDesignated = isPermanentlyDesignated)
--- a/app/controllers/serviceLearning/routes.py
+++ b/app/controllers/serviceLearning/routes.py
@@ -108,6 +108,18 @@ def getInstructors():
     instructorsDict["instructors"] = instructorObjectList
     return jsonify({"Success": True}), 200
 
+@serviceLearning_bp.route('/updateInstructorPhone', methods=['POST'])
+def updateInstructorPhone():
+    try:
+        instructorData = request.get_json()
+        print(instructorData)
+        updateInstructorPhone = User.update(phoneNumber=instructorData[1]).where(User.username == instructorData[0]).execute()
+        flash("Instructor's phone number updated", 'success')
+        return ""
+    except Exception as e:
+        print(e)
+        flash("Failed to update phone number", 'warning')
+        return e
 @serviceLearning_bp.route('/serviceLearning/withdraw/<courseID>', methods = ['POST'])
 def withdrawCourse(courseID):
     try:
--- a/app/controllers/serviceLearning/routes.py
+++ b/app/controllers/serviceLearning/routes.py
@@ -26,13 +26,29 @@ def serviceCourseManagement(username=None):
     else:
         flash("Unauthorized to view page", 'warning')
         return redirect(url_for('main.events'))
+
 @serviceLearning_bp.route('/serviceLearning/editProposal/<courseID>', methods=['GET', 'POST'])
 def slcEditProposal(courseID):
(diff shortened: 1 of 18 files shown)


app/controllers/serviceLearning/routes.py
```diff
@@ -27,38 +28,72 @@ def serviceCourseManagement(username=None):
...
+    isRegularlyOccuring = ""
+    isAllSectionsServiceLearning = ""
+    isPermanentlyDesignated = ""
+
+    if courseData.course.isRegularlyOccuring:
+        isRegularlyOccuring = "checked"
+    if courseData.course.isAllSectionsServiceLearning:
+        isAllSectionsServiceLearning = "checked"
+    if courseData.course.isPermanentlyDesignated:
+        isPermanentlyDesignated = "checked"
```
BrianRamsay: 'checked' is a UI concern. The controller should just be concerned with True and False

app/controllers/serviceLearning/routes.py
```diff
@@ -27,38 +28,72 @@ def serviceCourseManagement(username=None):
...
+
+    terms = Term.select().where(Term.year >= g.current_term.year)
```
BrianRamsay: We have a function for getting terms. Can you use that or is a different set needed here?

app/controllers/serviceLearning/routes.py
```diff
@@ -27,38 +28,72 @@ def serviceCourseManagement(username=None):
...
+            for instructor in instructorsDict["instructors"]:
+                CourseInstructor.create(course=course, user=instructor.username)
+            return redirect('/serviceLearning/courseManagement')
+        else:
+            updateCourse(request.form.copy(), instructorsDict)
+            return redirect('/serviceLearning/courseManagement')
     terms = Term.select().where(Term.year >= g.current_term.year)
-    return render_template('serviceLearning/slcNewProposal.html', terms=terms)
+    courseData = None
+    return render_template('serviceLearning/slcNewProposal.html', terms=terms, courseData = courseData)
```
BrianRamsay, line 76: I would get this logic (create or update) into a function outside the controller so it's easier to test.

app/logic/courseManagement.py
```diff
@@ -41,3 +43,25 @@ def getinstructorData(courseIds):
...
+    course = Course.update(
+        courseName=courseData["courseName"],
+        courseAbbreviation=courseData["courseAbbreviation"],
+        courseCredit=courseData["credit"],
+        isRegularlyOccuring=1 if "on" in courseData["regularOccurenceToggle"] else 0,
+        term=courseData['term'],
+        status=status,
+        isAllSectionsServiceLearning=1 if "on" in courseData["slSectionsToggle"] else 0,
+        serviceLearningDesignatedSections=courseData["slDesignation"],
+        isPermanentlyDesignated=1 if "on" in courseData["permanentDesignation"] else 0,
```
BrianRamsay: More of a preference thing, but these boolean fields could be assigned with something a little shorter that uses booleans and avoids if/else, like: 
`isPermanentlyDesignated=("on" in courseData["permanentDesignation"])

app/logic/searchUsers.py
```diff
@@ -15,6 +15,8 @@ def searchUsers(query):
         for participant in results:
             if participant not in resultsDict:
                 resultsDict[f"{participant.firstName} {participant.lastName} ({participant.username})"] = f"{participant.firstName} {participant.lastName} ({participant.username})"
+                if object=="true":
+                    resultsDict[f"{participant.username} phoneNumber"] = participant.phoneNumber
```
BrianRamsay: This isn't very intuitive to me. If object is true, we add a key with a phone number in it? It probably makes sense for how you are using it, but searchUsers is a very generic function. The use cases should be clear in both comments and code.

app/logic/searchUsers.py
```diff
@@ -23,4 +25,8 @@ def searchUsers(query):
                 for participant in results:
                     if participant not in resultsDict:
                         resultsDict[f"{participant.firstName} {participant.lastName} ({participant.username})"] = f"{participant.firstName} {participant.lastName} ({participant.username})"
+                        if object=="true":
+                            resultsDict[f"{participant.username} phoneNumber"] = participant.phoneNumber
+
+    print("resultsDict", resultsDict)
```
BrianRamsay: debug output!

app/static/js/slcNewProposal.js
```diff
@@ -72,7 +72,7 @@ function validateForm() {
   if (valid) {
     $(".step")[currentTab].className += " finish"
   }
-  return valid;
+  return true;
```
BrianRamsay: Are we wanting our validation to always be true?

app/static/js/slcNewProposal.js
```diff
@@ -86,17 +86,42 @@ function fixStepIndicator(navigateTab) {
 
 // TODO: empty the courseInstructor input after an instructor has been added to the table.
 function callback() {
-  let instructor = $("#courseInstructor").val();
+  let data = $("#courseInstructor").val();
+  data = data.split(",");
```
BrianRamsay: Rather than reassign data from text to an array, I think the assignment line should do the split().

app/static/js/slcNewProposal.js
```diff
@@ -86,17 +86,42 @@ function fixStepIndicator(navigateTab) {
...
 function callback() {
-  let instructor = $("#courseInstructor").val();
+  let data = $("#courseInstructor").val();
+  data = data.split(",");
+  let instructor = data[0]
+  let username = instructor.split('(').pop().split(')')[0]; // returns 'two'
+  let phone = data[1]
   let tableBody = $("#instructorTable").find("tbody");
   let lastRow = tableBody.find("tr:last");
   let newRow = lastRow.clone();
```
BrianRamsay, line 98: you young'uns and your `let`s. Back in my day we only had `var`, and sometimes we didn't even use that!

tests/code/test_search.py
```diff
@@ -29,33 +29,33 @@ def test_searchStudents():
...
+    searchResults = searchUsers(query, False)
     assert len(searchResults) == 0
 
     #test with multiple users matching query
     with mainDB.atomic() as transaction:
         User.get_or_create(username = 'sawconc', firstName = 'Candace', lastName = 'Sawcon', bnumber = '021556782', isStudent = True)
         query = 'sa'
-        searchResults = searchUsers(query)
+        searchResults = searchUsers(query, False)
         assert len(searchResults) == 2
```
BrianRamsay, line 62: I don't see any tests with the new parameter being True. You could also add a default value, which would let you not update all of the existing function calls.

BrianRamsay (changes requested):

app/controllers/serviceLearning/routes.py
tylerpar99: We use the "Checked" keyword because we pass it to the font end using jinja and if it says "Checked" that is a keyword the HTML uses to check the box if it is not "Checked" HTML will ignore it. We could make it True or False and then uses jinja logic to make it checked or not but I figured it would be better to do the logic on the Python side.

app/static/js/slcNewProposal.js
tylerpar99: Hila had made this function to validate the proposal form however she mentioned that it would be more effective if we created a more general validation and I assumed that would fall under a separate issue.

app/static/js/slcNewProposal.js
tylerpar99: Hila had made this function to validate the proposal form however she mentioned that it would be more effective if we created a more general validation and I assumed that would fall under a separate issue.

app/static/js/slcNewProposal.js
tylerpar99: Don't hate the player hate the game

app/controllers/serviceLearning/routes.py
BrianRamsay: What if you changed it to not be a checkbox? Then "checked" no longer makes sense, and you're having to change python code for a UI concern. Let jinja (the View) handle UI decisions.

app/controllers/serviceLearning/routes.py
BrianRamsay: I meant all of it, including course creation.

app/logic/searchUsers.py
```diff
@@ -15,6 +15,8 @@ def searchUsers(query):
         for participant in results:
             if participant not in resultsDict:
                 resultsDict[f"{participant.firstName} {participant.lastName} ({participant.username})"] = f"{participant.firstName} {participant.lastName} ({participant.username})"
+                if phoneNumber=="true":
```
BrianRamsay: Why are we using the string "true" and not a boolean?

BrianRamsay: Also, the variable should have a different name that indicates what it controls. A variable called `phoneNumber` should just contain a phone number.

app/logic/searchUsers.py
```diff
@@ -23,4 +25,7 @@ def searchUsers(query):
                 for participant in results:
                     if participant not in resultsDict:
                         resultsDict[f"{participant.firstName} {participant.lastName} ({participant.username})"] = f"{participant.firstName} {participant.lastName} ({participant.username})"
+                        if phoneNumber=="true":
+                            resultsDict[f"{participant.username} phoneNumber"] = participant.phoneNumber
```
BrianRamsay: I'm still a little confused why we have a dictionary with weird keys like "first last username" and "username phone", instead of just a dictionary with "username" as a key and a dictionary of all of the user's properties as the value.

app/static/js/slcNewProposal.js
BrianRamsay: I would leave the specific one if it works. We can replace it once the general validation exists.

app/static/js/slcNewProposal.js
BrianRamsay: like `let data = $("#courseInstructor").val().split(",");`

app/static/js/slcNewProposal.js
```diff
@@ -86,17 +73,42 @@ function fixStepIndicator(navigateTab) {
...
+   $.ajax({
+     url: "/updateInstructorPhone",
+     data: JSON.stringify(instructorData),
+     type: "POST",
+     contentType: "application/json",
+     success: function(response) {
+         msgFlash("Instructor's phone number updated", "success")
+     },
+     error: function(request, status, error) {
+       console.log(status,error);
```
BrianRamsay: You don't want to directly output errors, in general. This could be ok in the javascript, except the controller in this case is passing the exception error straight through. That's too much information exposed. The controller should be sending back a more generic message that couldn't be used to get information about the system. Obviously it's not a huge worry in this application but for general web development it's an important principle. If you really wanted to get all of the error info, the controller could selectively send information based on whether we are in a development environment, or debug = True

tests/code/test_search.py
```diff
@@ -29,34 +29,36 @@ def test_searchStudents():
...
+    searchResults = searchUsers(query, False)
     assert len(searchResults) == 0
 
     #test with multiple users matching query
     with mainDB.atomic() as transaction:
         User.get_or_create(username = 'sawconc', firstName = 'Candace', lastName = 'Sawcon', bnumber = '021556782', isStudent = True)
         query = 'sa'
-        searchResults = searchUsers(query)
+        searchResults = searchUsers(query, False)
+        searchPhoneNumber = searchUsers(query, 'true')
```
BrianRamsay: You can see the mismatch here in datatypes. Is that parameter a string or is it Boolean? Of course, we can mix types in Python and some languages, but this isn't a good case for that, as we have complete control over what is passed.

BrianRamsay (changes requested):

app/controllers/serviceLearning/routes.py
zachneill: Fixed in next push

app/static/js/slcNewProposal.js
zachneill: Fixed in next push

app/static/js/slcNewProposal.js
zachneill: We just took this bad boy out

app/controllers/serviceLearning/routes.py
```diff
@@ -27,38 +29,51 @@ def serviceCourseManagement(username=None):
...
-                questionContent=request.form.get(f"{i}"),
-                questionNumber=i
-            )
-        for instructor in instructorsDict["instructors"]:
-            CourseInstructor.create(course=course, user=instructor.username)
-        flash("New service-learning course successfully created.", "success")
-        return redirect('/serviceLearning/courseManagement')
-
+        courseExist = Course.get_or_none(Course.id == request.form.get('courseID'))
+        if not courseExist:
```
BrianRamsay: Now we're getting nitpicky, but it's generally better to have positive statements in conditionals rather than negatives. Sometimes it's not avoidable, but in this case you can just swap the blocks and do `if courseExists`

BrianRamsay (changes requested):
Just one minor change! 

I like your change to searchUser.

BrianRamsay (approved):

PR #235, opened 2022-04-01
Volunteer profile fix

BrianRamsay:
A branch for resolving the merge conflicts with development


tests/code/test_participants.py
```diff
@@ -213,9 +213,10 @@ def test_sendUserData():
     # Tests the Kiosk
     # user is banned
     signedInUser, userStatus = sendUserData("B00739736", 2, 1)
-    assert userStatus == "banned"
-    with pytest.raises(DoesNotExist):
-        EventParticipant.get(EventParticipant.user==signedInUser, EventParticipant.event==2)
+    assert signedInUser, userStatus == "banned"
```
BrianRamsay: This is confusing. How does `assert` handle this? Are you testing that `signedInUser` is true, that `userStatus == "banned"`, or both?

Even if it works it's not clear what the intention is. If you want to test both, use two assert statements.

BrianRamsay (commented):
So close! Just one comment on a test.

tests/code/test_participants.py
gastonjarju: Added a transaction to the test function and also realized that the test database was missing some endDates for the banned user. Thus, I added a ban endDate to it as well.

BrianRamsay: This isn't actually resolved. They are still both on the same line. Do you need to commit and push some changes?

BrianRamsay (commented):
I think you might be missing a recent change.

tests/code/test_participants.py
gastonjarju: Commit and pushed!

BrianRamsay:
Woohoo! Merging.

Issue #205, opened 2021-11-22
Ordering on Event List in the Training and Education Tab


bryantal:
The ordering of events in the Training and Education tab needs to be ordered by what is upcoming/required.
First, All CELTS Training
Then, program-specific trainings
Then, optional educational events

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #205: Ordering on Event List in the Training and Education Tab](https://trello.com/c/sbzIrnro/372-celts-205-ordering-on-event-list-in-the-training-and-education-tab)

PR #236, opened 2022-04-15
Order event list 205

tylerpar99:
# This PR fixes issue #205 

## Summary: 
We have adjusted a back-end query that will now get whichever event is linked to the most programs in a given term. It will take the id of that link and another query that grabs all training will ensure that when the id of all training appears it will be first in the object and from there on will order by descending start date.

## To Test: 
You can check this out by getting the term for an all training event and switching over to the training tab in that term and you will see that all training appears first and afterward the events will appear by descending dates.

## Files changed: 
`logic/events`

## UI:
image


app/logic/events.py
```diff
@@ -118,11 +110,19 @@ def getStudentLedProgram(term):
...
+                                        .order_by(fn.COUNT(1).desc()))
+                                        .where(Event.term == term).get())
+    trainingEvents = ((Event.select(Event)
+			               .order_by((Event.id == allTrainingsEvent.event.id).desc(), Event.startDate.desc())
                            .where(Event.isTraining,
-                                  Event.term == term))
+                                  Event.term == term)))
+
+    for t in trainingEvents:
+        print(t)
```
BrianRamsay: We don't want the debug output

app/logic/events.py
```diff
@@ -118,11 +110,19 @@ def getStudentLedProgram(term):
...
 def getTrainingProgram(term):
 
-    trainingEvents = (Event.select(Event, Program.id.alias("program_id"))
-                           .join(ProgramEvent)
-                           .join(Program)
+    allTrainingsEvent = ((ProgramEvent.select(ProgramEvent.event, fn.COUNT(1).alias('num_programs'))
+                                        .join(Event)
+                                        .group_by(ProgramEvent.event)
+                                        .order_by(fn.COUNT(1).desc()))
+                                        .where(Event.term == term).get())
```
BrianRamsay, line 128: I'd put a comment above this query describing how it works, since some students won't be familiar with the aggregate functions

app/logic/events.py
```diff
@@ -165,7 +165,7 @@ def getUpcomingEventsForUser(user,asOf=datetime.datetime.now()):
 
 def getAllFacilitators():
 
-    facilitators = User.select(User).where((User.isFaculty == 1) | (User.isCeltsAdmin == 1) | (User.isCeltsStudentStaff == 1)).order_by(User.username) #ordered because of the tests
+    facilitators = User.select(User).where((User.isFaculty == 1) | (User.isCeltsAdmin == 1) | (User.isCeltsStudentStaff == 1))
```
BrianRamsay: You probably still want this to be ordered, right?

BrianRamsay (changes requested):

app/logic/events.py
```diff
@@ -31,9 +29,6 @@ def deleteEvent(eventId):
     event = Event.get_or_none(Event.id == eventId)
     if event:
         event.delete_instance(recursive = True, delete_nullable = True)
-        if event.startDate:
-            createLog(f"Deleted event: {event.name}, which had a startdate of {datetime.datetime.strftime(event.startDate, '%m/%d/%Y')}")
-
```
BrianRamsay, line 36: Don't we still want to log event deletion?

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #237, opened 2022-04-15
Rework sidebar structure


tylerpar99:
## To Do: 
- Currently, the sidebar has names that are no longer suitable for the purpose of the page. 
- Also, it would be more efficient to have a drop-down that will include all of the admin-specific toolings.
- Student staff does not need access to admin logs, course management, or admin management

## Current UI:
![image](https://user-images.githubusercontent.com/67805799/163592276-332b3c84-fcbc-48e3-8a31-a29f10c2f497.png)

tylerpar99:
https://trello.com/c/cDM579b5/384-celts-237-rework-sidebar-structure

PR #240, opened 2022-04-19
Rework sidebar 237

tylerpar99:
This PR fixes issue #237 

## Summary:
The sidebar structure had become outdated due to new changes to development. Now the sidebar will have an admin drop-down that will include Activity Logs (prev. Admin Logs), Course Proposals (prev. Course Proposal Management), and Settings (prev. Admin Management). This drop-down is not accessible as a Student Admin.  

## To Test:
Go to the CELTS web app as an admin and you will see the admin dropdown and the options held under it. Pay close attention to what happens after you click as the tab will remain actively highlighted but the drop-down will close and hide the options this can easily be reopened but I am not sure your feelings regarding this. Also, the clickable area is noticeably smaller than the regular nav links. Not sure if these differences are okay or if you would like to approach this differently.

## Previous UI:
image

### DIFF
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -212,7 +212,9 @@ def courseManagement(term = None):
                             term = term)
 @admin_bp.route('/adminLogs', methods = ['GET', 'POST'])
 def adminLogs():
-    allLogs = AdminLogs.select().order_by(AdminLogs.createdOn.desc())
-
-    return render_template("/admin/adminLogs.html",
-                            allLogs = allLogs)
+    if g.current_user.isCeltsAdmin:
+        allLogs = AdminLogs.select().order_by(AdminLogs.createdOn.desc())
+        return render_template("/admin/adminLogs.html",
+                                allLogs = allLogs)
+    else:
+        abort(403)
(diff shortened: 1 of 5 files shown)


app/logic/events.py
```diff
@@ -119,11 +112,19 @@ def getStudentLedProgram(term):
...
+                                        .order_by(fn.COUNT(1).desc()))
+                                        .where(Event.term == term).get())
+    trainingEvents = ((Event.select(Event)
+			               .order_by((Event.id == allTrainingsEvent.event.id).desc(), Event.startDate.desc())
                            .where(Event.isTraining,
-                                  Event.term == term))
+                                  Event.term == term)))
+
+    for t in trainingEvents:
+        print(t)
```
BrianRamsay: debug output

BrianRamsay (changes requested):
Let's add a few nice things. 
1. the Admin menu should load in an expanded position if we are on one of the pages in the submenu. 
2. an open/closed indicator would be nice (or at least a closed one) so it is obvious that it is expandable.
3. there's a bit of a gap between the items in the submenu and the Admin header

BrianRamsay (approved):
Haha, wow, you changed those things a lot faster than I thought you would!

app/logic/events.py
tylerpar99: I think this disappeared when I merged development in

PR #248, opened 2022-04-25
Updating Test Data and Adding Faculty User to Switch User Dropdown

bryantal:
This PR changes the test data in CELTS to make the current term Spring 2022, change events to be in that term, adds a new user to be a faculty for testing, gives that faculty a new course, and adds that faculty user to the switch user dropdown.


BrianRamsay (changes requested):
You broke a bunch of tests with these changes. You should always have the tests running when you are developing (especially with test data).

bryantal:
> You broke a bunch of tests with these changes. You should always have the tests running when you are developing (especially with test data).

Right, yeah I can remove most of the database changes.  I was making them because I thought we were going to demo certain things, and we ended up not needing to.  I'll remove the changes to events, but leave the background check name changes and the addition of a Faculty into the switch user dropdown, and then fix any test failures from there.

BrianRamsay (approved):

Issue #243, opened 2022-04-21
SL Courses Created Don't Show Up in List of Faculty's Courses from Faculty View


bryantal:
When a faculty creates an SL course, once the page refreshes and it goes back to the list view of the faculty's SL courses, the new course doesn't show up.  I think this has to do with the user creating the course being able to enter in a name for the faculty and continue to the next page without the instructor search bar returning results.  Needs more investigation.
![image](https://user-images.githubusercontent.com/28715731/164542583-821ec207-7879-4b49-b84b-6c60907d1aae.png)

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #243 SL Courses Created Don't Show Up in List of Faculty's Courses from Faculty View](https://trello.com/c/t4kjUs1D/388-celts-243-sl-courses-created-dont-show-up-in-list-of-facultys-courses-from-faculty-view)

PR #249, opened 2022-04-26
List view 243

tylerpar99:
This PR fixes issue #243 

## Summary: 
An issue was occurring in which the SLC proposal form was not properly validating that an instructor was actually added to the course. This was causing courses to be created that would not have an instructor and would not show up under the faculty's list of proposals. To fix this we added a validation check that will count the `<td>` tags and ensure there are more than 5 tags, that way we know that at least one instructor has been added to the course. If there are less than 5 we block the next page until an instructor is added.

## To test:
If you create a course and the instructor is not added attempt to hit the next button and you will see it is blocked. If you then proceed to add at least one instructor you will see it will not allow you to proceed.


app/logic/searchUsers.py
```diff
@@ -2,6 +2,7 @@
 from app.models.user import User
 def searchUsers(query):
     '''Accepts user input and queries the database returning results that matches user search'''
+    print(query)
```
BrianRamsay: debug output

BrianRamsay (changes requested):
one change and a merge conflict

BrianRamsay (approved):

PR #269, opened 2022-05-10
Updated titles on all the pages Issue #259

### LINKED ISSUE #259: Update Page Titles
When demoing to CELTS, we noticed that the page titles are all "Welcome to CELTS."  This might be confusing when users are using the back and forward buttons or viewing their recent history to see what pages they were on.  Let's give each page a unique title relevant to the page

image

Karina-Agliullova:
**Problem:**  The page titles are all "Welcome to CELTS." This is confusing when users are using the back and forward buttons or viewing their recent history to see what pages they were on

**Solution:** We added titles to all the HTML pages so that users can view their pages and recent history with specific names of the pages on CELTS

**Testing:** We tested it by visiting all the pages, using back and forward buttons, and then we checked the recent history and it was showing the updated page titles.

Issue #259 
Closes #259


BrianRamsay:
The title tag should go inside the `<head>` tag (in the base template `base.html`)

BrianRamsay (changes requested):
Merge development into the branch and see if that takes care of the extra admin log changes (it should, if those were already merged into development).

app/templates/serviceLearning/slcManagment.html
```diff
@@ -1,3 +1,4 @@
+{%set title = "Service Learning Courses Management"%}
```
BrianRamsay: I would abbreviate all of the service learning pages to SLC Title. like, "SLC Management"

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #244, opened 2022-04-22
Error Flash When Volunteer Marks Interest


bryantal:
When marking interest as a volunteer on the volunteer profile, the user is able to mark interest and have it show up correctly in the database.  However, an error flash appears, and the user is not able to see the events that are upcoming in the event table.  An error flash appears above as well.  
Here is the error message that shows up in the console:
![image](https://user-images.githubusercontent.com/28715731/164772478-2cecd7db-085a-46e0-80b2-98fabd72cc7c.png)

This looks like it has to do with the add and remove interest functions in _app/logic/users.py_ both returning true when they probably shouldn't.

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #244: Error Flash When Volunteer Marks Interest #244](https://trello.com/c/2x97cGUf/389-celts-244-error-flash-when-volunteer-marks-interest-244)

PR #273, opened 2022-06-03
Got rid of the error flash when user marked interest in volunteer pro…

AndersonStettner:
Fixes issue #244


app/logic/users.py
```diff
@@ -26,7 +26,7 @@ def addUserInterest(program_id, username):
     username: username of the user showing interest
     """
     Interest.get_or_create(program = program_id, user = username)
-    return True
+    return ""
```
BrianRamsay: These functions are supposed to return True if they succeed. The error is in what comes back from the controller.

BrianRamsay (changes requested):

app/controllers/main/routes.py
```diff
@@ -174,7 +176,9 @@ def removeInterest(program_id, username):
     username: unique value of a user to correctly identify them
     """
     try:
-        return removeUserInterest(program_id, username)
+        failed = removeUserInterest(program_id, username)
```
BrianRamsay: It's not a failure if they remove the user interest. That's what we are wanting to do!

app/controllers/main/routes.py
```diff
@@ -160,7 +160,9 @@ def addInterest(program_id, username):
     username: unique value of a user to correctly identify them
     """
     try:
-        return addUserInterest(program_id, username)
+        success = addUserInterest(program_id, username)
+        if success == True:
```
BrianRamsay: success is already a boolean, so `success == True` is redundant in an `if` condition. Just use your boolean you already have, e.g., `if success:`

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #265, opened 2022-05-03
Add Custom Error Pages for 404, 403, and 500 Errors


bryantal:
We need to create custom error pages for errors.  Right now we just use the flask functionality of aborting.  We should reroute to custom error pages once we have created them.  Look for any time a controller calls abort() and make it reroute to the correlating new error page once they are created.

PR #286, opened 2022-06-13
Added custom error pages and routed abort calls to associated error page 

martinj2-dot:
Issue #265  

Created custom HTML template files for error codes: 403, 404, and 500. Routed all abort() calls associated with the 403, 404, and 500 errors through main/routes.py using @app.errorhandler function calls. 

How to test:
- 403: enter a page as a student to a page that students do not have access to.
- 404: enter a url that does not exist within the application.
- 500: add an abort(500) call in one of the pages or break the internal logic of the app.

HTML files can be located at the following directory: /app/templates/errors

Error handling can be located in the following file: /app/controllers/main/routes.py


app/templates/errors/403error.html
```diff
@@ -0,0 +1,17 @@
+<!--Custom HTML error template for Error Code 403-->
+
+{% extends "base.html"%}
+
+{% block app_content %}
+<div class="text-center">
+    <h1 class="display-1">Error 403</h1>
+
+    <p class="">You do not have permissions to view this page. If you believe this is a mistake, please contact the system administrator.</p>
```
BrianRamsay: If we tell them to contact the system administrator, we should give them the information they need. Not sure if we have a System email address in config anywhere yet.

app/templates/errors/500error.html
```diff
@@ -0,0 +1,15 @@
+<!--Custom HTML error template for Error Code 500-->
+
+{% extends "base.html"%}
+
+
+{% block app_content %}
+<div class="text-center">
+    <h1 class="display-1">Error 500</h1>
+
+    <p class="">Something went wrong on our end, please try again later. If the problem persists, please contact the system administrator.</p>
```
BrianRamsay: I'd change 'later' to 'in a few minutes'

app/controllers/main/routes.py
```diff
@@ -287,3 +287,20 @@ def getAllCourseIntructors(term=None):
...
+def handle_bad_request(e):
+    return render_template("/errors/403error.html")
+
+@app.errorhandler(404)
+def handle_bad_request(e):
+    return render_template("/errors/404error.html")
+
+@app.errorhandler(500)
+def handle_bad_request(e):
+    return render_template("/errors/500error.html")
```
BrianRamsay: If this is app-level error handling it shouldn't be in `controllers/main/routes.py`, it should be in `app/__init__.py` (probably. somewhere more global in any case)

BrianRamsay (changes requested):

app/templates/errors/403error.html
```diff
@@ -0,0 +1,18 @@
+<!--Custom HTML error template for Error Code 403-->
+
+{% extends "base.html"%}
+
+{% block app_content %}
+<div class="text-center">
+    <h1 class="display-1">Error 403</h1>
+
+    <p class="">You do not have permissions to view this page. If you believe this is a mistake,
+                <a href="mailto:heggens@berea.edu,ramsayb2@berea.edu">please contact the system administrators.</a></p>
```
BrianRamsay: Having to put this in 3 different places is a good clue that it needs to be refactored. The support email is support@bereacollege.onmicrosoft.com, and it should be in a config item.

app/templates/errors/404error.html
```diff
@@ -0,0 +1,15 @@
...
+{% extends "base.html"%}
+
+{% block app_content %}
+<div class="text-center">
+    <h1 class="display-1">Error 404: Page Not Found</h1>
+
+    <p class="">The page you tried to access does not exist. If you believe this is a mistake,
+                <a href="mailto:heggens@berea.edu,ramsayb2@berea.edu">please contact the system administrators.</a></p>
+    <br>
+    <a class="btn btn-primary" href="/" role=button>Return Home</a>
```
BrianRamsay: all attribute values need to have quotes around them

BrianRamsay (changes requested):

BrianRamsay (changes requested):
I think the font-size for the error messages is a bit dramatic. Let's use something not quite so huge.

Also, for the individual error titles themselves, the users don't need to see the error codes. 404 should be "Page Not Found", 403 should be "Access Denied", and 500 should be "System Error"

BrianRamsay (approved):

Issue #272, opened 2022-05-23
Faculty View in CELTS


bryantal:
If the user is a faculty, the landing page they get is the event list, and they do not have access to their course proposal management page in the sidebar.  We should change this to make the course proposal management page the page they see when they log in, and that is the only page they should have access to in the sidebar.

![image](https://user-images.githubusercontent.com/28715731/169845207-70d68988-a0e9-4bdd-895e-2040eb894cb3.png)

Boorclark:
PR 272 solved this issue and this has been merged. Closing issue.

PR #288, opened 2022-06-13
fixes issue 272 allows faculty to view only course proposal

qasema:
fixes issue #272 allowing faculty to only view course proposals!

### DIFF
--- a/app/templates/sidebar.html
+++ b/app/templates/sidebar.html
@@ -17,6 +17,7 @@
         </a>
       </li>
     {% endif %}
+    {% if not g.current_user.isFaculty %}
     <li class="nav-item">
       <a href="/eventsList/{{g.current_term}}"
          class="nav-link text-white {{'active' if 'eventsList' in request.path}}"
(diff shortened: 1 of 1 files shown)


app/templates/sidebar.html
```diff
@@ -16,6 +16,22 @@
...
       </li>
+    {% endif %} 
+    {% if g.current_user.isFaculty %}
+    <li>
+      <a href="/profile/{{g.current_user}}"
+         class="nav-link text-white {{'active' if 'profile' in request.path }}"
+         {{"aria-current='page'" if 'profile' in request.path }}>
+        My Profile
+      </a>
+    </li>
```
BrianRamsay, line 18: Since this is identical to lines 12-18 above, we should think about just including 'my profile' for everyone. I'm leaning towards yes. That would get it out of the 'if' statements

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #262, opened 2022-05-03
Program Manager Box is Missing for Admins


bryantal:
On the volunteer profile, the section where admins can make the volunteer a profile manager is not showing up.  This is probably and error with the Jinja in the html file or the controller for the volunteer profile.  This section should also not appear if the volunteer is not a student staff in our database, which it is not making this distinction now.

routes: http://[YOUR-IP]:8080/profile/[VOLUNTEER-USERNAME]
files: /app/templates/volunteer/volunteerProfile.html
        /app/controllers/main/routes.py

![image](https://user-images.githubusercontent.com/28715731/166537204-bf73d5a8-14ff-4031-ad31-d2b2af6e7a1e.png)

PR #297, opened 2022-06-15
submits check boxes that makes student staff as program manager, rela…

qasema:
this is related to issue 262. The check boxes that allow the admin to add a student staff to be a program manager do not update that in the database. 

fixes issue #262


app/controllers/admin/volunteers.py
```diff
@@ -109,3 +110,14 @@ def updateBackgroundCheck():
         type = eventData['bgType']
         setUserBackgroundCheck(user,type, checkPassed)
         return " "
+
+@admin_bp.route('/updateProgramManager', methods=["POST"])
+def updateProgramManager():
+    if g.current_user.isCeltsAdmin:
```
BrianRamsay, line 116: There should be an `else` clause that throws a 403 if they aren't an admin, `abort(403)`

app/logic/volunteers.py
```diff
@@ -1,8 +1,12 @@
+from tkinter import INSERT
+from sqlalchemy import insert
```
BrianRamsay, line 2: These two imports are definitely not used

app/static/js/volunteerProfile.js
```diff
@@ -109,3 +109,15 @@ $(document).ready(function(){
   });
 
 });
+
+function updateManagers(el, volunteerusername, action ){// retrieve the data of the studnet staff and program id if the boxes are checked or not
```
BrianRamsay, line 113: `volunteerusername` should be `volunteer_username` to be consistent with your other variables

app/templates/main/volunteerProfile.html
```diff
@@ -78,7 +78,7 @@ <h2><p class="text-dark" align="center">Choose interests below to be notified ab
         {% endif %}
           <tr>
             <td id="programNameTable">{{row.program.programName}}</td>
-            <td><input class="form-check-input" type="checkbox" data-programid="{{row.program.id}}" data-username="{{volunteer.username}}" {{checked}}></td>
+            <td><input class="form-check-input" type="checkbox" value="{{row.program.id}}" data-username="{{volunteer.username}}" {{checked}}></td>
```
BrianRamsay, line 81: Are you sure this attribute should be changed from a data attribute to a value?

app/logic/volunteers.py
```diff
@@ -92,3 +96,19 @@ def setUserBackgroundCheck(user,bgType, checkPassed):
     update = BackgroundCheck.create(user=user, type=bgType, passBackgroundCheck=checkPassed, datePassed=today)
     update.save()
     createLog(f"Updated {user.firstName} {user.lastName}'s background check for {bgType} to {bool(checkPassed)}.")
+
+def setPromgramManager(user_name, program_id, action):
```
BrianRamsay: Misspelling in function name

app/templates/main/volunteerProfile.html
```diff
@@ -122,7 +122,7 @@ <h2>Which programs do you want to make {{volunteer.firstName}} {{volunteer.lastN
                   {% set action = "remove" %}
                   {% set checked = "checked" %}
             {% endif %}
-                  <input  type="checkbox" id="mgr-{{ program.id }}" data-programid="{{program.id}}" {{checked}}  onclick="updateManagers(this,'{{volunteer.username}}','{{action}}')">
+                  <input class="program_checkboxes" type="checkbox" id="mgr-{{ program.id }}" data-programid="{{program.id}}" {{checked}}  onclick="updateManagers(this,'{{volunteer.username}}','{{action}}')">
```
BrianRamsay, line 125: I don't see where you define or reference this class. If you do still need it, it should be singular, and use a dash instead of an underscore: `program-checkbox`

BrianRamsay (changes requested):

BrianRamsay:
There was a bad conflict resolution. Check out the list of file changes. A lot of those are reversions of work others have done. We'll talk about how to fix it on Friday.

BrianRamsay:
There are still outstanding changes needed, @qasema and @solijonovam

Issue #268, opened 2022-05-04
Ensure Admin Pages Check Permissions Properly


bryantal:
Some admin pages in CELTS also restrict access if the user is a student staff when they shouldn't.  We need to make sure CELTS student staff can access the event pages they are program managers for.  Right now, program managers can edit the event they are program managers for, but they cannot view the track volunteers page.  Check the database and make your user is a program manager for certain program, so you can test if they can view the corresponding event pages.

files: /app/controllers/admin/volunteers.py

It would also be a good idea to check all of the controller files in /app/controllers/admin to see if they are checking permissions correctly.

Boorclark:
This issue was fixed with PR 299 and it has been merged. Closing issue.

PR #299, opened 2022-06-16
Issue#268 admin check permessions

martinj2-dot:
Issue #268 

Created a test suite for getProgramManagerForEvent
Created a function getProgramManagerForEvent to allow Program managers to: return a peewee query that contains the database entry that matches a user to a program as managers.

Test suite location: /tests/code/test_volunteers.py
Function location: /app/logic/volunnteers.py
Added functionality location: app/controllers/admin/routes.py

Functionality allowed program managers to view and edit events they are the manager of.

Non-mangers can still view events in general.


martinj2-dot:
We think this issue is ready for review. We are still working on reducing the amount of queries in the test cases and making them self sufficient (Creating test data and deleting it once the tests finish.). We also would like to know where we could implement this function maybe that could be a separate issue?

martinj2-dot:
Used test data to create temporary rows in the database to test our function. Reduced the amount of queries from 12 to 8 and encapsulated it in a transaction to delete the data once test finishes. Added comments and cleaned up the code to improve readability.

app/controllers/admin/routes.py
```diff
@@ -62,7 +62,7 @@ def templateSelect():
 @admin_bp.route('/eventTemplates/<templateid>/create', methods=['GET','POST'])
 @admin_bp.route('/eventTemplates/<templateid>/<programid>/create', methods=['GET','POST'])
 def createEvent(templateid, programid=None):
-    if not g.current_user.isAdmin:
+    if not (g.current_user.isAdmin or getProgramManagerForEvent(g.current_user, programId = programid)):
```
BrianRamsay: What happens here if there is no program id from the route?

app/logic/volunteers.py
```diff
@@ -92,3 +97,34 @@ def setUserBackgroundCheck(user,bgType, checkPassed, datePassed):
...
+
+def getProgramManagerForEvent(user, event= None, programId = None):
+    """
+    This function checks to see if a user is a program manager for a program.
+    NOTE: this function needs either the event or programId parameters to work
+    user: expects a user peewee object
+    event: expects an event peewee object
+    programId: expects an int that is the primary ID of a program in the database.
+    Returns: bool whether the appropriate information is given.
+    """
```
BrianRamsay, line 111: I like the description of the function

app/logic/volunteers.py
```diff
@@ -92,3 +97,34 @@ def setUserBackgroundCheck(user,bgType, checkPassed, datePassed):
     update = BackgroundCheck.create(user=user, type=bgType, passBackgroundCheck=checkPassed, datePassed=datePassed)
     update.save()
     createLog(f"Updated {user.firstName} {user.lastName}'s background check for {bgType} to {bool(checkPassed)}.")
+
+def getProgramManagerForEvent(user, event= None, programId = None):
```
BrianRamsay: Two issues here:

1. The name is a bit off. It makes me think this is a function for figuring out who is the program manager for an event. If it returns a boolean, it should be something like `isProgramManagerForEvent`.
2. We already have a function to check if a user has permissions to mess with the program, `hasPrivilege` (also not a great name, admittedly). this one should just take a user and event, and call hasPrivilege to check the program.

BrianRamsay (changes requested):

app/logic/volunteers.py
```diff
@@ -92,3 +98,26 @@ def setUserBackgroundCheck(user,bgType, checkPassed, datePassed):
     update = BackgroundCheck.create(user=user, type=bgType, passBackgroundCheck=checkPassed, datePassed=datePassed)
     update.save()
     createLog(f"Updated {user.firstName} {user.lastName}'s background check for {bgType} to {bool(checkPassed)}.")
+
+def isProgramManagerForEvent(user, event= None):
```
BrianRamsay: Since you are not using program id here anymore, and event is required, you can remove the default value.

app/logic/volunteers.py
```diff
@@ -92,3 +98,26 @@ def setUserBackgroundCheck(user,bgType, checkPassed, datePassed):
...
+    This function checks to see if a user is a program manager for a program.
+    NOTE: this function needs the event parameter to work
+    user: expects a user peewee object
+    event: expects an event peewee object
+    Returns: bool whether the appropriate information is given.
+    """
+
+    #neither event nor programId are passed
+    if not (event):
+        raise ValueError("Not enough parameters given to this function.")
```
BrianRamsay: Once you've removed the default value None, this bit isn't necessary anymore, as python will enforce passing the argument

app/logic/volunteers.py
```diff
@@ -92,3 +98,26 @@ def setUserBackgroundCheck(user,bgType, checkPassed, datePassed):
...
+
+    #if event is passed but no programId is passed
+    if event:
+        programIdQuery = (ProgramEvent.get(ProgramEvent.event == event)).program
```
BrianRamsay: The Event class has a property `singleProgram`, which you can use to get the program id from an event or None if there are multiple programs or no programs

app/logic/volunteers.py
```diff
@@ -92,3 +98,26 @@ def setUserBackgroundCheck(user,bgType, checkPassed, datePassed):
...
+        programManagerResult = hasPrivilege(user, programIdQuery)
+    if programManagerResult == True:
+        return True
+    else:
+        # print(e)
+        return False
```
BrianRamsay: Instead of saying essentially "if true equals true return true", just return the result. It's already True or False.

tests/code/test_volunteers.py
```diff
@@ -136,3 +143,116 @@ def test_backgroundCheck():
...
+
+        #gives event but no programId
+        ## user is manager of program
+        studentManager = isProgramManagerForEvent(programManager, test_event)
+        print(studentManager)
+        assert studentManager == True
+
+        ## user is not manager of program
+        studentManager = isProgramManagerForEvent(student, test_event)
+        print(studentManager)
```
BrianRamsay: Remove the print statements

BrianRamsay (changes requested):

app/controllers/admin/routes.py
```diff
@@ -62,7 +62,7 @@ def templateSelect():
 @admin_bp.route('/eventTemplates/<templateid>/create', methods=['GET','POST'])
 @admin_bp.route('/eventTemplates/<templateid>/<programid>/create', methods=['GET','POST'])
 def createEvent(templateid, programid=None):
-    if not g.current_user.isAdmin:
+    if not (g.current_user.isAdmin):
```
BrianRamsay: The program manager should be able to create events for their programs, right? I thought this was already in there (and I see you are doing it down in edit event).

app/controllers/admin/volunteers.py
```diff
@@ -37,8 +38,9 @@ def trackVolunteersPage(eventID):
 
     trainedParticipantsList = trainedParticipants(program, g.current_term)
     eventParticipants = getEventParticipants(event)
+    programManagerResult = isProgramManagerForEvent(g.current_user, event)
```
BrianRamsay: it's a good idea to name boolean variables so that they make expressions clearer. In this case, like `isProgramManager` or `canManageProgram`

app/logic/volunteers.py
```diff
@@ -92,3 +98,17 @@ def setUserBackgroundCheck(user,bgType, checkPassed, datePassed):
...
+def isProgramManagerForEvent(user, event):
+    """
+    This function checks to see if a user is a program manager for a program.
+    NOTE: this function needs the event parameter to work
+    user: expects a user peewee object
+    event: expects an event peewee object
+    Returns: bool whether the appropriate information is given.
+    """
+
+    programIdQuery = event.singleProgram
```
BrianRamsay: I would eliminate the variable here. It doesn't make it clearer (to me)

BrianRamsay (changes requested):

app/controllers/admin/routes.py
martinj2-dot: This is the only one we are confused about. Scott and I looked at this and don't know where to check if a user is a program manager for an event. Should it be located in validateNewEventData where all event data in the template is validated or should it check in attemptSaveEvent?

app/controllers/admin/routes.py
```diff
@@ -191,7 +191,7 @@ def addParticipants():
 @admin_bp.route('/adminLogs', methods = ['GET', 'POST'])
 def adminLogs():
     if g.current_user.isCeltsAdmin:
-        allLogs = AdminLogs.select(AdminLogs, User).join(User).order_by(AdminLogs.createdOn.desc())
+        allLogs = AdminLogs.select().order_by(AdminLogs.createdOn.desc())
```
BrianRamsay: This reverts the fix for the Admin Logs excess queries

BrianRamsay: Not sure how this snuck in there.

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #242, opened 2022-04-21
Event Time Selector Throws and Error When Selecting Minutes


bryantal:
When creating a new event, if you select the time to have minutes (such as 10:30 instead of 10:00), there will be an error and the event will not be created.  If you mark the event as requiring RSVP, then the event will get created.  This will need more investigation.
Here is the console error it gives:
![image](https://user-images.githubusercontent.com/28715731/164533460-bb4a4a75-b0c2-4334-a4b2-3a28d3f463f1.png)

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #242: Event Time Selector Throws and Error When Selecting Minutes](https://trello.com/c/B9jZQtAn/387-celts-242-event-time-selector-throws-and-error-when-selecting-minutes)

Boorclark:
Note:
- We came to the conclusion that the root of this issue is storing the am vs pm on the time slots. If you select two times in the am or pm it's fine but when you do am and pm an error is thrown.

Boorclark:
Note:
 - New update is that some of the PMs in the DB are converting to AMs and being stored not in 24hr time. We have to look into what is causing this error.

mohamudm1:
Note: 

On the createSingleEvents.html file, there might be an issue with the time formatting. Eg. Instead of using %I on the ("%I: %m %p") it might be a good idea to use %H which represents a 24-hour format!

Boorclark:
This issue was fixed with PR 301 and it has been merged. Closing issue.

PR #301, opened 2022-06-16
#242 Event Time Selector - Addressed Changes Requested

Boorclark:
Things Completed:
 - Fixed the formatting for event time handling
 - Fixed how this data is passed to the DB
 - Fixed the HTML logic for time formatting (%I - %H)  -> %H represents 24 hour time
 Fixed Issue #242 
image


app/logic/events.py
```diff
@@ -14,7 +14,7 @@
 from app.models.eventTemplate import EventTemplate
 from app.models.programEvent import ProgramEvent
 from app.logic.adminLogs import createLog
-
+import datetime
```
BrianRamsay: To avoid `datetime.datetime`, you can import like `from datetime import datetime`

app/logic/events.py
```diff
@@ -178,6 +178,15 @@ def getAllFacilitators():
     facilitators = User.select(User).where((User.isFaculty == 1) | (User.isCeltsAdmin == 1) | (User.isCeltsStudentStaff == 1)).order_by(User.username) # ordered because of the tests
     return facilitators
 
+def getStartTime(data):
```
BrianRamsay: A more descriptive name would be something like `get24HourTime`

```
def get24HourTime(time_str):
    return datetime.strptime(time_str, "%I:%M %p").strftime("%H:%M")
```

app/logic/events.py
```diff
@@ -186,14 +195,15 @@ def validateNewEventData(data):
 
         Returns 3 values: (boolean success, the validation error message, the data object)
     """
+    # timeEnd = datetime.time(data['timeEnd'])
```
BrianRamsay: remove extra commented code

app/logic/events.py
BrianRamsay: This function would be good in `logic/utils.py`, and make sure there is a test for a few different use cases.

BrianRamsay: A better name than `get24HourTime` would be something like `format24HourTime`

BrianRamsay (changes requested):

app/static/js/createEvents.js
```diff
@@ -31,13 +31,16 @@ $(document).ready(function() {
     }
   });
 
-  if (navigator.userAgent.indexOf("Firefox") != -1 || navigator.userAgent.indexOf("Safari") != -1) {
+  // everything except Chrome
+  if (navigator.userAgent.indexOf("Chrome") == -1) {
+    console.log ("But we have Chrome")
+    console.log(navigator.userAgent)
```
zachneill: Remove test console.log statements

app/logic/utils.py
```diff
@@ -42,3 +43,10 @@ def getStartofCurrentAcademicYear(currentTerm):
         fallTerm = Term.select().where(Term.year==currentTerm.year-1, Term.description == f"Fall {currentTerm.year-1}").get()
         return fallTerm
     return currentTerm
+
+def format24HourTime(time_str):
+    if type(time_str) is str:
+        return datetime.strptime(time_str, "%I:%M %p").strftime("%H:%M") # Converts string to datetime and formats correctly
+    else:
+        print("I am here")
```
zachneill: Remove test print statements

app/static/js/createEvents.js
Boorclark: We need print statements and console log statements for monday. They will be removed once issue is resolved.

app/logic/events.py
```diff
@@ -193,7 +193,7 @@ def validateNewEventData(data):
     if data['isRecurring'] and data['endDate']  <  data['startDate']:
         return (False, "Event start date is after event end date")
 
-    if data['endDate'] ==  data['startDate'] and data['timeEnd'] <=  data['timeStart']:
+    if data['endDate'] ==  data['startDate'] and datetime.datetime.strptime(data['timeEnd'], "%I:%M %p") <= datetime.datetime.strptime(data['timeStart'], "%I:%M %p"):
```
BrianRamsay: I'm confused about this, because this function should always be called after the function that processes the event data. And that function makes sure that startTime and endTime are in 24 hour time. But this `strptime` formatting is for 12 hour time.

tests/code/test_events.py
```diff
@@ -523,3 +520,20 @@ def test_userWithNoInterestedEvent():
...
+
+@pytest.mark.integration
+def test_format24HourTime():
+
+    # tests valid "input times"
+    assert format24HourTime('08:00 AM')
+    assert format24HourTime('8:00 AM')
+    assert format24HourTime('05:00 PM')
+    assert format24HourTime('07:30 PM')
+    assert format24HourTime('12:00 PM')
```
BrianRamsay: You should be checking that the function actually does this properly. right now you are just making sure there is no exception thrown.

app/logic/events.py
```diff
@@ -277,6 +276,12 @@ def preprocessEventData(eventData):
         except DoesNotExist:
             eventData['term'] = ''
 
+    if 'startTime' in eventData:
+        eventData['startTime'] = format24HourTime(eventData['startTime'])
+
+    if 'endTime' in eventData:
+        eventData['endTime'] = format24HourTime(eventData['endTime'])
```
BrianRamsay: I'm a little confused if we are passing around a string of 24 hour time everywhere, or a datetime object

BrianRamsay: The code in the view assumes it is an object

BrianRamsay (changes requested):

app/logic/events.py
Lyisus: That was a fix for the test data that still passed 12 hour format. We've looked through the tests that failed when using 24 hour format and updated them to 24 hour format. They're datetime objects to be able to do the comparison as data['timeEnd'] and data['timeStart'] both hold strings that are later added to the database.

tests/code/test_events.py
Boorclark: We corrected and expanded on the tests.

app/logic/events.py
Boorclark: This function is expecting a string in a 12-hour format because this is where we are processing the data that comes from the user (js). However, we did fix the if statements because it is supposed to be 'timeStart' and 'timeEnd'.

BrianRamsay:
I think you broke something when you were fixing it up for firefox. It's now expecting the 12 hour time string, instead of a date time.

When I created a new event:
`ValueError: time data '15:58' does not match format '%I:%M %p'`

```
File "/home/brian/dev/berea/celts/app/controllers/admin/routes.py", line 105, in createEvent
File "/home/brian/dev/berea/celts/app/logic/events.py", line 281, in preprocessEventData
eventData['timeStart'] = format24HourTime(eventData['timeStart'])
File "/home/brian/dev/berea/celts/app/logic/utils.py", line 53, in format24HourTime
time = datetime.strptime(timeStr, "%I:%M %p").strftime("%H:%M") # Converts string to datetime and formats correctly
```

app/logic/events.py
```diff
@@ -202,7 +203,7 @@ def validateNewEventData(data):
     if data['isRecurring'] and data['endDate']  <  data['startDate']:
         return (False, "Event start date is after event end date")
 
-    if data['endDate'] ==  data['startDate'] and data['timeEnd'] <=  data['timeStart']:
+    if data['endDate'] ==  data['startDate'] and datetime.datetime.strptime(data['timeEnd'], "%H:%M") <= datetime.datetime.strptime(data['timeStart'], "%H:%M"):
```
BrianRamsay: It would shorten this conditional if you just compared the dates. This function assumes that `preprocessEventData` has been run, so those dates will be 24 hour time, and you can compare 24 hour strings with < and >

BrianRamsay (changes requested):

app/logic/utils.py
```diff
@@ -42,3 +43,15 @@ def getStartofCurrentAcademicYear(currentTerm):
         fallTerm = Term.select().where(Term.year==currentTerm.year-1, Term.description == f"Fall {currentTerm.year-1}").get()
         return fallTerm
     return currentTerm
+
+def format24HourTime(timeStr):
```
BrianRamsay: To save ourselves some potential pain later, let's also handle if timeStr is a datetime object.

BrianRamsay (changes requested):

app/logic/utils.py
```diff
@@ -42,3 +43,20 @@ def getStartofCurrentAcademicYear(currentTerm):
...
+    unformattedTime: expects a string with format HH:mm AM/PM or HH:mm OR a datetime object
+    returns: a string in 24 hour format HH:mm
+    """
+    if type(unformattedTime) == str:
+        try:
+            formattedTime = datetime.strptime(unformattedTime, "%I:%M %p").strftime("%H:%M") # Converts string to datetime then back to string and formats correctly
+            return formattedTime
+        except ValueError:
+            formattedTime = datetime.strptime(unformattedTime, "%H:%M")
+            return unformattedTime
```
BrianRamsay, line 60: I'd leave a comment in this case about how you are calling strptime here to explicitly raise an exception if it wasn't properly in 24 hour format, since it might not be clear why that line is there.

BrianRamsay (changes requested):

app/logic/utils.py
```diff
@@ -42,3 +43,20 @@ def getStartofCurrentAcademicYear(currentTerm):
         fallTerm = Term.select().where(Term.year==currentTerm.year-1, Term.description == f"Fall {currentTerm.year-1}").get()
         return fallTerm
     return currentTerm
+
+def format24HourTime(unformattedTime):
```
BrianRamsay, line 47: I really like how you changed the variable name to be clear what data it contains

app/logic/utils.py
```diff
@@ -42,3 +43,20 @@ def getStartofCurrentAcademicYear(currentTerm):
...
+    """
+    Turns a time string or datetime object into a string with a time in 24 hour format
+    unformattedTime: expects a string with format HH:mm AM/PM or HH:mm OR a datetime object
+    returns: a string in 24 hour format HH:mm
+    """
```
BrianRamsay, line 52: This function description is really nice.

BrianRamsay (changes requested):

app/logic/utils.py
Boorclark: Added the comment.

app/logic/events.py
```diff
@@ -264,6 +264,7 @@ def preprocessEventData(eventData):
         - facilitators should be a list of objects. Use the given list of usernames if possible
           (and check for a MultiDict with getlist), or else get it from the existing event
           (or use an empty list if no event)
+        - times should exist be strings in 12 hour format example: 2:40 PM
```
BrianRamsay: This comment isn't true, because you are making them 24 hour format, not 12 hour format

app/logic/utils.py
BrianRamsay: I'm only adding this nitpicky one because there is one other fix to make anyway, but this new comment should be on a new line after the `except ValueError:` line, because it's not an explanation of why you are catching the exception, it's an explanation for why you are calling `datetime.strptime` but not using the result.

BrianRamsay (changes requested):

BrianRamsay (approved):
woohoo!

Issue #258, opened 2022-04-29
Add "All Eligible" Option to Email Modal


bryantal:
When admins are setting who the email for an event is sent to, there should be option to choose eligible volunteers to email in the dropdown. (This is everyone who has completed the All-Volunteer Training, completed and passed all background checks, and are not banned)

![image](https://user-images.githubusercontent.com/28715731/165982287-36ae023a-3c9d-415f-acc7-8e84b4fdaa5d.png)

Boorclark:
Notes:
  - Look through all users in event participant DB and in the background check DB
  - Look at event Ids and look for 14 (AVT)
  - If users passed FBI, CAN, and SHS then they can be displayed to admin

Boorclark:
Never mind we goofed. This is the same as Issue #201 and Anderson is working on this same issue. This issue can be deleted.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #258: Add "All Eligible" Option to Email Modal.](https://trello.com/c/FoH17sME/401-celts-258-add-all-eligible-option-to-email-modal)

PR #302, opened 2022-06-16
Eligible students

AndersonStettner:
fixes issue #258 

Created a new recipient option, Eligible Students, for a user to select when sending out emails about an event. 

For the new recipient option:

- Created a query to make sure that all the possible recipients are not banned the date the email is sent out by looking at the date a user has been unbanned in _programBan_. 

- Created a query to make sure that all the possible recipients have completed the All Volunteer Training.

- Created a new flag in the database _isAllVolunteerTraining_ that will make sure that all the people who have received the All Volunteer Training will receive the email and included it throughout the files that needed it to work correctly. 

- Created a query to combine the other two queries to: only send an email to possible recipients that have an unban date before the user is trying to send out the email and that they have completed the All Volunteer Training. 

- Created tests make sure the Eligible Students was working correctly: 
1. Makes sure someone who is NOT banned and IS trained will be sent an email
2.  Makes sure someone who is NOT banned and NOT trained will not be sent an email
3. Makes sure someone who is NOT banned and wasn't trained but BECOMES trained will be sent an email
4.  Makes sure someone who IS banned and IS trained will not be sent an email
6.  Makes sure someone who WAS banned becomes UNBANNED and they ARE trained will be sent an email
7.  Makes sure a program that does not have any banned users and nobody trained will not send out any emails


app/logic/emailHandler.py
```diff
@@ -97,6 +100,11 @@ def retrieve_recipients(self, recipients_category):
                 .join(EventRsvp)
                 .where(EventRsvp.event==self.event.id))
 
+        if recipients_category == "Eligible Students":
+            bannedUsers = ProgramBan.select(ProgramBan.user_id).where((ProgramBan.endDate > datetime.now()) | (ProgramBan.endDate is None), ProgramBan.program_id.in_([p.id for p in self.program_ids]))
+            allVolunteer = Event.select().where(Event.isAllVolunteerTraining == True)
```
BrianRamsay: We need to make sure we get the all volunteer event for this particular academic year

app/models/programBan.py
```diff
@@ -9,3 +9,6 @@ class ProgramBan(baseModel):
     endDate = DateField(null=True)
     banNote = ForeignKeyField(Note, null=False)
     unbanNote = ForeignKeyField(Note, null=True)
+
+    def __str__(self):
+        return f"{self.user.username}"
```
BrianRamsay: `self.user.username` is the same here as just `self.user_id`. I'm nervous that `self.user.username` is running a separate query when it doesn't need to. Not sure if peewee is smart enough or not.

Also, the string representation should include more info. Like the program and/or end date.

tests/code/test_emailHandler.py
```diff
@@ -119,6 +126,65 @@ def test_email_log():
...
+            newTrainedStudent.delete_instance()
+
+            # Unban khatts while they have All Volunteer Training: NOT banned IS trained
+            ProgramBan.update(endDate = parser.parser("2022-6-23")).where(ProgramBan.user == "khatts").execute()
+            newTrainedStudent = EventParticipant.create(user = "khatts", event = 14)
+            email.process_data()
+            assert email.recipients == [User.get_by_id("partont"), User.get_by_id("khatts")]
+            newTrainedStudent.delete_instance()
+
+            transaction.rollback()
```
BrianRamsay, line 176: This rollback should go at the end. It doesn't affect the current tests, but it would be easy for someone to come in and add tests after the one below, not realizing they were missing the rollback

BrianRamsay (changes requested):

tests/code/test_emailHandler.py
```diff
@@ -2,14 +2,22 @@
 from flask_mail import Message
 from urllib.parse import urlparse
 from flask import request, g
-from datetime import datetime
+from datetime import datetime, date, timedelta
```
BrianRamsay: `timedelta` isn't used

tests/code/test_emailHandler.py
```diff
@@ -119,6 +127,98 @@ def test_email_log():
             assert emailLog.sender == User.get_by_id("ramsayb2")
             transaction.rollback()
 
+@pytest.mark.integration
+def test_recipients_category():
```
BrianRamsay: I'd probably call this something like `test_recipients_eligible_students`, since it is only worried about that case.

BrianRamsay (changes requested):

tests/code/test_emailHandler.py
```diff
@@ -119,6 +127,98 @@ def test_email_log():
...
+
+            # Change the term that the All Volunteer Training takes place so that
+            # it is in the past
+            firstTerm = Term.select().order_by(Term.id).get()
+            allVolunteerEvent = Event.get_by_id(14)
+            allVolunteerEvent.term = firstTerm
+            allVolunteerEvent.save()
+
+            # Update the current term in the database
+            Term.update(isCurrentTerm = Term.id[6]).where(Term.isCurrentTerm == True).execute()
```
BrianRamsay: What the heck is `Term.id[6]`? That's probably just a character.

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #187, opened 2021-11-05
Add functionality to "last email sent to" in the email modal


Sandesh-L:
![image](https://user-images.githubusercontent.com/46611742/140556991-e9e71293-423f-48f7-bedf-5906ad4df5f4.png)
It should show the last email sent for that event. And the date. clicking on the email button on the event_list page will always show the same modal. That means that there needs to be an ajax request sent after the email button has been clicked.

bledsoef:
Missing log for sender of email in emailLog.py. Need to insert the sender whenever a new email is sent by checking who the user was who was logged in. Insert new spot in email template front end for information to go.

bledsoef:
We have updated emailLog and emailHandler to include a sender field. Now we need to test that the emails are being stored properly, and then write a function that has a sort query to get the last sent email for a particular event_id.

Boorclark:
This issue was merged into development in #305. Closing issue.

PR #305, opened 2022-06-17
Last email sent issue 187

nartehi:
Fixed issue #187 

Edited the model for the emailLog to address the issue to include the sender field.
Created a function with a query to retrieve the last email function.
Formatted the return string in the fetchEmailLogData
Created a test function


app/static/js/emailModal.js
```diff
@@ -40,12 +40,16 @@ async function fetchEmailLogData() {
     url: `/fetchEmailLogData/${eventId}`,
     type: 'GET',
     success: function(emailLog) {
+      console.log(emailLog)
```
zachneill: Remove test console.log statements

app/controllers/events/email.py
```diff
@@ -16,7 +17,14 @@ def retrieveEmailTemplate():
 
 @main_bp.route('/fetchEmailLogData/<eventId>', methods=['GET'])
 def fetchEmailLogData(eventId):
-    if (EmailLog.select().where(EmailLog.event==eventId)).exists():
-        emailLog = EmailLog.select().where(EmailLog.event==eventId).order_by(EmailLog.dateSent.desc()).get()
-        return {'recipients': emailLog.recipientsCategory, 'dateSent': emailLog.dateSent.strftime('%m/%d/%Y')}
-    return {'exists': False}
+    last_email = EmailHandler.retrieve_last_email(eventId)
+    if last_email != None:
```
BrianRamsay: This can be just `if last_email:`

app/controllers/events/email.py
```diff
@@ -16,7 +17,14 @@ def retrieveEmailTemplate():
...
-    return {'exists': False}
+    last_email = EmailHandler.retrieve_last_email(eventId)
+    if last_email != None:
+        return {'last_log': "The last email was sent to " + last_email.recipientsCategory + " on " + last_email.dateSent.strftime('%m/%d/%Y') + " by " + last_email.sender.email  + "." , 'last_log2': " Subject: " + last_email.subject}
+        # {'recipients':last_email.recipientsCategory,'dateSent': last_email.dateSent.strftime('%m/%d/%Y'),'subject':last_email.subject,'sender':last_email.sender.email}
+
+
+    # if (EmailLog.select().where(EmailLog.event==eventId)).exists():
+    #     emailLog = EmailLog.select().where(EmailLog.event==eventId).order_by(EmailLog.dateSent.desc()).get()
+    #     return {'recipients': emailLog.recipientsCategory, 'dateSent': emailLog.dateSent.strftime('%m/%d/%Y')}
```
BrianRamsay: Remove the old code when yours is working, rather than just comment it out.

app/logic/emailHandler.py
```diff
@@ -169,7 +171,7 @@ def send_email(self):
                         reply_to=self.reply_to,
                         sender = ("Sandesh", 'bramsayr@gmail.com')
                     ))
-            self.store_sent_email(subject, template_id)
+            self.store_sent_email(subject, template_id, g.current_user)
```
BrianRamsay: We don't want email handler to be dependent on the application session. Add an initialization argument for user and use that here. Then the calling code can pass in `g.current_user`.

app/logic/emailHandler.py
```diff
@@ -187,3 +189,11 @@ def update_email_template(self):
         except Exception as e:
             print("Error updating email template record: ", e)
             return False
+
+    def retrieve_last_email(event_id):
+        try:
+            get_query = EmailLog.select().where(EmailLog.event==event_id).order_by(EmailLog.dateSent.desc()).get()
+            return get_query
+        except Exception as e:
```
BrianRamsay: If we are using exceptions as a normal condition we have to only treat the expected exception as normal. Catching ALL exceptions means if something goes wrong with our application it will just print "No last email found" and continue on.

database/test_data.py
```diff
@@ -864,6 +867,56 @@
...
 
+emailLogs = [
+    {
+    'event': 5,
+    'subject': 'Location Change for {event_name}',
+    'templateUsed': 2,
+    'recipientsCategory': "RSVP'd",
+    'recipients': 'neillz',
+    'dateSent': datetime.strptime("2022 5 7","%Y %m %d"),
+    'sender': User.get_by_id("neillz")
```
BrianRamsay: In the test data don't retrieve the user object, just use the id. In this case, `neillz`

BrianRamsay (changes requested):

app/controllers/events/email.py
```diff
@@ -1,6 +1,9 @@
 from app.models.emailTemplate import EmailTemplate
+from flask import g
+from urllib.parse import urlparse
```
BrianRamsay: Do you use these two imports? I don't see that they are used in this file

app/logic/emailHandler.py
```diff
@@ -10,13 +10,14 @@
 from app.models.emailTemplate import EmailTemplate
 from app.models.emailLog import EmailLog
 from app.models.event import Event
-
+from peewee import DoesNotExist
```
BrianRamsay: Put this import up with the other library imports. I like to keep them sorted.

app/logic/emailHandler.py
```diff
@@ -187,3 +189,10 @@ def update_email_template(self):
         except Exception as e:
             print("Error updating email template record: ", e)
             return False
+
+    def retrieve_last_email(event_id):
+        try:
+            get_query = EmailLog.select().where(EmailLog.event==event_id).order_by(EmailLog.dateSent.desc()).get()
```
BrianRamsay: instead of calling this `get_query`, call it what it is. `last_email`, or `last_log`. Or, don't even use a variable and return the query result directly.

tests/code/test_emailHandler.py
```diff
@@ -114,5 +116,18 @@ def test_email_log():
...
+def test_get_last_email():
+    query = EmailHandler.retrieve_last_email(5)
+    assert query.sender.username == "neillz"
+    assert query.subject == "Time Change for {event_name}"
+    assert query.templateUsed.subject == "Test Email 2"
+    assert query.recipientsCategory == "RSVP'd"
+    assert query.recipients == "ramsayb2"
+
+    query = EmailHandler.retrieve_last_email(37)
+    assert query is None
```
BrianRamsay: What this variable is storing is not a query, so it's not a good variable name.

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #292, opened 2022-06-14
Activity Logs running too many queries


BrianRamsay:
The Activity Log page is running a separate query to get user information for every row in the log. This is inefficient, and will start causing problems when there are thousands of records in the log.

We should either cache the user info so we only run one query per user rather than one query per row, or edit the admin log query to get the user information we need along with the log information without needing any other queries at all. The second option is better, in my opinion.

PR #310, opened 2022-06-21
Activity log query issue 292

bledsoef:
Fixed the problem where the activity log was spitting out a query for every row in the table. Updated query so that it joined User and AdminLogs to not require an extra query upon iteration through the table.
Fixes issue #292 
Closes issue #292


BrianRamsay:
I made a change to this one because I finally figured out something about Peewee

BrianRamsay (approved):

Issue #245, opened 2022-04-22
Too Many Database Queries on Volunteer Profile View


bryantal:
When going to a volunteers profile, the database is queried around 80 times.  Let's investigate how we can reduce the amount of queries since we are only viewing the page.

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #245: Too Many Database Queries on Volunteer Profile View](https://trello.com/c/Akc2lOy9/390-celts-245-too-many-database-queries-on-volunteer-profile-view)

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #246: Background Check Table Below Footer #246](https://trello.com/c/H06Ung5P/391-celts-246-background-check-table-below-footer-246)

Lyisus:
Made pull request #314 in order to address this issue but could still be worked on further in the future.

Boorclark:
Issue Merged. Closing

PR #314, opened 2022-06-22
Reduced Queries in Volunteer Profile by 25% 

Boorclark:
Issue #245

Reduced the number of queries in the Volunteer Profile page. Two queries in trainedParticipants() in participants.py were merged together as they both accessed the same table. Added a property to term objects called academicYearStartingTerm that returns the first term in the academic year they belong to and stores it in a cache. If the term calling the property is the starting term, it'll return and save itself to the cache. If there is something in the cache already, it will return the term object in the cache without accessing the database. Deleted the function that did the same thing as the property and modified the test suite accordinly to account for the change.

Files Changed:
- app/logic/participants.py
- app/logic/utils.py
- app/models/term.py
- tests/code/test_utils.py


app/models/term.py
```diff
@@ -1,8 +1,19 @@
...
     description = CharField()
     year = IntegerField()
     academicYear = CharField()
     isSummer = BooleanField(default=False)
     isCurrentTerm = BooleanField(default=False)
+
+    _cache = None
+
+    @property
+    def academicYearStartingTerm(self):
```
BrianRamsay, line 12: Give a brief description of what this function does. It's short, but it might not be clear at first glance

BrianRamsay (changes requested):
Looking pretty good. We'll have to wait until the time selector changes are in to see if that gets rid of those changes in here.

app/models/term.py
Boorclark: The issue has been addressed. Resolving conversation.

BrianRamsay:
A lot of comments to resolve. Let's take a look

BrianRamsay:
This needs a merge from development and conflict resolution before review, @Boorclark , @Lyisus

BrianRamsay (changes requested):
This table was renamed in development

BrianRamsay (approved):

Issue #255, opened 2022-04-29
Add Continuity Between Recurring Events


bryantal:
There should be an option to import volunteers from the last time the event was created when they go into add volunteers to the track volunteers table.  This will mainly be for when the admins create recurring events, and then have to go into the events to add the volunteers.

bledsoef:
Create an ajax request in JavaScript using the volunteers.py addPastVolunteers controller method. Call the JavaScript function onclick in our button in HTML to get the data from the controller onto the webpage.

Boorclark:
Issue Merged. Closing

PR #315, opened 2022-06-23
Recurring event issue 255

solijonovam:
- got rid of isRecurring from the Events db table and added recurring_id which gives numerical value of a recurring event
-made isRecurring a property
-added a unit test to check if the function that returns a new maximum recurring id is working properly
-made a function that gets the information from the previous events
-added a unit test to check that it gives back the correct information
-added a button that appears if an event is recurring so that the admin is able to add past participants
-the recurring event page was not working properly, so we were able to debug and fix it. 

issue #255

### DIFF
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -56,8 +56,10 @@ def saveEventToDb(newEventData):
     isNewEvent = ('id' not in newEventData)
 
     eventsToCreate = []
+    recurringSeriesId = None
     if isNewEvent and newEventData['isRecurring']:
         eventsToCreate = calculateRecurringEventFrequency(newEventData)
+        recurringSeriesId = calculateNewRecurringId()
     else:
         eventsToCreate.append({'name': f"{newEventData['name']}",
                                 'date':newEventData['startDate'],
@@ -73,7 +75,7 @@ def saveEventToDb(newEventData):
                     "timeStart": newEventData['timeStart'],
                     "timeEnd": newEventData['timeEnd'],
                     "location": newEventData['location'],
-                    "isRecurring": newEventData['isRecurring'],
+                    "recurring_id": newEventData['recurring_id'],
                     "isTraining": newEventData['isTraining'],
                     "isRsvpRequired": newEventData['isRsvpRequired'],
                     "isService": newEventData['isService'],
@@ -214,6 +216,9 @@ def validateNewEventData(data):
     data['valid'] = True
 
     return (True, "All inputs are valid.")
+
+def calculateNewRecurringId():
+        return Event.select(fn.MAX(Event.recurring_id)).scalar() + 1
 
 def calculateRecurringEventFrequency(event):
     """
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -75,7 +75,7 @@ def saveEventToDb(newEventData):
                     "timeStart": newEventData['timeStart'],
                     "timeEnd": newEventData['timeEnd'],
                     "location": newEventData['location'],
-                    "recurring_id": recurringSeriesId,
+                    "recurringid": recurringSeriesId,
                     "isTraining": newEventData['isTraining'],
                     "isRsvpRequired": newEventData['isRsvpRequired'],
                     "isService": newEventData['isService'],
@@ -218,7 +218,7 @@ def validateNewEventData(data):
     return (True, "All inputs are valid.")
 
 def calculateNewRecurringId():
-    recurringId = Event.select(fn.MAX(Event.recurring_id)).scalar()
+    recurringId = Event.select(fn.MAX(Event.recurringid)).scalar()
     if recurringId:
         return recurringId + 1
     else:
@@ -226,7 +226,7 @@ def calculateNewRecurringId():
 
 def getPreviousRecurringEventData(recurringId, startDate):
     return list(User.select(User.username).join(EventParticipant).join(Event)
-    .where(Event.recurring_id==recurringId, Event.startDate<startDate))
+    .where(Event.recurringid==recurringId, Event.startDate<startDate))
 
 def calculateRecurringEventFrequency(event):
     """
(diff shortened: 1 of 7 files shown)


app/logic/events.py
```diff
@@ -1,10 +1,12 @@
+from reprlib import recursive_repr
```
BrianRamsay: this doesn't appear to be used

app/logic/events.py
```diff
@@ -215,6 +218,17 @@ def validateNewEventData(data):
 
     return (True, "All inputs are valid.")
 
+def calculateNewRecurringId():
+    recurringId = Event.select(fn.MAX(Event.recurring_id)).scalar()
+    if recurringId != None:
```
BrianRamsay: Avoid negatives in conditionals

app/logic/events.py
```diff
@@ -215,6 +218,17 @@ def validateNewEventData(data):
...
+        return recurringId + 1
+    else:
+        return 0
+
+def getPreviousRecurringEventData(recurring_id, start_date):
```
BrianRamsay: We should be consistent with case. `startDate` and `recurringId`

app/logic/events.py
```diff
@@ -215,6 +218,17 @@ def validateNewEventData(data):
...
+    return list(User.select(User.username).join(EventParticipant).join(Event).where(Event.recurring_id==recurring_id, Event.startDate<start_date))
+    #Event.select(EventParticipant.user.username).join(EventParticipant).where(Event.recurring_id==recurring_id, Event.startDate<start_date)
```
BrianRamsay: remove old commented code

app/logic/events.py
```diff
@@ -215,6 +218,17 @@ def validateNewEventData(data):
...
 
+def calculateNewRecurringId():
+    recurringId = Event.select(fn.MAX(Event.recurring_id)).scalar()
+    if recurringId != None:
+        return recurringId + 1
+    else:
+        return 0
+
+def getPreviousRecurringEventData(recurring_id, start_date):
+    return list(User.select(User.username).join(EventParticipant).join(Event).where(Event.recurring_id==recurring_id, Event.startDate<start_date))
```
BrianRamsay: bring the where clause onto a new line to make it easier to read

app/templates/events/trackVolunteers.html
```diff
@@ -37,6 +37,9 @@ <h1 class="text-center">{{event.name}}</h1>
 <div class="flex-container" style="padding: 5px;">
   <div class='form-group d-inline-flex'>
     <a class="btn btn-success" role="button" href="#" data-bs-toggle="modal" data-bs-target="#addVolunteerModal" style="margin:5px;">Add volunteer</a>
+    {% if isRecurring or isNotFirstRecurringEvent%}
```
BrianRamsay: These are properties of the event, and aren't passed in separately from the controller.

Go ahead and change `isPastEvent` as well

app/logic/events.py
```diff
@@ -215,6 +218,17 @@ def validateNewEventData(data):
 
     return (True, "All inputs are valid.")
 
+def calculateNewRecurringId():
+    recurringId = Event.select(fn.MAX(Event.recurring_id)).scalar()
+    if recurringId != None:
+        return recurringId + 1
+    else:
+        return 0
```
BrianRamsay: If it is the first recurring event, start at `1`

app/models/event.py
```diff
@@ -35,4 +35,16 @@ def isPast(self):
...
         startTimePassed = self.timeStart < currentTime.time() and self.startDate == currentTime.date()
-        return startDatePassed or startTimePassed 
+        return startDatePassed or startTimePassed
+
+    @property
+    def isRecurring(self):
+        return bool(self.recurring_id)
+
+    @property
+    def isNotFirstRecurringEvent(self):
```
BrianRamsay: Remove the negative and flip the return values.

app/models/event.py
```diff
@@ -35,4 +35,16 @@ def isPast(self):
...
+        firstRecurringEvent = Event.select().where(Event.recurring_id==self.recurring_id).order_by(Event.startDate).get()
+        if firstRecurringEvent.id == self.id:
+            return False
+        else:
+            return True
```
BrianRamsay: Once the function is 'positive', you can just return this boolean value: ` return firstRecurringEvent.id == self.id`

BrianRamsay (changes requested):

app/logic/events.py
```diff
@@ -215,6 +217,17 @@ def validateNewEventData(data):
...
+def calculateNewRecurringId():
+    recurringId = Event.select(fn.MAX(Event.recurringid)).scalar()
+    if recurringId:
+        return recurringId + 1
+    else:
+        return 1
+
+def getPreviousRecurringEventData(recurringId, startDate):
+    return list(User.select(User.username).join(EventParticipant).join(Event)
+    .where(Event.recurringid==recurringId, Event.startDate<startDate))
```
BrianRamsay: indent this to right below `.select` (or close)

app/models/event.py
```diff
@@ -10,12 +10,12 @@ class Event(baseModel):
     timeStart = TimeField()
     timeEnd = TimeField()
     location = CharField()
-    isRecurring = BooleanField(default=False)
     isTraining = BooleanField(default=False)
     isRsvpRequired = BooleanField(default=False)
     isService = BooleanField(default=False)
     startDate = DateField()
     endDate = DateField(null=True)
+    recurringid = IntegerField(null=True)
```
BrianRamsay: This should be camelCase like the other variables `recurringId`

app/templates/events/trackVolunteers.html
```diff
@@ -18,7 +18,7 @@
 
 <div class="text-center mt-5">
     <h1 class="text-center">{{event.name}}</h1>
-    {% if isPastEvent %}
+    {% if event.isPastEvent %}
```
BrianRamsay: The correct property name is `isPast`

BrianRamsay (changes requested):
In addition to the comments, the recurring event dates in the UI should be formatted m/d/Y when showing the recurring events that will be created.

BrianRamsay (approved):
Done!

Issue #257, opened 2022-04-29
Import All Faculty Into Instructor Table


bryantal:
We'll need all faculty to be in the instructor table in our database.  We'll need to import them from somewhere else, or query somewhere else (Tracy?  LSF?)

Boorclark:
PR merged closing issue

Boorclark:
PR merged closing issue

PR #319, opened 2022-06-24
Issue257 accessing Tracy

Boorclark:
Issue #257 

Created a new file called tracy_data.py. In this file, we access Tracy and pull faculty and student data to insert into the User table in the CELTS database. We run these functions once every 24 hours, however, this is subject to change as there may end up being a script that updates multiple parts of the database regularly. Additionally, we created a method to get the username from the email as usernames are not accessible in Tracy. This is also subject to change once we have access to AD. Lastly, the user model has been changed to allow phone numbers to be null. 

Files Changed:
- app/models/user.py
- database/tracy_data.py (New File)


database/tracy_data.py
```diff
@@ -0,0 +1,104 @@
...
+    This function loops through a users email until the index is an "@" and creates the
+    username.
+    This function will most likely not be needed once we access data from another DB.
+    """
+    username = ''
+    for i in emailStr:
+        if i != '@':
+            username += i
+        else:
+            return username
```
BrianRamsay: this is a good use case for a regular expression

database/tracy_data.py
```diff
@@ -0,0 +1,104 @@
...
+                    "phoneNumber": None,
+                    "firstName": row[2],
+                    "lastName": row[3],
+                    "isStudent": True,
+                    "isFaculty": False,
+                    "isCeltsAdmin":False,
+                    "isCeltsStudentStaff": False
+            }
+            User.insert(user).execute()
+        except Exception as e:
```
BrianRamsay: If you are catching duplicate entries and that is expected, you should have separate `except` blocks, one for that case and one for the general case

BrianRamsay (changes requested):
It's looking good. We need to reorganize the script so that it updates when the script is run. It doesn't need to worry about it's own scheduling.

Also, I would name it something like import_users.py. If you use `git mv` git will be able to track the history.

BrianRamsay (approved):

Issue #318, opened 2022-06-24
Fix text alignment in Volunteer Ban UI 


BrianRamsay:
The text in each row of the Volunteer Ban UI should be aligned vertically, and it is not. The Edit link doesn't match.

![alignment-issue](https://user-images.githubusercontent.com/96464/175563660-3826eb2d-6c2b-4751-b1d8-a67d1340e15b.png)

bledsoef:
Pull request merged.

PR #322, opened 2022-06-24
Changed edit button centering on Ban UI

Boorclark:
Issue #318 

Added relative positioning to button as the button border is actually larger than the words. We may eventually want to get rid of the "highlight" so it doesn't extend over the other events and it looks a bit odd.

File Changed:

image
image


BrianRamsay (changes requested):
It's not a good idea to use style attributes, and it is also good to avoid specifying things with 'px'. In this case I couldn't figure out a good way to do the last one, though.

To fix this problem I think we should remove the `<p>` around Banned to get rid of its bottom padding, and then we can fix the edit links with `margin-top: -8px` (in a class). Still hacky, and still uses 'px', but at least we don't have to mess with position.

BrianRamsay:
Something weird happened in the merge with development. You have undone a few changes in the volunteer profile. You'll need to reset back to the commit before the merge and try again, either with this branch or a new one (to preserve this one)

BrianRamsay (approved):

PR #327, opened 2022-06-27
Add slinstructor

mohamudm1:
Changes made:
* Removed the search instructor search option from the Manage Service-Learning Faculty page and cleaned up all the code that was doing that functionality across all the files.
*  We modified the code so that it is adaptable and responsive to the new functionality

### DIFF
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -26,8 +26,6 @@ from app.logic.transcript import *
 from app.logic.manageSLFaculty import getCourseDict
 from app.logic.courseManagement import pendingCourses, approvedCourses
 from app.logic.utils import selectSurroundingTerms
-from app.logic.userManagement import addSlInstructor
-
 
 @main_bp.route('/', methods=['GET'])
 def redirectToEventsList():
@@ -255,11 +253,10 @@ def serviceTranscript(username):
                             userData = user)
 
 @main_bp.route('/searchUser/<query>', methods = ['GET'])
-@main_bp.route('/searchInstructor/<query>', methods = ['GET'])
 def searchUser(query):
 
     data= request.args.get("searchvalue")
-   
+
     '''Accepts user input and queries the database returning results that matches user search'''
     try:
         query = query.strip()
@@ -271,25 +268,9 @@ def searchUser(query):
         print(e)
         return "Error in searching for user", 500
 
-
-@main_bp.route('/searchInstructor/<query>', methods = ['GET'])
-def searchInstructor(query):
-    '''Accepts user input and queries the database returning results that matches user search'''
-    try:
-        query = query.strip()
-        search = query.upper()
-        splitSearch = search.split()
-        searchResults = searchInstructors(query)
-        return searchResults
-    except Exception as e:
-        print(e)
-        return "Error in searching for user", 500
-        
-
 @main_bp.route('/contributors',methods = ['GET'])
 def contributors():
     return render_template("/contributors.html")
-
 
 @main_bp.route('/manageServiceLearning', methods = ['GET', 'POST'])
 @main_bp.route('/manageServiceLearning/<term>', methods = ['GET', 'POST'])
@@ -324,4 +305,3 @@ def AddSLInstructor():
     addSlInstructor(Data["username"])
 
     return ""
-
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -257,13 +257,15 @@ def serviceTranscript(username):
 @main_bp.route('/searchUser/<query>', methods = ['GET'])
 @main_bp.route('/searchInstructor/<query>', methods = ['GET'])
 def searchUser(query):
-    print(request.path)
+
+    data= request.args.get("searchvalue")
+   
     '''Accepts user input and queries the database returning results that matches user search'''
     try:
         query = query.strip()
         search = query.upper()
         splitSearch = search.split()
-        searchResults = searchUsers(query)
+        searchResults = searchUsers(query,data)
         return searchResults
     except Exception as e:
         print(e)
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -26,6 +26,7 @@ from app.logic.transcript import *
 from app.logic.manageSLFaculty import getCourseDict
 from app.logic.courseManagement import pendingCourses, approvedCourses
 from app.logic.utils import selectSurroundingTerms
+from app.logic.userManagement import addSlInstructor
 
 
 @main_bp.route('/', methods=['GET'])
@@ -254,7 +255,9 @@ def serviceTranscript(username):
                             userData = user)
 
 @main_bp.route('/searchUser/<query>', methods = ['GET'])
+@main_bp.route('/searchInstructor/<query>', methods = ['GET'])
 def searchUser(query):
+    print(request.path)
     '''Accepts user input and queries the database returning results that matches user search'''
     try:
         query = query.strip()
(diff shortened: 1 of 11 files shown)


app/logic/searchUsers.py
```diff
@@ -1,6 +1,6 @@
 from playhouse.shortcuts import model_to_dict
 from app.models.user import User
-def searchUsers(query):
+def searchUsers(query, searchvalue):
```
BrianRamsay: this should be category

app/static/js/courseManagement.js
```diff
@@ -6,3 +6,25 @@ function formSubmit(el) {
   $("#termSelector").attr('action', '/manageServiceLearning/' + el);
   $("#termSelector").submit()
 }
+function openForm() {
+  $("#myForm").modal("toggle");
+}
```
BrianRamsay: Doesn't look like this function is getting called

app/static/js/courseManagement.js
```diff
@@ -6,3 +6,25 @@ function formSubmit(el) {
...
+  console.log("We're here");
+  $.ajax({
+    method: "POST",
+    url:  "/addInstructor",
+    data: {"username": $("#uname").val(),
+           "bnumber":$("#bnum").val(),
+           "email":$("#email").val(),
+           "phoneNumber":$("#pnum").val(),
+           "firstName":$("#fname").val(),
+           "lastName":$("#lname").val()
```
BrianRamsay: We don't need to pass all of the user information around.Only the username is needed to add the instructor (on the backend.

app/controllers/main/routes.py
```diff
@@ -296,3 +297,11 @@ def getAllCourseIntructors(term=None):
                                 term = term)
     else:
         abort(403)
+
+
+@main_bp.route('/addSLInstructor', methods = ['POST'])
```
BrianRamsay: This route isn't necessary anymore

app/static/js/manageServiceLearningFaculty.js
```diff
@@ -1,3 +1,4 @@
+import searchUser from './searchUser.js'
```
BrianRamsay: No longer necessary to import this

app/static/js/courseManagement.js
```diff
@@ -6,3 +6,25 @@ function formSubmit(el) {
   $("#termSelector").attr('action', '/manageServiceLearning/' + el);
   $("#termSelector").submit()
 }
+function openForm() {
+  $("#myForm").modal("toggle");
+}
+
+function addInstructorS() {
```
BrianRamsay: Not used anymore

BrianRamsay (changes requested):

app/controllers/main/routes.py
```diff
@@ -255,12 +254,15 @@ def serviceTranscript(username):
 
 @main_bp.route('/searchUser/<query>', methods = ['GET'])
 def searchUser(query):
+
+    data= request.args.get("searchCategory")
```
BrianRamsay: Still not a fan of `data` as a variable name. Way too generic. `category` would be good.

app/controllers/main/routes.py
BrianRamsay: This is still here, and not resolved.

app/logic/searchUsers.py
```diff
@@ -11,15 +11,21 @@ def searchUsers(query):
     lastName = " ".join(splitSearch[1:]) +"%"
 
     if len(splitSearch) == 1: #search for first or last name
-        results = User.select().where(User.isStudent & (User.firstName ** firstName | User.lastName ** firstName))
+        if category =="searchCategory":
```
BrianRamsay: This is confusing to me. "searchCategory" should not be a value. You aren't searching categories, you are searching instructors.

app/logic/searchUsers.py
```diff
@@ -11,15 +11,21 @@ def searchUsers(query):
...
     else:
-        for searchTerm in splitSearch: #searching for specified first and last name
+        for searchTerm in splitSearch: #searching for both first and last names
             if len(searchTerm) > 1:
                 searchTerm += "%"
-                results = User.select().where(User.isStudent & User.firstName ** firstName & User.lastName ** lastName)
+                if category =="searchCategory":
+                    results = User.select().where( User.isFaculty & (User.firstName ** firstName | User.lastName ** firstName))
+                else:
+                    results = User.select().where(User.isStudent & (User.firstName ** firstName | User.lastName ** firstName))
```
BrianRamsay: You changed these queries so they are the same as the other queries. Searching for "Brian Rams" would not work anymore.

app/static/js/courseManagement.js
BrianRamsay: Not resolved. Still not needed.

app/static/js/manageServiceLearningFaculty.js
BrianRamsay: Still not needed

app/static/js/searchUser.js
```diff
@@ -1,4 +1,4 @@
-export default function searchUser(inputId, callback, clear=false, parentElementId=null, columnRequested=null){
+export default function searchUser(inputId, callback, clear=false, parentElementId=null, columnRequested=null, searchUser){
```
BrianRamsay: if I didn't know what these changes were about, I would have no idea why there was a `searchUser` parameter. This is the category, remember?

BrianRamsay (changes requested):

app/logic/searchUsers.py
mohamudm1: Yes, Brian Rams would not show up because he is neither staff/faculty nor student but he is a CELTS admin. Therefore, none of the situations applies to him.

BrianRamsay: I just mean any "Firstname La" search wouldn't work

app/controllers/main/routes.py
```diff
@@ -255,12 +254,15 @@ def serviceTranscript(username):
 
 @main_bp.route('/searchUser/<query>', methods = ['GET'])
 def searchUser(query):
+
+    data= request.args.get("category")
```
BrianRamsay: `data` is a bad variable name here. This is the category.

app/static/js/searchUser.js
```diff
@@ -27,7 +28,11 @@ export default function searchUser(inputId, callback, clear=false, parentElement
                 // value: "{column: response, column2: response2, ...}"
                 // must JSON.parse
                 label: (item[1]["firstName"]+" "+item[1]["lastName"]+" ("+item[0]+")"),
-                value: JSON.stringify(columnDict)
+                dictvalue:JSON.stringify(columnDict),
+                
+                
+                
+
```
BrianRamsay: Remove all the extra lines

app/static/js/searchUser.js
```diff
@@ -27,7 +28,11 @@ export default function searchUser(inputId, callback, clear=false, parentElement
                 // value: "{column: response, column2: response2, ...}"
                 // must JSON.parse
                 label: (item[1]["firstName"]+" "+item[1]["lastName"]+" ("+item[0]+")"),
-                value: JSON.stringify(columnDict)
+                dictvalue:JSON.stringify(columnDict),
```
BrianRamsay, line 31: Is there a reason this was changed?

BrianRamsay (changes requested):

app/static/js/searchUser.js
mohamudm1: Yes, that has solved the issue of the whole JSON dictionary showing in the search bar.

BrianRamsay (approved):

Issue #267, opened 2022-05-04
Add Program Settings with Option to Edit Email Address and Sender Name


bryantal:
We need a program settings section in our sidebar where admins can go in and change various settings related to CELTS programs.  Right now, the one we know they will need is the ability to edit the email address for the program and the sender name that the volunteers will get emails about the program from.

Boorclark:
Notes for future development/thought process: 

 - In program.py there is not a column for program email addresses and sender names. This may mean that we need to add the sections to the DB in order to edit them. This may be different if these are hard coded. I think this website may help as far as concepts for this in the future: https://rockset.com/blog/change-data-capture-what-it-is-and-how-to-use-it/
 - Unless I am reading this wrong this does not seem like a good first issue as it will require an entire interface unless we wanted to add this to the setting page. I think it would make more sense to have its own interface though because it relates directly to the celts information. 
 - It may also make more sense to have this be a subsection of the actual event page. For example on the Adoption 101 page, it would make more sense that this even could be edited within its even information maybe as a subset of "Events List".
 - Will this have to interact directly with an email system? I don't know exactly how these systems work and if they are automated.

mohamudm1:
Notes: 
I think this problem is not a good first issue and the reason I say that is because it will need to create a whole new interface.

bledsoef:
Pull request merged.

PR #331, opened 2022-06-28
Issue#267 create program setting

bledsoef:
Created an accordion UI to allow the users to access program settings.
Created drop down for users to choose which program to edit.
Made text boxes that allow users to change the sender name and reply-to-email associated with each program.
Made the text boxes prefill with what is already in the database.

Fixes issue #267 
Closes issue #267


app/controllers/admin/userManagement.py
```diff
@@ -64,12 +66,27 @@ def removeProgramManagers():
         flash('Error while removing a manager.','warning')
         abort(500,"Error while trying to remove a manager.")
 
+@admin_bp.route('/admin/addNewProgramInfo', methods=['POST'])
+def addProgramInfo():
```
BrianRamsay: This action is more of an update than an add. I would rename to something more like updateProgramInfo wherever relevant.

app/controllers/admin/userManagement.py
```diff
@@ -64,12 +66,27 @@ def removeProgramManagers():
...
+    """Grabs info from JavaScript and then outputs it to logic function"""
+    programInfo = request.form #grabs user inputs from js
```
BrianRamsay: The webserver really doesn't know anything about where the data comes from, and you really shouldn't assume you know anyway. I'd remove these references to javascript in the comments

app/controllers/admin/userManagement.py
```diff
@@ -64,12 +66,27 @@ def removeProgramManagers():
...
+    try:
+        return addNewProgramInfo(programInfo["senderName"],  #calls logic function to add data to database
+                                programInfo["replyToEmail"],
+                                programInfo["programId"])
+    except Exception as e:
+        print(e)
+        flash('Error while adding program info to database.','warning')
+        abort(500,'Error while adding program info to database.')
```
BrianRamsay: same comment as above, 'updating' instead of 'adding'

app/controllers/admin/userManagement.py
BrianRamsay: This is an admin function so the method should ensure that only admins can update

app/static/js/userManagement.js
```diff
@@ -106,4 +111,28 @@ function addNewTerm(){
...
+function addNewProgramInfo(){
+  var programInfo = {senderName: $("#senderName").val(),
+                    replyToEmail: $("#replyToEmail").val(),
+                    programId: $("#programSelect").val()};
+  $.ajax({   // sends ajax request to controller with programInfo containing user input
+    url: "/admin/addNewProgramInfo",
+    type: "POST",
+    data: programInfo,
+    success: function(s){
+      location.reload()
```
BrianRamsay: we shouldn't reload the page on success, because then the accordion changes back to the default, and it is a little jarring.

app/templates/admin/programSettings.html
```diff
@@ -0,0 +1,11 @@
...
+{% extends "base.html" %}
+
+{% block scripts %}
+  {{super()}}
+  <script type="module" src="/static/js/programSettings.js"></script>
+
+{% endblock %}
+{% block app_content %}
+<h1 class="text-center mt-5 mb-5">Program Settings</h1>
+{% endblock %}
```
BrianRamsay: This template isn't used

app/templates/admin/userManagement.html
```diff
@@ -19,48 +17,109 @@ <h1 class="text-center mt-5 mb-5">Admin Management</h1>
...
+      <button class="accordion-button collapsed" type="button" data-bs-toggle="collapse" data-bs-target="#collapseTwo" aria-expanded="false" aria-controls="collapseTwo">
+          Term Management Settings
+        </button>
+        </h3>
+        <div id="collapseTwo" class="accordion-collapse collapse" aria-labelledby="headingTwo" data-bs-parent="#adminManagement">
+          <div class="accordion-body">
+              <div class="row">
+                <div class="container-fluid mt-5 col-md-4">
+                  <div class="text-center">
+                    <div class="card">
```
BrianRamsay, line 63: Can some of these classes be combined into one div?

app/templates/admin/userManagement.html
```diff
@@ -19,48 +17,109 @@ <h1 class="text-center mt-5 mb-5">Admin Management</h1>
...
+                      <div class="card-header"><strong>Select the Current Term</strong></div>
+                        <ul class="list-group" name="currentTermList" id="currentTermList">
+                        {% for term in terms %}
+                        <li class="term-btn list-group-item list-group-item-action {{ 'active' if g.current_term == term }}"
+                          id="termFormID_{{ loop.index }}"
+                          aria-label="{{'Current term' if g.current_term == term}}"
+                          value='{{ term.id }}'
+                          tabindex="0">
```
BrianRamsay: The indent on these attribute should be a little deeper, so they stand apart from the contents of the `<li>`

app/templates/admin/userManagement.html
```diff
@@ -19,48 +17,109 @@ <h1 class="text-center mt-5 mb-5">Admin Management</h1>
...
+        <div class="accordion-body">
+          <table>
+            <tr>
+              <th>Program Select</th>
+              <th>Reply-To-Email</th>
+              <th>Sender Name</th>
+            </tr>
+            <tr>
+            <form action="" id="programSettings">
+                      <div>
```
BrianRamsay: The div and form tags shouldn't be in the table row (but outside of the table cell). Definitely for the div, but the form feels weird as well

app/controllers/admin/userManagement.py
```diff
@@ -64,12 +66,27 @@ def removeProgramManagers():
...
         abort(500,"Error while trying to remove a manager.")
 
+@admin_bp.route('/admin/addNewProgramInfo', methods=['POST'])
+def addProgramInfo():
+    """Grabs info from JavaScript and then outputs it to logic function"""
+    programInfo = request.form #grabs user inputs from js
+    try:
+        return addNewProgramInfo(programInfo["senderName"],  #calls logic function to add data to database
+                                programInfo["replyToEmail"],
+                                programInfo["programId"])
```
BrianRamsay: You should have a flash for success as well as failure, as there is no other UI indication that it succeeded.

BrianRamsay (changes requested):
There is an extra empty js file, and you should add a test for your new function

app/templates/admin/userManagement.html
bledsoef: It does not seem that way.

app/models/program.py
```diff
@@ -8,3 +8,5 @@ class Program(baseModel):
     partner = ForeignKeyField(Partner, null=True)
     isStudentLed = BooleanField(default=False)
     isBonnerScholars = BooleanField(default=False)
+    replyToEmail = CharField()
+    senderName = CharField()
```
BrianRamsay: I should have caught this before, but `senderName` should have a clearer name. `Program.senderName` doesn't have any indication that it's about emailing. And if we are going with a good name like ` emailSenderName`, we should change the other to `emailReplyTo`, so it matches.

Global find and replace will be your friend.

app/static/js/userManagement.js
```diff
@@ -106,4 +111,31 @@ function addNewTerm(){
...
+                    programId: $("#programSelect").val()};
+  $.ajax({   // sends ajax request to controller with programInfo containing user input
+    url: "/admin/updateProgramInfo",
+    type: "POST",
+    data: programInfo,
+    success: function(s){
+      $('body').prepend('<div id="flash" style="display:none"></div>');
+      $('#flash').html("Successfully updated information.");
+      $('#flash').slideDown('slow');
+      $('#flash').click(function () { $('#flash').toggle('highlight') });
```
BrianRamsay: Use the `msgFlash` method we already have, so that the behavior is consistent.

BrianRamsay (changes requested):
You've got an error in a test as well:

`tests/code/test_userManagement.py:58: AssertionError: assert 'New Test Name' == 'testName'`

tests/code/test_emailHandler.py
```diff
@@ -10,7 +10,7 @@
 from app.models.eventRsvp import EventRsvp
 from app.models.user import User
 from app.models import mainDB
-from app.logic.emailHandler import EmailHandler
+from app.logic.emailHandler import *
```
zachneill: The asterisk is a wild card with possible consequences later on. Not pressing, but consider importing what you need explicitly.

BrianRamsay: I agree. In general I'm not a fan of blanket imports

app/controllers/events/email.py
```diff
@@ -22,3 +23,9 @@ def fetchEmailLogData(eventId):
         return {'last_log': "The last email was sent to " + last_email.recipientsCategory + " on " + last_email.dateSent.strftime('%m/%d/%Y') + " by " + last_email.sender.email  + "." , 'last_log2': " Subject: " + last_email.subject}
     else:
         return {'exists': False}
+
+@main_bp.route("/getProgramSender/", methods=['POST'])
+def getProgramSender():
+    programInfo = request.form
+    programId = programInfo['programId']
+    return str((EmailHandler.getEmailSender(programId)))
```
BrianRamsay: `getEmailSender` doesn't exist in this branch right now. Also, casting to string is probably not what you want. If you just want the username but have a User object, use `.username`

app/logic/emailHandler.py
```diff
@@ -197,3 +197,11 @@ def retrieve_last_email(event_id):
             return last_email
         except DoesNotExist:
             return None
+
+    def getProgramEmailInfo(programId): #query to get sender info
+        programInfo = Program.get_by_id(programId)
+        return programInfo.emailSenderName
+
+    def setEmailReplyTo(programId):
```
BrianRamsay: The name doesn't match what the function does. This is getting, not setting.

app/static/js/emailModal.js
```diff
@@ -52,6 +56,23 @@ async function fetchEmailLogData() {
   })
 }
 
+function fetchProgramSender() { // gets the sender name based on what is in the database
+  var programInfo = {programId:$(".modal-body #programID").val()};
+  $.ajax({
+
+   url: "/getProgramSender/",
+   type: "POST",
```
BrianRamsay: This should be a GET, not a POST

app/static/js/emailModal.js
```diff
@@ -52,6 +56,23 @@ async function fetchEmailLogData() {
...
 
+function fetchProgramSender() { // gets the sender name based on what is in the database
+  var programInfo = {programId:$(".modal-body #programID").val()};
+  $.ajax({
+
+   url: "/getProgramSender/",
+   type: "POST",
+   data: programInfo,
+   success: function(s) {
+     var selectEmail = document.getElementById("senderEmail")
```
BrianRamsay: Use jquery methods, for consistency. `$("#senderEmail")`. In the line below as well.

tests/code/test_userManagement.py
```diff
@@ -47,3 +48,16 @@ def test_modifyCeltsStudentStaff():
             addCeltsStudentStaff("asdf")
         with pytest.raises(DoesNotExist):
             removeCeltsStudentStaff("1234")
+
+@pytest.mark.integration
+def test_changeProgramInfo():
+    programId = 3
+    emailSenderName = "New Test Name"
+    emailReplyTo = 'newtest@email'
+    currentProgramInfo = Program.select().where(Program.id==programId).get()
```
BrianRamsay: a lot shorter to use `Program.get_by_id` here instead of select, where, and get

BrianRamsay (changes requested):

app/static/js/emailModal.js
bledsoef: Our code breaks when it is switched to a GET and I was under the impression that they can both do the same thing.

BrianRamsay: The only difference is you need to get your data differently (no `request.form`

app/static/js/userManagement.js
```diff
@@ -106,4 +110,29 @@ function addNewTerm(){
...
+  var programInfo = {emailSenderName: $("#emailSenderName").val(),
+                    emailReplyTo: $("#emailReplyTo").val(),
+                    programId: $("#programSelect").val()};
+  $.ajax({   // sends ajax request to controller with programInfo containing user input
+    url: "/admin/updateProgramInfo",
+    type: "POST",
+    data: programInfo,
+    success: function(s){
+      $("#flash_container").prepend('<div class=\"w-50 position-absolute top-0 start-50 translate-middle-x alert alert-'+ "success" + '\" role="alert" id="flasher">' + "Successfully updated program info" + '</div>');
+      $("#flasher").delay(5000).fadeOut(); //kept getting errors when importing, no one could figure it out, so had to do it unconventionally.
```
BrianRamsay: You should use `msgFlash` for this, to avoid duplicating code

app/controllers/events/email.py
```diff
@@ -22,3 +23,10 @@ def fetchEmailLogData(eventId):
         return {'last_log': "The last email was sent to " + last_email.recipientsCategory + " on " + last_email.dateSent.strftime('%m/%d/%Y') + " by " + last_email.sender.email  + "." , 'last_log2': " Subject: " + last_email.subject}
     else:
         return {'exists': False}
+
+@main_bp.route("/getProgramSender/", methods=['POST'])
```
BrianRamsay: This also needs to be switched to GET if the ajax call is using GET. You should get your data with `request.args.get("programId")` instead of with `request.form`

BrianRamsay (changes requested):

app/logic/emailHandler.py
```diff
@@ -158,6 +156,11 @@ def build_email(self):
 
     def send_email(self):
         template_id, subject, body = self.build_email()
+        if len(self.program_ids) == 1:
+            self.reply_to = self.program_ids[0].emailReplyTo
+            sender = (self.program_ids[0].emailSenderName, self.program_ids[0].emailReplyTo)
+        else:
+            sender = ("Sandesh", 'bramsayr@gmail.com')
```
BrianRamsay: We need to make sure the defaults are set if the program doesn't have a sender name or reply to set.

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #340, opened 2022-06-28
Add flash message for "Invalid User Search"


Boorclark:
Currently on the volunteer search page, if you search for a user that does not exist it will redirect you to a page saying "user does not exist". We will want this to go away and instead be replaced with a flash message that says "User does not exist." 
![image](https://user-images.githubusercontent.com/97637517/176260075-53421ac6-6d10-45d8-8dff-41ce53fc53be.png)

nartehi:
This issue has been resolved.

PR #343, opened 2022-06-28
Added new flash message for search student page

Boorclark:
Fixes Issue #340

Added a new flash message that displays when an invalid user is searched. This replaces the 404 page that was originally and awkwardly there.  Additionally, I had it return a redirect for this page as this one is not "jarring" to refresh. 
image

File(s) Changed:

### DIFF
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -19,7 +19,6 @@ from app.models.eventRsvp import EventRsvp
 from app.models.note import Note
 from app.models.programManager import ProgramManager
 from app.controllers.main import main_bp
-from app.controllers.admin.routes import studentSearchPage
 from app.logic.users import addUserInterest, removeUserInterest, banUser, unbanUser, isEligibleForProgram
 from app.logic.participants import userRsvpForEvent, unattendedRequiredEvents, trainedParticipants
 from app.logic.events import *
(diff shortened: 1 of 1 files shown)


app/controllers/main/routes.py
```diff
@@ -66,7 +67,8 @@ def viewVolunteersProfile(username):
     try:
         volunteer = User.get(User.username == username)
     except Exception as e:
-        return "User does not exist", 404
+        flash(f"{username} does not exist! ", category='danger')
+        return redirect(url_for('admin.studentSearchPage'))
```
BrianRamsay: One thing we need to check for is if a user has permissions to the search page before redirecting them there. The search page does check, but typing in an invalid username in the url and then getting redirected to a page you don't even have permission to would be weird. In that case, a 404 on the profile page makes more sense. For now, we can just use `isAdmin` on the current user.

BrianRamsay (changes requested):
I like this change!

app/controllers/main/routes.py
```diff
@@ -66,7 +67,11 @@ def viewVolunteersProfile(username):
     try:
         volunteer = User.get(User.username == username)
     except Exception as e:
-        return "User does not exist", 404
+        if g.current_user.isAdmin or g.current_user.isCeltsStudentStaff:
```
BrianRamsay: `isAdmin` is all you need. that includes celts admins and celts student staff

BrianRamsay (changes requested):

app/controllers/main/routes.py
```diff
@@ -18,6 +18,7 @@
 from app.models.note import Note
 from app.models.programManager import ProgramManager
 from app.controllers.main import main_bp
+from app.controllers.admin.routes import studentSearchPage
```
BrianRamsay: is that import needed for the redirect?

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #326, opened 2022-06-27
Refactor program manager permissions


BrianRamsay:
We should move our program manager permission functions into methods of the User object.

So instead of calling `hasPrivilege(user, program)`, we would call `user.hasPrivilege(program)`. Same with `isProgramManagerForEvent`.

We should also rename `hasPrivilege`. We can make it more clear by calling it `isProgramManagerFor` or `canManageProgram`

Boorclark:
PR merged closing issue

PR #345, opened 2022-06-29
refactored ProgramManager permissions

qasema:
fixes issue #326 refactor PromgramManager permissions method and have hasPrivilege method as a method of the User object.


app/controllers/admin/routes.py
```diff
@@ -103,7 +103,10 @@ def createEvent(templateid, programid=None):
 
     # make sure our data is the same regardless of GET or POST
     preprocessEventData(eventData)
-    isProgramManager = hasPrivilege(g.current_user,programid)
+    
+    user = User.get_by_id(g.current_user)
```
BrianRamsay: you already have a user in `g.current_user`. you don't need a database query to get the same user.

app/models/user.py
```diff
@@ -15,3 +16,13 @@ class User(baseModel):
     @property
     def isAdmin(self):
         return (self.isCeltsAdmin or self.isCeltsStudentStaff)
+    
+    def isProgramManagerFor(self, program):
+        from app.models.programManager import ProgramManager  # Must defer import until now to avoid circular reference
+        if ProgramManager.select().where(ProgramManager.user == self.username, ProgramManager.program == program).exists():
```
BrianRamsay: You can just use `self` instead of `self.username` (to make it a bit shorter), and then instead if `if <boolean> return True else return False`, you can just return the boolean you get from the `.exists()` query directly. `return ProgramManager.select...`

tests/code/test_users.py
```diff
@@ -170,3 +170,16 @@ def test_unbanUser():
...
+
+@pytest.mark.integration
+def test_userpriv():
+    user = User.get_by_id("khatts")
+    prg = Program.get_by_id(1)
+    assert user.isProgramManagerFor(prg)
+
+    user = User.get_by_id("mupotsal")
+    prg = Program.get_by_id(12)
+    assert not user.isProgramManagerFor(prg)
```
BrianRamsay, line 182: adding a test! nice.

BrianRamsay (changes requested):

BrianRamsay:
Don't forget this PR!

app/controllers/admin/routes.py
qasema: I thought it is supposed to check different users not only the current user!

qasema: I fixed this

app/controllers/admin/routes.py
```diff
@@ -141,8 +142,7 @@ def editEvent(eventId):
     userHasRSVPed = EventRsvp.get_or_none(EventRsvp.user == g.current_user, EventRsvp.event == event)
     isPastEvent = (datetime.now() >= datetime.combine(event.startDate, event.timeStart))
     program = event.singleProgram
-
-    isProgramManager = hasPrivilege(g.current_user,program)
+    isProgramManager =   g.current_user.isProgramManagerFor(Program)
```
BrianRamsay: You're passing a class instead of a program object here.

app/models/user.py
BrianRamsay: not addressed yet.

BrianRamsay (changes requested):

BrianRamsay (approved):

BrianRamsay:
I've approved the PR but there are conflicts to resolve

BrianRamsay (dismissed):

BrianRamsay:
Ready to merge but there is a conflict

qasema:
I just resolved it

BrianRamsay (changes requested):
There's an import of the old function that breaks in `app/logic/volunteers.py`

BrianRamsay:
Lots of windows line endings. It looks like development hasn't been merged in, so do that to see if it takes care of the issues.

BrianRamsay (commented):
Merge conflict to resove

BrianRamsay (approved):

Issue #337, opened 2022-06-28
Email Attachments 


AndersonStettner:
In _emailHandler.py_:
Complete the _attach_attachments_ function. 
![image](https://user-images.githubusercontent.com/26262045/176255036-28df8f66-84be-4ece-bbfb-11d5294a57c9.png)

Boorclark:
Notes:
- Add "Add File" section to the Email Template HTML --> https://getbootstrap.com/docs/4.6/components/input-group/#custom-file-input
- We don't want to store it as a blob (maybeee) and instead want to save it to the database and access it via the "server directory."
- Look into how images are stored in a DB and what a model would look like
- Link for information on storing images in a file system --> https://softwareengineering.stackexchange.com/questions/164905/
how-best-to-store-images-on-a-file-system

PR #346, opened 2022-06-29
Added Attachment Functionality to Emails

Boorclark:
Fixes Issue #337 

Things Changed:
- Created functionality for attach_attachments() to create a directory to the attachment in the new path
- Added a field to the Email Log model to save the attachment path
- Created the "Choose File" UI on the email page
- Created a directory in static for email attachments --> static/files/attachments/{attachmentName}
image
Files Changed:
 - app/logic/emailHandler.py
 - app/models/emailLog.py
 - app/static/files/attachments/__init__.py
 - app/templates/emailModal.html


app/logic/emailHandler.py
```diff
@@ -11,6 +11,8 @@
 from app.models.emailTemplate import EmailTemplate
 from app.models.emailLog import EmailLog
 from app.models.event import Event
+from PIL import Image
```
BrianRamsay: unused import

app/logic/emailHandler.py
```diff
@@ -133,22 +134,36 @@ def retrieve_and_modify_email_template(self):
...
+        :returns: directory path for attachment
+        """
+        attachmentObject = request.files['attachmentObject']
+        attachmentName = attachmentObject.filename
+        attachment_path = 'app/static/files/attachments'
+        try:  # tries to create the attachment_path directory and passes if already created
+            os.mkdir(attachment_path)
+        except:
+            pass
+        attachmentDirectory = os.path.join(attachment_path, attachmentName) # joins attachmnet_path with attachment name to create directory
```
BrianRamsay: this isn't a directory, this is the full path of the location

app/logic/emailHandler.py
```diff
@@ -133,22 +134,36 @@ def retrieve_and_modify_email_template(self):
...
-            dateSent=date_sent,
-            sender=self.sender)
+            event = self.event.id,
+            subject = subject,
+            templateUsed = template_id,
+            recipientsCategory = self.recipients_category,
+            recipients = ", ".join(recipient.email for recipient in self.recipients),
+            dateSent = date_sent,
+            sender = self.sender,
+            attachmentDirectory = self.attach_attachments())
```
BrianRamsay: we don't need to go through the attachment and file-saving process just to tell the log what files have been saved

app/logic/emailHandler.py
```diff
@@ -158,21 +173,24 @@ def build_email(self):
...
+            #         conn.send(Message(
+            #             subject,
+            #             # [recipient.email],
+            #             [self.override_all_mail],
+            #             email_body,
+            #             reply_to=self.reply_to,
+            #             file_attachment=self.attach_attachments(),
+            #             sender = ("Sandesh", 'bramsayr@gmail.com')
+            #         ))
+            # self.store_sent_email(subject, template_id)
```
BrianRamsay: need to restore this function

app/logic/emailHandler.py
```diff
@@ -133,22 +134,36 @@ def retrieve_and_modify_email_template(self):
...
-        # Q: how would this work?
-        pass
+        """
+        This creates the directory/path for the object from the "Choose File" input in the emailModal.html file.
+        Additionally, it saves the attachment in the app/static/files/attachments/ directory.
+        :returns: directory path for attachment
+        """
+        attachmentObject = request.files['attachmentObject']
+        attachmentName = attachmentObject.filename
+        attachment_path = 'app/static/files/attachments'
```
BrianRamsay: This should be a config item

app/logic/emailHandler.py
BrianRamsay: We should also probably do some scrubbing of the attachment name to make sure we have valid and consistent filenames.

BrianRamsay (changes requested):

app/logic/emailHandler.py
Boorclark: Issue resolved.

app/logic/emailHandler.py
Boorclark: Updated the comment. Leaving the second part to be addressed later when we know what we want to do in regards to "saving images".

app/logic/emailHandler.py
Boorclark: This may be something that is addressed later when we want to figure out if and when we are saving files sent in the database. The attach attachments function currently is creating the path as well as saving the attachments to this path that is created. If we do want to save attachments this can be renamed to "getAttachementDirectory"

BrianRamsay (dismissed):
Nice! We'll save multiple file submission for later

BrianRamsay:
Merge conflict with the recent email handler changes from Paw and Finn

app/logic/emailHandler.py
```diff
@@ -1,7 +1,7 @@
 from datetime import datetime
 from peewee import DoesNotExist
 from flask_mail import Mail, Message
-from flask import g, session
+from flask import g, session, request
```
BrianRamsay: None of these should be used in EmailHandler

app/logic/emailHandler.py
```diff
@@ -27,6 +28,8 @@ def __init__(self, raw_form_data, url_domain, sender_object):
         self.program_ids = None
         self.recipients = None
         self.sl_course_id = None
+        self.attachment_path = app.config['email']['attachment_path']
+        self.attachmentFullPath = os.path.join(self.attachment_path, request.files['attachmentObject'].filename)
```
BrianRamsay: We need to find another way without directly referencing the `request` module

app/logic/emailHandler.py
```diff
@@ -173,6 +190,7 @@ def send_email(self):
                         # [recipient.email],
                         [self.override_all_mail],
                         email_body,
+                        file_attachment=self.attach_attachments(),
```
BrianRamsay: This is still the old function name

app/models/emailLog.py
```diff
@@ -11,3 +11,4 @@ class EmailLog(baseModel):
     recipients = CharField()
     dateSent = DateTimeField()
     sender = ForeignKeyField(User)
+    attachmentFullPath = CharField()
```
BrianRamsay: we probably don't want to store the full path here. We want the path relative to the project root

BrianRamsay (changes requested):

app/logic/emailHandler.py
Boorclark: resolved

app/logic/emailHandler.py
Boorclark: resolved

app/logic/emailHandler.py
Boorclark: resolved

app/models/emailLog.py
Boorclark: resolved --> changed to attachmentName

app/logic/emailHandler.py
```diff
@@ -130,38 +131,53 @@ def retrieve_and_modify_email_template(self):
...
-            dateSent=date_sent,
-            sender=self.sender)
+            event = self.event.id,
+            subject = subject,
+            templateUsed = template_id,
+            recipientsCategory = self.recipients_category,
+            recipients = ", ".join(recipient.email for recipient in self.recipients),
+            dateSent = date_sent,
+            sender = self.sender,
+            attachmentName = self.attachment_file.filename)
```
BrianRamsay: This will break if there is no attachment for an email.

app/logic/emailHandler.py
```diff
@@ -130,38 +131,53 @@ def retrieve_and_modify_email_template(self):
...
+        attachmentFullPath = os.path.join(self.attachment_path, self.attachment_file.filename)
+        try:  # tries to create the full path of the files location and passes if already created
+            os.mkdir(self.attachment_path)
+        except:
+            pass
+        return attachmentFullPath
+
+    def saveAttachment(self):
+        """ Saves the attachment in the app/static/files/attachments/ directory """
+        self.attachment_file.save(self.getAttachmentFullPath()) # saves attachment in directory
```
BrianRamsay: This will break if there is no attachment for the email

app/logic/emailHandler.py
BrianRamsay: it looks like it's still there for some reason

app/logic/emailHandler.py
```diff
@@ -130,38 +131,53 @@ def retrieve_and_modify_email_template(self):
...
-        # TODO for later
-        # retrieve attachments, attach it to the email
-        # Q: how would this work?
-        pass
+    def getAttachmentFullPath(self):
+        """
+        This creates the directory/path for the object from the "Choose File" input in the emailModal.html file.
+        :returns: directory path for attachment
+        """
+        attachmentFullPath = os.path.join(self.attachment_path, self.attachment_file.filename)
```
BrianRamsay: This will break if there is no attachment for the email

BrianRamsay (changes requested):

app/logic/emailHandler.py
BrianRamsay: This is the last issue.

app/logic/emailHandler.py
```diff
@@ -145,38 +146,62 @@ def retrieve_and_modify_email_template(self):
...
-            dateSent=date_sent,
-            sender=self.sender)
+            event = self.event.id,
+            subject = subject,
+            templateUsed = template_id,
+            recipientsCategory = self.recipients_category,
+            recipients = ", ".join(recipient.email for recipient in self.recipients),
+            dateSent = date_sent,
+            sender = self.sender,
+            attachmentName = self.attachment_file.filename)
```
BrianRamsay: This line will still break if there is no attachment.

BrianRamsay (changes requested):

app/logic/emailHandler.py
nartehi: Resolved issue now.

BrianRamsay (approved):

PR #352, opened 2022-07-01
 Removed the recurring event tap from All Volunteer Training

mohamudm1:
Removed the recurring event choice from the All-Volunteer Training


app/templates/admin/createSingleEvent.html
```diff
@@ -98,7 +98,7 @@ <h1 id="title">{{page_title}}</h1>
             <p>{{eventData.location}}</p>
           {%endif%}
         </div>
-        {% if isNewEvent %}
+        {% if isNewEvent and page_title != 'Create All Volunteer Training' %}
```
BrianRamsay: Don't use the page title, use the event template tag. Because the page title could change and that would break this code, which we wouldn't expect

BrianRamsay (changes requested):

app/templates/admin/createSingleEvent.html
BrianRamsay: I'm not sure if that template tag is here in the page, actually.

BrianRamsay (changes requested):

app/templates/admin/createSingleEvent.html
```diff
@@ -166,10 +166,10 @@ <h1 id="title">{{page_title}}</h1>
           </table>
         </div>
       </div>
-      <div class="col-md-6 form-group">
+      <div style="margin: -25px 0px 0px 0px" class="col-md-6 form-group">
```
BrianRamsay: This inline style messes up the appearance when creating an event that isn't the all-volunteer event.

BrianRamsay (changes requested):

app/templates/admin/createSingleEvent.html
BrianRamsay: Experiment with using the bootstrap class for the margin, like you did with the padding below. And only add the negative margin if it's the all volunteer page.

BrianRamsay (changes requested):

BrianRamsay (changes requested):
I have an idea that might be a better alternative. Let's move the yes/no options to below the facilitators. That way if they aren't there we don't have to mess with any positions. And I think it looks better with Description always up top across from Name anyway.

BrianRamsay (changes requested):
Two issues:
- The Save button should appear centered
- Right now there is no Save button for the all volunteer training

BrianRamsay:
Looks good, but there are windows line endings in the template file still. Make sure your editor is set to always use LF

BrianRamsay (changes requested):
There are a couple of merge conflicts to resolve

BrianRamsay (changes requested):
You are resetting indentation a lot. Does it look like they are all in a line in your code? I'm confused how this keeps happening.

BrianRamsay (approved):

Issue #356, opened 2022-07-01
Track Volunteers Button/Page is redirected to wrong page


Boorclark:
The events in the "Other Events" category on the Event List have an error regarding the "Track Volunteers" tab. When you hit this button you are redirected to a page that states, "TODO: What do we do for no programs or multiple programs?" instead of the normal track volunteers section. 
![image](https://user-images.githubusercontent.com/97637517/176962146-345c49d8-92ae-45fa-940d-f7d5062276fd.png)
![image](https://user-images.githubusercontent.com/97637517/176962239-eae14d0c-f311-434a-8523-fedba01b55c1.png)

PR #359, opened 2022-07-04
Got rid of if not program conditional

Boorclark:
Fixes Issue #356 

Changes Made:
 - Got rid of "if not program" conditional as non-program events still have volunteers that need to be tracked. 

image

File(s) Changed:
 - app/controllers/admin/volunteers.py

### DIFF
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -32,11 +32,7 @@ def trackVolunteersPage(eventID):
         abort(404)
 
     program = event.singleProgram
-
-    # TODO: What do we do for no programs or multiple programs?
-    if not program:
-        return "TODO: What do we do for no programs or multiple programs?"
-
+    
     trainedParticipantsList = trainedParticipants(program, g.current_term)
     eventParticipants = getEventParticipants(event)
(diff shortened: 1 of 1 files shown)


app/controllers/admin/volunteers.py
```diff
@@ -32,11 +32,7 @@ def trackVolunteersPage(eventID):
         abort(404)
 
     program = event.singleProgram
-
-    # TODO: What do we do for no programs or multiple programs?
-    if not program:
-        return "TODO: What do we do for no programs or multiple programs?"
-
+    
     trainedParticipantsList = trainedParticipants(program, g.current_term)
```
BrianRamsay, line 36: Does this function handle `program` being `None`? Also, there are two cases to consider: events with no programs and events with multiple programs (probably only the All Volunteer Training)

BrianRamsay (changes requested):

app/controllers/admin/volunteers.py
Boorclark: image
image
I believe so because "Track Volunteers" works for both. I added screenshots to show.

BrianRamsay (commented):
The best PR is one that removes code!

BrianRamsay (approved):

PR #361, opened 2022-07-05
fixed broken profile page due to type

qasema:
there was just a type in the name of the table. so it was not defined.


nartehi:
This has already been done in another Pull Request.

qasema:
sad :(

Issue #342, opened 2022-06-28
Edit Service Learning Course duplicates data


Lyisus:
In the edit Service Learning Course proposal interface, if somebody submits edits, waits until the page refreshes, then goes back to the previous page, clicks on the previous button, and submits again will cause for data to be duplicated in the database, primarily entries in the courseinstructor table. This issue needs some more investigating to prevent the data duplication.

Boorclark:
PR merged closing issue

PR #362, opened 2022-07-05
fixed the data duplicate of the course proposal edit

qasema:
fix data duplicate in the data when adding the same instructor multiple times. fixes issue #342

### DIFF
--- a/app/logic/courseManagement.py
+++ b/app/logic/courseManagement.py
@@ -96,4 +96,7 @@ def updateCourse(courseData, instructorsDict):
                     .where((CourseQuestion.questionNumber == i) & (CourseQuestion.course==courseData["courseID"])).execute())
     removeInstructors = CourseInstructor.delete().where(CourseInstructor.course == courseData["courseID"]).execute()
     for instructor in instructorsDict["instructors"]:
-        addInstructors = CourseInstructor.create(course=courseData["courseID"], user=instructor)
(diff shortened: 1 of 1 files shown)


app/logic/courseManagement.py
```diff
@@ -96,4 +96,7 @@ def updateCourse(courseData, instructorsDict):
                     .where((CourseQuestion.questionNumber == i) & (CourseQuestion.course==courseData["courseID"])).execute())
     removeInstructors = CourseInstructor.delete().where(CourseInstructor.course == courseData["courseID"]).execute()
     for instructor in instructorsDict["instructors"]:
-        addInstructors = CourseInstructor.create(course=courseData["courseID"], user=instructor)
+        if not CourseInstructor.select().where(CourseInstructor.course==courseData["courseID"], CourseInstructor.user==instructor).exists():
+            addInstructors = CourseInstructor.create(course=courseData["courseID"], user=instructor)
+        else:
+            flash("The Insructor has already been added before.", "danger")
```
BrianRamsay: If the user is already there we can just ignore it, instead of flash a warning. The goal is for the instructor to be tied to the course, and that is happening, so that's still a success.

BrianRamsay (changes requested):

app/logic/courseManagement.py
qasema: shall we just remove the flash message

BrianRamsay (approved):

Issue #367, opened 2022-07-05
Saving an edited event 


AndersonStettner:
If the user edits an existing event on the Event List page and they save it they are not redirected back to the Event List page but instead they stay on the edit event page. They should be sent back to the event list page.

PR #370, opened 2022-07-05
made the save event button redirect to events list

bledsoef:
The save event button now redirects back to the home page instead of just refreshing the page. 

Fixes issue #367


BrianRamsay (changes requested):
Group decision (or split decision with a tie-breaker), to not redirect. And also to change the Create Event functionality to go to the edit page instead of the events list.

BrianRamsay (approved):

Issue #382, opened 2022-07-06
Send to Registrar Button: needs to not be an option on past semesters 


AndersonStettner:
In the "Service Learning Course Proposal Management" section in the Admin user: 
![image](https://user-images.githubusercontent.com/26262045/177570685-7a243dd3-2f87-4bef-9be6-cb3b4d36d22d.png)

The user is able to approved courses from previous semesters to the registrar. They should only be able to send them to the registrar they are in the current term or a future term

PR #390, opened 2022-07-07
Send to Registrar button to only show for current or future terms

gahimbaref:
#382 

- If the selected term is a current term or future term, the "Send to Registrar" button shows up. 
- If it is a past term, the button does not show up.
- A property for term object called isFutureTerm has been created, which returns True, if the term is a future term, or False, if the term is not a future term.


BrianRamsay (changes requested):
I like it. But there are enough different cases in `isFutureTerm` that you should add a test for it to make sure the logic is doing what we expect in all the different cases.

BrianRamsay:
There is a conflict to resolve.

tests/code/test_utils.py
```diff
@@ -194,3 +194,51 @@ def test_getStartofCurrentAcademicYear():
...
+    with mainDB.atomic() as transaction:
+        dbCurrentTerm = Term.select().where(Term.isCurrentTerm == True).get()
+        dbCurrentTerm.isCurrentTerm = False
+        dbCurrentTerm.save()
+        testCurrentTerm = Term.create(description = "Summer 1900",
+                                    year = 1900,
+                                    academicYear = "1899-1900",
+                                    isSummer = True,
+                                    isCurrentTerm = True)
+        testCurrentTerm.save()
```
BrianRamsay: The `.save()` calls after you create the terms are unnecessary.

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #202, opened 2021-11-22
Send Reminder Emails for Recurring Events


bryantal:
We want the emailer to be able to automatically send reminder emails for events that are recurring once the date happens each week.  We would want this to possibly send to outside participants, too.

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #202: Send Reminder Emails for Recurring Events](https://trello.com/c/LRZkQJd0/369-celts-202-send-reminder-emails-for-recurring-events)

Boorclark:
Notes:
- After talking with Anderson we most likely will not be using an API like Twilio and instead be searching the DB (event participant I am assuming) for emails that have already been sent to the participants of events. I think this may be helpful in the future: https://stackoverflow.com/questions/4455282/call-a-javascript-function-at-a-specific-time-of-day
https://stackoverflow.com/questions/52022134/how-do-i-schedule-an-email-to-send-at-a-certain-time-using-cron-and-smtp-in-pyt

Boorclark:
Notes:
 - As far as actual code goes, I believe an emailer should be able to search the event_id in the event participant DB. From here, they can get all participants with that event_id and send them an automated email. I think the best way for outside participants is to allow the emailer to choose events that may relate to the one for which they are sending the email. Also just a note for the future, I believe the only person who would have access to this page would be the accounts that manage certain events and admins. 

- Adding a section for reoccurring into the following picture may be the best way to go about this. We may add a section in "To" to something along the lines of "All previous Participants"
![image](https://user-images.githubusercontent.com/97637517/173676225-7d67bd0f-6bff-473c-b727-8be4feb39ceb.png)

Boorclark:
Notes: 
 - Once Issue #187 has been resolved we may be able to look into how they stored the "latest" email and apply that to saving a recurring email. 
 -  Function to view event participants for reminder emails
 -  Potentially use smtplib module for automation with a date module

PR #391, opened 2022-07-07
Automated recurring emails 202

bledsoef:
Created a script that when executed will check to see if there are any events the next day and if there are, send an email to all of the recipients of each of the events.

Fixes issue #202


app/logic/events.py
```diff
@@ -318,3 +319,10 @@ def preprocessEventData(eventData):
         eventData['facilitators'] = list(User.select().join(EventFacilitator).where(EventFacilitator.event == event))
 
     return eventData
+
+
+def getTomorrowsEvents():
+    """Grabs each event that occurs tomorrow"""
+    tomorrowDate = date.today() + timedelta(days=1)
+    events = list(Event.select().where(Event.startDate==tomorrowDate))
```
zachneill, line 327: See if you can change list() to the [ ] notation. It actually calculates faster and is easier to read.

BrianRamsay: WRONG! Actually, I'm not sure, but when I try using `[]` around a select I end up with a list with a select object in it, not a list of the results.

app/scripts/sendEventReminderEmails.py
```diff
@@ -0,0 +1,35 @@
...
+        counter = 0
+        currentTerm = Term.select(Term.id).where(Term.isCurrentTerm==1)
+        template = EmailTemplate.get(purpose = "Reminder")
+        templateSubject = template.subject
+        templateBody = template.body
+        for event in events:
+            programId = event.singleProgram
+            emailData = {"eventID":event.id,  # creates the email data
+                            "programID":programId,
+                            "term":currentTerm,
```
BrianRamsay: currentTerm here doesn't have all the data in it which is risky if EmailHandler needed something other than id.

app/scripts/sendEventReminderEmails.py
```diff
@@ -0,0 +1,35 @@
...
+                            "recipientsCategory":"Interested",
+                            "subject":templateSubject,
+                            "body":templateBody}
+            sendEmail = EmailHandler(emailData, "172.31.3.239:8080", User.get_by_id("ramsayb2"))
+            sendEmail.send_email()
+            counter+=1
+        return counter
+    except (TypeError) as e:
+        print("No events were found.")
+        return 0
```
BrianRamsay, line 11: I'm skeptical that this is necessary

BrianRamsay: And if it is necessary, we should figure out why

BrianRamsay (changes requested):

app/logic/events.py
bledsoef: Nice try Zach.

zachneill: _Lame_

BrianRamsay (approved):
Tada!

Issue #374, opened 2022-07-05
 BGC Completion Date Input


Boorclark:
Only the admin should have access to change the date completed for the Background Check. It should display the date to everyone else but without the input box. 
![image](https://user-images.githubusercontent.com/97637517/177414671-f0ed2878-db39-4f25-a0d5-36e32a24a521.png)

Boorclark:
PR merged closing issue

PR #392, opened 2022-07-07
Updated Background Check Date Display #374

Boorclark:
Fixes Issue #374  

Changes Made:
 - Updated completedBackgroundCheck to include date completed
 - Redid UI for Background Check so only Admin can edit, CELTS Student Staff can view, but Students/Faculty cannot see or edit
 - Changed formating for a date to be M-D-Y on display
 - If there is no completion date selected, it will display " - " as we determined this looks better than blank or "n/a"

image
image

Files Changed:


app/controllers/main/routes.py
```diff
@@ -86,7 +86,7 @@ def viewVolunteersProfile(username):
         permissionPrograms = [entry.program.id for entry in programManagerPrograms]
 
         allUserEntries = BackgroundCheck.select().where(BackgroundCheck.user == volunteer)
-        completedBackgroundCheck = {entry.type.id: entry.passBackgroundCheck for entry in allUserEntries}
+        completedBackgroundCheck = {entry.type.id: [entry.passBackgroundCheck, entry.dateCompleted] for entry in allUserEntries}
```
BrianRamsay: There's not much reason to create this dictionary value as a list with two elements rather than just the entry object itself. Then the jinja file has all of the event information without having to use subscripts.

app/templates/main/volunteerProfile.html
```diff
@@ -158,40 +158,53 @@ <h2>{{volunteer.firstName}} {{volunteer.lastName}} is a manager for the followin
...
+      <table class="table">
+         <thead>
+           <tr>
+             <th scope="col">Description</th>
+             <th scope="col">Date Completed</th>
+             <th scope="col">Completed</th>
+           </tr>
+         </thead>
+         <tbody>
+           {% for entry in backgroundTypes %}
```
BrianRamsay: let's call this `type` or `bgType` instead of `entry`

BrianRamsay (changes requested):
Let's avoid all of the expressions like `if entry.id in completedBackgroundCheck` by just having `completedBackgroundCheck` contain all of the types. That way you can just use the dictionary without having to check for the presence or absence of a key.

Boorclark:
Current UI for Student Staff:
image

app/templates/main/volunteerProfile.html
```diff
@@ -158,40 +158,39 @@ <h2>{{volunteer.firstName}} {{volunteer.lastName}} is a manager for the followin
...
-               Yes
-           {% endif%}
-        {% endif %}
-         </td>
-       </tr>
-       {%endfor%}
-     </tbody>
-  </table>
-</div>
+{% if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff %}
```
BrianRamsay: you can just use `isAdmin`

BrianRamsay (changes requested):

BrianRamsay (changes requested):
Two changes to make:

- There should be an indication in the UI that the data was saved
- You shouldn't be able to save without a date picked

app/static/js/volunteerProfile.js
```diff
@@ -86,17 +86,31 @@ $(document).ready(function(){
...
+        return
+    }else if(bgDate == '' && checkPassed != 0){
+        displayMsg.html("Date<br>Empty!").addClass("text-danger")
+        setTimeout(function() {displayMsg.html("").removeClass("text-danger")}, 2000)
+        return
+    }
+    else{
+    displayMsg.html("Saved!").addClass("text-success")
+    setTimeout(function() {displayMsg.html("").removeClass("text-success")}, 2000)
+    }
```
BrianRamsay: Let's consolidate this "display a message and then clear it after 2 seconds" code into a function. That will make the code a little clearer, and if we need to modify it in the future it will be easier.

BrianRamsay (changes requested):

app/static/js/volunteerProfile.js
Boorclark: resolved

app/static/js/volunteerProfile.js
```diff
@@ -86,17 +86,27 @@ $(document).ready(function(){
...
+
+    if (checkPassed == 0 && bgDate != '' ) {
+        displayMessage("Y/N<br>Empty!", "danger")
+        return
+    } else if(bgDate == '' && checkPassed != 0) {
+        displayMessage("Date<br>Empty!", "danger")
+        return
+    }
+    else {
+        displayMessage("Saved!", "success")
```
BrianRamsay: This one is premature. We don't want to send a success message before we've even tried the ajax request. We have `success` and `error` functions on the ajax request object for this.

BrianRamsay (changes requested):

app/static/js/volunteerProfile.js
Boorclark: resolved

BrianRamsay (approved):

Issue #232, opened 2022-03-16
Service Learning Courses Admin Review Interface


bryantal:
Admins should be able to go into a separate interface for reviewing course proposals that are in the pending table in the SL course management interface.  They'll go to this interface through the Service Learning Course Proposal Management interface.  (There currently isn't a link to this page in the sidebar so the route is: <ipAddress>:8080/courseManagement/<term>)

![image](https://user-images.githubusercontent.com/28715731/158688950-9ce9be6a-475b-4585-b523-f77abdb7bcac.png)

Here is the wireframe for this interface (the slides titled 4. Admin Review): https://balsamiq-wireframes.appspot.com/?state=%7B%22ids%22:%5B%221etrefTLsvWQntesh5_TDdC66TRVzRpHf%22%5D,%22action%22:%22open%22,%22userId%22:%22109665407847548855125%22,%22resourceKeys%22:%7B%7D%7D

bryantal:
![](https://github.trello.services/images/mini-trello-icon.png) [CELTS #232: Service Learning Courses Admin Review Interface](https://trello.com/c/M9u2Wtp8/382-celts-232-service-learning-courses-admin-review-interface)

Boorclark:
Issue Merged. Closing

PR #419, opened 2022-07-08
Review button modal

qasema:
fixes issue #232 enable the review button to show the modal of the course proposal information and allow approve them
location: http://<IP>/manageServiceLearning


BrianRamsay (changes requested):
There are multiple files with windows line endings

app/templates/serviceLearning/reviewproposal.html
```diff
@@ -0,0 +1,32 @@
+  <div class="modal fade" tabindex="-1" id="myReview">
```
BrianRamsay: Looks like this file is old. Should it be removed?

app/controllers/serviceLearning/routes.py
```diff
@@ -129,5 +129,10 @@ def approveCourse():
         course.execute() # Executes the query and approves course in the database
         flash("Course approved!", "success")
     except:
-        flash("Course not approved", "warning")
+        courseID=request.form #has only the course id
```
BrianRamsay: don't name this variable courseID. It is form data.

app/controllers/serviceLearning/routes.py
```diff
@@ -129,5 +129,10 @@ def approveCourse():
...
+        if courseID: #if only approving a pending course
+            course = Course.update(status = 2).where(Course.id == courseID["course_id"])
+            course.execute() # Executes the query and approves course in the database
+        else:
+            flash("Course not approved", "warning")
```
BrianRamsay: This code is basically the same as the above, just without the call to `updateCourse`. Figure out how to repeat less code.

app/static/js/courseManagement.js
```diff
@@ -1,8 +1,34 @@
...
+    $('form').submit();
+};
+
+function formSubmit(el) {
+  $("#termSelector").attr('action', '/manageServiceLearning/' + el);
+  $("#termSelector").submit()
+};
+
+function reviewCourses(el) {
+  var courseID=$(el).attr('data-id');
```
BrianRamsay: if you are using data attributes, you can use the jquery method `.data('id')`

app/static/js/courseManagement.js
```diff
@@ -1,8 +1,34 @@
...
+    data:{"course_id":courseID},
+    success: function(modal_html) {
+      $("#review-modal").html(modal_html)
+      $("#myReview").modal('show')
+    }
+  })
+}
+
+function approve_proposal(el){
+  var courseID=$(el).attr('course-id');
```
BrianRamsay: I would use data attributes here. This works because you can set arbitrary attributes, but those should be in data attributes, really.

app/templates/main/manageServiceLearningFaculty.html
```diff
@@ -2,74 +2,72 @@
...
 
 {% block scripts %}
-  {{super()}}
-  <script type="module" src="/static/js/manageServiceLearningFaculty.js"></script>
-  <script src="/static/js/courseManagement.js"></script>
-  <script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/1.11.3/js/jquery.dataTables.js"></script>
+{{super()}}
+<script type="module" src="/static/js/manageServiceLearningFaculty.js"></script>
+<script src="/static/js/courseManagement.js"></script>
+<script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/1.11.3/js/jquery.dataTables.js"></script>
```
BrianRamsay: dataTables should be imported before the files that use it. Just in case.

app/templates/main/manageServiceLearningFaculty.html
```diff
@@ -2,74 +2,72 @@
...
+            <th scope="col">Courses</th>
+            <th scope="col">Faculty</th>
+            <th scope="col">Term</th>
+            <th scope="col">Status</th>
+            <th scope="col">Review</th>
+            <th scope="col">Edit</th>
+          </tr>
+        </thead>
+        <tbody>
+          {% for data in pendingCourses %}
```
BrianRamsay: `data` is too generic a name

app/templates/main/reviewproposal.html
```diff
@@ -0,0 +1,51 @@
+  <div class="modal fade" tabindex="-1" id="myReview">
+    <div id="myDIV1">Hello baby</div>
```
BrianRamsay: Not production text

BrianRamsay (changes requested):

app/static/js/courseManagement.js
qasema: But the course Id is stored in an attribute called it "data-id"

app/templates/main/reviewproposal.html
qasema: lol

app/controllers/main/routes.py
```diff
@@ -279,6 +280,18 @@ def searchUser(query):
 def contributors():
     return render_template("/contributors.html")
 
+@main_bp.route('/proposalReview/', methods = ['GET', 'POST'])
+def reviewProposal():
+    """
+    this function gets the pending course id and returns the its data to the review proposal modal 
+    """
+    courseID=request.form
+    course_data=Course.get_by_id(courseID["course_id"])
```
BrianRamsay: This variable should just be `course` (in my opinion), since it is a course object. And then later you should use this variable instead of referencing the form data again.

app/controllers/serviceLearning/routes.py
```diff
@@ -1,3 +1,4 @@
+import re
```
BrianRamsay: You aren't using regular expressions, so you don't need this module.

app/controllers/serviceLearning/routes.py
```diff
@@ -121,13 +122,17 @@ def approveCourse():
...
-        updateCourse(request.form.copy(), instructorsDict) # Updates database with the completed fields
-        # The next line creates the query to approve the course
-        course = Course.update(status = 2).where(Course.id == request.form['courseID'])
-        course.execute() # Executes the query and approves course in the database
-        flash("Course approved!", "success")
-    except:
-        flash("Course not approved", "warning")
+    """ 
+    if len(request.form)==1:
+        course_data = request.form['courseID'] # if only course is reviewed pass the course ID
```
BrianRamsay: This isn't `course_data`, it's a `course_id`

app/logic/courseManagement.py
```diff
@@ -70,12 +70,14 @@ def createCourse(courseData, instructorsDict):
...
     for instructor in instructorsDict["instructors"]:
         CourseInstructor.create(course=course, user=instructor.username)
+    return course
 
 def updateCourse(courseData, instructorsDict):
     """
         This function will take in courseData for the SLC proposal page and a dictionary
         of instuctors assigned to the course and update the information in the db.
     """
+    print("---------------------------------------")
```
BrianRamsay: debug output

app/controllers/serviceLearning/routes.py
```diff
@@ -121,13 +122,17 @@ def approveCourse():
...
+    else:
+        try:
+            course_data=createCourse(request.form.copy(), instructorsDict) # creat course first and get its ID to approve next 
+        except:
+            course_upate = updateCourse(request.form.copy(), instructorsDict) # if edit course, Updates database with the completed fields and get course ID
+            course_data=course_upate['courseID']
```
BrianRamsay: `course_upate` is misspelled. Also, doesn't `course_upate['courseID']` break? what does `updateCourse` return?

You are mixing the data that is in `course_data` in the different cases, which can be very confusing later when you are using it.

app/controllers/serviceLearning/routes.py
```diff
@@ -121,13 +122,17 @@ def approveCourse():
...
+
+    course = Course.update(status = 2).where(Course.id == course_data)
+    course.execute() # Executes the query and approves course in the database
```
BrianRamsay: Since you should already have a course object from the above code, you can just use the `course.status = 2`, `course.save()` syntax for updating. Both work, but this is shorter

app/static/js/courseManagement.js
BrianRamsay: That's what `.data('id')` looks for.

BrianRamsay: `.attr()` works, but usually is used for manipulating standard attributes. you use `.data()` when you are using data attributes

BrianRamsay (changes requested):

qasema:
@BrianRamsay, you can take a look now!

app/controllers/serviceLearning/routes.py
BrianRamsay: Doesn't look like you use the `count` module either

app/controllers/serviceLearning/routes.py
```diff
@@ -1,4 +1,7 @@
+from itertools import count
+import re
 from flask import request, render_template, g, abort, json, redirect, jsonify, flash, session
+from app.models import course
```
BrianRamsay: This is an unused import also. You might want to turn off whatever PyCharm setting is auto-adding imports

app/logic/courseManagement.py
```diff
@@ -70,32 +70,39 @@ def createCourse(courseData, instructorsDict):
...
-    ).where(Course.id == courseData['courseID']).execute()
-    for i in range(1, 7):
-        (CourseQuestion.update(questionContent=courseData[f"{i}"])
-                    .where((CourseQuestion.questionNumber == i) & (CourseQuestion.course==courseData["courseID"])).execute())
-    removeInstructors = CourseInstructor.delete().where(CourseInstructor.course == courseData["courseID"]).execute()
-    for instructor in instructorsDict["instructors"]:
-        if not CourseInstructor.select().where(CourseInstructor.course==courseData["courseID"], CourseInstructor.user==instructor).exists():
-            addInstructors = CourseInstructor.create(course=courseData["courseID"], user=instructor)
+    try:
+        coursebyid= Course.get_by_id(courseData['courseID'])
```
BrianRamsay: I would just call this one `course`, and then for confusion just remove the existing `course` variable from the `.update()`. It wasn't being used, anyway.

app/templates/main/manageServiceLearningFaculty.html
```diff
@@ -2,73 +2,71 @@
...
 
 {% block scripts %}
-  {{super()}}
-  <script type="module" src="/static/js/manageServiceLearningFaculty.js"></script>
-  <script src="/static/js/courseManagement.js"></script>
-  <script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/1.11.3/js/jquery.dataTables.js"></script>
+{{super()}}
+<script type="text/javascript" charset="utf8" src="https://cdn.datatables.net/1.11.3/js/jquery.dataTables.js"></script>
+<script type="module" src="/static/js/manageServiceLearningFaculty.js"></script>
+<script src="/static/js/courseManagement.js"></script>
```
BrianRamsay: Let's put these indents back in.

app/templates/main/manageServiceLearningFaculty.html
```diff
@@ -2,73 +2,71 @@
...
 {% endblock %}
 
 {% block styles %}
-    {{super()}}
-    <link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/1.11.3/css/jquery.dataTables.css">
+{{super()}}
+<link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/1.11.3/css/jquery.dataTables.css">
```
BrianRamsay: put these indents back as well

app/templates/serviceLearning/reviewproposal.html
BrianRamsay: still relevant

app/controllers/main/routes.py
BrianRamsay: also still relevant

BrianRamsay (changes requested):

app/controllers/main/routes.py
qasema: done

app/templates/serviceLearning/reviewproposal.html
qasema: Just deleted it

BrianRamsay (approved):
Woohoo!

Issue #389, opened 2022-07-07
Redesign the student view event info page


bledsoef:
The student view page for the event info page looks very barren and gross, we need to redesign it to make it look a little bit better.

bledsoef:
Pull request merged.

PR #421, opened 2022-07-11
Make new student view 389

thawpt:
made a new page and a new controller method for the student event view

fixes issue #389


app/templates/eventStudentView.html
```diff
@@ -0,0 +1,97 @@
+{% set title = "View Event" %}
```
BrianRamsay: Let's make this the name of the event instead of something general

app/templates/eventStudentView.html
```diff
@@ -0,0 +1,97 @@
...
+{% extends "base.html"%}
+
+{% block styles %}
+    {{super()}}
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-1BmE4kWBq78iYhFldvKuhfTAU6auU8tT94WrHftjDbrCEXSU1oBoqyl2QvZ6jIW3" crossorigin="anonymous">
```
BrianRamsay: We already have bootstrap included in base.html, right?

app/templates/eventStudentView.html
```diff
@@ -0,0 +1,97 @@
...
+    {% else %}
+        <p><font size="+2"><b>{{eventData.name}}</b> will take place on <b>{{eventData.startDate}}</b> from <b>{{eventData.timeStart}}</b> to <b>{{eventData.timeEnd}}</b> and will be located in <b>{{eventData.location}}</b>.</font></p>
+    {% endif %}
+            <div>
+                <div>
+                    <font size="+2"><h3><br>Description:</h3>{{eventData.description}}<br><br>
+                    <div>
+                        {% if eventData.isService %}
+                            This event earns service hours.
+                        {% elif not eventData.isService %}
```
BrianRamsay: You can just use an `else` here and below

app/templates/eventStudentView.html
```diff
@@ -0,0 +1,97 @@
...
+                                    {{programManager.user.firstName}} {{programManager.user.lastName}}<br>
+                                    {{programManager.user.email}}
+                                </div>
+                        {% endif %}
+                    </div>
+                </div>
+                <div>
+                    <div class="row">
+                      <div class="col-sm">
+                          {% if program != None %}
```
BrianRamsay: Just use `if program`. It's clearer than "if program not equal to None"

app/templates/events/event_list.html
```diff
@@ -65,7 +65,11 @@ <h1 class="text-center mt-5">Events List for {{selectedTerm.description}}</h1>
     <tbody>
         {% for event in events %}
           <tr>
-            <td> <a href="/eventsList/{{event.id}}/edit" class="link-primary">{{event.name}}</a></td>
+              {% if user.isCeltsAdmin or user.isCeltsStudentStaff %}
```
BrianRamsay: student staff can only edit if they are the program manager for the program the event is in

app/controllers/admin/routes.py
```diff
@@ -150,11 +152,53 @@ def editEvent(eventId):
     return render_template("admin/createSingleEvent.html",
                             eventData = eventData,
                             allFacilitators = getAllFacilitators(),
-                            futureTerms = futureTerms,
+                            futureTerms=futureTerms,
                             isPastEvent = isPastEvent,
                             userHasRSVPed = userHasRSVPed,
                             isProgramManager = isProgramManager)
 
+@admin_bp.route('/eventsList/<eventId>/view', methods=['GET'])
```
BrianRamsay: Let's try using the same controller method as the 'edit' page, since you are providing almost all of the same information.

BrianRamsay (changes requested):

app/templates/eventStudentView.html
Boorclark: "if program" doesn't work here so we are keeping this the same.

app/templates/eventStudentView.html
Boorclark: resolved

app/templates/eventStudentView.html
Boorclark: resolved

app/templates/eventStudentView.html
Boorclark: resolved

app/templates/events/event_list.html
Boorclark: resolved but we need to edit it once the new property is merged

app/controllers/admin/routes.py

app/controllers/admin/routes.py
```diff
@@ -141,19 +143,43 @@ def editEvent(eventId):
...
+        userParticipatedEvents = {}
+        for training in listOfProgramTrainings:
+            eventParticipants = getEventParticipants(training.id)
+            if training.startDate>date.today():
+                didParticipate = None
+            elif g.current_user.username in eventParticipants.keys():
+                didParticipate = True
+            else:
+                didParticipate = False
+            userParticipatedEvents[training.name] = didParticipate
```
BrianRamsay: This logic to get the list of trainings participated in should be its own function

BrianRamsay: We should combine the render_template calls, so we don't accidentally break one view by forgetting to add a keyword in two places.

BrianRamsay (changes requested):

app/controllers/admin/routes.py

Boorclark: working on test suite now

BrianRamsay (approved):

PR #427, opened 2022-07-11
Add past participants issue#255

solijonovam:
issue #255

Controller:
added a functionality to add recurring participants to EventRSVP  and Event Participants table so that once the "Add  Past Participant" is clicked the volunteers will be added to the Track Volunteers table to be displayed. 
Inside the Modal:
-added 3 buttons; close, select all, and confirm
-added checkboxes and displayed the participants first and last name
-displayed the participants as a vertical list


app/controllers/admin/volunteers.py
```diff
@@ -32,7 +34,11 @@ def trackVolunteersPage(eventID):
         abort(404)
 
     program = event.singleProgram
-    
+
+    # TODO: What do we do for no programs or multiple programs?
+    if not program:
+        return "TODO: What do we do for no programs or multiple programs?"
+
```
BrianRamsay: Wasn't this removed at some point? Either way, we are past the point where we can have these TODOs, so these cases should work

app/controllers/admin/volunteers.py
```diff
@@ -78,22 +89,56 @@ def updateVolunteerTable(eventID):
...
+            else:
+                addVolunteerToEventRsvp(user, eventId)
+                EventParticipant.create(user = user, event = eventId)
+                successfullyAddedRecurringVolunteer = True
     else:
-        flash("Error when adding volunteer", "danger")
-    return ""
+        username = volunteer.strip("()").split('(')[-1]
+        user = User.get(User.username==username)
```
BrianRamsay: This stuff should go up above the loop, so that the same code is used to add the volunteer in both the single and the multi cases.

app/controllers/admin/volunteers.py
```diff
@@ -78,22 +89,56 @@ def updateVolunteerTable(eventID):
...
+        addVolunteerToEventRsvp(user, eventId)
+        EventParticipant.create(user=user, event=eventId)
+        successfullyAddedVolunteer = True
+
+    if (successfullyAddedRecurringVolunteer or successfullyAddedVolunteer == True) :
```
BrianRamsay: remove the `== True`

app/logic/events.py
```diff
@@ -231,15 +231,17 @@ def validateNewEventData(data):
     return (True, "All inputs are valid.")
 
 def calculateNewrecurringId():
+                 #gets the highest recurring Id so that a new recurring Id can be assigned
```
BrianRamsay: adjust the indentation on this comment

app/static/js/trackVolunteers.js
```diff
@@ -12,66 +12,74 @@ $(document).ready(function() {
...
+  }
+
+  $("#selectVolunteerButton").prop('disabled', true);
 
+  $("#addVolunteerInput").on("input", function() {
+    searchUser("addVolunteerInput", callback, false, "addVolunteerModal");
+  });
+
+  $(".removeVolunteer").on("click", function() {
+    let username =  $(this)[0].id;
```
BrianRamsay: if you know you are only wanting to reference one element on a page (and there is only one of them), use an id instead of a class. Since you need to get the username still, you could use a data attribute to store the username instead of the button id

app/templates/events/trackVolunteers.html
```diff
@@ -145,4 +145,30 @@ <h5 class="modal-title" id="exampleModalLabel">Add a volunteer</h5>
...
+      <form action="/addVolunteersToEvent/{{event.id}}" method="POST">
+          <div class="modal-header">
+            <h5 class="modal-title">Add Past Volunteers</h5>
+            <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+          </div>
+              <div class="modal-body pl-5">
+                  {%for participant in recurringVolunteers%}
+                    <table>
+                      <input class="m-2" type="checkbox" name="volunteer[]" value= {{participant}} />
+                      <label class="form-check-label" for="PastVolunteers"> {{participant.firstName}} {{participant.lastName}}  </label>
```
BrianRamsay: Right now this label `for` attribute isn't connecting the label and input. You may need an id on the input since they all have the same name. This matters because right now you can't click the name to control the checkbox.

app/templates/events/trackVolunteers.html
```diff
@@ -145,4 +145,30 @@ <h5 class="modal-title" id="exampleModalLabel">Add a volunteer</h5>
...
+                    </table>
+                  {% endfor %}
+              </div>
+          <div class="modal-footer" >
+            <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Close </button>
+            <button type="button" id = "selectAllVolunteers" class="btn btn-secondary" >Select All</button>
+            <button id="AddPastVolunteerButton" type="submit" class="btn btn-primary" >Confirm Selection</button>
```
BrianRamsay: Let's make this button label "Add Volunteers", because that is what our button click is doing

BrianRamsay (changes requested):
This is really nice!

One UI note, if there are no past participants the modal should say that there are no past participants, and the Select All and confirm button should be disabled.

app/static/js/trackVolunteers.js
martinj2-dot: There are multiple remove buttons, they are generated with each volunteer on the page.

app/templates/events/trackVolunteers.html
martinj2-dot: When adding an ID I run into the same problem where when I click the label for any checkbox after the first one that's generated It will only toggle the first checkbox.

app/static/js/trackVolunteers.js
BrianRamsay: Ah yes. That makes sense. Do you need to use [0] then? `this` would be each individual element I would think

app/templates/events/trackVolunteers.html
BrianRamsay: There can only be one unique id per page, so you would need to generate an id that is different for each item in the loop. Usually some fixed text + something that changes, like the username

app/controllers/admin/volunteers.py
```diff
@@ -1,14 +1,16 @@
+import re
```
BrianRamsay: You don't need this module

app/controllers/admin/volunteers.py
```diff
@@ -67,33 +73,60 @@ def updateVolunteerTable(eventID):
...
+    usernameList = []
+    if volunteer == None:
+        usernameList = request.form.getlist("volunteer[]")
     else:
-        flash("Error when adding volunteer", "danger")
-    return ""
+        username = volunteer.strip("()").split('(')[-1]
+        user = User.get(User.username==username)
+        usernameList.append(user)
+    eventParticipants = getEventParticipants(eventId)
```
BrianRamsay: Let's add a newline above this to separate the sections of the code

app/controllers/admin/volunteers.py
```diff
@@ -67,33 +73,60 @@ def updateVolunteerTable(eventID):
...
+def addVolunteer(eventId, volunteer = None):
+    successfullyAddedVolunteer = False
+    usernameList = []
+    if volunteer == None:
+        usernameList = request.form.getlist("volunteer[]")
     else:
-        flash("Error when adding volunteer", "danger")
-    return ""
+        username = volunteer.strip("()").split('(')[-1]
+        user = User.get(User.username==username)
```
BrianRamsay: If you wanted, you could get rid of this line, since you are doing a `.get()` on each value in the loop

app/controllers/admin/volunteers.py
```diff
@@ -67,33 +73,60 @@ def updateVolunteerTable(eventID):
...
+
+        if len(eventParticipants) == 0:
+            addVolunteerToEventRsvp(user, eventId)
+            EventParticipant.create(user = user, event = eventId)
+            successfullyAddedVolunteer = True
+
+        elif (EventRsvp.select().where(EventRsvp.user==user, EventRsvp.event_id == eventId) and
+              EventParticipant.select().where(EventParticipant.user == user, EventParticipant.event_id == eventId)):
+            duplicateVolunteerMessage = "This volunteer has already been added to this event."
+            successfullyAddedVolunteer = False
```
BrianRamsay: I think it should still be successfully added. The user wants to add a volunteer, so if they are already in the db, we can consider that a success. It does mean something weird is going on, perhaps, but the end result is still a success.

app/controllers/admin/volunteers.py
```diff
@@ -67,33 +73,60 @@ def updateVolunteerTable(eventID):
...
+
+        else:
+            addVolunteerToEventRsvp(user, eventId)
+            EventParticipant.create(user = user, event = eventId)
+            successfullyAddedVolunteer = True
```
BrianRamsay, line 107: This case is exactly the same as the the first one in the `if` statement. Can we combine them?

app/static/js/trackVolunteers.js
```diff
@@ -12,66 +12,74 @@ $(document).ready(function() {
...
-  let username =  $(this)[0].id;
-  let eventId = $('#eventID').val()
-  $.ajax({
-    url: `/removeVolunteerFromEvent/${username}/${eventId}`,
-    type: "POST",
-    success: function(s) {
-      location.reload();
-    },
-    error: function(request, status, error) {
+    if ($(this)[0].checked) {
```
BrianRamsay: Let's go ahead and just use `this` instead of `$(this)[0]` for consistency

app/static/js/trackVolunteers.js
```diff
@@ -12,66 +12,74 @@ $(document).ready(function() {
...
       }
     });
   });
-});
 
-function callback(selected) {
-  $("#selectVolunteerButton").prop('disabled', false);
-}
+$(".attendanceCheck").on("change", function() {
+    let username =  $(this)[0].name.substring(9) //get everything after the 9th character;
```
BrianRamsay: Let's go ahead and just use `this` instead of `$(this)[0]` for consistency

app/controllers/admin/volunteers.py
```diff
@@ -67,33 +73,60 @@ def updateVolunteerTable(eventID):
...
+        usernameList.append(user)
+    eventParticipants = getEventParticipants(eventId)
+    for user in usernameList:
+        user = User.get(User.username==user)
+
+        if len(usernameList) == 0:
+            noVolunteersSelectedMessage = "No volunteers selected, please select a volunteer."
+            successfullyAddedVolunteer = False
+            noVolunteerSelected = True
+
```
BrianRamsay, line 101: It will never reach this `if`, because it won't get into the `for` loop at all.

BrianRamsay (changes requested):

app/controllers/admin/volunteers.py
martinj2-dot: So this is interesting. This is how we handled adding duplicate entries into the database. Without this check here when adding a volunteer who is already part of an event it will add a duplicate entry into the the EventRsvp and EventParticipant table. I can however just disable and uncheck the boxes in the front end if that would be better practice.

BrianRamsay (approved):
Woohoo!

Issue #411, opened 2022-07-08
Allow Student Staff to search volunteers. Also, when Student Staff are looking at a volunteer they should only be able to see the Upcoming Events tab and not the Service Transcript tab.


AndersonStettner:
Student Staff currently do not have permission to go to a volunteers profile through the Volunteer Search but they do have permission to view the profile itself if they type in the url due to a permissions check on the Volunteer Search bar. Change the check to allow the Student Staff to use the Volunteer Search bar. 

When a Student Staff is looking at another user the only thing they should be able to see is the Upcoming Events of that user and not the Service Transcript.

PR #432, opened 2022-07-11
hid service transcript from student staff

gahimbaref:
#411 
- Added a conditional to only show "Service Transcript" for celts admins

Files modified:

### DIFF
--- a/app/templates/main/volunteerProfile.html
+++ b/app/templates/main/volunteerProfile.html
@@ -23,7 +23,7 @@
      <li class="nav-item" role="presentation">
        <a class="nav-link active" aria-current="page" href="#">Upcoming Events</a>
      </li>
-     {% if g.current_user.isCeltsAdmin or g.current_user.username in volunteer.username %}
+     {% if g.current_user.isCeltsAdmin or g.current_user.username == volunteer.username %}
      <li class="nav-item">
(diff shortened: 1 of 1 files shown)


app/templates/main/volunteerProfile.html
```diff
@@ -23,9 +23,11 @@ <h6 align="center">{{volunteer.phoneNumber}}</h6>
      <li class="nav-item" role="presentation">
        <a class="nav-link active" aria-current="page" href="#">Upcoming Events</a>
      </li>
+     {% if g.current_user.isCeltsAdmin %}
```
BrianRamsay: Students should be able to see their own transcript

BrianRamsay (changes requested):

app/templates/main/volunteerProfile.html
```diff
@@ -23,9 +23,11 @@ <h6 align="center">{{volunteer.phoneNumber}}</h6>
      <li class="nav-item" role="presentation">
        <a class="nav-link active" aria-current="page" href="#">Upcoming Events</a>
      </li>
+     {% if g.current_user.isCeltsAdmin or g.current_user.username in volunteer.username %}
```
BrianRamsay: We should use `==` instead of `in`, to make sure the user is actually the same user

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #394, opened 2022-07-08
Add what event the training/education is for in the Training and Education tab of the Events List page


AndersonStettner:
The user should be able to see what event a specific training is for on the Events List page
![Annotation 2022-07-08 092516](https://user-images.githubusercontent.com/26262045/178000807-516a5a90-6ba5-450e-b040-ccd727c94edd.png)

PR #433, opened 2022-07-11
Add training purpose 394

mohamudm1:
Fixed issue #394 

Changes made:
We have added a new purpose field under the Training and Education sector which identifies what program an event belongs to.


qasema:
has conflicts

Issue #407, opened 2022-07-08
Make it so that long course names, descriptions, and/or other user inserted texts don't break the application.


AndersonStettner:
![image](https://user-images.githubusercontent.com/26262045/178018871-e9ddaa94-786b-4f66-b152-b3ca5f08d82f.png)

![image](https://user-images.githubusercontent.com/26262045/178019121-b4d49a32-1424-4055-b80f-85360fc015a7.png)
![image](https://user-images.githubusercontent.com/26262045/178019160-a9095a30-b97b-45c0-ac1d-bfe1e1f6fffc.png)

PR #437, opened 2022-07-11
Handle Large Texts Across the Application

Lyisus:
#407 
Large strings for names, inputs, descriptions, addresses, etc; break how the interface look by going off the page and requiring scrolling to see everything. This PR should make it so that every text that is overly large will wrap and go to the next line. However, really long texts in tables will give those fields more space in the table compared to the other fields, sometimes wrapping those smaller fields in awkward ways. Though, the awkward wrappings seem to be minimal with fields of about 60 characters.

Files affected:
- app/static/css/base.css


AndersonStettner (changes requested):
In all the tables where text is squished we will want to have a nowrap attribute so that they are not squished. EX: Service Transcript and Event List
image

BrianRamsay (changes requested):
See Anderson's comment

BrianRamsay (approved):

Issue #405, opened 2022-07-08
On Create event and Edit event pages: the Start Time and End Time should not take the current time into consideration at all. 


AndersonStettner:
The times should default to 12:00pm when the user goes to select a time instead of the users current time.
![Annotation 2022-07-08 100119](https://user-images.githubusercontent.com/26262045/178007107-38102cc2-693f-4234-9db1-4e6cc0ab9ab6.png)

PR #443, opened 2022-07-12
set time picker to default time 12 pm

qasema:
fixes issue #405 

Set the start and end times for a default value of 12 pm when creating an event

### DIFF
--- a/app/templates/admin/createSingleEvent.html
+++ b/app/templates/admin/createSingleEvent.html
@@ -157,7 +157,7 @@
                 {% if eventData.timeEnd %}
                   {%set end_time = eventData.timeEnd%}
                 {%else%}
-                  {%set end_time = "23:59" %}
+                  {%set end_time = "12:00" %}
                 {% endif %}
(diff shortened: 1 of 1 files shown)


BrianRamsay (changes requested):
Don't repeat the html. Just make a string with the time you want for each case and put it in the same html

app/templates/admin/createSingleEvent.html
```diff
@@ -149,8 +154,13 @@ <h1 id="title">{{page_title}}</h1>
               <label class="form-label me-2" for="pickEndTime"><strong>End Time</strong></label>
               {% if isEditable %}
               <div class="input-group">
-                <input autocomplete="off" type="time" class="form-control timepicker readonly" value="{{eventData.timeEnd}}" name="timeEnd" aria-describedby="timeIcon2" placeholder="Pick an end time" onchange="update(this)" id=endTime required />
-                <span class="input-group-text timeIcons" id="timeIcon2" hidden><i class="bi bi-clock"></i></span>
+                {% if eventData.timeEnd %}
+                  {%set end_time = eventData.timeEnd%}
+                {%else%}
+                  {%set end_time = "23:59" %}
```
BrianRamsay: I think this should also be an even hour, so our time picker doesn't default to 59 minutes

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #417, opened 2022-07-08
On Create Event page: change the Facilitators selector into a search bar. The user that is creating the event should still have the ability to select multiple facilitators.


AndersonStettner:
![image](https://user-images.githubusercontent.com/26262045/178047338-d0291cef-741d-4648-8cc5-fd61042f13f4.png)

PR #447, opened 2022-07-12
Changed the facilitators selector into a search bar

mohamudm1:
Solves issue #417. 
Changed the facilitators selector in the created events page into a search bar where facilitators can be searched.


BrianRamsay:
merge conflict to resolve

app/logic/events.py
```diff
@@ -325,16 +323,39 @@ def preprocessEventData(eventData):
...
-        eventData['facilitators'] = list(User.select().join(EventFacilitator).where(EventFacilitator.event == event))
+    if 'facilitators' in eventData:
+        eventData['facilitators'] = getFacilitatorsFromList(eventData['facilitators'])
+    # try:
+    #     if type(eventData) == MultiDict and type(eventData['facilitators']) is not list:
+    #         eventData['facilitators'] = eventData.getlist('facilitators')
+    #     eventData['facilitators'] = [User.get_by_id(f) for f in eventData['facilitators']]
+    # except Exception as e:
+    #     event = eventData.get('id', -1)
+    #     eventData['facilitators'] = list(User.select().join(EventFacilitator).where(EventFacilitator.event == event))
```
BrianRamsay: Don't leave a bunch of commented out code.

app/logic/events.py
```diff
@@ -325,16 +323,39 @@ def preprocessEventData(eventData):
...
+    """
+    This function takes in a list of usernames or a string with the usernames of
+        facilitators separated by commas (,) and returns a list of facilitator
+        objects that match the usernames in the list
+    facilitatorList: expected to be a list with facilitator usernames as strings
+    return: list of facilitator objects matching the usernames in facilitatorList
+    """
+    if type(facilitatorList) == str:
+        facilitatorList = facilitatorList.split(',')
+    print(facilitatorList)
```
BrianRamsay: remove print

app/logic/events.py
```diff
@@ -325,16 +323,39 @@ def preprocessEventData(eventData):
...
+    #     eventData['facilitators'] = list(User.select().join(EventFacilitator).where(EventFacilitator.event == event))
 
     return eventData
 
+def getFacilitatorsFromList(facilitatorList):
+    """
+    This function takes in a list of usernames or a string with the usernames of
+        facilitators separated by commas (,) and returns a list of facilitator
+        objects that match the usernames in the list
+    facilitatorList: expected to be a list with facilitator usernames as strings
```
BrianRamsay, line 336: Your function also handles a comma separated list of usernames

BrianRamsay: Make sure your comment is consistent

app/logic/events.py
```diff
@@ -325,16 +323,39 @@ def preprocessEventData(eventData):
...
+    return: list of facilitator objects matching the usernames in facilitatorList
+    """
+    if type(facilitatorList) == str:
+        facilitatorList = facilitatorList.split(',')
+    print(facilitatorList)
+    finalFacilitatorList = []
+    for i in facilitatorList:
```
BrianRamsay: Use a descriptive name instead of `i`. `i` is only for simple incrementing numbers

app/logic/events.py
```diff
@@ -325,16 +323,39 @@ def preprocessEventData(eventData):
...
+        if i:
+            facilitatorToAdd = User.select().where(User.username == i).get()
```
BrianRamsay: This can just be `User.get_by_id(username)`

app/logic/events.py
BrianRamsay: This is a good case for a list comprehension instead of a for loop

app/static/js/createEvents.js
```diff
@@ -127,4 +129,39 @@ $(document).ready(function() {
...
+        newRow.attr("id", username);
+        newRow.prop("hidden", false);
+        lastRow.after(newRow);
+        $("#hiddenFacilitatorArray").attr("value", facilitatorArray);
+    }
+  }
+
+  $("#eventFacilitator").on('input', function() {
+    // To retrieve specific columns into a dict, create a [] list and put columns inside
+    searchUser("eventFacilitator", callback, true, null, ["firstName", "lastName", "username"]);
```
BrianRamsay: That last parameter isn't necessary anymore. Also, don't you need a category here?

app/static/js/searchStudent.js
```diff
@@ -4,5 +4,5 @@ function callback(selected) {
 }
 
 $("#searchStudentsInput").on("input", function() {
-  searchUser("searchStudentsInput", callback);
+  searchUser("searchStudentsInput", callback, undefined, undefined, undefined,"students");
```
BrianRamsay: it searches students by default

app/templates/admin/createEvent.html
```diff
@@ -13,6 +13,8 @@
 {% endblock %}
 
 {% block app_content %}
+{% set isAdmin = g.current_user.isCeltsAdmin or isProgramManager %}
+{% set isEditable = isAdmin %}
```
BrianRamsay: I think this is irrelevant now that there is a separate view template

app/templates/admin/createEvent.html
```diff
@@ -152,17 +154,58 @@ <h1 id="title">{{page_title}}</h1>
...
+            <input
+              id="eventFacilitator"
+              class="form-control"
+              type="search"
+              placeholder="Search facilitator"
+              value="" />
+            <table
+              class="table mb-3"
+              id="facilitatorTable"
+              name="facilitatorTable">
```
BrianRamsay: if the attributes fit on one row, leave them on one row

app/templates/admin/createEvent.html
```diff
@@ -152,17 +154,58 @@ <h1 id="title">{{page_title}}</h1>
...
+                {% endfor %}
+              </tbody>
+            </table>
+            {% else %}
+              {% if not eventData.facilitators|count %}
+                <p>None</p>
+                {%endif%}
+                {% for f in eventData.facilitators %}
+                <p>{{f.firstName}} {{f.lastName}}</p>
+                {% endfor %}
```
BrianRamsay: This bit should go in the view template

BrianRamsay (changes requested):

app/templates/admin/createEvent.html
Lyisus: This part should be kept here for when the event edited. If there are facilitators, the person editing the event should be able to see them. However, this currently does not work with the new search bar, but I think Mostaphe created an issue for it to keep this PR from getting overly large.

app/logic/events.py
```diff
@@ -325,16 +323,27 @@ def preprocessEventData(eventData):
...
+    This function takes in a list of usernames or a string with the usernames of
+        facilitators separated by commas (,) and returns a list of facilitator
+        objects that match the usernames in the list
+    facilitatorList: expected to be a list with facilitator usernames as strings
+                        or a string of usernames separated by commas (,)
+    return: list of facilitator objects matching the usernames in facilitatorList
+    """
+    if facilitatorList and type(facilitatorList) == str:
+        facilitatorList = facilitatorList.split(',')
+    finalFacilitatorList = []
```
BrianRamsay: This isn't necessary anymore

app/logic/events.py
```diff
@@ -325,16 +323,27 @@ def preprocessEventData(eventData):
...
+    # for each facilitator in facilitatorList, if it is not a blank string of null
+    # add a user object matching said facilitator to finalFacilitatorList
+    finalFacilitatorList = [User.get_by_id(facilitator) for facilitator in facilitatorList if facilitator]
```
BrianRamsay, line 344: Is there a case where we need that `if facilitator`? We probably want to have an error if there is weird data being passed in

app/templates/admin/createEvent.html
BrianRamsay: This page is always editable. That's why there aren't any more isEditable references

BrianRamsay (changes requested):

app/templates/admin/createEvent.html
```diff
@@ -152,17 +154,40 @@ <h1 id="title">{{page_title}}</h1>
...
+              </thead>
+              <tbody>
+                <tr hidden>
+                  <td>
+                    <p class="facilitatorName" required></p>
+                  </td>
+                  <td><button id="remove" type="button" class="btn btn-danger">Remove</button></td>
+
+                </tr>
+                {% for instructor in courseInstructor %}
```
BrianRamsay, line 177: Facilitators aren't showing up on the create event page. I can see it was added by looking at View, but it's not there on Edit

app/templates/admin/createEvent.html
```diff
@@ -152,17 +154,40 @@ <h1 id="title">{{page_title}}</h1>
...
+              id="eventFacilitator"
+              class="form-control"
+              type="search"
+              placeholder="Search facilitator"
+              value="" />
+            <table class="table mb-3" id="facilitatorTable" name="facilitatorTable">
+              <thead>
+                <tr>
+                  <th scope="col">Facilitator</th>
+                  <th scope="col">Remove</th>
```
BrianRamsay: Let's remove the header text above the buttons

BrianRamsay (changes requested):

app/templates/admin/createEvent.html
Lyisus: We made that a separate issue to keep this PR smaller. We'll get to that one once this one is merged.

app/static/js/createEvents.js
```diff
@@ -127,4 +129,39 @@ $(document).ready(function() {
       $("#checkIsRequired").prop('disabled', false);
     }
   });
+
+  var facilitatorArray = []
+  function callback(selectedFacilitator) {
+    // JSON.parse is required to de-stringify the search results into a dictionary.
```
BrianRamsay, line 135: This comment isn't true anymore

BrianRamsay (approved):

Issue #397, opened 2022-07-08
On the Admin Service Learning page: get the Email All button working and move it to the top right of the table, across from the search bar. 


AndersonStettner:
![Annotation 2022-07-08 093207](https://user-images.githubusercontent.com/26262045/178001953-7945b736-888b-46c5-a861-56e8128184f5.png)

bledsoef:
PR merged, closing issue.

PR #449, opened 2022-07-12
Email all issue397

thawpt:
created a list of instructor emails and inserted in the href so that all service learning faculty will be emailed. when button is clicked it will redirect to outlook and set the recipients as faculty from the SL faculty table.

only half the email all button was working so we edited the style 

fixes issue #397


BrianRamsay:
It looks like this was created from the branch with the add past participants change. I will look at this after that one is reviewed and merged.

app/templates/main/manageServiceLearningFaculty.html
```diff
@@ -102,12 +102,21 @@ <h2>Approved</h2>
...
         <button class="btn btn-primary">Send to Registrar</button>
       </div>
-      {% endif %}
+     {% endif %}
       </div>
     </div>
 
 <div id="wrapper">
     <h1 class="text-center pt-1 pb-4">Service-Learning Faculty</h1>
+    <div style="text-align:right; width:100%; padding:0; margin-bottom: -42px; display: inline-block" >
```
zachneill: Change as many of the style tags to their BS5 class equivalents. E.g. p-0 and w-100 are BS classes that should do the same thing. Look at the utilities found here: https://getbootstrap.com/docs/5.0/utilities/spacing/

app/templates/events/trackVolunteers.html
```diff
@@ -39,16 +39,19 @@ <h1 class="text-center">{{event.name}}</h1>
...
   <div class='form-group d-inline-flex'>
-    <a class="btn btn-success" role="button" href="#" data-bs-toggle="modal" data-bs-target="#addVolunteerModal" style="margin:5px;">Add volunteer</a>
+    <a class="btn btn-success" role="button" href="#" data-bs-toggle="modal" data-bs-target="#addVolunteerModal" style="margin:5px;">Add Volunteer</a>
     {% if event.isRecurring and not event.isFirstRecurringEvent%}
-      <a class="btn btn-success" id="pastVolunteers" role="button" data-bs-toggle="modal" style="margin:5px;">Add previous volunteers</a>
-    {% endif %}
+        {% if recurringVolunteers | count == 0  %}
+            {% set disabled = "disabled" %}
+        {% endif %}
+        <button type="button" class="btn btn-success" id="pastVolunteers" data-bs-toggle="modal" data-bs-target= "#addPastVolunteerModal" style="margin:5px;" {{disabled}}>Add Past Volunteers</button>
```
zachneill: Avoid style tags with margin if you can do a similar bootstrap utility. https://getbootstrap.com/docs/5.0/utilities/spacing/

zachneill (changes requested):
Look into Bootstrap 5 spacing utilities https://getbootstrap.com/docs/5.0/utilities/spacing/

app/templates/main/manageServiceLearningFaculty.html
BrianRamsay: One trouble they had is that we'd have to compile our own bootstrap css with negative margin enabled to use their negative margin classes. And we haven't done that yet

BrianRamsay (changes requested):
Are there unpushed commits? It doesn't look like the z-index fix is in.

BrianRamsay (approved):

Issue #398, opened 2022-07-08
In the Service Learning Course Proposal: take out the hyphen in Service-Learning everywhere it appears, Improve the name of the page "New Designated Service-Learning Course", and fix both of the "service learning course guidelines" links


AndersonStettner:
Link 1:
![image](https://user-images.githubusercontent.com/26262045/178003349-57db6cf9-1db8-47c6-a492-5ac6527bdf83.png)

Link 2 and title to be changed:
![image](https://user-images.githubusercontent.com/26262045/178003392-df868157-4948-4a3f-84d1-541e51653244.png)

bledsoef:
PR merged, closing issue.

PR #451, opened 2022-07-12
Renaming page and fixing links

thawpt:
- removed the hyphens 

- edited the href so that it directs to the right page when the link is clicked 

fixes issue #398

### DIFF
--- a/app/templates/serviceLearning/slcProposal.html
+++ b/app/templates/serviceLearning/slcProposal.html
@@ -1,6 +1,6 @@
 {%set title ="SLC Proposal"%}
-<h2 class="text-center pt-4 pb-2"> New Proposal For Service Learning Course</h2>
-<p class="text-center fw-bold"> Please review the <a href="/serviceLearning/editProposal/{{courseID}}">service learning course guidelines</a> before completing this form </p>
+<h2 class="text-center pt-4 pb-2"> New Proposal For Service-Learning Course</h2>
+<p class="text-center fw-bold"> Please review the <a href="/serviceLearning/editProposal/{{courseID}}">Service-Learning course guidelines</a> before completing this form </p>
 <div class="container-fluid">
   <div class="row">
     <div class="p-2 col-md-6">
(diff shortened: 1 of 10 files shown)


app/templates/serviceLearning/slcProposal.html
```diff
@@ -1,6 +1,6 @@
 {%set title ="SLC Proposal"%}
-<h2 class="text-center pt-4 pb-2"> New Proposal For Service-Learning Course Designation</h2>
-<p class="text-center fw-bold"> Please review the <a href="/serviceLearning/newProposal">service learning course guidelines</a> before completing this form </p>
+<h2 class="text-center pt-4 pb-2"> New Proposal For Service Learning Course</h2>
+<p class="text-center fw-bold"> Please review the <a href="/serviceLearning/editProposal/{{courseID}}">service learning course guidelines</a> before completing this form </p>
```
BrianRamsay: Does this work if they are on a new proposal?

BrianRamsay (changes requested):
I think the original ticket had an incorrect task. We want to make "service learning" vs "service-learning" consistent, and since they have "service-learning" everywhere in the proposal description that is probably the right way to go. So the new task should be to change all of the text and headings to use the hyphenated version.

If you want to do this without having to manually fix the ones you changed, you can undo just that commit with `git revert`

app/templates/serviceLearning/slcProposal.html
```diff
@@ -1,6 +1,6 @@
 {%set title ="SLC Proposal"%}
-<h2 class="text-center pt-4 pb-2"> New Proposal For Service-Learning Course Designation</h2>
-<p class="text-center fw-bold"> Please review the <a href="/serviceLearning/newProposal">service learning course guidelines</a> before completing this form </p>
+<h2 class="text-center pt-4 pb-2"> New Proposal For Service-Learning Course</h2>
+<p class="text-center fw-bold"> Please review the <a href="/serviceLearning/editProposal/{{courseID}}">Service-Learning course guidelines</a> before completing this form </p>
```
BrianRamsay: This breaks if they are creating a new proposal. There are two different cases to handle in this link.

BrianRamsay (changes requested):
We don't want to change the capitalization everywhere, just the hyphens. So everywhere that the words would normally be lowercase, it should still be "service-learning"

BrianRamsay (dismissed):

BrianRamsay (approved):

Issue #458, opened 2022-07-13
Make a new function that determines what templates a user can access and update the getPrograms function


bledsoef:
Make a new function called getTemplates that determines what templates a user can access and update the getPrograms function to make the admins get all programs.

Boorclark:
This has been merged. Closing issue.

PR #459, opened 2022-07-13
Made separate function for getTemplates and updated getPrograms.

bledsoef:
Adjusted the templateSelect controller method to utilize getPrograms more and made a function getTemplates toi remove any unnecessary conditionals.

Fixes issue #458

### DIFF
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -26,7 +26,7 @@ from app.logic.participants import getEventParticipants, getUserParticipatedEven
 from app.controllers.admin import admin_bp
 from app.controllers.admin.volunteers import getVolunteers
 from app.controllers.admin.userManagement import manageUsers
-from app.logic.userManagement import getPrograms, getTemplates
+from app.logic.userManagement import getAllowedPrograms, getAllowedTemplates
 
 
 @admin_bp.route('/switch_user', methods=['POST'])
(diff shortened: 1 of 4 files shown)


BrianRamsay (changes requested):
We need to add tests for the new functionality in `getPrograms` and for `getTemplates`. And while we're at it, we should rename them to `getAllowedPrograms` and `getAllowedTemplates` (or better names if you think of better ones).

app/controllers/admin/routes.py
```diff
@@ -44,13 +44,8 @@ def switchUser():
...
     if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff:
-        allprograms = []
-        if g.current_user.isCeltsStudentStaff:
-            allprograms = getPrograms(g.current_user)
-            visibleTemplates = []
-        else:
-            allprograms = Program.select().order_by(Program.programName)
-            visibleTemplates = EventTemplate.select().where(EventTemplate.isVisible==True).order_by(EventTemplate.name)
+        allprograms = getPrograms(g.current_user)
+        visibleTemplates = getTemplates(g.current_user)
```
BrianRamsay: This no longer works.

BrianRamsay (changes requested):
Get in the habit of testing your pages after you make changes. This breaks just clicking on Create Event

app/controllers/admin/routes.py

BrianRamsay (approved):

Issue #445, opened 2022-07-12
Add functionality to allow multiple volunteers to be added at once.


martinj2-dot:
Add volunteer modal only allows user to add one volunteer at a time when form submits. Fix it to where you can add multiple volunteers at once.

![image](https://user-images.githubusercontent.com/54441608/178514449-bf15df05-8eb4-49b5-8a2d-a92c263f695c.png)

file paths:

app/templates/events/trackVolunteers.html

bledsoef:
PR merged, closing issue

PR #471, opened 2022-07-14
Issue#445 add multiple volunteers

martinj2-dot:
#445 
This feature allows users to select multiple volunteers to be added to volunteerTable. Implemented within the Add Volunteer Modal and utilizes a checkbox to allow users to toggle which volunteers they want to add.


app/controllers/admin/volunteers.py
```diff
@@ -78,12 +78,19 @@ def updateVolunteerTable(eventID):
...
-@admin_bp.route('/addVolunteerToEvent/<volunteer>/<eventId>', methods = ['POST'])
-def addVolunteer(volunteer, eventId):
-    username = volunteer.strip("()").split('(')[-1]
-    user = User.get(User.username==username)
-    successfullyAddedVolunteer = addVolunteerToEventRsvp(user, eventId)
-    EventParticipant.create(user=user, event=eventId) # user is present
+@admin_bp.route('/addVolunteerToEvent/<eventId>', methods = ['POST'])
+def addVolunteer(eventId):
+    volunteerDict = request.form
+    volunteerList = volunteerDict.getlist("volunteer[]")
```
BrianRamsay: I don't think `volunteerDict` adds anything here. It's longer than just using `request.form` in the first place, and is only needed in that one spot.

app/controllers/admin/volunteers.py
```diff
@@ -78,12 +78,19 @@ def updateVolunteerTable(eventID):
...
+    for volunteerUsername in volunteerList:
+        successfullyAddedVolunteer = False
+        user = User.get(User.username==volunteerUsername)
+        if EventParticipant.select().where(EventParticipant.user == user, EventParticipant.event == eventId).exists():
+            successfullyAddedVolunteer = False
+        else:
+            addVolunteerToEventRsvp(user, eventId)
+            EventParticipant.create(user=user, event=eventId) # user is present
+            successfullyAddedVolunteer = True
     if successfullyAddedVolunteer:
```
BrianRamsay: We tell them success or failure based only on the last volunteer. We need to think about multiples

app/controllers/admin/volunteers.py
```diff
@@ -78,12 +78,19 @@ def updateVolunteerTable(eventID):
...
-    EventParticipant.create(user=user, event=eventId) # user is present
+@admin_bp.route('/addVolunteerToEvent/<eventId>', methods = ['POST'])
+def addVolunteer(eventId):
+    volunteerDict = request.form
+    volunteerList = volunteerDict.getlist("volunteer[]")
+    for volunteerUsername in volunteerList:
+        successfullyAddedVolunteer = False
+        user = User.get(User.username==volunteerUsername)
+        if EventParticipant.select().where(EventParticipant.user == user, EventParticipant.event == eventId).exists():
+            successfullyAddedVolunteer = False
```
BrianRamsay: We talked on past participants about treating an existing record as a success

BrianRamsay (changes requested):

app/static/js/trackVolunteers.js
```diff
@@ -21,32 +21,53 @@ $(document).ready(function() {
...
+    $("#selectVolunteerButton").click(function(){
+        let user = $("#addVolunteerInput").val()
+        let eventId = $("#eventID").val()
+        let checkboxlist = $("#addVolunteerModal input[type=checkbox]")
+        let volunteerList = []
+            $.each(checkboxlist, function(index, value){
+                if(value["checked"] == true){
+                    volunteerList.push(value["value"])
+                }
+            })
```
BrianRamsay: This block should be unindented one level

app/static/js/trackVolunteers.js
```diff
@@ -21,32 +21,53 @@ $(document).ready(function() {
...
-  }
+  // Adding the new volunteer to the user database table
+    $("#selectVolunteerButton").click(function(){
+        let user = $("#addVolunteerInput").val()
+        let eventId = $("#eventID").val()
+        let checkboxlist = $("#addVolunteerModal input[type=checkbox]")
+        let volunteerList = []
+            $.each(checkboxlist, function(index, value){
+                if(value["checked"] == true){
+                    volunteerList.push(value["value"])
```
BrianRamsay: I wouldn't use this variable name `value`. They are checkbox elements.

app/static/js/trackVolunteers.js
```diff
@@ -21,32 +21,53 @@ $(document).ready(function() {
...
+      })
+    })
 
+    var userlist = []
+    function callback(selected) {
+      $("#selectVolunteerButton").prop('disabled', false);
+      let user = $("#addVolunteerInput").val()
+      if(userlist.includes(selected["username"]) == false){
+          userlist.push(user)
+          $("#addVolunteerList").append("<table>")
```
BrianRamsay: Let's not add tables for every username. You're not even using rows and columns for the table

app/static/js/trackVolunteers.js
```diff
@@ -21,32 +21,53 @@ $(document).ready(function() {
...
+          for(let i = 0; i < 1 ; i++){
```
BrianRamsay: I'm not sure why this `for` loop is necessary at all. You are getting your index from `userlist.length`

app/static/js/trackVolunteers.js
BrianRamsay: `#addVolunteerList` is a `<ul>`, so you should be creating and appending list items, not tables

BrianRamsay (changes requested):

app/static/js/trackVolunteers.js
```diff
@@ -21,32 +21,50 @@ $(document).ready(function() {
...
+    })
 
+    var userlist = []
+    function callback(selected) {
+      $("#selectVolunteerButton").prop('disabled', false);
+      let user = $("#addVolunteerInput").val()
+      if(userlist.includes(selected["username"]) == false){
+          userlist.push(user)
+          let i = userlist.length;
+          $("#addVolunteerList").append("<li id= addVolunteerElements"+i+"> </li>")
```
BrianRamsay: Now that we are using list items, we should style them so they look a little better. We don't need bullets.

BrianRamsay: Let's also make sure that attribute value has quotes around it, even though it does work as is.

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #454, opened 2022-07-13
In the Program Settings of the admin settings: remove drop down and just have a list of programs. Make the program names editable.


AndersonStettner:
![image](https://user-images.githubusercontent.com/26262045/178783136-9c3a654c-bcfb-47c4-b5ff-e073a2350c0e.png)
Instead of "Program Select" it should be a list of the program names that the user is able to update.

bledsoef:
PR merged, closing issue.

PR #483, opened 2022-07-15
Issue#454 edit program settings

qasema:
fixes issue #454 by reformatting the program setting and making the editable


app/controllers/admin/userManagement.py
```diff
@@ -1,3 +1,4 @@
+from imp import reload
```
BrianRamsay: You need to turn off the auto-import feature. This import isn't used.

app/controllers/admin/userManagement.py
```diff
@@ -66,20 +67,24 @@ def removeProgramManagers():
...
+                                    programInfo["Sender"],
+                                    programID)
+            flash("Program updated", "success")
+            return redirect(url_for("admin.userManagement"))     
         except Exception as e:
             print(e)
             flash('Error while updating program info.','warning')
             abort(500,'Error while updating program.')
     abort(403)
+    return ""
```
BrianRamsay: This return will never be reached, as `abort` will exit the function

app/templates/admin/userManagement.html
```diff
@@ -99,25 +99,18 @@ <h3 class="accordion-header" id="headingThree">
...
-                <td><input type="text" name="Sender" value="" id="emailSenderName"></td>
-                <td>
-                  <button type="button" class="btn btn-success w-101" id="addNewProgramInfo" role ="button">Update</button>
-                </td>
+        <div class="programData" id="programSelect" form="programSettings">
+            {% for program in programs %}
+            <form action="/admin/updateProgramInfo/{{program.id}}" method="POST">
+                <tr class="p-1"> 
+                    <td> <input type="text" name="Name" value ="{{program.programName}}" id="programName_{{program.id}}"> </td>
+                    <td><input type="text" name="Email" value="{{program.emailReplyTo}}"emailReplyTo_{{program.id}}" ></td>
```
BrianRamsay: Some extra text in here, and mismatched quotes

app/templates/admin/userManagement.html
```diff
@@ -99,25 +99,18 @@ <h3 class="accordion-header" id="headingThree">
...
+                    <td><input type="text" name="Sender" value="{{program.emailSenderName}}" id=emailSenderName_{{program.id}}"></td>
+                    <td><button type="submit" class="btn btn-success w-101" id="button_{{program.id}}" role ="button">Update</button> </td>
```
BrianRamsay: We also can drop all of the id attributes from these elements

BrianRamsay (changes requested):

app/templates/admin/userManagement.html
```diff
@@ -99,25 +99,18 @@ <h3 class="accordion-header" id="headingThree">
...
-                    <option value ="{{program.id}}" data-replytoemail="{{program.emailReplyTo}}" data-sendername="{{program.emailSenderName}}">{{program.programName}}</option>
-                    {% endfor %}
-                  </select>
-                </td>
-                <td><input type="text" name="Email" id="emailReplyTo" value=""></td>
-                <td><input type="text" name="Sender" value="" id="emailSenderName"></td>
-                <td>
-                  <button type="button" class="btn btn-success w-101" id="addNewProgramInfo" role ="button">Update</button>
-                </td>
+        <div class="programData" id="programSelect" form="programSettings">
```
BrianRamsay: Divs around the table elements but not the whole table aren't a great idea. Do you need this div for anything? Could you put the id and class on other elements?

BrianRamsay (changes requested):

app/templates/admin/userManagement.html
```diff
@@ -92,32 +92,23 @@ <h3 class="accordion-header" id="headingThree">
         </button>
       </h3>
       <div id="collapseThree" class="accordion-collapse collapse" aria-labelledby="headingThree" data-bs-parent="#adminManagement">
-        <div class="accordion-body">
+        <div id="programSelect" form="programSettings" class="accordion-body programData">
```
BrianRamsay: I don't think any of these attributes are necessary anymore. Are they?

BrianRamsay (changes requested):
A few considerations:
- There is quite a bit of javascript that isn't relevant any more
- Using a table for layout means the display isn't responsive anymore
- The program name should be not null in the database, and we don't want them to clear it.
- It would be nice if the page reload kept the Program Settings accordion open

app/templates/admin/userManagement.html
qasema: We are going to look at the display responsivity later

app/templates/admin/userManagement.html
```diff
@@ -3,7 +3,12 @@
 
 {% block scripts %}
   {{super()}}
+<<<<<<< HEAD
+  <script src="/static/js/userManagement.js"></script>
+
+=======
   <script type="module" src="/static/js/userManagement.js"></script>
+>>>>>>> origin/development
```
BrianRamsay: Merge conflicts in the file

app/templates/admin/userManagement.html
```diff
@@ -117,13 +122,26 @@ <h3 class="accordion-header" id="headingThree">
...
+                <tr class="p-1"> 
+                    <td> <input type="text" name="Name" value ="{{program.programName}}" required> </td>
+                    <td><input type="text" name="Email" value="{{program.emailReplyTo}}" required></td>
+                    <td><input type="text" name="Sender" value="{{program.emailSenderName}}" required></td>
+                    <td><button type="submit" class="btn btn-success w-101" role ="button">Update</button> </td>
+                </tr>
+                </form>
+            {% endfor %}
+        </table>
+=======
```
BrianRamsay: more merge conflicts

BrianRamsay (changes requested):

BrianRamsay (changes requested):
You didn't update the test for `changeProgramInfo`.

Remember, when you are modifying a function, change the test first, before you touch the function. That way you won't forget.

app/controllers/admin/userManagement.py
```diff
@@ -1,4 +1,5 @@
-from flask import Flask, render_template,request, flash, g, json, abort, redirect, url_for
+from this import d
```
BrianRamsay: turn off auto-adding imports!!!!!!!

BrianRamsay (changes requested):
You didn't update the test for `changeProgramInfo`.

Remember, when you are modifying a function, change the test first, before you touch the function. That way you won't forget.

app/controllers/admin/userManagement.py
```diff
@@ -64,15 +65,18 @@ def removeProgramManagers():
...
         try:
-            return changeProgramInfo(programInfo["emailReplyTo"],  #calls logic function to add data to database
-                                    programInfo["emailSenderName"],
-                                    programInfo["programId"])
+            changeProgramInfo(programInfo["Name"],  #calls logic function to add data to database
+                                    programInfo["Email"],
+                                    programInfo["Sender"],
+                                    programID)
+            flash("Program updated", "success")
+            return redirect(url_for("admin.userManagement", programID=programID))
```
BrianRamsay: Let's pass this like `showSettingsPane="program"`

BrianRamsay (changes requested):

BrianRamsay (changes requested):
The ideal behavior after saving the program is that the page reloads and we are left looking at the program settings again.

Right now, both program settings and user management are open, and we are viewing the top of the page.

BrianRamsay:
Some conflicts to resolve

BrianRamsay (approved):

Issue #376, opened 2022-07-05
Student Staff: the way we show the programs they are managing should be clearer


AndersonStettner:
Change how the programs that a student staff is managing are displayed.  A colon also needs to be added at the end of "User is a manager for the following programs"

bledsoef:
I improved the look of this but here are a couple of other potential fixes:
- making the list of programs an accordion so that when you click on it the events for the program display.
- for admins, make the list of programs divided into two columns and make it a table that will make it look much better.

bledsoef:
PR merged, closing issue.

PR #484, opened 2022-07-15
Cleaned up staff and admin program selection/view 376

bledsoef:
Part of issue #376 

Added page breaks and made the list of programs a student staff manages to be a table.

### DIFF
--- a/app/templates/main/volunteerProfile.html
+++ b/app/templates/main/volunteerProfile.html
@@ -147,7 +147,17 @@
   <div class="row">
     <h2><br>{{volunteer.firstName}} {{volunteer.lastName}} is the manager of:</h2>
     <div>
-      <table class="table">
+        <h5><br><u>Program(s)</u></h5>
+        <ul>
+        {% for program in programs %}
+          {% set status = '' %}
+            {% if program.id in permissionPrograms %}
+               <li>{{program.programName}}</li>
+            {% else %}
+            {% endif %}
+        {% endfor %}
+        </ul>
+      <!-- <table class="table">
           <thead>
              <tr>
                <th scope="col">Program</th>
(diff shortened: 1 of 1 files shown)


BrianRamsay (changes requested):
We don't want to use a table for data that isn't tabular. That is, there's only one column here. There are better html cases

app/templates/main/volunteerProfile.html
```diff
@@ -140,21 +140,26 @@ <h2>Which programs do you want to make {{volunteer.firstName}} {{volunteer.lastN
...
+    <h2><br>{{volunteer.firstName}} {{volunteer.lastName}} is the manager of:</h2>
     <div>
-      {% for program in programs %}
-        {% set status = '' %}
-          {% if program.id in permissionPrograms %}
-            <p  style="padding-right:50px;">{{program.programName}}</p>
-          {% else %}
-          {% endif %}
-      {% endfor %}
+        <h5><br><u>Program(s)</u></h5>
```
BrianRamsay: I think it is understood that this is a list of programs. I'd say leave this header out.

app/templates/main/volunteerProfile.html
```diff
@@ -140,21 +140,26 @@ <h2>Which programs do you want to make {{volunteer.firstName}} {{volunteer.lastN
           {% endfor %}
         </div>
       </div>
+      <br>
     {% endif %}
 
 {% elif g.current_user.isCeltsStudentStaff %}
```
BrianRamsay: It's not the current user we want to check here, but the volunteer

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #478, opened 2022-07-15
Need to be able to edit/add phone numbers on the Volunteer Profile page


AndersonStettner:

bledsoef:
PR merged, closing issue

PR #486, opened 2022-07-18
Update phone#478

qasema:
fixes issue #478  allow updating the phone number of the volunteer in volunteer profile page

### DIFF
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -148,6 +148,11 @@ def updateProgramManager():
         abort(403)
 @admin_bp.route("/updatePhone", methods=["POST"])
 def updatePhone():
+    ''' 
+    This will update the phone Number of user 
+    para: None
+    
+    '''
     newinfo=request.form
     User.update(phoneNumber=newinfo["phoneNumber"]).where(User.username==newinfo["user_name"]).execute()
     return ""
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -146,3 +146,8 @@ def updateProgramManager():
         return ""
     else:
         abort(403)
+@admin_bp.route("/updatePhone", methods=["POST"])
+def updatePhone():
+    newinfo=request.form
+    User.update(phoneNumber=newinfo["phoneNumber"]).where(User.username==newinfo["user_name"]).execute()
+    return ""
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -148,11 +148,6 @@ def updateProgramManager():
         abort(403)
 @admin_bp.route("/updatePhone", methods=["POST"])
 def updatePhone():
-    ''' 
-    This will update the phone Number of user 
-    para: None
-    
-    '''
     newinfo=request.form
-    User.update(phoneNumber=newinfo["phoneNumber"]).where(User.username==newinfo["user_name"]).execute()
+    User.update(phoneNumber=newinfo["phoneNumber"]).where(User.username==newinfo["username"]).execute()
     return ""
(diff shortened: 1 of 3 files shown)


app/static/js/volunteerProfile.js
```diff
@@ -114,6 +114,16 @@ $(document).ready(function(){
...
     })
   });
+  $("#updatePhone").click(function(){
+    user=$(this).data("username") 
+    phonNo=$("#newinput").val()
+    $.ajax({
+      method:"POST",
+      url:"/updatePhone",
+      data:{"user_name":user,
+            "phoneNumber":phonNo}
```
BrianRamsay: `user_name` should be `username`, and `phonNo` is either misspelled or poorly abbreviated.

app/controllers/admin/volunteers.py
```diff
@@ -146,3 +146,13 @@ def updateProgramManager():
         return ""
     else:
         abort(403)
+@admin_bp.route("/updatePhone", methods=["POST"])
+def updatePhone():
+    ''' 
+    This will update the phone Number of user 
+    para: None
+    
+    '''
```
BrianRamsay: This is a simple enough function we can leave off the docstring comment. We aren't generating this for api documentation or anything

BrianRamsay (changes requested):

app/templates/main/volunteerProfile.html
```diff
@@ -15,7 +15,16 @@
...
-    <h6 align="center">{{volunteer.phoneNumber}}</h6>
+    <div align="center">
+      <tr>
+        <td>
+          <input id="newinput" type="text" value="{{volunteer.phoneNumber}}">
+        </td>
+        <td>
+          <button class=" btn-success bi-pincel h-2" id="updatePhone" data-username="{{volunteer.username}}">Update</button>
+        </td>
+    </tr>
```
BrianRamsay: Figure out how to do layout without tables. Especially when you aren't even including a `<table>` tag

BrianRamsay (changes requested):
Let's make the input size fit the size of a phone number a little better.

BrianRamsay (changes requested):
Since we are displaying our phone numbers straight from the database, we need to standardize the phone numbers before we insert. We need a function that will format a phone number like (###) ###-####, or maybe better, a function that updates the phone number that formats it that way.

BrianRamsay (changes requested):
We should leave feedback that the database has been updated. Also, it should not let me save an invalid phone number.

app/static/js/volunteerProfile.js
```diff
@@ -114,6 +114,24 @@ $(document).ready(function(){
       }
     })
   });
+  $("#updatePhone").click(function(){
+    userName=$(this).data("username") 
+    phoneNumber=$("#newinput").val()
+    let isvalid =phoneNumber.replace(/\D/g,"").length
```
BrianRamsay: isValid should be boolean, not the length. So either you do the check on this line and keep it as `isValid`, or you change your variable name

app/templates/main/volunteerProfile.html
```diff
@@ -15,9 +16,11 @@
 <div>
     <h1 align="center">{{volunteer.firstName}} {{volunteer.lastName}}</h1>
     <h6 align="center">{{volunteer.email}}</h6>
-    <h6 align="center">{{volunteer.phoneNumber}}</h6>
+    <div align="center">
+          <input size="14" id="newinput" type="text"  class="bfh-phone" value="{{volunteer.phoneNumber}}">
```
BrianRamsay: This should be something like `phoneInput`, not `newInput`

app/static/js/volunteerProfile.js
```diff
@@ -114,6 +114,24 @@ $(document).ready(function(){
...
+    if (isvalid ===10){
+      $.ajax({
+        method:"POST",
+        url:"/updatePhone",
+        data:{"username":userName,
+              "phoneNumber":phoneNumber}
+      })
+      msgFlash("Phone Number is updated", "success")
```
BrianRamsay: Let's use ajax success and error functions to make sure we are only flashing a success message if we were successful

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #408, opened 2022-07-08
On the My Profile page on the Upcoming Events tab: display the users upcoming events and make it clear that the "Choose interests below..." header is related to the table. 


AndersonStettner:
![Annotation 2022-07-08 111519](https://user-images.githubusercontent.com/26262045/178021081-c557306f-fe18-4a75-91bc-9b0099bccb04.png)
At the top of the page the events that the user is going to be participating in should be displayed. If there are none the user should be told this.

Also the "Choose interests below..." header should be more clearly related to the table it is about.

bledsoef:
The code in the html works fine, I think the query is messed up in logic because it is never printing any upcoming events.

PR #487, opened 2022-07-18
Display upcoming events 408

Boorclark:
Fixes Issue #408

Changes Made:
-  Changed Event query 
-  Updated HTML table headers
-  Got rid of unneeded breaks

image
image

Files Changed:
 - app/logic/events.py

### DIFF
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -190,8 +190,8 @@ def getUpcomingEventsForUser(user,asOf=datetime.datetime.now()):
                             .join(ProgramEvent)
                             .join(Interest, on=(ProgramEvent.program == Interest.program))
                             .where(Interest.user == user,
-                            Event.startDate >= asOf,
-                            Event.timeStart > asOf.time())
+                                   Event.startDate >= asOf,
+                                   Event.timeStart > asOf.time())
                             .distinct() # necessary because of multiple programs
                             .order_by(Event.startDate, Event.name) # keeps the order of events the same when the dates are the same
                             )
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -170,7 +170,7 @@ def getOtherEvents(term):
                            Event.isTraining == False,
                            Event.isAllVolunteerTraining == False,
                            ((ProgramEvent.program == None) |
-                            (Program.isStudentLed == False) & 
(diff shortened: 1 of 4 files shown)


app/logic/events.py
```diff
@@ -189,9 +189,9 @@ def getUpcomingEventsForUser(user,asOf=datetime.datetime.now()):
     events = (Event.select(Event)
                             .join(ProgramEvent)
                             .join(Interest, on=(ProgramEvent.program == Interest.program))
-                            .where(Interest.user == user)
-                            .where(Event.startDate >= asOf)
-                            .where(Event.timeStart > asOf.time())
+                            .where(Interest.user == user,
+                            Event.startDate >= asOf,
+                            Event.timeStart > asOf.time())
```
BrianRamsay: Let's line these up with `Interest`, not `.where`

BrianRamsay (changes requested):
Let's also reduce the header size of the message when there are no events

app/logic/events.py
Boorclark: resolved

app/templates/main/volunteerProfile.html
```diff
@@ -53,14 +53,11 @@ <h6 align="center">{{volunteer.phoneNumber}}</h6>
            </tr>
        {% endfor %}
      {% else %}
-       <h2><p class="text-dark" align="center">Choose interests below to be notified about upcoming events.</p><h2>
+       <h4><p class="text-dark" align="center"> No upcoming events. </p></h4>
+       <h4><p class="text-dark" align="center">Choose interests below to be notified about upcoming events.</p><h4>
```
BrianRamsay: Is there a reason we are using a paragraph tag inside a header tag? It seems a little backwards. You should be able to get the styles you want with just a header or just a paragraph

BrianRamsay: The size seems good right now. I'm not sold on the center alignment, though.

BrianRamsay (changes requested):
I added an upcoming Empty Bowls event for myself and checked notify for all the programs, but I don't see any upcoming events table

BrianRamsay (approved):

PR #489, opened 2022-07-18
Fixing test_userManagement 

### LINKED ISSUE #480: Fix test_UserManagement
If you run the tests individually _test_UserManagement_ will break permanently.

AndersonStettner:
Moved the addProgramManager function out of UserManagement and into User. Created a test for addProgramManager in test_users. Also fix referances for addProgramManager from its old location to its new one.


app/models/user.py
```diff
@@ -17,11 +17,33 @@ class User(baseModel):
     @property
     def isAdmin(self):
         return (self.isCeltsAdmin or self.isCeltsStudentStaff)
-    
+
+    def addProgramManager(self, program):
+        # Makes a user a Program Manager
+        from app.models.programManager import ProgramManager
+
+        addManager = ProgramManager.create(user = self, program = program)
```
BrianRamsay: If we aren't using this value, no need to make a variable for it.

app/models/user.py
```diff
@@ -17,11 +17,33 @@ class User(baseModel):
...
+
+        return (f' {self} added as Program Manager')
+
+    def removeProgramManager(self, program):
+        # Removes an existing Program Manager from being a Program Manager
+        from app.models.programManager import ProgramManager
+
+        removeManager = ProgramManager.delete().where(ProgramManager.user == self, ProgramManager.program == program).execute()
```
BrianRamsay: Same here

app/models/user.py
```diff
@@ -17,11 +17,33 @@ class User(baseModel):
...
+        # Looks to see who is the Program Manager for a program
         from app.models.programManager import ProgramManager  # Must defer import until now to avoid circular reference
         return ProgramManager.select().where(ProgramManager.user == self, ProgramManager.program == program).exists()
-    
 
- 
+    def isProgramManagerForEvent(self, event):
+        # Looks to see who the Program Manager for a specific event is
+        from app.models.event import Event
+        from app.models.programManager import ProgramManager
```
BrianRamsay: We don't use either of these imports in this method

tests/code/test_users.py
```diff
@@ -172,3 +174,165 @@ def test_userpriv():
...
+#
+#         noPM = list(User.select(User.username).join(ProgramManager).where(ProgramManager.program_id == 2))
+#
+#
+#         # Make sure what is expected is returned
+#         assert removePM == (f'{user} removed from Program Manager')
+#         # Check that the Program Manager is actually removed from the Program Manager table
+#         assert user not in noPM
+#
+#         transaction.rollback()
```
BrianRamsay: we don't want this code still in here

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #490, opened 2022-07-18
Minor example code removal

zachneill:
This PR just removes a couple bootstrap 5 example tags. Found a duplicate id reference as well, although it wasn't causing noticeable issues.

### DIFF
--- a/app/templates/events/trackVolunteers.html
+++ b/app/templates/events/trackVolunteers.html
@@ -50,7 +50,7 @@
   </div>
 </div>
 <form method="post">
-  <label class='d-none' id="eventLength">{{eventLength}}</label>
+  <label class='d-none'>{{eventLength}}</label>
   <input class="form-control" id="eventID" name="event" value="{{event.id}}" type='hidden'>
   <table class="table" id="trackVolunteerstable">
     <thead>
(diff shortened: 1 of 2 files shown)


app/templates/main/reviewproposal.html
```diff
@@ -6,7 +6,7 @@ <h5  class="col-11 modal-title text-center">Review Course Proposal </h5>
           <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
         </div>
         <div class="modal-body">
-          <table class="table table-striped ml-4" id="myTable">
```
BrianRamsay, line 9: This id is used by js

app/templates/events/trackVolunteers.html
```diff
@@ -50,7 +50,7 @@ <h1 class="text-center">{{event.name}}</h1>
   </div>
 </div>
 <form method="post">
-  <label class='d-none' id="eventLength">{{eventLength}}</label>
+  <label class='d-none'>{{eventLength}}</label>
```
BrianRamsay, line 53: rather than just remove the id, we should clean up the other duplicate code around it and make sure it still works. Both the label and input after it are duplicated elsewhere. Do we know if the hidden elements inside the form are needed or the ones outside?

BrianRamsay (changes requested):

app/templates/main/reviewproposal.html
bledsoef: among-us-sus

PR #511, opened 2022-07-19
Service Learning Course Proposal Extra Page Pop-up

mohamudm1:
Changed made:
We got rid of a quick extra window that was popping up for a few seconds after the user clicks on the submit button and before they are redirected to the course proposal management page


mohamudm1:
Closing this PR because it was built on top of the previous branch accidently

Issue #499, opened 2022-07-19
User Management page should list admins and student staff


BrianRamsay:
I think a better UI for the user management section of settings would be to list the current admins and the current student staff. Each item could have a remove button next to it to remove the user from the list. The two Add search boxes would still operate the same way, below the lists

bledsoef:
PR merged, closing issue.

PR #515, opened 2022-07-20
Fixed User Management UI To Show Student Staff and Admins

Boorclark:
Fixes Issue #499 

Changes Made:
 - Fixed "flash" conditionals and got rid of unneeded code
 - Adjusted JavaScript submit request function and conditionals
 - Changed UI for user management accordion tab
 - Updated formatting on the lower half of HTML code because it was funky looking

image

Files Changed:
 - app/controllers/admin/userManagement.py
 - app/static/js/userManagement.js
 - app/templates/admin/userManagement.html


app/static/js/userManagement.js
```diff
@@ -27,6 +27,13 @@ $(document).ready(function() {
   $("#programSelect").on("change",function(){
     displayProgramInfo();
   });
+  $(".removeAdmin").on("click",function(){
+    submitRequest("removeCeltsAdmin", ($(this).val()), 'button');
+  });
+  $(".removeStudentStaff").on("click",function(){
+    submitRequest("removeCeltsStudentStaff", ($(this).val()), 'button');
```
BrianRamsay: remove the parentheses around the value parameter. It's a little weirder in js than it is in python

app/static/js/userManagement.js
```diff
@@ -39,13 +46,21 @@ $(document).ready(function() {
...
+          method: method,
+          user: identifier,
+          from: "ajax"
+      }
+  } else {
+      data = {
+          method: method,
+          user: $(identifier).val(),
+          from: "ajax"
+      }
```
BrianRamsay: What if, instead of passing in `identifier` and `from`, we just pass in the value every time instead of just in the 'button' case. Then we don't have to have two different cases here.

app/templates/admin/userManagement.html
```diff
@@ -31,24 +29,51 @@ <h3 class="accordion-header" id="headingOne">
...
+                      <tr>
+                        <th scope="col">Current Admin</th>
+                        <th scope="col"></th>
+                      </tr>
+                  </thead>
+                  <tbody>
+                      {% for admin in currentAdmins %}
+                         <tr>
+                            <td id="{{admin.username}}">{{admin.firstName}} {{admin.lastName}}</td>
+                            <td><button value="{{admin.username}}" type="button" class="btn btn-danger view removeAdmin" val="{{admin.username}}">Remove</button></td>
```
BrianRamsay: Why do you have val and value on the button? Let's use a data attribute to store the username like we do elsewhere.

app/templates/admin/userManagement.html
```diff
@@ -31,24 +29,51 @@ <h3 class="accordion-header" id="headingOne">
...
+                </table>
             </div>
-            <div class="row d-flex justify-content-center">
-              <div class="col-md-6 mb-3">
-                {{createInputsButtons("removeCeltsAdminInput","removeCeltsAdmin","Remove","danger","Remove Celts Admin")}}
-              </div>
-              <div class="col-md-6">
-                {{createInputsButtons("removeCeltsStudentStaffInput","removeCeltsStudentStaff","Remove","danger", "Remove Celts Student Staff")}}
+            <div class="col-md-6 mb-3">
+                <div>
```
BrianRamsay, line 53: Does this div do anything?

app/templates/admin/userManagement.html
```diff
@@ -100,26 +125,24 @@ <h3 class="accordion-header" id="headingThree">
...
-                    {% endfor %}
-                  </select>
-                </td>
-                <td><input type="text" name="Email" id="emailReplyTo" value=""></td>
-                <td><input type="text" name="Sender" value="" id="emailSenderName"></td>
-                <td>
-                  <button type="button" class="btn btn-success w-101" id="addNewProgramInfo" role ="button">Update</button>
-                </td>
-              </div>
+                <div>
```
BrianRamsay: Don't put divs inside a table mixed in with the rows

BrianRamsay: actually, why are we messing with this section at all? You will have conflicts with ala's changes to the program selection if all you are changing is indentation here.

BrianRamsay (changes requested):

app/static/js/userManagement.js
Boorclark: resolved

app/static/js/userManagement.js
Boorclark: resolved

app/templates/admin/userManagement.html
Boorclark: resolved

app/templates/admin/userManagement.html
Boorclark: resolved

BrianRamsay (approved):

Issue #514, opened 2022-07-19
New SL Course Proposal Cancel button throws the user to an "Access Denied" page.


AndersonStettner:
When a user goes to create a new SLC proposal and they hit cancel on the first page, they are thrown out to an "Access Denied" page. They should instead be thrown back out to the "Course Proposals" page.

PR #523, opened 2022-07-20
Redirected the user to the course proposals page when the they click cancel

mohamudm1:
Solves issue #514 

Changes made: Redirected the faculty user to the course proposal window if they click cancel after the second page.


app/static/js/slcNewProposal.js
```diff
@@ -57,7 +57,7 @@ $("#nextButton").on("click", function() {
 });
 
 $("#cancelButton").on("click", function() {
-  window.location.replace("/manageServiceLearning");
+        window.location.replace("/serviceLearning/courseManagement");
```
BrianRamsay: Doesn't this depend on whether it is an admin creating the new proposal or a faculty member?

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #516, opened 2022-07-20
Remove the add button on admin management


bledsoef:
Remove the add button on admin management and just add the user upon the enter key or clicking the drop down. PR #515 is currently in the process of being merged which reformats the page so ideally this shouldn't be completed until that is resolved.

![image](https://user-images.githubusercontent.com/89262886/179995887-98386be5-7d90-4981-a180-62c65d30f58d.png)

bledsoef:
PR merged, closing issue

PR #526, opened 2022-07-20
Removed add button and moved functionality to drop down and enter key.

bledsoef:
Fixes issue #516 

Removed the add button from the user management portion of program settings and allowed the user dropdown to be clickable or for enter to submit.

### DIFF
--- a/app/controllers/admin/userManagement.py
+++ b/app/controllers/admin/userManagement.py
@@ -22,7 +22,7 @@ def manageUsers():
     except Exception as e:
         print(e)
         flash(username + " is an invalid user.", "danger")
-        return ("danger")
+        return ("danger", 500)
 
     if method == "addCeltsAdmin":
         if user.isStudent:
--- a/app/controllers/admin/userManagement.py
+++ b/app/controllers/admin/userManagement.py
@@ -20,10 +20,10 @@ def manageUsers():
 
     if method == "addCeltsAdmin":
         if user.isStudent:
-            flash(username + " cannot be added as a Celts admin", 'danger')
+            flash(user.firstName + " " + user.lastName + " cannot be added as a Celts admin", 'danger')
         else:
             if user.isCeltsAdmin:
-                flash(username + " is already a Celts Admin", 'danger')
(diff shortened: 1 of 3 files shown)


app/controllers/admin/userManagement.py
```diff
@@ -16,14 +16,20 @@ def manageUsers():
...
     method = eventData['method']
     username = re.sub("[()]","", (user.split())[-1])
-    user = User.get_by_id(username)
+
+    try:
+        user = User.get_by_id(username)
+    except Exception as e:
+        print(e)
+        flash(username + " is an invalid user.", "danger")
+        return ("danger")
```
BrianRamsay: If it's an error, return an error code. Like `return ("danger"), 500`

app/static/js/userManagement.js
```diff
@@ -1,22 +1,16 @@
...
-    ['removeCeltsAdminInput',       'removeCeltsAdmin',       'admin'],
-    ['removeCeltsStudentStaffInput','removeCeltsStudentStaff','studentstaff']
-  ];
-  $.each(searchElements, function(i,arr) {
-      let [inputId, btnId, category] = arr
-      $("#"+inputId).on("input", () => searchUser(inputId, callback, false, null, category))
-      $("#"+btnId).on("click", () => submitRequest(btnId, $("#"+inputId).val()))
-  });
+  $("#searchCeltsAdminInput").on("input", () => searchUser("searchCeltsAdminInput", callbackAdmin, false, null, 'instructor'))
+  $("#searchCeltsStudentStaffInput").on("input", () => searchUser("searchCeltsStudentStaffInput", callbackStudentStaff, false, null, "student"))
```
BrianRamsay: Let's reformat these to match the other event handlers, now that they are the same as the others

BrianRamsay (changes requested):

app/controllers/admin/userManagement.py
Boorclark: resolved

app/static/js/userManagement.js
Boorclark: resolved

BrianRamsay (approved):

Issue #461, opened 2022-07-14
Give a temporary course ID to New Proposal so that a user is able to go back to their page without losing the work they put in


solijonovam:

Boorclark:
PR merged, closing issue.

PR #527, opened 2022-07-20
New course

solijonovam:
-assigns a new course ID as soon as 'create course proposal' so that it directs to an editing page using the course ID
-we added a new button called 'save and continue'/'save for later' so that it saves the changes
----> we used an ajax request to update the course
we also added a space between 'add course' and 'send to faculty' buttons
-essentially got rid of the newProposal page

addresses issue #461


app/controllers/serviceLearning/routes.py
```diff
@@ -68,6 +68,27 @@ def slcEditProposal(courseID):
                                 isAllSectionsServiceLearning = isAllSectionsServiceLearning,
                                 isPermanentlyDesignated = isPermanentlyDesignated,
                                 courseID=courseID)
+@serviceLearning_bp.route('/serviceLearning/saveProposal', methods=['POST'])
+def slcSaveContinue():
+    """Will update the the course proposal and return an empty string since ajax request needs a response"""
+    courseExist = Course.get_or_none(Course.id == request.form.get('courseID'))
+    if courseExist:
+        updateCourse(request.form.copy(), instructorsDict)
```
BrianRamsay: Where does `instructorsDict` come from?

app/controllers/serviceLearning/routes.py
```diff
@@ -68,6 +68,27 @@ def slcEditProposal(courseID):
...
+    return ""
+
+@serviceLearning_bp.route('/serviceLearning/createCourse/', methods=['POST'])
+def slcCreateCourse():
+    """will give a new course ID so that it can redirect to an edit page"""
+    status = CourseStatus.get(CourseStatus.status == "Pending")
+    courseID = Course.create(
+        status=status)
+    id = Course.get_by_id(courseID)
```
BrianRamsay: This line is redundant. You are already getting a course object when you do `Course.create`. Also, your variable isn't a courseID, it's a course object.

app/templates/main/manageServiceLearningFaculty.html
```diff
@@ -34,7 +34,13 @@ <h2> Pending </h2>
...
             <div class="float-end">
               <button type="button" class="btn btn-primary btn-sm">Send To Faculty Fellow</button>
-              <a type="button" href="/serviceLearning/newProposal" class="btn btn-success btn-sm">Add Course</a>
+            </div>
+            <div class =AddnewCourse align="right">
+              <form method="post" action="/serviceLearning/createCourse/" >
+                <button class="btn btn-success btn-sm" type = 'Submit' style='margin-right:16px'>Add Course </button>
+              </form>
+            </div>
+            <div class="space">
```
BrianRamsay: What does this class do?

app/templates/main/manageServiceLearningFaculty.html
```diff
@@ -34,7 +34,13 @@ <h2> Pending </h2>
           <div class="col">
             <div class="float-end">
               <button type="button" class="btn btn-primary btn-sm">Send To Faculty Fellow</button>
-              <a type="button" href="/serviceLearning/newProposal" class="btn btn-success btn-sm">Add Course</a>
+            </div>
+            <div class =AddnewCourse align="right">
```
BrianRamsay: Put quotes around the class, and fix the name to be camelCase

app/templates/serviceLearning/slcNewProposal.html
```diff
@@ -32,9 +32,10 @@
       <span class="step"></span>
     </div>
     <div class="col" align="right">
+      <button class="btn btn-primary" type="button" id="saveContinue">Save and Continue</button>
       <button class="btn btn-primary" type="button" id="nextButton">Next</button>
-      {% if g.current_user.isCeltsAdmin %}
-        <button class="btn btn-success" type="button" id="saveAndApproveButton">Save and Approve</button>
+      {% if g.current_user == isCeltsAdmin %}
```
BrianRamsay: This would only work if you pass in isCeltsAdmin, and that's not necessary since you can use `g.current_user`

BrianRamsay (changes requested):

app/controllers/serviceLearning/routes.py
```diff
@@ -67,6 +67,32 @@ def slcEditProposal(courseID):
                                 isAllSectionsServiceLearning = isAllSectionsServiceLearning,
                                 isPermanentlyDesignated = isPermanentlyDesignated,
                                 courseID=courseID)
+@serviceLearning_bp.route('/serviceLearning/saveProposal', methods=['POST'])
+def slcSaveContinue():
+    """Will update the the course proposal and return an empty string since ajax request needs a response"""
+    course = Course.get_by_id(request.form.get('courseID'))
+    updateCourse(request.form.copy(), instructorsDict)
+    if len(request.form) == 1:
```
BrianRamsay: What is this conditional checking for? Does it need to be there?

app/controllers/serviceLearning/routes.py
```diff
@@ -67,6 +67,32 @@ def slcEditProposal(courseID):
...
+        except:
+            flash("Course not incomplete!", "danger")
+    return ""
+
+@serviceLearning_bp.route('/serviceLearning/createCourse/', methods=['POST'])
+def slcCreateCourse():
+    """will give a new course ID so that it can redirect to an edit page"""
+    course = Course.create(
+        status=CourseStatus.INCOMPLETE)
+    id = Course.get_by_id(course)
```
BrianRamsay: You don't need to get the course from the database, because you already have the course object in the line above

app/templates/main/manageServiceLearningFaculty.html
```diff
@@ -34,7 +34,11 @@ <h2> Submitted </h2>
           <div class="col">
             <div class="float-end">
               <button type="button" class="btn btn-primary btn-sm">Send To Faculty Fellow</button>
-              <a type="button" href="/serviceLearning/newProposal" class="btn btn-success btn-sm">Add Course</a>
+            </div>
+            <div class = 'addNewCourse' align="right">
+              <form method="post" action="/serviceLearning/createCourse/" >
+                <button class="btn btn-success btn-sm" type = 'Submit' style='margin-right:16px'>Add Course </button>
```
BrianRamsay: Put styles in a class instead of in a style attribute. That way we could use them in other places where we have two buttons next to each other

app/templates/main/manageServiceLearningFaculty.html
```diff
@@ -34,7 +34,11 @@ <h2> Submitted </h2>
           <div class="col">
             <div class="float-end">
               <button type="button" class="btn btn-primary btn-sm">Send To Faculty Fellow</button>
-              <a type="button" href="/serviceLearning/newProposal" class="btn btn-success btn-sm">Add Course</a>
+            </div>
+            <div class = 'addNewCourse' align="right">
```
BrianRamsay: You don't reference this class (as far as I can tell)

app/templates/serviceLearning/slcManagement.html
```diff
@@ -8,8 +8,10 @@
 {% block app_content %}
 <input id="username" type="hidden" value='{{user.username}}'>
 <h1 class="text-center mt-5 mb-2">{{user.firstName}} {{user.lastName}}'s SLC Proposals</h1>
-<div align="right">
-  <a class="btn btn-primary" href="/serviceLearning/newProposal" role="button">Create Course Proposal</a>
+<div class =newCourse align="right">
```
BrianRamsay: You don't reference this class (as far as I can tell)

app/templates/serviceLearning/slcNewProposal.html
```diff
@@ -32,9 +32,10 @@
       <span class="step"></span>
     </div>
     <div class="col" align="right">
+      <button class="btn btn-primary" type="button" id="saveContinue">Save and Continue</button>
       <button class="btn btn-primary" type="button" id="nextButton">Next</button>
       {% if g.current_user.isCeltsAdmin %}
-        <button class="btn btn-success" type="button" id="saveAndApproveButton">Save and Approve</button>
+        <button class="btn btn-success" type="button" id="approveButton">Save and Approve</button>
```
BrianRamsay: Let's rename this "Submit and Approve", to make it clearer that it is the same kind of thing as Submit

BrianRamsay (changes requested):

app/controllers/serviceLearning/routes.py
thawpt: it is not properly creating the course id and is throwing an error saying "index out of range" when we remove  
 id = Course.get_by_id(courseID)

app/controllers/serviceLearning/routes.py
BrianRamsay: Still relevant

app/static/css/manageServiceLearningFaculty.css
```diff
@@ -0,0 +1,3 @@
+.mr{
+    margin-right: 16px;
```
BrianRamsay, line 2: Don't use `px` for spacing because it might look weird when the screen size changes

BrianRamsay (approved):

PR #534, opened 2022-07-20
Fixing Upcoming events and refactoring its test

### LINKED ISSUE #521: Upcoming events are only showing up when user has an interest.

AndersonStettner:


tests/code/test_events.py
```diff
@@ -519,33 +522,92 @@ def test_getAllFacilitators():
...
+                                description = "Test upcoming program event.",
+                                timeStart = "18:00:00",
+                                timeEnd = "21:00:00",
+                                location = "The sun",
+                                startDate = 2021-12-12,
+                                endDate = 2021-12-13)
+
+        # Create a new Program to create the new Program Event off of so the
+        # user can mark interest for it
+        ProgramForInterest = Program.create(id = 13,
```
BrianRamsay: Don't create variables with an initial capital. Those names are for classes

tests/code/test_events.py
```diff
@@ -519,33 +522,92 @@ def test_getAllFacilitators():
...
+                                            programName = "BOO",
+                                            isStudentLed = False,
+                                            isBonnerScholars = False,
+                                            emailReplyTo = "test@email",
+                                            emailSenderName = "testName")
+
+        ProgramEventForInterest = ProgramEvent.create(program = 13, event = newProgramEvent)
+
+        program_id = 13
```
BrianRamsay: Don't duplicate this magic number. Just use the id in your program object

BrianRamsay: Or just use the program object itself in the query

tests/code/test_events.py
```diff
@@ -519,33 +522,92 @@ def test_getAllFacilitators():
...
+        # User has not RSVPd and is Interested
+        addInterest = addUserInterest(program_id, user)
+        eventsInUserInterestedProgram = getUpcomingEventsForUser(user, asOf = testDate)
+
+        eventsInProgram = list((Event.select()
+                                     .join(ProgramEvent)
+                                     .join(Program)
+                                     .where(Event.startDate >= testDate,
+                                            ProgramEvent.event_id == Event.id,
+                                            ProgramEvent.program_id == 13)))
```
BrianRamsay: You created the events here for a new program, so you shouldn't need to do a big select like this

tests/code/test_events.py
```diff
@@ -519,33 +522,92 @@ def test_getAllFacilitators():
...
+        # user has RSVPd and is Interested
+        addUserRsvp = addVolunteerToEventRsvp(user, noProgram.id)
+        eventsInUserInterestAndRsvp = getUpcomingEventsForUser(user, asOf = testDate)
+
+        eventsInProgramAndRsvp = list((Event.select()
+                                        .join(ProgramEvent, JOIN.LEFT_OUTER)
+                                        .join(Interest, JOIN.LEFT_OUTER, on=(ProgramEvent.program == Interest.program))
+                                        .join(EventRsvp, JOIN.LEFT_OUTER, on=(Event.id == EventRsvp.event))
+                                        .where(Event.startDate >= testDate,
+                                              (Interest.user == user) | (EventRsvp.user == user))))
```
BrianRamsay: Same comment here. You don't need to duplicate the query because you know what will be included

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #493, opened 2022-07-19
Investigate and fix the pages that have lost mobile responsiveness.


AndersonStettner:

bledsoef:
Pages that need to be refactoring:
- main events list table
- Make the z index of the sidebar icon mega big and keep it in the same place
- Kind of both service learning course proposal tables
- Program settings in admin management

bledsoef:
PR merged, closing issue.

PR #536, opened 2022-07-21
Updates UI for Mobile Responsiveness 

Boorclark:
Fixes Issue #493 

Changes Made:
 - Made tables responsive by adding a table-responsive bootstrap class
 - Fixes the Program Settings so it doesn't go off screen
 - Changed the Z-index of the hamburger Icon to be 1 less than the CELTS menu

image
image

### DIFF
--- a/app/templates/main/volunteerProfile.html
+++ b/app/templates/main/volunteerProfile.html
@@ -31,7 +31,7 @@
    </ul>
  </div>
 
- <div class="row">
+ <div class="row table-responsive">
    <table class="table">
      {% if upcomingEvents|length > 0 %}
        <thead class="tableHeader">
@@ -59,7 +59,7 @@
    </table>
  </div>
  <!-- Start Program Eligibility Table -->
- <div class="row">
+ <div class="row table-responsive">
    <table class="table banTable">
       <thead>
         <tr>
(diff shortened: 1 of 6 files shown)


app/templates/main/volunteerProfile.html
```diff
@@ -129,10 +129,12 @@ <h2>Which programs do you want to make {{volunteer.firstName}} {{volunteer.lastN
                   {% set action = "remove" %}
                   {% set checked = "checked" %}
             {% endif %}
+            <div class="table-responsive">
               <table>
                     <input  type="checkbox" id="mgr-{{ program.id }}" data-programid="{{program.id}}" {{checked}}  onclick="updateManagers(this,'{{volunteer.username}}')">
                     <label  style="padding-left: 15px;" for="mgr-{{ program.id }}">{{program.programName}}</label>
               </table>
+            </div>
```
BrianRamsay: Rather than wrap this with table-responsive we should get rid of the table. It's weird, anyway. Also, there was already an empty div wrapping this (for some reason)

BrianRamsay (changes requested):

app/templates/main/volunteerProfile.html
Boorclark: resolved

BrianRamsay (approved):

Issue #535, opened 2022-07-21
Spamming the create event button keeps you on the create event page.


bledsoef:
Clicking save on the create event page should direct you to edit event but if you click it more than once it keeps you there.

bledsoef:
PR merged, closing issue

PR #537, opened 2022-07-21
Made the save event button disable after form submit

bledsoef:
Fixes issue #535 
Made the save event button disable after form submit because it was causing issues.

### DIFF
--- a/app/static/js/createEvents.js
+++ b/app/static/js/createEvents.js
@@ -35,9 +35,9 @@ function format24to12HourTime(timeStr){
  * Run when the webpage is ready for javascript
  */
 $(document).ready(function() {
-  $("#saveEventBtn").click(function() {
-      $("#saveEvent").submit()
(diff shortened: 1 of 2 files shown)


app/static/js/createEvents.js
```diff
@@ -35,6 +35,11 @@ function format24to12HourTime(timeStr){
  * Run when the webpage is ready for javascript
  */
 $(document).ready(function() {
+  $("#saveEventBtn").click(function() {
+      $("#saveEvent").submit()
```
BrianRamsay: Doesn't this double submit? When you click a submit button the form will submit automatically. Your onclick should just disable the button

BrianRamsay (changes requested):

app/static/js/createEvents.js
bledsoef: No, disabling the button somehow makes the form not submit.

BrianRamsay (approved):

Issue #510, opened 2022-07-19
Buttons should look like buttons and links should look like links.


AndersonStettner:
There are buttons that look like links and links that look like buttons. Go through the application and fix all of the formattings so that they are consistent.

PR #544, opened 2022-07-21
Stadarize Buttons and Links Accross the app

Lyisus:
#510 
Standarizes buttons to look like buttons and links to look like links. We only found the one Edit link in the volunteer profile that looked like a link but its functionality would was more fitting for a button.


BrianRamsay:
@Andersonstettner do you want to check this one out and see if this was the only instance you were thinking of?

AndersonStettner (approved):

Issue #541, opened 2022-07-21
Hovering over the completed trainings should show all trainings for program


bledsoef:
Hovering over the completed trainings icon should show all trainings for program. 
![image](https://user-images.githubusercontent.com/89262886/180252953-0b92a8da-be80-4d7b-ba99-1f616565f3bd.png)

bledsoef:
PR merged, closing issue.

PR #547, opened 2022-07-21
Show trainings upon on hover

bledsoef:
Fixes issue #541  
When hovering over the "view" tag next to the x's and check's the program trainings for each program shows up along with whether you have attended them. 
We also changed the logic function getUserParticipatedEvents and the test suite accordingly to only get the current term programs or current academic year.

### DIFF
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -113,7 +113,7 @@ def viewVolunteersProfile(username):
                                               ProgramBan.program == program,
                                               ProgramBan.endDate > datetime.datetime.now())
 
-            userParticipatedEvents = getUserParticipatedEvents(program, g.current_user)
+            userParticipatedEvents = getUserParticipatedEvents(program, g.current_user, g.current_term)
             noteForDict = notes[-1].banNote.noteContent if notes else ""
             eligibilityTable.append({"program": program,
                                    "completedTraining": (volunteer.username in trainedParticipants(program, g.current_term)),
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -120,7 +120,6 @@ def viewVolunteersProfile(username):
                                    "trainingList": userParticipatedEvents,
                                    "isNotBanned": True if not notes else False,
                                    "banNote": noteForDict})
-        print(eligibilityTable)
         return render_template ("/main/volunteerProfile.html",
                 programs = programs,
                 programsInterested = programsInterested,
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -22,7 +22,7 @@ from app.models.courseStatus import CourseStatus
 from app.controllers.main import main_bp
 from app.logic.loginManager import logout
 from app.logic.users import addUserInterest, removeUserInterest, banUser, unbanUser, isEligibleForProgram
-from app.logic.participants import userRsvpForEvent, unattendedRequiredEvents, trainedParticipants
+from app.logic.participants import userRsvpForEvent, unattendedRequiredEvents, trainedParticipants, getUserParticipatedEvents
 from app.logic.events import *
 from app.logic.searchUsers import searchUsers
 from app.logic.transcript import *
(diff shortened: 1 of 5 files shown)


app/controllers/main/routes.py
```diff
@@ -113,12 +113,14 @@ def viewVolunteersProfile(username):
...
-            eligibilityTable.append({"program" : program,
-                                   "completedTraining" : (volunteer.username in trainedParticipants(program, g.current_term)),
-                                   "isNotBanned" : True if not notes else False,
+            eligibilityTable.append({"program": program,
+                                   "completedTraining": (volunteer.username in trainedParticipants(program, g.current_term)),
+                                   "trainingList": userParticipatedEvents,
+                                   "isNotBanned": True if not notes else False,
                                    "banNote": noteForDict})
-
+        print(eligibilityTable)
```
BrianRamsay: remove debug output

app/logic/participants.py
```diff
@@ -22,9 +22,9 @@ def trainedParticipants(programID, currentTerm):
 
     otherTrainingEvents = (Event.select(Event.id).join(ProgramEvent)
             .where(
-                ProgramEvent.program==programID,
-                Event.isTraining==True,
-                (((Event.name == "All Celts Training") | (Event.name == "All Volunteer Training")) & (Event.term == ayStart)) | (Event.term==currentTerm))
+                ProgramEvent.program == programID,
+                Event.isTraining == True,
+                (((Event.name == "All Celts Training") | (Event.name == "All Volunteer Training")) & (Event.term == ayStart)) | (Event.term == currentTerm))
```
BrianRamsay: We have a flag for the volunteer training event now, so we don't need to check names. I know all you did was change spaces here, but congratulations, now you get to fix it.

app/logic/participants.py
```diff
@@ -96,23 +96,31 @@ def unattendedRequiredEvents(program, user):
...
     :returns: trainings for program and if the user participated
     """
-    programTrainings = Event.select().join(ProgramEvent).where(Event.isTraining == 1, ProgramEvent.program == program)
+    ayStart = currentTerm.academicYearStartingTerm
+
+    programTrainings = (Event.select()
+                               .join(ProgramEvent)
+                               .where(Event.isTraining == True,
+                                      ProgramEvent.program == program,
+                                      (Event.term == ayStart) | (Event.term == currentTerm))
```
BrianRamsay: If our current term is Summer, this will query will only be checking Fall and Summer

app/static/js/volunteerProfile.js
```diff
@@ -113,6 +113,18 @@ $(document).ready(function(){
...
+    var requiredTraining = $(".trainingPopover");
+    requiredTraining.popover({
+      trigger: "hover",
+      sanitize: false,
+      html: true,
+      content: function() {
+                     return $(this).attr('data-content');
+        }
+    });
+  });
```
BrianRamsay, line 145: When is this called? I'm confused why it is in the jquery function

tests/code/test_participants.py
```diff
@@ -301,13 +301,23 @@ def test_getEventParticipantsWithWrongParticipant():
...
     with mainDB.atomic() as transaction:
-        allProgramTrainings = Event.select().join(ProgramEvent).where(Event.isTraining == 1, ProgramEvent.program == Program.get_by_id(2))
+        currentTerm = Term.get_by_id(3)
+        ayStart = currentTerm.academicYearStartingTerm
+
+        allProgramTrainings = (Event.select()
+                                   .join(ProgramEvent)
+                                   .where(Event.isTraining == True,
+                                          ProgramEvent.program == Program.get_by_id(2),
+                                          (Event.term == ayStart) | (Event.term == currentTerm))
```
BrianRamsay: Same comment here as before. Also, we don't want to just repeat the query. We probably should set up data that we know will work (or not work) regardless of what is in test data

BrianRamsay (changes requested):

app/controllers/main/routes.py
bledsoef: Resolved

app/logic/participants.py
bledsoef: Resolved

app/logic/participants.py
bledsoef: Resolved

app/static/js/volunteerProfile.js
bledsoef: We need this to actually display the trainings if I'm not mistaken, the return happens upon hover.

tests/code/test_participants.py
bledsoef: Resolved

app/static/js/volunteerProfile.js
BrianRamsay: What I mean is, is this supposed to be in the onready handler, instead of in `$(function() { })`. Clearly it works, I'm just a little confused as to how. I don't know when this function runs

BrianRamsay (changes requested):

BrianRamsay (changes requested):
A couple of usability enhancements to consider:
- Let's link to the event page for each of the training events.
- Let's put the event date next to the name if the training event is in the future.

bledsoef:
> A couple of usability enhancements to consider:
> 
> * Let's link to the event page for each of the training events.
> * Let's put the event date next to the name if the training event is in the future.

We would have to make this no longer a hover if you want us to have clickable links on the popover.

BrianRamsay (approved):

Issue #550, opened 2022-07-21
Sending emails does not seem to work anymore


bledsoef:
Even though sending emails wasn't working in the first place, it now breaks development if you try to send an email due to an error with attachments.

Boorclark:
PR merged. Closing issue

PR #552, opened 2022-07-21
added exception handling for IsADirectoryError

Boorclark:
Fixes Issue #550 

Added exception handling for IsADirectoryError. This occurs when we try to interact with a directory as if it were a file

### DIFF
--- a/app/logic/emailHandler.py
+++ b/app/logic/emailHandler.py
@@ -157,7 +157,7 @@ class EmailHandler:
 
         except AttributeError:  # will pass if there is no attachment to save
             pass
-        except FileExistsError:
+        except FileExistsError: # will pass if the file already exists
             pass
 
         return attachmentFullPath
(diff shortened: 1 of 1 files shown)


app/logic/emailHandler.py
```diff
@@ -168,6 +168,8 @@ def saveAttachment(self):
             self.attachment_file.save(self.getAttachmentFullPath()) # saves attachment in directory
         except AttributeError: # will pass if there is no attachment to save
             pass
+        except IsADirectoryError: # will pass if we try to handle a directory as a file
+            pass
```
BrianRamsay: Seems like we should prevent `getAttachmentFullPath` from returning only a directory path in the first place.

BrianRamsay (changes requested):

app/logic/emailHandler.py
Boorclark: resolved

BrianRamsay (approved):

Issue #546, opened 2022-07-21
Remove <font> tags from User 


BrianRamsay:
In general, tags shouldn't be used for styling. E.g., `<font>`, or `<small>`

We should use css for styling. Also, it's fine if the initial text on that view is larger, but the rest should be normally sized

bledsoef:
PR merged, closing issue

PR #554, opened 2022-07-21
Got rid of the tags used for styling and used CSS instead

mohamudm1:
Solved issue #546 

Removed the font tags used to determine the size of some of the headings and utilized CSS instead

### DIFF
--- a/app/templates/eventView.html
+++ b/app/templates/eventView.html
@@ -36,7 +36,7 @@
     {% endif %}
             <div>
                 <div>
-                    <h3 class="heading-size"><br>Description:</h3>{{eventData.description}}<br><br>
+                    <h3><br>Description:</h3>{{eventData.description}}<br><br>
                     <div>
                         {% if eventData.isService %}
                             This event earns service hours.
@@ -60,7 +60,7 @@
             <div class="container">
               <div class="row">
                 <div class="col-sm">
-                    <h3 class="heading-size"><br>Facilitators:</h3>
+                    <h3><br>Facilitators:</h3>
                         {% if eventFacilitatorNames %}
                             <ul>
                                 {% for facilitator in eventFacilitatorNames%}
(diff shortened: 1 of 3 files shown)


app/templates/eventView.html
```diff
@@ -30,13 +30,13 @@ <h1 id="title">{{page_title}}</h1>
...
-        <p><font size="+2"><b>{{eventData.name}}</b> took place on <b>{{eventData.startDate}}</b> from <b>{{eventData.timeStart}}</b> to <b>{{eventData.timeEnd}}</b> and was located in <b>{{eventData.location}}</b>.</font></p>
+        <p class="heading-size"><b>{{eventData.name}}</b> took place on <b>{{eventData.startDate}}</b> from <b>{{eventData.timeStart}}</b> to <b>{{eventData.timeEnd}}</b> and was located in <b>{{eventData.location}}</b>.</p>
     {% else %}
-        <p><font size="+2"><b>{{eventData.name}}</b> will take place on <b>{{eventData.startDate}}</b> from <b>{{eventData.timeStart}}</b> to <b>{{eventData.timeEnd}}</b> and will be located in <b>{{eventData.location}}</b>.</font></p>
+        <p class="heading-size"><b>{{eventData.name}}</b> will take place on <b>{{eventData.startDate}}</b> from <b>{{eventData.timeStart}}</b> to <b>{{eventData.timeEnd}}</b> and will be located in <b>{{eventData.location}}</b>.</p>
     {% endif %}
             <div>
                 <div>
-                    <font size="+2"><h3><br>Description:</h3>{{eventData.description}}<br><br>
+                    <h3><br class="heading-size">Description:</h3>{{eventData.description}}<br><br>
```
BrianRamsay: having the class on the break tag is kind of weird. There's no text in the break.

BrianRamsay (changes requested):

BrianRamsay (changes requested):
It looks good (and almost exactly the same) with the plan `<h3>` tags, without the class applied. Do you see the same thing?

mohamudm1:
I see that! Actually not so sure why I applied the class to that title. Maybe I was thinking of uniformity!!

BrianRamsay (approved):

Issue #562, opened 2022-07-22
Adding/Deleting Volunteer Buttons needs to be disabled after one click


Boorclark:
When you spam while adding or deleting a volunteer in the "Track Volunteers" section too many flash messages will appear. This can be fixed by disabling the buttons once they have been clicked as we have done for other buttons in the program. 
![image](https://user-images.githubusercontent.com/97637517/180462929-e9130524-75c7-4921-8fe0-81493476bd47.png)
![image](https://user-images.githubusercontent.com/97637517/180463046-795d3843-6860-49fd-84f6-67e592976c75.png)

Boorclark:
PR merged, closing issue.

PR #566, opened 2022-07-22
Disabled the remove and add volunteer buttons after click

bledsoef:
Fixes issue #562 
Disabled the remove and add volunteer buttons after click.

### DIFF
--- a/app/static/js/trackVolunteers.js
+++ b/app/static/js/trackVolunteers.js
@@ -24,6 +24,7 @@ $(document).ready(function() {
 
   // Adding the new volunteer to the user database table
     $("#selectVolunteerButton").click(function(){
+        $("#selectVolunteerButton").prop("disabled", true)
         let user = $("#addVolunteerInput").val()
         let eventId = $("#eventID").val()
         let checkboxlist = $("#addVolunteerModal input[type=checkbox]")
@@ -68,6 +69,7 @@ $(document).ready(function() {
   });
 
   $(".removeVolunteer").on("click", function() {
+    $(".removeVolunteer").prop("disabled", true)
     let username =  this.id;
     let eventId = $('#eventID').val()
     $.ajax({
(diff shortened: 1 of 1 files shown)


app/static/js/trackVolunteers.js
```diff
@@ -68,6 +69,7 @@ $(document).ready(function() {
   });
 
   $(".removeVolunteer").on("click", function() {
+    $(".removeVolunteer").prop("disabled", true)
```
BrianRamsay: Make sure if there is an error and the page doesn't reload that we can click the button again

BrianRamsay (changes requested):

app/static/js/trackVolunteers.js
```diff
@@ -68,12 +69,14 @@ $(document).ready(function() {
...
     let username =  this.id;
     let eventId = $('#eventID').val()
     $.ajax({
       url: `/removeVolunteerFromEvent/${username}/${eventId}`,
       type: "POST",
       success: function(s) {
+         $("#selectVolunteerButton").prop("disabled", true)
```
BrianRamsay: This isn't the right button for here.

BrianRamsay (changes requested):

app/static/js/trackVolunteers.js
Boorclark: resolved

app/static/js/trackVolunteers.js
Boorclark: resolved

app/static/js/trackVolunteers.js
```diff
@@ -74,6 +75,7 @@ $(document).ready(function() {
       url: `/removeVolunteerFromEvent/${username}/${eventId}`,
       type: "POST",
       success: function(s) {
+         $(".removeVolunteer").prop("disabled", true)
```
BrianRamsay: This disable is unnecessary, as the page is about to reload (and you already disabled it on click anyway). What you need is to re-enable the button if you don't reload the page

BrianRamsay (changes requested):

app/static/js/trackVolunteers.js
bledsoef: resolved

BrianRamsay (approved):

Issue #497, opened 2022-07-19
Facilitators should be displayed on events that have facilitators assigned to them 


mohamudm1:
P.S. This issue is waiting on #447 :)

When an event with facilitators is created, the facilitator's names will not show up if you try to review the event. Reviewers should be able to see the facilitators list. 

![Facilitators](https://user-images.githubusercontent.com/97688286/179789237-7efab6af-a55b-443a-9809-db0c6e6d8d9d.PNG)

PR #568, opened 2022-07-22
Display Event Facilitators

Lyisus:
#497 
Events that have facilitators assigned to them, display their facilitators in the table when editing the event and the backend handles the added functionality.

### DIFF
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -104,7 +104,6 @@ def createEvent(templateid, programid=None):
             template = template,
             eventData = eventData,
             futureTerms = futureTerms,
-            # eventFacilitators = getAllFacilitators(),
             isProgramManager = isProgramManager)
 
 @admin_bp.route('/eventsList/<eventId>/view', methods=['GET'])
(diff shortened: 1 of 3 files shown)


app/controllers/admin/routes.py
```diff
@@ -98,14 +98,13 @@ def createEvent(templateid, programid=None):
...
-
     futureTerms = selectSurroundingTerms(g.current_term, prevTerms=0)
-
+    # eventFacilitators = EventFacilitator.select().where(EventFacilitator.event == event)
     return render_template(f"/admin/{template.templateFile}",
             template = template,
             eventData = eventData,
             futureTerms = futureTerms,
-            allFacilitators = getAllFacilitators(),
+            # eventFacilitators = getAllFacilitators(),
```
BrianRamsay: Remove these instead of comment them out if they are not necessary. Are they necessary?

BrianRamsay (dismissed):

BrianRamsay (changes requested):
There is some commented out code

BrianRamsay:
On hold in case we add back our facilitator search

Issue #567, opened 2022-07-22
Redesign the View Event Page 


mohamudm1:
The Event view page looks strange and does not have the same design pattern as the rest of the app pages.

![Even View Page Window ](https://user-images.githubusercontent.com/97688286/180505544-93f6d8cb-d781-4017-94f6-5ec82c3204f4.PNG)

PR #590, opened 2022-07-25
Redesign event view page

mohamudm1:
Solved issue #567 

Changes made:
Redesigned the view event page complete. You can see what it looked like before and how it looks not below
Even View Page Window
After

### DIFF
--- a/app/static/css/base.css
+++ b/app/static/css/base.css
@@ -34,10 +34,6 @@
 
 }
 
-.font-size{
-    font-size: 1.2em;
-}
-
 .skipContent {
   left:0px;
   position:absolute;
--- a/app/static/css/base.css
+++ b/app/static/css/base.css
@@ -71,3 +71,7 @@
 .sidebar-icon {
     z-index: 1049;
 }
+
+.rsvp-btn{
+    padding-top: 4em;
+}
--- a/app/static/css/base.css
+++ b/app/static/css/base.css
@@ -34,7 +34,7 @@
 
 }
 
-.size{
+.font-size{
     font-size: 1.2em;
 }
 
--- a/app/static/css/base.css
+++ b/app/static/css/base.css
@@ -43,7 +43,7 @@
   z-index:0;
 }
 .heading-size{
-    font-size: 1.5em;
+    font-size: 1.2em;
 }
 
 .skipContent:focus, .skipContent:active {
--- a/app/static/css/base.css
+++ b/app/static/css/base.css
@@ -31,6 +31,11 @@
 
 .footerlink {
   font-weight: bold;
+
+}
+
+.size{
+    font-size: 1.2em;
 }
 
 .skipContent {
--- a/app/templates/eventView.html
+++ b/app/templates/eventView.html
@@ -30,18 +30,18 @@
          </div>
     {% endif %}
             <div class="row">
-                    <div style="padding-right:150px;"class="col-md-6">
+                    <div class="col-md-6">
                     <h3><br>Description</h3>
                     {% if isPastEvent %}
                         <p class="heading-size"><b>{{eventData.name}}</b> took place on <b>{{eventData.startDate}}</b> from <b>{{eventData.timeStart}}</b> to <b>{{eventData.timeEnd}}</b> and was located in <b>{{eventData.location}}</b>.</p>
                     {% else %}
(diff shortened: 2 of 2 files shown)


app/templates/eventView.html
```diff
@@ -29,82 +29,94 @@ <h1 id="title">{{page_title}}</h1>
...
-        <p class="heading-size"><b>{{eventData.name}}</b> took place on <b>{{eventData.startDate}}</b> from <b>{{eventData.timeStart}}</b> to <b>{{eventData.timeEnd}}</b> and was located in <b>{{eventData.location}}</b>.</p>
-    {% else %}
-        <p class="heading-size"><b>{{eventData.name}}</b> will take place on <b>{{eventData.startDate}}</b> from <b>{{eventData.timeStart}}</b> to <b>{{eventData.timeEnd}}</b> and will be located in <b>{{eventData.location}}</b>.</p>
-    {% endif %}
-            <div>
-                <div>
-                    <h3><br>Description:</h3>{{eventData.description}}<br><br>
-                    <div>
+            <div class="row">
+                    <div style="padding-right:150px;"class="col-md-6">
```
zachneill: The padding-right here gets a bit noticeable on smaller screens and especially mobile. Taking out the padding looked okay. If you still want padding, I'd suggest using rem or similar, not pixels.

app/static/css/base.css
```diff
@@ -31,6 +31,11 @@
 
 .footerlink {
   font-weight: bold;
+
+}
+
+.size{
```
zachneill: The class name "size" is pretty unindicative. A slightly more obvious name would be helpful

app/templates/eventView.html
```diff
@@ -29,82 +29,94 @@ <h1 id="title">{{page_title}}</h1>
...
-                                    {{programManager.user.firstName}} {{programManager.user.lastName}}<br>
-                                    {{programManager.user.email}}
-                                </div>
-                        {% endif %}
-                    </div>
-                </div>
-                <div>
-                    <div class="row">
+
+                <div class=" col-md-6">
```
zachneill: If you made it this far, might as well remove the added spaces in front of your col-md-6 's. I'm assuming you removed a class in front of it...

zachneill (dismissed):

BrianRamsay:
Let's try putting that first descriptive sentence above the Description and Facilitators headers. I also think the font-size class can go. The smaller text is fine, since that's the default text size we use in the rest of the application

app/templates/eventView.html
```diff
@@ -27,84 +27,89 @@ <h1 id="title">{{page_title}}</h1>
...
-                                            <div><b>{{training}}</b>: Attended</div>
-                                        {% elif participated == 0 %}
-                                            <div><b>{{training}}</b>: Not attended</div>
-                                        {% else %}
-                                            <div><b>{{training}}</b>: Upcoming</div>
-                                        {% endif%}
-                                    {% endif %}
-                                {% endfor %}
+                          {% if programManager %}
+                              <div><h3><br>Program Manager</h3></div>
```
BrianRamsay: No reason for a div around the header that I can see.

app/templates/eventView.html
```diff
@@ -27,84 +27,89 @@ <h1 id="title">{{page_title}}</h1>
...
-                <div>
-                    <h3><br>Description:</h3>{{eventData.description}}<br><br>
+            <div class="row">
+                    <div class="col-md-6">
+                    <h3><br>Description</h3>
+                    {{eventData.description}}<br><br>
                     <div>
                         {% if eventData.isService %}
                             This event earns service hours.
+                            <br />
```
BrianRamsay: Put all of these in `<p>` tags, instead of using breaks

app/templates/eventView.html
```diff
@@ -27,84 +27,89 @@ <h1 id="title">{{page_title}}</h1>
...
                           {% endif %}
-                        </div>
-                    </div>
-                </div>
+                      </div>
+                  </div>
             </div>
-            <div class="text-center">
+
+            <div style="padding-top:100px;"class="text-center">
```
BrianRamsay: We want to avoid style tags, and avoid defining things with pixels

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #555, opened 2022-07-21
Fix Term Change Popups 


AndersonStettner:
The term change popup should be centered over Admin Management not on the page.

![image](https://user-images.githubusercontent.com/26262045/180319854-578a71a9-5657-4931-bccc-e1dc82eb2b0b.png)

PR #592, opened 2022-07-25
 centered flash messages on admin management page and  on other pages

nartehi:
#555 
centered flash messages 

files changed: app/static/js/base.js

-----------------------------------------------------------------------------------------------------------------------------------------
image


BrianRamsay:
So we know that `position: relative` fixes the centering, but it also pushes everything in the page down when it shows up. We need to figure out how to get both centering, and positioning over other elements on the page.

BrianRamsay (approved):
I made a couple of changes to simplify the centering, and to make it work in all cases.

Issue #588, opened 2022-07-25
Program Management email Sender Name


AndersonStettner:
The sender name in the Program Management section of Admin Management auto populates to "testName". It would be best to have it be an empty field that is required to be filled in so that there is no accidental emails from someone named "testName". 
![image](https://user-images.githubusercontent.com/26262045/180795843-7a8a5fc5-ca23-4023-b8d8-1736fc9737eb.png)

PR #600, opened 2022-07-25
removed testName from sender Name field in Program Management

gahimbaref:
#588 
- removed "testName" from sender Name field in Program Management and made it blank by default"
- kept the sender name field required 

Files changed:
- app/templates/admin/userManagement.html

### DIFF
--- a/app/templates/admin/userManagement.html
+++ b/app/templates/admin/userManagement.html
@@ -133,7 +133,7 @@
             <form action="/admin/updateProgramInfo/{{program.id}}" method="POST">
                 <tr class="p-1">
                     <td> <input type="text" name="Name" value ="{{program.programName}}" required> </td>
-                    <td><input type="text" name="Email" value="{{program.emailReplyTo}}" required></td>
+                    <td><input type="text" name="Email" value="" required></td>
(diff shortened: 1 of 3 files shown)


app/templates/admin/userManagement.html
```diff
@@ -134,7 +134,7 @@ <h3 class="accordion-header" id="headingThree">
                 <tr class="p-1">
                     <td> <input type="text" name="Name" value ="{{program.programName}}" required> </td>
                     <td><input type="text" name="Email" value="{{program.emailReplyTo}}" required></td>
-                    <td><input type="text" name="Sender" value="{{program.emailSenderName}}" required></td>
+                    <td><input type="text" name="Sender" value="" required></td>
```
AndersonStettner: Now the input box is always empty after the name is updated. We want the box to be empty when they haven't inserted a name yet but if they have one in there it should show them as the sender.

AndersonStettner (dismissed):

app/templates/admin/userManagement.html
BrianRamsay: It also shouldn't be required. They might just want to update the program name

BrianRamsay (approved):

Issue #602, opened 2022-07-26
Need Modal Warning For Deleting Events


Boorclark:
When deleting an event, there is no popup for "verification" that this is what the user wants to do.

Issue #593, opened 2022-07-25
Admin Edit Event tab stacks


AndersonStettner:
When an admin goes to edit an event on mobile or on a smaller screen all of the tabs stack on top of each other yet if they are either of the other two tabs they are side-by-side. We want them to always be consistent so the tabs should either always be stacked or side-by-side on smaller/squished screens.
 
![image](https://user-images.githubusercontent.com/26262045/180815377-6f8e86cf-ced4-44d6-8335-aa09c947e885.png)
![image](https://user-images.githubusercontent.com/26262045/180815402-1955a76f-0d1d-4633-89b5-3fc0ec3fb7af.png)

PR #604, opened 2022-07-26
Showed modal on delete button click to confirm and moved delete button

bledsoef:
Fixes issues #602 and #593 
Showed modal on delete button click to confirm and moved delete button to bottom with save to fix the issue where tabs are inconsistent.

### DIFF
--- a/app/static/js/createEvents.js
+++ b/app/static/js/createEvents.js
@@ -168,5 +168,8 @@ $(document).ready(function() {
      $(this).closest("tr").remove();
   });
 
+  $('#deleteEventModal').on('click', function() {
+      $("#deleteWarning").modal("show")
+  })
 
 });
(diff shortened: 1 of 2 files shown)


app/static/js/createEvents.js
```diff
@@ -168,5 +168,8 @@ $(document).ready(function() {
      $(this).closest("tr").remove();
   });
 
+  $('#deleteEventModal').on('click', function() {
+      $("#deleteWarning").modal("show")
+  })
```
BrianRamsay: I think you can have a button trigger a modal automatically with something bootstrappy in the html. That lets us avoid a little javascript

BrianRamsay (changes requested):

app/static/js/createEvents.js
bledsoef: did the bootsrappy thing

BrianRamsay (approved):

Issue #383, opened 2022-07-06
Allow multiple file uploads for events and email attachments


BrianRamsay:
Users should be allowed to attach multiple files to events and emails. We should also implement a little bit more validation over the files themselves (don't worry about file contents validation). See: https://blog.miguelgrinberg.com/post/handling-file-uploads-with-flask

martinj2-dot:
By file validation do you mean choose what types of files we are allowing users to upload using accept?

martinj2-dot:
Relevant file paths:

controllers/events/routes.py
templates/emailModal.html

Notes: 
"multiple" attribute allows multiple files to be sent, need to fix controller to handle multiple files.

bledsoef:
PR's merged, closing issue.

PR #606, opened 2022-07-26
Issue#383 allow multiple attachments

qasema:
fixes issue #383 handles attachments for emails and events. A new data table (EventFile) and event file sub-directories were added.


app/config/default.yml
```diff
@@ -19,7 +19,9 @@ tracy:
     password: "password"
 
 email:
-  attachment_path: 'app/static/files/attachments'
+  email_attachment_path: 'app/static/files/emailattachments'
+event:
```
BrianRamsay: Let's put these both under a `files:` heading

app/controllers/admin/routes.py
```diff
@@ -210,3 +217,13 @@ def adminLogs():
...
     else:
         abort(403)
+
+@admin_bp.route("/deleteFile", methods=["POST"])
+def deleteFIle():
+    fileData= request.form
+    eventfile=FileHandler()
+    eventfile.deleteFile(fileData["fileId"],fileData["eventId"])
+    print(fileData["fileId"])
+    print(fileData["eventId"])
```
BrianRamsay: Remove debug output

app/controllers/admin/routes.py
```diff
@@ -72,16 +74,17 @@ def createEvent(templateid, programid=None):
     # Get the data for the form, from the template or the form submission
     eventData = template.templateData
     if request.method == "POST":
+        attachmentFiles = request.files.getlist("attachmentObject")
+        eventData = request.form.copy()
```
BrianRamsay: This line is redundant with the line below it

BrianRamsay: And also will stomp on the template data

app/controllers/admin/routes.py
```diff
@@ -147,7 +153,8 @@ def eventDisplay(eventId):
                                 futureTerms=futureTerms,
                                 isPastEvent = isPastEvent,
                                 userHasRSVPed = userHasRSVPed,
-                                isProgramManager = isProgramManager)
+                                isProgramManager = isProgramManager,
+                                paths = paths)
```
BrianRamsay: I would call this filepaths in the jinja template

app/logic/emailHandler.py
```diff
@@ -161,25 +162,26 @@ def getAttachmentFullPath(self):
...
             pass
 
     def store_sent_email(self, subject, template_id):
         """ Stores sent email in the email log """
         date_sent = datetime.now()
 
-        attachmentName = ''
-        if self.attachment_file:
-            attachmentName = self.attachment_file.filename
+        attachmentName = []
```
BrianRamsay: This is plural now

app/logic/emailHandler.py
```diff
@@ -208,6 +210,7 @@ def send_email(self):
             if self.program_ids[0].emailSenderName:
                 defaultEmailInfo["senderName"] = self.program_ids[0].emailSenderName
 
+        self.store_sent_email(subject, template_id)
```
BrianRamsay: We want to store this after sending, so if it breaks during sending it doesn't show up as a sent email

app/logic/fileHandler.py
```diff
@@ -0,0 +1,65 @@
...
+
+from flask import redirect, url_for
+from app import app
+from app.models.eventFile import EventFile
+
+class FileHandler:
+    def __init__(self,files=None):
+        self.files=files
+        self.path= app.config['event']['event_attachment_path']
+
```
BrianRamsay: Let's get rid of some of the whitespace here

app/logic/fileHandler.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        try:
+            # tries to create the full path of the files location and passes if
+            # the directories already exist or there is no attachment
+            if eventId:
+                filePath=(os.path.join(self.path, str(eventId), newfile.filename))
+                os.mkdir(self.path+"/"+ str(eventId))
+            else:
+                filePath=(os.path.join(self.path, newfile.filename))
+                os.mkdir(self.path+"/"+ str(eventId))
```
BrianRamsay: This line is the same in both cases. If we move it after the conditional we won't have to repeat it

app/logic/fileHandler.py
```diff
@@ -0,0 +1,65 @@
...
+        except AttributeError:  # will pass if there is no attachment to save
+            pass
+        except FileExistsError:
+            pass
+        return filePath
+
+    def saveFile(self, eventId):
```
BrianRamsay: Since we can handle multiple files this name should be plural `saveFiles`

app/logic/fileHandler.py
```diff
@@ -0,0 +1,65 @@
...
+        """ Saves the attachment in the app/static/files/eventattachments/ directory """
+
+
+        try:
+            for file in self.files:
+                if not EventFile.select().where(EventFile.event == eventId, EventFile.fileName == file.filename).exists():
```
BrianRamsay: Let's store this query result in a variable so the conditional is a little clearer

app/logic/fileHandler.py
```diff
@@ -0,0 +1,65 @@
...
+                    EventFile.create(event = eventId, fileName = file.filename)
+                    file.save(self.getFileFullPath(eventId, file)) # saves attachment in directory
```
BrianRamsay: Since these parameters are keyword parameters in the method, we should use the keywords here.

app/logic/fileHandler.py
```diff
@@ -0,0 +1,65 @@
...
+        except AttributeError: # will pass if there is no attachment to save
```
BrianRamsay, line 40: I don't understand this exception case anymore. Is there a chance that the files in the list wouldn't exist?

app/logic/fileHandler.py
```diff
@@ -0,0 +1,65 @@
...
+        for file in files:
+            pathDict[file.fileName] = ((self.path+"/"+ str(eventId) +"/"+ file.fileName)[3:], file)
+        return pathDict
+
+    def deleteFile(self, fileId, eventId):
+        """
+        Deletes attachmant from the app/static/files/eventattachments/ directory
+        """
+        try:
+            File = EventFile.get_by_id(fileId)
```
BrianRamsay: Our variable names should start lowercase

app/logic/fileHandler.py
```diff
@@ -0,0 +1,65 @@
...
+            return False
+            pass
+
+    def retrievePath(self,files, eventId = None):
+        pathDict={}
+        for file in files:
+            pathDict[file.fileName] = ((self.path+"/"+ str(eventId) +"/"+ file.fileName)[3:], file)
+        return pathDict
+
+    def deleteFile(self, fileId, eventId):
```
BrianRamsay: Let's indicate that this method is only for event files by naming it something like `deleteEventFile` or `deleteFileForEvent`. We should probably also rename the matching `saveFile` method

app/models/eventFile.py
```diff
@@ -0,0 +1,6 @@
+from app.models import*
+from app.models.event import Event
+
+class EventFile(baseModel):
+    event = ForeignKeyField(Event)
+    fileName = CharField()
```
BrianRamsay, line 6: Do we want to use the default peewee id, or should we make this a composite primary key?

app/static/js/createEvents.js
```diff
@@ -168,5 +169,23 @@ $(document).ready(function() {
      $(this).closest("tr").remove();
   });
 
+  $(".removeAttachment").on("click", function(){
 
+    let fileId=  $(this).data("id")
+      let fileData = {fileId : fileId,
+                      eventId:this.id}
+      $("#attachment_"+fileId).remove()
```
BrianRamsay: I wouldn't remove this until the ajax request was successful. If it fails for some reason, we don't want that row to disappear

app/templates/admin/createEvent.html
```diff
@@ -1,15 +1,28 @@
...
+
 {% block scripts %}
   {{super()}}
   <script type="module" src="/static/js/createEvents.js"></script>
   <script src="//cdnjs.cloudflare.com/ajax/libs/timepicker/1.3.5/jquery.timepicker.min.js"></script>
+  <script src="https://cdn.jsdelivr.net/gh/kartik-v/bootstrap-fileinput@5.5.0/js/plugins/buffer.min.js" type="text/javascript"></script>
+  <script src="https://cdn.jsdelivr.net/gh/kartik-v/bootstrap-fileinput@5.5.0/js/plugins/filetype.min.js" type="text/javascript"></script>
+  <script src="https://cdn.jsdelivr.net/gh/kartik-v/bootstrap-fileinput@5.5.0/js/plugins/piexif.min.js" type="text/javascript"></script>
+  <script src="https://cdn.jsdelivr.net/gh/kartik-v/bootstrap-fileinput@5.5.0/js/plugins/sortable.min.js" type="text/javascript"></script>
+  <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/js/bootstrap.bundle.min.js" crossorigin="anonymous"></script>
```
BrianRamsay: This looks like we are including a duplicate copy of bootstrap. We already have bootstrap included in base.html

app/templates/admin/createEvent.html
```diff
@@ -1,15 +1,28 @@
...
+
 {% endblock %}
 
 {% block styles %}
-  {{super()}}
-  <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/timepicker/1.3.5/jquery.timepicker.min.css">
+{{super()}}
+<link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/timepicker/1.3.5/jquery.timepicker.min.css">
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css" crossorigin="anonymous">
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap-icons@1.5.0/font/bootstrap-icons.min.css" crossorigin="anonymous">
```
BrianRamsay: Same deal with the bootstrap css files

app/templates/admin/createEvent.html
```diff
@@ -146,6 +160,17 @@ <h1 id="title">{{page_title}}</h1>
...
+        <div class="d-none" id="recurringTableDiv">
+          <table class="table" id="recurringEventsTable" name="recurringEventsTable">
+            <thead><tr><th>Event Name</th><th>Date</th></tr></thead>
+            <tbody></tbody>
+          </table>
+        </div>
+        <div class="d-none">
+            {{attachment.fileName}}
+        </div>
+        {% endfor %}
```
BrianRamsay: I'm confused by this block. How are attachments related to recurring events?

BrianRamsay (changes requested):

BrianRamsay (changes requested):
Nice work getting the plugin working!

I think we need to do some workflow improvements though. You should be able to have the existing files displayed in the same widget without having to maintain your own list. Also, if you drag and drop files separately, they should all show up in the list to upload. I also wonder if we can get rid of the Upload button - it's a little confusing to me to have it show up as well as having a Submit button on the page.

There are some examples here https://plugins.krajee.com/file-advanced-usage-demo

BrianRamsay (approved):
Approving. We will improve the user experience by switching the plugin to ajax mode with the next round of changes.

Issue #585, opened 2022-07-22
Events List appearance 


AndersonStettner:
Currently, when the Events List page gets narrower/is on mobile, the tabs bunch together. 
![image](https://user-images.githubusercontent.com/26262045/180568375-83aa0cfa-6fd4-44e4-8fa8-4d46ba1bdd82.png)

We want them to collapse into a column instead.

PR #608, opened 2022-07-26
Makes the event list tabs responsive to different screen sizes

mohamudm1:
Solves issue #585 

Changes made:
Used bootstrap classes to specify how the tabs respond accordingly on different screen sizes

### DIFF
--- a/app/templates/events/event_list.html
+++ b/app/templates/events/event_list.html
@@ -32,19 +32,19 @@
 </div>
 <div class="row">
   <ul class="nav nav-tabs nav-fill mx-3 mb-3" id="pills-tab" role="tablist">
-  <li class="col-lg-3 col-sm-6 col-12 nav-item" role="presentation">
+  <li class="col-md-3 col-sm-12 col-12 nav-item" role="presentation">
     <button class="nav-link active" id="studentLedEvents"
     data-bs-toggle="pill" data-bs-target="#pills-student-led" type="button" role="tab" aria-controls="pills-student-led" aria-selected="true">Student Led Service</button>
   </li>
-  <li class="col-lg-3 col-sm-6 col-12 nav-item" role="presentation">
+  <li class="col-md-3 col-sm-12 col-12 nav-item" role="presentation">
     <button class="nav-link" id="trainingEvents"
     data-bs-toggle="pill" data-bs-target="#pills-training" type="button" role="tab" aria-controls="pills-training" aria-selected="false">Training and Education</button>
   </li>
-  <li class="col-lg-3 col-sm-6 col-12 nav-item" role="presentation">
+  <li class="col-md-3 col-sm-12 col-12 nav-item" role="presentation">
     <button class="nav-link" id="bonnerScholarsEvents"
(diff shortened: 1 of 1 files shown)


app/templates/events/event_list.html
```diff
@@ -30,23 +30,22 @@ <h1 class="text-center mt-5">Events List for {{selectedTerm.description}}</h1>
     </div>
   </div>
 </div>
-<div>
+<div class="row">
   <ul class="nav nav-tabs nav-fill mx-3 mb-3" id="pills-tab" role="tablist">
-  <li class="nav-item" role="presentation">
+  <li class="col-lg-3 col-md-6 col-sm-12 col-xs-12 nav-item" role="presentation">
```
BrianRamsay: We don't need to specify all of these, since we only need to break at a certain screensize, and they should all be equal width. It should work to give each of them `col-sm`, `col-md`, or `col-lg`, depending on which one of those causes the break at the right place.

BrianRamsay (changes requested):

app/templates/events/event_list.html
```diff
@@ -30,23 +30,22 @@ <h1 class="text-center mt-5">Events List for {{selectedTerm.description}}</h1>
     </div>
   </div>
 </div>
-<div>
+<div class="row">
   <ul class="nav nav-tabs nav-fill mx-3 mb-3" id="pills-tab" role="tablist">
-  <li class="nav-item" role="presentation">
+  <li class="col-lg-3 col-sm-6 col-12 nav-item" role="presentation">
```
BrianRamsay: I think we should only break from 3 or 4 tabs per row to 1. This would be my suggested class list `text-nowrap col-12 col-md-3`. And I suppose we should decide if that should be `col-md-4` if we only are showing 3 tabs (no Bonners)

BrianRamsay (changes requested):

app/templates/events/event_list.html
```diff
@@ -30,18 +30,18 @@ <h1 class="text-center mt-5">Events List for {{selectedTerm.description}}</h1>
...
     data-bs-toggle="pill" data-bs-target="#pills-student-led" type="button" role="tab" aria-controls="pills-student-led" aria-selected="true">Student Led Service</button>
   </li>
-  <li class="nav-item" role="presentation">
+  <li class="col-md-3 col-sm-12 col-12 nav-item" role="presentation">
     <button class="nav-link" id="trainingEvents"
     data-bs-toggle="pill" data-bs-target="#pills-training" type="button" role="tab" aria-controls="pills-training" aria-selected="false">Training and Education</button>
   </li>
   {% if user.isCeltsAdmin or user.isCeltsStudentStaff or user.isBonnerScholar %}
-    <li class="nav-item" role="presentation">
+    <li class="col-md-3 col-sm-12 col-12 nav-item" role="presentation">
```
BrianRamsay: The `col-sm-12` class is redundant here, because it's already size 12 (from `col-12`). It goes from smallest to largest, and you need the class where the breakpoint is. In this case, at medium.

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #614, opened 2022-07-26
Display Next Event Data of Recurring Events

Lyisus:
Waiting on PR #590 
In the event view page, next to the sentence that specifies if an event is a recurring event, there is a short sentence showing when the next event in the series in, as well as a link to the event view page of said next event.

### DIFF
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -156,8 +156,9 @@ def eventDisplay(eventId):
         eventData["startDate"] = event.startDate.strftime("%m/%d/%Y")
         # List below is to identify the last event in the series
         eventSeriesList = list(Event.select().where(Event.recurringId == event.recurringId))
-        if event.recurringId and not eventSeriesList[-1] == event:
-            eventData["nextRecurringEvent"] = Event.select().where(Event.id > event.id, Event.recurringId == event.recurringId).get()
+        eventIndex = eventSeriesList.index(event)
+        if event.recurringId and len(eventSeriesList) != (eventIndex + 1):
+            eventData["nextRecurringEvent"] = eventSeriesList[eventIndex + 1]
         programManager = ProgramManager.get_or_none(program=program)
         userParticipatedEvents = getUserParticipatedEvents(program, g.current_user, g.current_term)
         return render_template("eventView.html",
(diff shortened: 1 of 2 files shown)


app/controllers/admin/routes.py
```diff
@@ -154,6 +154,10 @@ def eventDisplay(eventId):
         eventData['timeStart'] = event.timeStart.strftime("%-I:%M %p")
         eventData['timeEnd'] = event.timeEnd.strftime("%-I:%M %p")
         eventData["startDate"] = event.startDate.strftime("%m/%d/%Y")
+        # List below is to identify the last event in the series
+        eventSeriesList = list(Event.select().where(Event.recurringId == event.recurringId))
+        if event.recurringId and not eventSeriesList[-1] == event:
+            eventData["nextRecurringEvent"] = Event.select().where(Event.id > event.id, Event.recurringId == event.recurringId).get()
```
BrianRamsay: You already have the list of recurring events, you don't need to select from the database again to find the next one. Try `list.index()` here, and for your conditional

BrianRamsay (changes requested):

app/controllers/admin/routes.py
```diff
@@ -154,6 +154,11 @@ def eventDisplay(eventId):
         eventData['timeStart'] = event.timeStart.strftime("%-I:%M %p")
         eventData['timeEnd'] = event.timeEnd.strftime("%-I:%M %p")
         eventData["startDate"] = event.startDate.strftime("%m/%d/%Y")
+        # List below is to identify the last event in the series
```
BrianRamsay: This comment says it is getting the last event, but it is really getting the next event

app/templates/eventView.html
```diff
@@ -49,7 +49,14 @@ <h3><br>Description</h3>
                         <p>This event does not require RSVP.</p>
                     {% endif %}
                     {% if eventData.recurringId %}
-                        <p>This event is a recurring event.</p>
+                        <p>This event is a recurring event.
+                            {% if eventData.nextRecurringEvent %}
+                            The next event in the series occurs on {{eventData.nextRecurringEvent.startDate.strftime("%m/%d/%Y")}}.
```
BrianRamsay: Since this is in a sentence, let's make the date a date in english without the year. Like "occurs on December 5."

BrianRamsay (changes requested):

BrianRamsay (approved):
Looks good!

Issue #583, opened 2022-07-22
Replace Facilitator on event


AndersonStettner:
Facilitator should be replaced with a contact name and contact info for the program.

bledsoef:
PR merged, closing issue

Issue #616, opened 2022-07-26
Remove Program Manager from event view


BrianRamsay:
The program manager info should not be on the event view. We will be showing the contact information for the event there, and the program managers can stay private.

bledsoef:
We will take care of this as we add the contact info to the view page.

bledsoef:
PR merged, closing issue

PR #619, opened 2022-07-27
Replace Facilitator Search 583

Boorclark:
Fixes Issues #583 and #616

Changes Made:
 - Got rid of all instances of facilitators throughout the application
 - Added "contactName" and "contactEmail" into event model
 - Change program model: emailSenderName --> contactName  &  senderEmail --> contactEmail
 - Changed Create Event and View Event view 
 - Removed program manager
 - Updated tests


app/controllers/admin/routes.py
```diff
@@ -120,17 +121,20 @@ def eventDisplay(eventId):
     except DoesNotExist as e:
         print(f"Unknown event: {eventId}")
         abort(404)
-
     eventData = model_to_dict(event, recurse=False)
     if request.method == "POST": # Attempt to save form
         eventData = request.form.copy()
+        print(eventData)
```
BrianRamsay: debug output

app/templates/admin/createEvent.html
```diff
@@ -144,41 +144,25 @@ <h1 id="title">{{page_title}}</h1>
...
+          {% else %}
+              <div class="form-group mb-4">
+                <label class="form-label" for='inputContactName'><strong>Contact Name</strong></label>
+                <input class="form-control" id='inputContactName' value="" placeholder="Enter name" name="contactName" required>
+              </div>
+              <div class="form-group mb-4">
+                <label class="form-label" for='inputContactEmail'><strong>Contact Email</strong></label>
+                <input class="form-control" id='inputContactEmail' value="" placeholder="Enter email" name="contactEmail" required>
+              </div>
+          {% endif %}
```
BrianRamsay: We definitely don't want to repeat all these blocks. Just make sure contact_info exists, and you can just use that dictionary. Or better yet, just use eventData like everything else

app/templates/admin/userManagement.html
```diff
@@ -126,15 +126,15 @@ <h3 class="accordion-header" id="headingThree">
...
             </tr>
             {% for program in programs %}
             <form action="/admin/updateProgramInfo/{{program.id}}" method="POST">
                 <tr class="p-1">
-                    <td> <input type="text" name="Name" value ="{{program.programName}}" required> </td>
-                    <td><input type="text" name="Email" value="{{program.emailReplyTo}}"></td>
-                    <td><input type="text" name="Sender" value="{{program.emailSenderName}}"></td>
+                    <td> <input type="text" name="programName" value ="{{program.programName}}" required> </td>
+                    <td><input type="text" name="contactEmail" value="{{program.contactEmail}}" required></td>
+                    <td><input type="text" name="contactName" value="{{program.contactName}}" required></td>
```
BrianRamsay: These are not required fields

BrianRamsay (changes requested):

Boorclark:
3bf50ff5ace3b828fbefe0ef08633d770213c7ed  -- Previous branch head

app/controllers/admin/routes.py

app/templates/admin/createEvent.html
Boorclark: resolved

app/templates/admin/userManagement.html
Boorclark: resolved

app/models/event.py
```diff
@@ -17,6 +17,8 @@ class Event(baseModel):
     startDate = DateField()
     endDate = DateField(null=True)
     recurringId = IntegerField(null=True)
+    contactEmail = CharField()
+    contactName = CharField()
```
BrianRamsay: These should be nullable

app/templates/eventView.html
```diff
@@ -57,49 +57,25 @@ <h3><br>Description</h3>
...
-                                    <li>{{facilitator.firstName}} {{facilitator.lastName}}</li>
-                                    {% endfor %}
-                              </ul>
-                              {% else %}
-                                  <p>This event has no facilitators.</p>
-                              {% endif %}
-                    </div>
-                    <div>
-                        <div>
+                        <h3><br>Contact Information:</h3>
```
BrianRamsay: Move the break tag before the header (if it is still necessary)

BrianRamsay: We should handle the case where there isn't any contact info saved for the event. (probably just don't show the header)

BrianRamsay (changes requested):
We should default contact info to the program name and email if it is empty when creating an event, and to an admin name and email if the program doesn't have anything.

app/models/event.py
Boorclark: resolved

app/templates/eventView.html
Boorclark: resolved

Boorclark: resolved

Boorclark:
@BrianRamsay We made the "default admin" be the CELTS director for now. We were going to make it be the CELTS email and number but it doesn't seem like they actually have one. Personally, I think it would be better to leave the placeholders for these blank if nobody is assigned so the director's phone isn't blown up.

app/controllers/admin/routes.py
```diff
@@ -75,9 +73,18 @@ def createEvent(templateid, programid=None):
...
 
+        if program.contactName and  program.contactEmail:
+            # TODO need to handle the multiple programs case
+            eventData['contactName'] = program.contactName
+            eventData['contactEmail'] = program.contactEmail
+
+        else:
+            # sets contact name and email to CELTS director (admin) if there is none set
+            eventData['contactName'] = 'Ashley Cochrane'
+            eventData['contactEmail'] = 'ashley_cochrane@berea.edu'
```
BrianRamsay: There is already a `celts_admin_contact` address in the config. Let's just use that.

BrianRamsay: Also, having this here will overwrite whatever they put in, won't it? because this is happening after eventData is updated with their submitted form data. We should only be filling these in when we first show them the create event page.

BrianRamsay (changes requested):

app/controllers/admin/routes.py
```diff
@@ -74,10 +72,19 @@ def createEvent(templateid, programid=None):
...
 
+        if program.contactName and program.contactEmail:
+            # TODO need to handle the multiple programs case
+            eventData['contactName'] = program.contactName
+            eventData['contactEmail'] = program.contactEmail
+
+        else:
+            # sets contact name and email to CELTS director (admin) if there is none set
+            eventData['contactName'] = "CELTS"
+            eventData['contactEmail'] = app.config['celts_admin_contact']
```
BrianRamsay: We want this general contact to be available for all events, not just ones with a program, so it needs to happen outside of the program check. Also, let's make the name "CELTS Admin"

app/models/program.py
```diff
@@ -22,5 +22,6 @@ def url(self):
                 7: baseUrl + "volunteer-opportunities/teen-mentoring-program/",
                 8: baseUrl + "volunteer-opportunities/hispanic-outreach-project/",
                 9: baseUrl + "volunteer-opportunities/people-who-care-program/",
+                10: baseUrl + "hunger-initiatives/hunger-hurts-food-drive/",
```
BrianRamsay: It looks like you have undone a change that was in development. food drive and empty bowls have been combined into hunger initiatives

app/templates/admin/createEvent.html
```diff
@@ -144,41 +144,14 @@ <h1 id="title">{{page_title}}</h1>
...
-            </table>
-           </div>
+          <div class="form-group mb-4">
+            <label class="form-label" for='inputContactName'><strong>Contact Name</strong></label>
+            <input class="form-control" id='inputContactName' value="{{eventData.contactName}}"  placeholder="Enter name"name="contactName">
+          </div>
+          <div class="form-group mb-4">
+            <label class="form-label" for='inputContactEmail'><strong>Contact Email</strong></label>
+            <input class="form-control" id='inputContactEmail' value="{{eventData.contactEmail}}" placeholder="Enter email" name="contactEmail">
+          </div>
```
BrianRamsay, line 162: Let's let these stack next to each other on the page, if there is room.

BrianRamsay (changes requested):

app/controllers/admin/routes.py

app/controllers/admin/routes.py

app/models/program.py
Boorclark: resolved

app/templates/admin/createEvent.html
Boorclark: resolved

app/controllers/admin/routes.py
```diff
@@ -74,9 +72,15 @@ def createEvent(templateid, programid=None):
     if request.method == "POST":
         eventData.update(request.form.copy())
 
-    if program:
-        # TODO need to handle the multiple programs case
+    if program and request.method == "GET":
         eventData["program"] = program
+        eventData['contactName'] = "CELTS Admin"
+        eventData['contactEmail'] = app.config['celts_admin_contact']
```
BrianRamsay, line 85: They are still dependent on a program existing. You'll have to split up your conditional so we can handle get request and program vs get request and no program

BrianRamsay (changes requested):
Just one thing!

BrianRamsay:
congratulations! merge conflicts

app/controllers/admin/routes.py
```diff
@@ -169,13 +170,16 @@ def eventDisplay(eventId):
...
                                 isPastEvent = isPastEvent,
                                 userHasRSVPed = userHasRSVPed,
                                 programTrainings = userParticipatedEvents,
+<<<<<<< HEAD
+                                isProgramManager = isProgramManager)
+=======
                                 programManager = programManager,
                                 isProgramManager = isProgramManager,
                                 filepaths = filepaths)
+>>>>>>> d23f0881b1c464e2f2dd981dd19e303684f4a70d
```
BrianRamsay: *sad trombone*

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #565, opened 2022-07-22
Background Check: Failed 


AndersonStettner:
The title of the column should be "Passed" and no should be a viable entry for cases where someone might fail a background check.

bledsoef:
PR merged, closing issue.

PR #621, opened 2022-07-27
Fixed conditionals and column name

Boorclark:
Fixes Issue #565 

Changes Made:
 - Changes JS conditionals 
 - Completed --> Passed

### DIFF
--- a/app/logic/volunteers.py
+++ b/app/logic/volunteers.py
@@ -99,13 +99,16 @@ def setUserBackgroundCheck(user, bgType, checkPassed, dateCompleted):
     user = User.get_by_id(user)
     deleteInstance = BackgroundCheck.delete().where(BackgroundCheck.user==user, BackgroundCheck.type==bgType)
     deleteInstance.execute()
-    if not dateCompleted:
-        dateCompleted = None
-    update = BackgroundCheck.create(user=user, type=bgType, passBackgroundCheck=checkPassed, dateCompleted=dateCompleted)
-    if bool(checkPassed):
-        createLog(f"Marked {user.firstName} {user.lastName}'s background check for {bgType} as completed.")
+    if checkPassed == '' and dateCompleted == '':
+        createLog(f"Marked {user.firstName} {user.lastName}'s background check for {bgType} as empty.")
     else:
-        createLog(f"Marked {user.firstName} {user.lastName}'s background check for {bgType} as incompleted.")
+        if not dateCompleted:
+            dateCompleted = None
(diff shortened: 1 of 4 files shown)


BrianRamsay (changes requested):
I think a better fix is to have 3 options in the select box, because we need to be able to tell a difference between No and never updated. Whether they are choosing yes or no, we still need a date.

app/logic/volunteers.py
```diff
@@ -99,13 +99,16 @@ def setUserBackgroundCheck(user, bgType, checkPassed, dateCompleted):
...
+        createLog(f"Marked {user.firstName} {user.lastName}'s background check for {bgType} as empty.")
     else:
-        createLog(f"Marked {user.firstName} {user.lastName}'s background check for {bgType} as incompleted.")
+        if not dateCompleted:
+            dateCompleted = None
+        update = BackgroundCheck.create(user=user, type=bgType, passBackgroundCheck=int(checkPassed), dateCompleted=dateCompleted)
+        if bool(checkPassed):
+            createLog(f"Marked {user.firstName} {user.lastName}'s background check for {bgType} as completed.")
+        else:
+            createLog(f"Marked {user.firstName} {user.lastName}'s background check for {bgType} as incompleted.")
```
BrianRamsay: They are marking them as "Passed" and "Not Passed" now, not complete or incomplete

app/static/js/volunteerProfile.js
```diff
@@ -87,10 +87,11 @@ $(document).ready(function(){
...
     let checkPassed = $("[data-id=" + bgCheckType + "]").val()
 
-    if (checkPassed == 0 && bgDate != '' ) {
-        displayMessage("Y/N<br>Empty!", "danger")
+    if (checkPassed == '' && bgDate != '') {
+        displayMessage("Passed<br>Empty!", "danger")
         return
-    } else if(bgDate == '' && checkPassed != 0) {
+    }
+    if (checkPassed && bgDate == '' ) {
```
BrianRamsay: Does this work with checkPassed = 0?

BrianRamsay (changes requested):

app/logic/volunteers.py
bledsoef: resolved

app/static/js/volunteerProfile.js
bledsoef: resolved

BrianRamsay (approved):

Issue #584, opened 2022-07-22
Check Admin Logs 


AndersonStettner:
Make sure the Admin Logs is saving the event creation/deletion, email logs, SLC creation/deletion, and SLC proposals status changes. If it isn't, make it.

bledsoef:
PR merged, closing issue

PR #625, opened 2022-07-28
Admin logs saves various actions across the interface 

mohamudm1:
Solves issue #584 Changes made:
 
Admin log saves event creation properly.


app/logic/courseManagement.py
```diff
@@ -77,5 +78,44 @@ def updateCourse(courseData):
     CourseInstructor.delete().where(CourseInstructor.course == course).execute()
     for instructor in instructorList:
         CourseInstructor.create(course=course, user=instructor)
+    createLog(f"Created SLC proposal: {courseData['courseName']}")
+    return course
+
+def updateCourse(courseData):
```
BrianRamsay: Where did this function come from? It's suspicious if your merge conflict result is adding a whole new method, and not removing one from somewhere else. You have two `updateCourse` methods in the file now

BrianRamsay (changes requested):

app/logic/courseManagement.py
```diff
@@ -49,33 +50,36 @@ def updateCourse(courseData):
...
-        CourseInstructor.create(course=course, user=instructor)
+        CourseInstructor.delete().where(CourseInstructor.course == course).execute()
+        for instructor in instructorList:
+            CourseInstructor.create(course=course, user=instructor)
+        createLog(f"Edited SLC proposal: {courseData['courseName']}")
+    except:
+        return False;
 
-    return Course.get_by_id(course.id)
+    return course
```
BrianRamsay: You've broken a recent fix in development by changing this line (by taking the old function from the merge conflict, really, instead of the new one). update course needs to return a course object with the updated data, not the original one that was passed in

BrianRamsay (changes requested):

app/logic/courseManagement.py
```diff
@@ -49,33 +50,36 @@ def updateCourse(courseData):
...
 
-    CourseInstructor.delete().where(CourseInstructor.course == course).execute()
-    for instructor in instructorList:
-        CourseInstructor.create(course=course, user=instructor)
+        CourseInstructor.delete().where(CourseInstructor.course == course).execute()
+        for instructor in instructorList:
+            CourseInstructor.create(course=course, user=instructor)
+        createLog(f"Edited SLC proposal: {courseData['courseName']}")
+    except:
+        return False;
```
BrianRamsay: Found one last thing (probably). The development version of this function got rid of the blanket try/except altogether. If you are going to add it back in, you need to print the exception so we know what went wrong.

BrianRamsay (changes requested):

BrianRamsay (dismissed):

BrianRamsay (approved):

PR #629, opened 2022-08-02
Refactored Service Transcript Page

Boorclark:
### **Changes Made:**
 - Refactored the view for viewing the service transcript
 - Rounded the hours for events to the hundredth place so there are not huge numbers
 - Fixed formatting for the print page
 - Added a description for each program in the program model
 - Updated tests to account for deletion of getBonnersScholarsEvents() and getTrainingTranscript()

### **How To Test**
1.  Create events for multiple programs via _Create Events_
2.  In the _Track Volunteers_ tab, assign a user (khatts) to any number of hours for each event
3.  Navigate to Khatts profile via the _Student Search_
4.  In the top right, click _View Service Transcript_
5.  To test printing, navigate to the bottom of the Transcript page and select the blue _Print_ button


AndersonStettner:
The same SLC is showing up twice in the table
image

It does appear twice in the database so investigate why since it means your code is probably working as it should.
image

app/logic/transcript.py
```diff
@@ -10,72 +10,82 @@
...
-
     """
-
     eventHours = EventParticipant.select(fn.SUM(EventParticipant.hoursEarned)).where(EventParticipant.user == username).scalar()
     courseHours =  CourseParticipant.select(fn.SUM(CourseParticipant.hoursEarned)).where(CourseParticipant.user == username).scalar()
 
-    return (eventHours or 0) + (courseHours or 0)
+    allHours = {"totalEventHours": (eventHours or 0),
+                "totalCourseHours": (courseHours or 0),
+                "totalHours": (eventHours or 0) + (courseHours or 0)}
```
BrianRamsay, line 92: I like this use of `or`

app/models/program.py
```diff
@@ -24,3 +24,17 @@ def url(self):
...
+                        9: "People Who Care (PWC) helps to connect Berea College students with organizations and opportunities that promote change through advocacy, education, action, and direct community service. Volunteers may serve at local shelters, work with the Fair Trade University Campaign, or help to raise awareness about local issues like domestic violence, homelessness, fair trade, and AIDS awareness education.",
+                        10: "The Berea Food Bank is sponsored by the Berea Faith Community Outreach. Local churches, individuals and regional food distribution agencies are tapped to provide staple groceries for a full week, in proportion to the size of the family.",
+                        12: "Berea Tutoring provides an encouraging atmosphere for local students who need help in achieving academic success, and for college volunteers who want to learn more about teaching or volunteering. Our mission is to increase conceptual understanding in academic subject areas, enrich educational experiences, and build self-confidence by providing college-aged tutors to local school children."}
```
BrianRamsay: program 10 has gone away. id 1 is now Hunger Initiatives

app/templates/main/serviceTranscript.html
```diff
@@ -4,100 +4,75 @@
...
-      <th>Accrued Hours</th>
-    </tr>
+{% if programs %}
     {% for program in programs %}
-    <tr>
-      <td> {{program.program.programName}}</td>
-      <td> {{program.event.term.description}}</td>
-      <td class="w-25"> {{program.hoursEarned}}</td>
-    </tr>
+        {% if not (programs[program][0][1] == None and programs[program]|length == 1) %}
```
BrianRamsay: This is a confusing conditional

tests/code/test_transcripts.py
```diff
@@ -177,20 +144,21 @@ def testingProgram():
     adminName = "ramsayb2"
     testingProgramExist = getProgramTranscript(username)
     testingProgramNotExist = getProgramTranscript(adminName)
```
BrianRamsay: Let's name these variables according to what is in them, not according to what we expect to find. I think it will make the assertions easier to follow.

app/templates/main/serviceTranscript.html
```diff
@@ -4,100 +4,75 @@
...
-      <td class="w-25"> {{programEvent.hoursEarned}} </td>
-    </tr>
+{% if otherData %}
+  <div class="card-body">
+    <h5 class="card-title">Other Celts Sponsored Events</h5>
+    {% for term in otherData %}
+        <h6 class="card-subtitle mb-2 text-muted"> {{term[0]}} - {{term[1]}} hour(s)</h6>
     {% endfor %}
-  </table>
+    <p class="card-text">These are events that are not assosiated with a Celts program but are still sponsored by Celts.</p>
```
BrianRamsay: I'm not sure this disclaimer is necessary. The title is pretty self-explanatory

app/templates/main/serviceTranscript.html
```diff
@@ -4,100 +4,75 @@
...
+    <a class="d-print-none" href="#" class="card-link">More Information</a>
+  </div>
```
BrianRamsay: Let's add this data to the list with the other data, so we can use the same code to print out the cards for both

app/logic/transcript.py
```diff
@@ -10,72 +10,82 @@
 from peewee import DoesNotExist, fn
 
 
-#FIXME: Needs to break hours down by program and term, not just program
+def getOtherEventsTranscript(username):
```
BrianRamsay, line 13: I don't see a test for this one.

app/logic/transcript.py
```diff
@@ -10,72 +10,82 @@
...
+                    .join(Event)
+                    .join(ProgramEvent)
+                    .join(Program)
+                    .where(EventParticipant.user == username,
+                           Event.isTraining == False,
+                           Event.isAllVolunteerTraining == False,
+                           ((ProgramEvent.program == None) |
+                            (Program.isStudentLed == False) &
+                            (Program.isBonnerScholars == False)))
+                    .group_by(Event.term)
```
BrianRamsay, line 26: I'm not sure how this works with all the extra stuff in the Select clause, since you are only grouping by event.term. Also, can this just be as simple as `.where(ProgramEvent.program == None)`? The other programs should show up in the list of programs

app/logic/transcript.py
```diff
@@ -10,72 +10,82 @@
...
+                  )
+
+    otherEvents = []
+
+    for term in otherEventsData:
+        otherEvents.append([term.event.term.description, term.hoursEarned])
```
BrianRamsay: A good case for a list comprehension

BrianRamsay: Also, you are currently doing a separate query to get the term description for every entry in the list

app/logic/transcript.py
```diff
@@ -10,72 +10,82 @@
...
     """
     Returns a SLCourse query object containing all the training events for
     current user.
     """
+
     slCourses = (Course
         .select(CourseParticipant.hoursEarned, Course)
         .join(CourseParticipant, on=(Course.id == CourseParticipant.course))
-        .where(CourseParticipant.user == username).distinct())
+        .where(CourseParticipant.user == username))
```
BrianRamsay: We shouldn't assume that the hours for each course are in one entry. We might need to sum them up into one.

BrianRamsay (changes requested):

AndersonStettner:
It would be good to have a better display for when a user has not participated in any events or SL courses. 

image

app/templates/main/serviceTranscript.html
zachneill: Also make sure you use CELTS, all caps. Celts is not used.

app/logic/transcript.py
bledsoef: Fixed the extra query but a list comprehension would likely not be ideal unless we use something like `[x.append('a') or x for x in a]`

app/logic/transcript.py
bledsoef: I talked with Anderson about this and since there is currently no way to add students to a course as a participant, whenever that functionality is added we should probably merge the participant instances there if duplicated exist.

app/logic/transcript.py
BrianRamsay: I was thinking `return [[row.event.term.description, row.hoursEarned] for row in otherEventsData]`

app/logic/transcript.py
BrianRamsay: I think we want to keep them separate in the database, if they are entering them separately. It may not end up being an issue that needs to be fixed.

app/models/program.py
BrianRamsay: This isn't right anymore. 1-8 and 10 are the ones that should be there

BrianRamsay (changes requested):

app/models/program.py
```diff
@@ -31,3 +31,15 @@ def url(self):
...
+                        8: "The Hispanic Outreach Program (HOP) is a service-learning effort which brings together CELTS, several community organizations, and the Department of Foreign Languages at Berea College. HOP aims to build bridges among the Spanish-speaking and English-speaking residents of Madison County.",
+                        10: "Berea Tutoring provides an encouraging atmosphere for local students who need help in achieving academic success, and for college volunteers who want to learn more about teaching or volunteering. Our mission is to increase conceptual understanding in academic subject areas, enrich educational experiences, and build self-confidence by providing college-aged tutors to local school children."}
+        return descriptions[self.id]
```
BrianRamsay, line 50: We should make sure all the programs have a value here, even if empty, and that we have the same sanity check that the url property does to avoid an index error

app/templates/main/serviceTranscript.html
```diff
@@ -4,102 +4,73 @@
...
+                {% if item[1] != None%}
+                    <h6 class="card-subtitle mb-2 text-muted"> {{item[0]}} - {{item[1]|round(2)}} hour(s)</h6>
+                {% endif %}
+            {% endfor %}
+            {% if program != "CELTS Sponsored Events" %}
+                <p class="card-text">{{program.description}}</p>
+                <a class="d-print-none" href="{{program.url}}" class="card-link">More Information</a>
+            {% endif %}
+          </div>
+        {% endif %}
```
BrianRamsay, line 23: It feels a little messy in this block. I think it's because we assume a lot about what the data format might be, and what indicates an empty list or not, or the program vs non program card.

That hour check is odd to me. Is there a way to avoid the triple array index and that particular boolean expression in the conditional? Like, why would there be a single term in the array if there were no hours for it?

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #595, opened 2022-07-25
Events List: remove Invitation column title and have conditional on RSVP button 


AndersonStettner:
We want the right column of the table (the one that has the RSVP button in it) to have no title. Also, if the event does not require and RSVP then the user should not be able to see the button, event if it is disabled.

PR #633, opened 2022-08-02
removed invitation column title and added conditionals on RSVP button

gahimbaref:
#595 
- removed invitation column title
- added conditionals on RSVP button
- if an event is past, or doesn't require an RSVP, the RSVP button doesn't show at all instead of being disabled
- if an event requires an RSVP and is not past, the button shows up.

Files Changed:
- app/templates/eventView.html
- app/templates/events/event_list.html


app/templates/eventView.html
```diff
@@ -106,7 +106,7 @@ <h3>Program Trainings</h3>
               {% if (not isPastEvent) and (not isNewEvent)%}
                   {% if userHasRSVPed %}
                     <button type="submit" class="btn btn-danger" value="{{eventData.id}}" id="removeRsvpBtn">Remove RSVP</button>
-                  {%else%}
+                  {% elif (not isPastEvent) and (eventData.isRsvpRequired) %}
```
BrianRamsay: we know the first part will always be True, so we can leave it out of the expression

app/templates/events/event_list.html
```diff
@@ -65,7 +65,6 @@ <h1 class="text-center mt-5">Events List for {{selectedTerm.description}}</h1>
         {% if type != "bonner" %}
             <th>Event Program</th>
         {% endif %}
-        <th scope="col">Invitation</th>
```
BrianRamsay, line 68: We need to be careful that we are ending up with the same number of headers and columns in every case. Usually removing the header cell altogether isn't what you want.

app/templates/events/event_list.html
```diff
@@ -98,11 +97,9 @@ <h1 class="text-center mt-5">Events List for {{selectedTerm.description}}</h1>
...
               </td>
             {% else %}
-              {% if (event.isPast) %}
-                <td><button type="button" class="btn btn-primary" disabled>RSVP</button></td>
-              {% elif event.id in rsvpedEventsID %}
+              {% if event.id in rsvpedEventsID %}
                 <td><button type="button" class="btn btn-danger" onclick="removeRsvpForEvent({{event.id}})">Remove RSVP</button></td>
-              {% else %}
+              {% elif (not event.isPast) and (event.isRsvpRequired) %}
                 <td><button type="button" class="btn btn-primary" onclick="rsvpForEvent({{event.id}})">RSVP</button></td>
```
BrianRamsay: There always needs to be a cell, even if no button shows up in it, so the `<td>` and `</td>` should be outside of the conditionals

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #640, opened 2022-08-03
Issue#383 allow multiple attachments

martinj2-dot:
Fixed an issue with creating events. Issue 383

### DIFF
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -7,6 +7,7 @@ from dateutil import parser
 
 from app import app
 from app.models.program import Program
+from app.models.programManager import ProgramManager
 from app.models.event import Event
 from app.models.eventParticipant import EventParticipant
 from app.models.eventRsvp import EventRsvp
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -93,9 +93,8 @@ def createEvent(templateid, programid=None):
     if request.method == "POST":
         try:
             if fileDoesNotExist:
-                saveSuccess, validationErrorMessage = attemptSaveEvent(eventData)
-            else:
-                saveSuccess, validationErrorMessage = attemptSaveEvent(eventData, attachmentFiles)
+                attachmentFiles = None
+            saveSuccess, validationErrorMessage = attemptSaveEvent(eventData, attachmentFiles)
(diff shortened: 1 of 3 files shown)


BrianRamsay (dismissed):

app/logic/events.py
```diff
@@ -64,8 +64,11 @@ def attemptSaveEvent(eventData, attachmentFiles = None):
         return False, validationErrorMessage
 
     try:
-        events = saveEventToDb(newEventData)
-        if attachmentFiles:
+        if  attachmentFiles == None:
+            saveEventToDb(newEventData)
+            return True, ""
+        else:
+            events = saveEventToDb(newEventData)
```
BrianRamsay: Let's reorganize this so that the attachmentFiles check just decides whether we do the addfile loop. Because everything else is the same

app/controllers/admin/routes.py
```diff
@@ -76,6 +76,7 @@ def createEvent(templateid, programid=None):
     eventData = template.templateData
     if request.method == "POST":
         attachmentFiles = request.files.getlist("attachmentObject")
+        fileDoesNotExist = attachmentFiles[0].content_type == "application/octet-stream"
```
BrianRamsay, line 80: Could we just make attachmentFiles = None here if there are no files? We wouldn't need a conditional later

BrianRamsay (changes requested):

AndersonStettner (dismissed):

app/controllers/admin/routes.py
```diff
@@ -91,6 +93,8 @@ def createEvent(templateid, programid=None):
     # Try to save the form
     if request.method == "POST":
         try:
+            if fileDoesNotExist:
+                attachmentFiles = None
```
BrianRamsay: Let's just do this up where I made the comment, and avoid having to do a conditional here at all

app/logic/events.py
```diff
@@ -65,7 +65,7 @@ def attemptSaveEvent(eventData, attachmentFiles = None):
 
     try:
         events = saveEventToDb(newEventData)
-        if attachmentFiles:
+        if  attachmentFiles != None:
```
BrianRamsay: Did `if attachmentFiles` not work? Seems like it would, and that is preferred over a double negative

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #635, opened 2022-08-02
Make flash messages fade properly


BrianRamsay:
Currently, flash messages from `msgFlash()` fade properly, but the ones from the server do not. We need to add javascript so that the flash messages fade with the same delay as the others.

An easy way to trigger both types is on the admin settings page. The current term selector uses `msgFlash()`, the program settings update uses server `flash()`.

PR #666, opened 2022-08-04
added js to make message fade

thawpt:
- made html flashes fade using js

fixes issue #635

### DIFF
--- a/app/static/js/base.js
+++ b/app/static/js/base.js
@@ -1,5 +1,3 @@
-const demoId = document.getElementById('paw');
-console.log(demoId);
 
 function msgFlash(flash_message, status){
     if (status === "success") {
@@ -18,7 +16,7 @@ $(document).ready(function() {
     $("#userSelect").on('change', function() {
         $("#userSelectForm").submit();
     });
-    $("#flashes").delay(5000).fadeOut();
+    $("#flash_container").delay(5000).fadeOut();
 
 });
 
--- a/app/static/js/base.js
+++ b/app/static/js/base.js
@@ -16,7 +16,7 @@ $(document).ready(function() {
     $("#userSelect").on('change', function() {
         $("#userSelectForm").submit();
     });
-    $("#flash_container").delay(5000).fadeOut();
+    $(".alert").delay(5000).fadeOut();
 
 });
 
--- a/app/static/js/base.js
+++ b/app/static/js/base.js
@@ -1,3 +1,6 @@
+const demoId = document.getElementById('paw');
+console.log(demoId);
+
 function msgFlash(flash_message, status){
     if (status === "success") {
         category = "success";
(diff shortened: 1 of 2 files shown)


app/templates/base.html
```diff
@@ -56,8 +56,7 @@
             {% with messages = get_flashed_messages(with_categories=true) %}
               {% if messages %}
                 {% for category, message in messages %}
-                  <div class="alert  alert-{{category}} alert-dismissible" role="alert">{{ message }}
-                    <button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close"></button>
+                  <div class="alert  alert-{{category}} alert-dismissible" role="alert" id="flashes">{{ message }}
```
BrianRamsay: You can't have more than one id on a page

app/static/js/base.js
```diff
@@ -1,3 +1,6 @@
+const demoId = document.getElementById('paw');
+console.log(demoId);
```
BrianRamsay: Remove debug output

BrianRamsay: (and debug code)

app/static/js/base.js
```diff
@@ -15,8 +18,11 @@ $(document).ready(function() {
     $("#userSelect").on('change', function() {
         $("#userSelectForm").submit();
     });
+    $("#flashes").delay(5000).fadeOut();
```
BrianRamsay: This is the right idea, but it will only work for one, because it assumes each id is unique

BrianRamsay (changes requested):

app/static/js/base.js
```diff
@@ -15,8 +16,11 @@ $(document).ready(function() {
     $("#userSelect").on('change', function() {
         $("#userSelectForm").submit();
     });
+    $("#flash_container").delay(5000).fadeOut();
```
BrianRamsay: The trouble with this is that no more flashes will be able to show up until the page is reloaded. You need to fade just the alerts, not the container that holds them

BrianRamsay (changes requested):

app/templates/base.html
BrianRamsay: Let's keep that close button.

app/templates/base.html
```diff
@@ -57,7 +57,6 @@
               {% if messages %}
                 {% for category, message in messages %}
                   <div class="alert  alert-{{category}} alert-dismissible" role="alert">{{ message }}
-                    <button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close"></button>
```
BrianRamsay: Let's keep that close button

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #637, opened 2022-08-03
Bonner training events should not show up for normal students


AndersonStettner:
Currently if a Bonner event is created as a training it will show up in the Training and Education tab for everyone. The Bonner trainings should have the same permissions to view as other Bonner events.

Lyisus:
![](https://github.trello.services/images/mini-trello-icon.png) [Bonner Training events should not show up for non-Bonner students #637](https://trello.com/c/MpfSRdHg/436-bonner-training-events-should-not-show-up-for-non-bonner-students-637)

PR #669, opened 2022-08-04
Makes sure Bonner program event trainings do not show up for some user

mohamudm1:
Solves issue #637 
Changes made:
- Bonner scholar program event training does not show up for students who are not Bonner Scholars and faculty members under the training and education tab on the event list interface


Lyisus:
[Bonner Training events should not show up for non-Bonner students #637](https://trello.com/c/MpfSRdHg/436-bonner-training-events-should-not-show-up-for-non-bonner-students-637)

app/logic/events.py
```diff
@@ -146,10 +147,19 @@ def getTrainingEvents(term):
...
+                                .join(Program)
+                                .order_by(Event.isAllVolunteerTraining.desc(), Event.startDate)
+                                .where(Event.isTraining,
+                                        Event.term == term,
+                                        not Program.isBonnerScholars))
+    else:
+        trainingEvents = (Event.select(Event)
+                               .order_by(Event.isAllVolunteerTraining.desc(), Event.startDate)
+                               .where(Event.isTraining,
+                                      Event.term == term))
```
BrianRamsay: This definitely breaks the test (and the new functionality isn't tested either). We shouldn't use the global variable inside this function. if the training events now depend on the user, we should pass a user into the function

BrianRamsay (changes requested):

app/logic/events.py
```diff
@@ -139,17 +139,29 @@ def getStudentLedEvents(term):
...
         returned is the All Trainings Event.
+        term: expected to be the ID of a term
+        user: expected to be the current user
+        return: a list of all trainings the user can view
     """
-    trainingEvents = (Event.select(Event)
-                           .order_by(Event.isAllVolunteerTraining.desc(), Event.startDate)
-                           .where(Event.isTraining,
-                                  Event.term == term))
+    if ((user.isStudent and not user.isBonnerScholar) or user.isFaculty):
```
BrianRamsay: This doesn't check for staff. A better way to structure this would be in the positive, so `(user.isStudent and user.isBonnerScholar) or user.isAdmin` would be the all trainings case `else` the exclude-bonner case

app/logic/events.py
```diff
@@ -139,17 +139,29 @@ def getStudentLedEvents(term):
...
+                                .join(Program)
+                                .order_by(Event.isAllVolunteerTraining.desc(), Event.startDate)
+                                .where(Event.isTraining,
+                                        Event.term == term,
+                                        Program.isBonnerScholars == False))
+    else:
+        trainingEvents = (Event.select(Event)
+                               .order_by(Event.isAllVolunteerTraining.desc(), Event.startDate)
+                               .where(Event.isTraining,
+                                      Event.term == term))
```
BrianRamsay: These queries are similar enough let's explore combining them and just adding the extra `.where` if needed. I just realized this contradicts the "make the conditional positive" comment I made above.

tests/code/test_event_list.py
```diff
@@ -69,10 +70,75 @@ def test_studentled_events(training_events):
...
+                            isCeltsAdmin = False,
+                            isCeltsStudentStaff = False,
+                            isBonnerScholar = False)
+        notBonnerList = [testNotBonnerTraining]
+        bonnerList = [testNotBonnerTraining, testBonnerTraining]
 
-    assert allTrainingPrograms == getTrainingEvents(2)
+        assert notBonnerList == getTrainingEvents(testTerm, userNotBonner)
+        assert bonnerList == getTrainingEvents(testTerm, userBonner)
+
```
BrianRamsay, line 182: We should add the student exclusion case as well

BrianRamsay (changes requested):

app/logic/events.py
```diff
@@ -139,17 +139,29 @@ def getStudentLedEvents(term):
...
+    else:
+        trainingEvents = (Event.select(Event)
+                                .join(ProgramEvent)
+                                .join(Program)
+                                .order_by(Event.isAllVolunteerTraining.desc(), Event.startDate)
+                                .where(Event.isTraining,
+                                        Event.term == term,
+                                        Program.isBonnerScholars == False))
```
BrianRamsay: This will return duplicate all volunteer trainings, since it is part of a bunch of programs

BrianRamsay: Also we should combine these queries. The only real difference is that we add an extra where expression in the second case.

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #684, opened 2022-08-12
Profile refactor

### LINKED ISSUE #663: Refactor Volunteer Profile page view
Break into sections with accordions. Make sure we can navigate to specific open accordion windows.

AndersonStettner:
Files changed:
/app
1. __init__.py

/app/controllers
- /admin
1.  userManagement.py
3. volunteers.py

- /main
1. routes.py

app/logic
1. events.py
2. users.py
3. volunteers.py

app/static
1. userProfile.js
2. volunteerIndicateInterest.js

app/templates
- /admin
1. userManagement.html

- /main
1. userProfile.html
------------------------------------------
Summary of changes:
- Refactored userProfile.py so that the contents of the page are all contained in their own accordions instead of everything being displayed all together and made sure the permissions to see specific accordions were correct. (i.e. The Program Manager is only visible on the profiles of Student Staff)

- Made it so that if a page is reloaded due to an action taken/done inside an accordion, the accordion where it occurred will be the one focused/displayed after the page reload. (Done for the User Profile page and the Admin Management page)

- Made it so that if an admin updates a background check, the newly submitted background check will be displayed accurately (with correct formatting) in the background history section on button click. (No page reload required)

- Renamed the variable checkPassed to bgStatus throughout the entire application since it no longer accurately represented what it was after the background check refactoring. It used to just be a bool value but after adding in the new statuses: Submitted, Passed, and Failed it was now a misleading variable name.

- Changed how Admin Management page displayed the correct/focused accordion after page reload so that it was consistent with the way it was being done on the User Profile page. 

- Added a function to __init__.py, load_visibleAccordion, which is used to keep track of the accordion that is supposed to be focused and it is able to be used throughout the entire application without needing to change the controllers. 

- Removed items out of a select in events.py that caused multiple entries to appear on a users Upcoming Events table due to it counteracting a .distinct() later on in the query. 

- Other small formatting changes throughout the application.


AndersonStettner:
[Refactor Volunteer Profile page view #663](https://trello.com/c/j4wn5uK0/462-refactor-volunteer-profile-page-view-663)

app/__init__.py
```diff
@@ -108,6 +108,13 @@ def load_currentTerm():
         term = getCurrentTerm()
         session['current_term'] = model_to_dict(term)
         g.current_term = term
+
+from flask import request
+@app.context_processor
+def load_visibleAccordion():
+    # gets the current accoridon and assigns it to g.visibleAccordion
```
BrianRamsay: This comment isn't accurate.

app/static/js/userProfile.js
```diff
@@ -108,6 +114,9 @@ $(document).ready(function(){
       data: data,
       success: function(s){
           displayMessage("Saved!", "success")
+          var date = new Date(`${data.bgDate}`).toLocaleDateString()
```
BrianRamsay: There's no reason to put your string variable into a string to pass it to `Date()`. Just use the variable.

app/templates/admin/userManagement.html
```diff
@@ -118,9 +118,7 @@ <h3 class="accordion-header" id="headingThree">
         Program Management
         </button>
       </h3>
-      {% if showSettingsPane %}
-        {% set show= "show" %}
-      {% endif %}
+      {% set show = "show" if visibleAccordion == "program" else "" %}
```
BrianRamsay, line 122: Go ahead and add this to the other accordion panes as well.

app/templates/admin/userManagement.html
```diff
@@ -24,7 +24,7 @@ <h3 class="accordion-header" id="headingOne">
         </button>
       </h3>
       {%set collapse="show" %}
-      {% if showSettingsPane %}
+      {% if visibleAccordion %}
         {% set collapse= "" %}
       {% endif %}
```
BrianRamsay: Let's make this check match the other one in terms of variable names and style

app/templates/main/userProfile.html
```diff
@@ -21,232 +21,268 @@ <h6 align="center">{{volunteer.email}}</h6>
...
+                   {% endif %}
+                 </tr>
+               </thead>
+               <tbody>
+                 {% for bgType in backgroundTypes %}
+                 <tr>
+                   <td>
+                       <h6>{{bgType.description}}</h6>
+                       <ul class="mb-0" id="bgHistory{{bgType.id}}" >
+                         {% if allBackgroundHistory[bgType.id][0] %}
```
BrianRamsay: This conditional is unnecessary. the for loop will either happen or not anyway, which is what we want.

BrianRamsay (changes requested):

app/static/js/userManagement.js
```diff
@@ -82,7 +82,13 @@ function addNewTerm(){
...
     type: "POST",
     success: function(s){
-      location.reload()
+      var origin = window.location.href;
+
+      if (origin.includes("?")){
+        origin = origin.slice(0, origin.indexOf("?"));
+      }
+
+      location.replace(origin + "?accordion=term")
```
BrianRamsay: We should make this a function so we can just call `reloadWithAccordion("term")`

BrianRamsay (changes requested):

BrianRamsay (approved):

Issue #697, opened 2022-08-19
Open accordion on profile pages should be relevant to user


AndersonStettner:
When an Admin or a Student Staff go to a profile page that is not their own the programs accordion should be the one open by default not the Upcoming Events accordion.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Admin and Student Staff default open accordion on profiles that are not theirs #697](https://trello.com/c/IEmsgg0L/502-admin-and-student-staff-default-open-accordion-on-profiles-that-are-not-theirs-697)

PR #707, opened 2022-08-25
fix accodion conditional

qasema:
fixes issue #697 open only relevant accordions to the current user in the profile page

### DIFF
--- a/app/templates/main/userProfile.html
+++ b/app/templates/main/userProfile.html
@@ -29,12 +29,12 @@
 <div class="accordion" id="userProfile">
   <div class="accordion-item">
     <h3 class="accordion-header" id="headingOne">
-      {% set focus = "open" if not visibleAccordion or visibleAccordion == "upcoming" else "collapsed" %}
+      <!-- {% set focus = "open" if not visibleAccordion or visibleAccordion == "upcoming" else "collapsed" %} -->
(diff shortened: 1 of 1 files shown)


app/templates/main/userProfile.html
```diff
@@ -29,12 +29,12 @@ <h6 align="center">{{volunteer.email}}</h6>
 <div class="accordion" id="userProfile">
   <div class="accordion-item">
     <h3 class="accordion-header" id="headingOne">
-      {% set focus = "open" if not visibleAccordion or visibleAccordion == "upcoming" else "collapsed" %}
+      <!-- {% set focus = "open" if not visibleAccordion or visibleAccordion == "upcoming" else "collapsed" %} -->
```
AndersonStettner: Without this piece the Upcoming Events accordion is focused despite it not being the open accordion. Make sure the accordion that is open is always the one that is focused. 
How the page currently looks on page load:
image

How the page should look on page load:
image

AndersonStettner (changes requested):

AndersonStettner (changes requested):
Close but still not entirely there. Now, after the page is reloaded following the user selecting an interest, both the upcoming accordion and the programs accordion are open. 
image

Karina-Agliullova:
It works. Good job!

Karina-Agliullova (approved):

AndersonStettner (approved):

Issue #656, opened 2022-08-04
The user should be able to type in a time when they are creating an event. 


AndersonStettner:

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [The user should be able to type in a time when they are creating an event. #656](https://trello.com/c/3ztaDjXI/455-the-user-should-be-able-to-type-in-a-time-when-they-are-creating-an-event-656)

PR #710, opened 2022-08-26
Typing in time when creating an event - Issue #656

Karina-Agliullova:
**Issue:** The user cannot modify time by typing it in.
**Solution:** I removed the "readonly" attribute to let the user edit the time if they want to type it in. Also, I changed the time format from 24-hour format to 12-hour format so that the user doesn't have to manually move to the other side when choosing time.
**Testing:** Tested by creating an event and typing in the time I wanted. 
**Issue:** https://github.com/BCStudentSoftwareDevTeam/celts/issues/656

### DIFF
--- a/app/static/js/createEvents.js
+++ b/app/static/js/createEvents.js
@@ -56,7 +56,7 @@ $(document).ready(function() {
   // everything except Chrome
   if (navigator.userAgent.indexOf("Chrome") == -1) {
     $('input.timepicker').timepicker({
-             timeFormat : 'h:mm p',
+             timeFormat : 'hh:mm p',
              scrollbar: true,
              dropdown: true,
              dynamic: true,
--- a/app/static/js/createEvents.js
+++ b/app/static/js/createEvents.js
@@ -56,7 +56,7 @@ $(document).ready(function() {
   // everything except Chrome
   if (navigator.userAgent.indexOf("Chrome") == -1) {
     $('input.timepicker').timepicker({
-             timeFormat : 'hh:mm p',
+             timeFormat : 'h:mm p',
              scrollbar: true,
              dropdown: true,
              dynamic: true,
--- a/app/templates/admin/createEvent.html
+++ b/app/templates/admin/createEvent.html
@@ -117,7 +117,7 @@
                 {%else%}
                   {%set start_time = "12:00" %}
                 {% endif %}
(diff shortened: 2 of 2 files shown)


app/templates/admin/createEvent.html
```diff
@@ -129,7 +129,7 @@ <h1 id="title">{{page_title}}</h1>
                 {%else%}
                   {%set end_time = "12:00" %}
                 {% endif %}
-                  <input autocomplete="off" type="time" class="form-control timepicker readonly" value="{{end_time}}" name="timeEnd" aria-describedby="timeIcon2" placeholder="Pick an end time" onchange="update(this)" id=endTime required />
+                  <input autocomplete="off" type="time" class="form-control timepicker" value="{{end_time}}" name="timeEnd" aria-describedby="timeIcon2" placeholder="Pick an end time" onchange="update(this)" id=endTime required />
```
AndersonStettner, line 132: Nice, this works but, if a user went to type something like 1:30, they would have to manually move over to the minute section with either a mouse click or a right arrow key. It would be nice to catch the : and move to the next input (the minutes) so the user doesn't have to manually move themselves.

AndersonStettner (changes requested):

app/templates/admin/createEvent.html
Karina-Agliullova: That's right. But I am a little confused. The user has to input two digits on the left and on the right of a colon symbol (:). That's why it doesn't let the user input 1:30, it has to be 01:30. Wouldn't removing the option of choosing two digits on the left result in the user not being able to choose time after 9, meaning 10:00, 11:00, and so on?

app/static/js/createEvents.js
```diff
@@ -56,7 +56,7 @@ $(document).ready(function() {
   // everything except Chrome
   if (navigator.userAgent.indexOf("Chrome") == -1) {
     $('input.timepicker').timepicker({
-             timeFormat : 'hh:mm p',
+             timeFormat : 'h:mm p',
```
AndersonStettner: This does not appear to have any effect. What was this change intended to do?

AndersonStettner (changes requested):

app/static/js/createEvents.js
Karina-Agliullova: This change lets you type in only one digit except when you start with 1 because it can be 10,11,12. Basically, I switched from 24-hour format to 12-hour format so that when you press anything more than 1, it will automatically take you to the minutes, so you don't have to press any keys or do anything manually

AndersonStettner (approved):
Looks good!

Issue #695, opened 2022-08-19
Fix View button on Eligibility Table


AndersonStettner:
If a user clicks the View button in the Eligibility table right now they are taken back to the top of the page, this should not happen. Remove the default behavior from the href

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix View button on Eligibility Table #695](https://trello.com/c/bgyXs2Mr/499-fix-view-button-on-eligibility-table-695)

PR #711, opened 2022-08-26
Fixing View Button in Eligibility Table - Issue #695

Karina-Agliullova:
**Issue:** When clicking on the **View** button, it takes the user to the top of the page.
**Solution:** In href for the **View** button, I prevented the default behavior of the link so that it doesn't go to the top of the page and it doesn't refresh the page  and it doesn't change the link's behavior.
**Testing:** Tested by clinking on the **View** button in the Eligibility table. It does not take the user to the top of the page anymore. Also, when expressing interest in one of the events, the functionality of the "View" button doesn't break anything.
Issue #695

### DIFF
--- a/app/templates/main/userProfile.html
+++ b/app/templates/main/userProfile.html
@@ -158,10 +158,10 @@
                    <td>
                        {% if row.completedTraining or not trainingList %}
                            <i class="bi bi-check-lg"></i> &nbsp&nbsp
-                           <a class="trainingPopover" id="{{row.program.id}}" href="#" data-toggle="popover" title="Program Trainings" data-content= "{{trainingListView()}}">View</a>
+                           <a class="trainingPopover" id="{{row.program.id}}" href="#!" data-toggle="popover" title="Program Trainings" data-content= "{{trainingListView()}}">View</a>
                        {% else %}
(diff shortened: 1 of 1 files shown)


app/templates/main/userProfile.html
```diff
@@ -158,10 +158,10 @@ <h3 class="accordion-header" id="headingThree">
                    <td>
                        {% if row.completedTraining or not trainingList %}
                            <i class="bi bi-check-lg"></i> &nbsp&nbsp
-                           <a class="trainingPopover" id="{{row.program.id}}" href="#" data-toggle="popover" title="Program Trainings" data-content= "{{trainingListView()}}">View</a>
+                           <a class="trainingPopover" id="{{row.program.id}}" href="#!" data-toggle="popover" title="Program Trainings" data-content= "{{trainingListView()}}">View</a>
```
AndersonStettner: Rather than adding something to the url we want to change the link behavior. Currently if a users first action on a page is to click the view button it breaks other interactions/notifications.

AndersonStettner (changes requested):

bledsoef (approved):
PR looks good, approved.

AndersonStettner (approved):

Issue #689, opened 2022-08-19
Phone number 'Edit' link on User Profile not changing 


AndersonStettner:
Phone number 'Edit' link on User Profile is not changing to Save when you click on the phone number to edit it. It still works on the SLC instructors

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Phone numbers are not actually being updated on the user profile #689](https://trello.com/c/JIFlAoYD/489-phone-numbers-are-not-actually-being-updated-on-the-user-profile-689)

PR #713, opened 2022-08-29
User profile phone input save/edit 

Boorclark:
### Fixes Issues #689 

> Added an onclick on phone number input to change between save/edit HTML

**How To Test**
1. Navigate to My Profile
2. At the top of the page, find the phone number input box
3. Click on the Box as if you were editing 
4. BAM RESOLVED


app/static/js/userProfile.js
```diff
@@ -128,7 +128,13 @@ $(document).ready(function(){
...
         }
     });
-
+  $("#phoneInput").on('click', function() {
+    if ($("#updatePhone").html() == "Save") {
+        $("#updatePhone").html('Edit');
+    } else {
+        $("#updatePhone").html('Save');
+    }
+  });
```
AndersonStettner: This is good but it does not work the way we would want it to in all cases. If a user has focused the input area to change the number have the outline around the box be red and if they click off the input area without hitting save the outline will stay red until they hit save and submit the change.

AndersonStettner (changes requested):

app/static/js/userProfile.js
bledsoef: base.js already included this functionality in the validatePhoneNumber function, it probably isn't working but I think that would be the place to change it instead. Making a new function probably isn't a great fix as we are duplicating that code.

Issue #686, opened 2022-08-18
Fix Admin Log for Background checks


AndersonStettner:
With the changes made to the Background check statuses (Submitted, Passed, and Failed), the way we are creating the admin logs no longer works since it is based off of a bool value.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Adding Admin Log for Background checks #686](https://trello.com/c/rCggbRgO/483-adding-admin-log-for-background-checks-686)

PR #715, opened 2022-08-30
 Saves the admin logs for background checks in accordance with the new status 

mohamudm1:
Solves issue #686 

Changes made: Refactored the code so that the admin logs for background checks save correctly in correspondence with the new statuses: submitted, passed and failed.


AndersonStettner (changes requested):
If a user inputs a date but leaves the status empty the error message they receive above the button is "Passed Empty!". This is from when the only options were Passed and Failed and it is no longer a helpful error message. Let's update it to something like "Status Empty!".

AndersonStettner (approved):

PR #718, opened 2022-08-31
Only show upcoming recurring event 575

### LINKED ISSUE #575: Only show the next recurring event on a users upcoming event
On the "My Profile" page: when we are displaying the recurring upcoming events for a user we should only show them the next event, not all of the upcoming events.

tylerpar99:
This PR fixes #575 

Summary: 
 Changed `getUpcomingEventsForUser` function to remove duplicates when the event listed is a recurring. This will now show only one recurring event for each Id and will always show the next recurring event in the series. 

To Test: 

  We have made tests in the test_events.py if you run the suite it should pass. You can also create a recurring event and show interest then when checking under the upcoming events tab on the user profile only one will show up.

Files changed: `logic/events.py` and `tests/code/test_events.py`

### DIFF
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -211,7 +211,21 @@ def getUpcomingEventsForUser(user, asOf=datetime.datetime.now()):
                     .order_by(Event.startDate, Event.name).execute() # keeps the order of events the same when the dates are the same
                     )
 
-    return events
+
+    events_list = []
+    curr_recurr_id = None
+
+    # removes all recurring events except for the next upcoming one
+    for event in events:
+        if event.recurringId:
+            if event.recurringId != curr_recurr_id:
+                curr_recurr_id = event.recurringId
+                events_list.append(event)
+
+        else:
+            events_list.append(event)
+
+    return events_list
 
 def getParticipatedEventsForUser(user):
     """
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -213,14 +213,14 @@ def getUpcomingEventsForUser(user, asOf=datetime.datetime.now()):
 
 
     events_list = []
-    curr_recurr_id = None
+    shown_recurring_event_list = []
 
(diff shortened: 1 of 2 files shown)


bledsoef:
[Only show the next recurring event #575](https://trello.com/c/oqWEPpuM/423-only-show-the-next-recurring-event-575)

app/logic/events.py
```diff
@@ -211,7 +211,21 @@ def getUpcomingEventsForUser(user, asOf=datetime.datetime.now()):
...
+
+    # removes all recurring events except for the next upcoming one
+    for event in events:
+        if event.recurringId:
+            if event.recurringId != curr_recurr_id:
+                curr_recurr_id = event.recurringId
+                events_list.append(event)
+
+        else:
+            events_list.append(event)
```
AndersonStettner, line 226: This is good but it does not account for multiple recurring events. Currently, you are only checking the recurring event id and making sure it isn't the same as the last recurring id encountered, it would be good to store the recurring ids somewhere so we know if we have encountered that recurring event before.

AndersonStettner (changes requested):

tests/code/test_events.py
```diff
@@ -511,6 +511,30 @@ def test_upcomingEvents():
...
+                                recurringId = 1)
+
+        newRecurringDifferentId = Event.create(name = "Recurring different Id",
+                                term = 2,
+                                description = "Test upcoming program event.",
+                                location = "The sun",
+                                startDate = datetime.date(2021,12,12),
+                                endDate = datetime.date(2021,12,13),
+                                recurringId = 2)
+
```
AndersonStettner, line 537: getUpcomingEventsForUser orders events by start date so you want to make sure the start dates on the recurring events you create are correct so the test functions correctly.

AndersonStettner (dismissed):

BrianRamsay (approved):

Issue #680, opened 2022-08-10
Need Program on event page


BrianRamsay:
The program for an event is not visible on any of the event pages. We should display it, perhaps under the title. "A Berea Tutoring Event", "A CELTS-Sponsored Event", etc

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Need Program on event page #680](https://trello.com/c/NzYMOa8I/486-need-program-on-event-page-680)

PR #737, opened 2022-09-08
Program on event page 680

thawpt:
Did: Added the program name for the event under the title for both the View Event and Edit Event page

Test: You can see the changes by going to the Events List on the sidebar and clicking on any event.

-Fixes issue #680

### DIFF
--- a/app/templates/admin/createEvent.html
+++ b/app/templates/admin/createEvent.html
@@ -20,20 +20,22 @@
 {% set isNewEvent = 'create' in request.path %}
 <!-- Program manager isCeltsAdmin and they are a manager for the program  -->
 <div class="text-center mb-3">
-    {% if isNewEvent %}
-      {% if template.tag == 'single-program' %}
-        {% set page_title = 'Create Event for ' + eventData["program"].programName %}
-      {% elif template.tag == 'all-volunteer' %}
-        {% set page_title = 'Create All Volunteer Training' %}
-      {% elif template.tag == 'no-program' %}
-        {% set page_title = 'Create Other CELTS-Sponsored Event' %}
-      {% else %}
-        {% set page_title = 'Create ' + template.name + ' Event' %}
-      {% endif %}
-      {% else %}
-        {% set page_title = eventData.name %}
-      {% endif %}
+  {% if isNewEvent %}
+    {% if template.tag == 'single-program' %}
+      {% set page_title = 'Create Event for ' + eventData["program"].programName %}
+    {% elif template.tag == 'all-volunteer' %}
+      {% set page_title = 'Create All Volunteer Training' %}
+    {% elif template.tag == 'no-program' %}
+      {% set page_title = 'Create Other CELTS-Sponsored Event' %}
+    {% else %}
+      {% set page_title = 'Create ' + template.name + ' Event' %}
+    {% endif %}
+    {% else %}
+      {% set page_title = eventData.name %}
+      {% set program_name = eventData["program"].programName %}
+    {% endif %}
     <h1 id="title">{{page_title}}</h1>
+    <h3>{{program_name}}</h3>
     {% if isPastEvent %}
       <div class="alert alert-danger" role="alert">This event is in the past.</div>
     {% endif %}
@@ -141,7 +143,7 @@
             <tbody></tbody>
           </table>
         </div>
-        
+
     </div>
         <div class="col-md-6 form-group">
           <div class="form-group mb-4">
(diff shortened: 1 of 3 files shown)


app/templates/admin/createEvent.html
```diff
@@ -20,20 +20,33 @@
...
+      {% set page_title = 'Create ' + template.name + ' Event' %}
+    {% endif %}
+    {% else %}
+      {% set chicken = eventData["program"].programName %}
+      {% if eventData.name == 'All Volunteer Training' and program == None %}
+        <h1>{{eventData.name}}</h1>
+        <h3> An All Volunteer Training Event</h3>
+      {% elif eventData.program != None %}
+        <h1>{{eventData.name}}</h1>
+        <h3>A {{chicken}} Event</h3>
```
Lyisus: Can these "chicken" variables have a more descriptive name? Something like "programName" or similar should good, as long as it doesn't clash with other variables in the file.

thawpt: my bad!! completely forgot to change it

app/templates/eventView.html
```diff
@@ -10,6 +10,13 @@
     <div class="text-center mb-3">
     {% set page_title = eventData.name %}
         <h1 id="title">{{page_title}}</h1>
+        {% if eventData.name == 'All Volunteer Training' and program == None%}
+          <h3> An All Volunteer Training Event</h3>
+        {% elif program != None%}
+          <h3> A {{program.programName}} Event</h3>
```
bledsoef: Take the A out because in the instance of Adopt-A-Grandparent it is "A Adopt-A-Grandparent Event" which is grammatically incorrect. I think "Adopt-A-Grandparent Event" looks good too.

Also, I think it may be redundant to say All Volunteer Training right under All Volunteer Training so you could remove that.

app/templates/admin/createEvent.html
```diff
@@ -20,20 +20,33 @@
...
+    {% if template.tag == 'single-program' %}
+      {% set page_title = 'Create Event for ' + eventData["program"].programName %}
+    {% elif template.tag == 'all-volunteer' %}
+      {% set page_title = 'Create All Volunteer Training' %}
+    {% elif template.tag == 'no-program' %}
+      {% set page_title = 'Create Other CELTS-Sponsored Event' %}
+    {% else %}
+      {% set page_title = 'Create ' + template.name + ' Event' %}
+    {% endif %}
+    {% else %}
```
bledsoef: I would unindent the else here as it currently looks like it is with the other conditional block which could be confusing.

bledsoef: I would also fix the indentation for all the code under it to clear up any confusion.

bledsoef (changes requested):

app/templates/eventView.html
sreynit02: Removed A before the program name.

Got rid of All Volunteer Training under All Volunteer Training Event

app/templates/admin/createEvent.html
sreynit02: fixed indentation

app/templates/admin/createEvent.html
```diff
@@ -20,20 +20,32 @@
...
+    {% elif template.tag == 'no-program' %}
+      {% set page_title = 'Create Other CELTS-Sponsored Event' %}
+    {% else %}
+      {% set page_title = 'Create ' + template.name + ' Event' %}
+    {% endif %}
+{% else %}
+     {% set programNameforEvent = eventData["program"].programName %}
+     {% if eventData.name == 'All Volunteer Training' and program == None %}
+        <h1>{{eventData.name}}</h1>
+     {% elif eventData.program != None %}
```
bledsoef: This conditional is a double negative and could be changed to just `{% elif eventData.program %}'` if possible.

bledsoef (changes requested):
This will require a rereview once the Celts-Sponsored Events becomes a program but this will still need to change then so I figured I'd add it.

bledsoef:
Making this PR a draft as it would have to be changed once Celts-Sponsored Events becomes a program, once that functionality is added, it will be changed and rereviewed.

AndersonStettner:
[Need Program on event page #680](https://trello.com/c/NzYMOa8I/486-need-program-on-event-page-680)

app/templates/admin/createEvent.html
sreynit02: fixed

AndersonStettner (changes requested):
When an All Volunteer Training is created, it labeled as a CELTS-Sponsored Event on the page:
image

PR #739, opened 2022-09-09
Add banned users indicators

### LINKED ISSUE #659: Add indicators when banned volunteers interact with event
- We should have a popup window notification when adding a banned user while in kiosk mode
- There should be a warning text when a banned volunteer is added through the add volunteer modal
- There should be a warning indicated in the volunteer/RSVP table

zachneill:
Fixes #659.

Added a flasher in kiosk mode that alerts when a banned user attempts to log in. Also added an eligible column in the track volunteers table. Finally, when the user searches someone in the track volunteers modal and clicks on them, a warning icon will display beside their name if they are banned from the program associated with that event.

To test:

Fix 1: Sign in with a banned user in kiosk mode. You should see the flasher. 
Fix 2: You should see an "Eligible" column in the volunteer table.
Fix 3: Select a banned user in the volunteer modal. You should see the icon.

_Note: In the test database, khatts is banned in Adoption 101._


app/controllers/admin/volunteers.py
```diff
@@ -13,7 +13,10 @@
 from app.models.eventRsvp import EventRsvp
 from app.models.backgroundCheck import BackgroundCheck
 from app.models.programManager import ProgramManager
+from app.models.program import Program
```
bledsoef: Remove unused import statements.

app/static/js/eventKiosk.js
```diff
@@ -61,8 +61,20 @@ function submitData(hitEnter = false){
...
+        } else if (status === "banned") {
+          message = result.user + " is Banned."
+          flasherStatus = "danger"
+        } else if (status === "does not exist") {
+          message = "User does not exist"
+          flasherStatus = "danger"
+        } else {
+          message = result.user + " Successfully Signed In!"
+        }
+        eventFlasher(message, flasherStatus);
```
bledsoef, line 77: This code does not work, no matter what the input is, whether the user does not exist or the user has already signed in, it still says "Successfully Signed In!"

app/templates/events/trackVolunteers.html
```diff
@@ -59,6 +59,7 @@ <h1 class="text-center">{{event.name}}</h1>
         <th>Email</th>
         <th>Phone Number</th>
         <th class="noSort">Completed Trainings?</th>
+        <th class="noSort">Eligible/Not Banned?</th>
```
bledsoef: We do not like this title, the slash is unnecessary, I would recommend changing it to either eligible or eligibility and having what's in the column be like is on the volunteer profile where it says eligible or banned.

app/logic/users.py
```diff
@@ -47,6 +49,19 @@ def removeUserInterest(program_id, username):
         interestToDelete.delete_instance()
     return True
 
+def getBannedUsers(program):
+    """
+    This function returns users banned from a program.
+    """
+    return ProgramBan.select().where(ProgramBan.program == program)
+
+def isBannedFromEvent(username, eventId):
```
bledsoef, line 57: Test your functions.

bledsoef (changes requested):

app/controllers/admin/volunteers.py
zachneill: removed Program, ProgramEvent, and isEligibleForProgram

app/templates/events/trackVolunteers.html
zachneill: Changed to Eligible

app/static/js/eventKiosk.js
zachneill: Fixed. idk how that happened

app/logic/users.py
zachneill: Added tests

app/logic/users.py
```diff
@@ -47,6 +49,19 @@ def removeUserInterest(program_id, username):
         interestToDelete.delete_instance()
     return True
 
+def getBannedUsers(program):
+    """
+    This function returns users banned from a program.
+    """
+    return ProgramBan.select().where(ProgramBan.program == program)
```
bledsoef: This query does not check if they have been unbanned.

app/controllers/admin/volunteers.py
```diff
@@ -102,6 +104,9 @@ def addVolunteer(eventId):
 
     return redirect(url_for('admin.trackVolunteersPage', eventID = eventId))
 
+@admin_bp.route('/addVolunteersToEvent/<username>/<eventId>/isBanned', methods = ['GET'])
+def isVolunteerBanned(username, eventId):
+    return "banned" if isBannedFromEvent(username, eventId) else "not banned"
```
bledsoef: I feel that this should be a boolean value to follow the format of our other "is" functions.

app/logic/users.py
```diff
@@ -47,6 +49,19 @@ def removeUserInterest(program_id, username):
...
+
+def isBannedFromEvent(username, eventId):
+    """
+    This function returns whether the user is banned from the program associated with an event.
+    """
+    program = ProgramEvent.get(ProgramEvent.event_id == eventId).program
```
bledsoef: You don't necessarily have to, but we have an event property `singleProgram` that could shorten that line and make it a little simpler.

tests/code/test_users.py
```diff
@@ -322,3 +321,27 @@ def test_getUserBGCheckHistory():
...
+            # Ban user from program 1
+            banUser(1, User.get_by_id("usrtst"), "nope", "2022-11-29", "ramsayb2")
+            # Test banned user
+            assert testusr in [user.user for user in getBannedUsers(1)]
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_isBannedFromEvent():
+    with mainDB.atomic() as transaction:
+        assert isBannedFromEvent("khatts", 3)
```
bledsoef: Create a new user to make sure this test always passes.

tests/code/test_users.py
```diff
@@ -322,3 +321,27 @@ def test_getUserBGCheckHistory():
 
 
         transaction.rollback()
+
+@pytest.mark.integration
+def test_getBannedUsers():
+    with mainDB.atomic() as transaction:
```
bledsoef, line 334: Remove this as it is unnecessary, please,

tests/code/test_users.py
```diff
@@ -322,3 +321,27 @@ def test_getUserBGCheckHistory():
...
+
+            # Create test user
+            testusr = User.create(username = 'usrtst',
+                                  firstName = 'Test',
+                                  lastName = 'User',
+                                  bnumber = '03522492',
+                                  email = 'usert@berea.deu',
+                                  isStudent = True)
+            # Ban user from program 1
+            banUser(1, User.get_by_id("usrtst"), "nope", "2022-11-29", "ramsayb2")
```
bledsoef: Make the date further in the future as this will break in 2 months.

bledsoef (changes requested):

tests/code/test_users.py
zachneill: Done

tests/code/test_users.py
zachneill: Done

tests/code/test_users.py
zachneill: Changed to 2050 which is a while from now

app/controllers/admin/volunteers.py
zachneill: Fixed

app/logic/users.py
zachneill: Fixed

app/logic/users.py
zachneill: Done

tests/code/test_users.py
```diff
@@ -322,3 +321,51 @@ def test_getUserBGCheckHistory():
 
 
         transaction.rollback()
+
+@pytest.mark.integration
+def test_getBannedUsers():
+    with mainDB.atomic() as transaction:
+        bannedUser = User.create(username = 'usrtst', # Test banned user
```
thawpt: I think the name should not be bannedUser because at this point the user has not been banned yet.

tests/code/test_users.py
```diff
@@ -322,3 +321,51 @@ def test_getUserBGCheckHistory():
...
+                              bnumber = '03522493',
+                              email = 'usert2@berea.deu',
+                              isStudent = True)
+        assert notBannedUser not in [user.user for user in getBannedUsers(1)]
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_isBannedFromEvent():
+    with mainDB.atomic() as transaction:
+        bannedUser = User.create(username = 'usrtst', # Test banned user
```
thawpt: the same comment as the previous function, the variable name bannedUser is not representative of what the code is doing.

tests/code/test_users.py
zachneill: Changed to userToBan

tests/code/test_users.py
zachneill: Changed to userToBan

app/logic/users.py
```diff
@@ -47,6 +48,19 @@ def removeUserInterest(program_id, username):
         interestToDelete.delete_instance()
     return True
 
+def getBannedUsers(program):
+    """
+    This function returns users banned from a program.
+    """
+    return ProgramBan.select().where(ProgramBan.program == program, ProgramBan.unbanNote == None)
```
AndersonStettner, line 55: We are already getting what users are banned from a program above this in isElegibleForProgram. Why are you getting the users banned from a program again?

app/templates/events/trackVolunteers.html
```diff
@@ -59,6 +59,7 @@ <h1 class="text-center">{{event.name}}</h1>
         <th>Email</th>
         <th>Phone Number</th>
         <th class="noSort">Completed Trainings?</th>
+        <th class="noSort">Eligible?</th>
```
AndersonStettner: I am not a big fan of the name here, is there anything else we could name it that might be a bit better? Also, I think we might want to take out the punition here, and on the header before.

AndersonStettner (changes requested):
Everything is looking pretty good. I think we might want to add a popup when hovering the red circle that appears to the right of a volunteers name when they are added manually to indicate what it means.

app/logic/users.py
zachneill: isElegibleForProgram is for a single user whereas getBannedUsers is for all users. That way we can filter through a list in the table in jinja. It's querying just once instead of separate queries every time there's a user in the table.

AndersonStettner (changes requested):
Almost there, lets move the "symbol" -> banned out of the footer and into the body (capitalize the b of Banned) and move the close button to the left where "symbol" -> banned is now (just add me-auto to the class of Close button). Also, on the Kiosk entry page: make the button primary instead of secondary so that is blue.

AndersonStettner (commented):
Just leaving as a comment since you haven't requested changes and are probably still working on it: The student search has stopped working when trying to add them as a participant/RSVP.

AndersonStettner (dismissed):

bledsoef (dismissed):

AndersonStettner (approved):

PR #740, opened 2022-09-09
Implement New Logo and Name

### LINKED ISSUE #736: Add new CELTS logo and change the application name to "CELTS Link"
We have been told that the name they came up with for the application is "CELTS Link" so we will want to change the title and we need to find a place to work in their new logo.

The new CELTS logo: 
CMYK_Horiz_Internal

zachneill:
Resolves #736 

Instead of the "CELTS" span, the sidebar uses CELTS's new logo. It is of size 230px (sidebar is size 280). Additionally, references of CELTS in the tab are renamed to CELTS Link. Although, I only found references in the base.html file. There is a favicon of the logo in the tab as well. 

We actually altered the font color from blue to white and saved the logo as a .png to make it transparent. The favicon is just a cropped version of the logo with only the design, no text. 

The original blue logo and design png aren't used, but they are in the static folder for reference.


tylerpar99 (changes requested):
PR looks good, with only one change. Don't know if it is necessary to keep the images that are not being used in the repo.

BrianRamsay:
> PR looks good, with only one change. Don't know if it is necessary to keep the images that are not being used in the repo.

I actually like the images in the repo because otherwise we will lose track of them

BrianRamsay (approved):
This looks really nice. Thanks Zach.

PR #750, opened 2022-09-14
Delete Recurring Events

### LINKED ISSUE #722: Make it possible to delete all recurring events in a series
Allow the user to delete all the recurring events in a series or individual events.

sreynit02:
Did
 - Wrote a deleteAllRecurringEvents function that deleted the event we are on and the recurring events after it.
 - Added a button in the modal that appears when the "Delete" button is pressed on edit event page.
 - Added a test suite for the deleteAllRecurringEvents function

Test
 - Run test
 - Create a recurring event.
 - Delete one of the recurring events by pressing the "delete" button on the edit event page. When modal pops up, click "Delete All Recurring" button. The event that you were on and all the recurring events after should be deleted.


app/controllers/admin/routes.py
```diff
@@ -203,7 +203,16 @@ def deleteRoute(eventId):
     except Exception as e:
         print('Error while canceling event:', e)
         return "", 500
+@admin_bp.route('/event/<eventId>/deleteRecurring', methods=['POST'])
+def deleteRecurringRoute(eventId):
+    try:
+        deleteAllRecurringEvents(eventId)
+        flash("Event successfully deleted.", "success")
```
AndersonStettner: Since we are deleting multiple events in a recurring series I think it would be better to make it plural: "Events successfully deleted."

app/logic/events.py
```diff
@@ -62,6 +62,18 @@ def deleteEvent(eventId):
...
+def deleteAllRecurringEvents(eventId):
+        """
+        Deletes a recurring event and all the recurring events after it.
+        """
+        event = Event.get_or_none(Event.id == eventId)
+        if event:
+            if event.recurringId:
+                recurringId = event.recurringId
+                recurringEvents = list(Event.select().where((Event.recurringId == recurringId) & (Event.startDate >= event.startDate))) 
+        for seriesEvent in recurringEvents:
```
AndersonStettner: Since it is a recurring series I think a name like recurringSeries would work a bit better than seriesEvent but that is mainly personal preference

AndersonStettner (changes requested):
After talking with Brian a bit we think it would be better to:
1.  Move the delete buttons to the main body of the modal instead of having them along the bottom.
2. Move the close button to the bottom left of the modal and add a new delete option, Delete all recurring events in a series, which would delete all the recurring events regardless of what event in the series they are on.
3. Rename the current "Delete All Recurring" button so it better matches its functionality (it deletes the one its on and all after in the series).

app/logic/events.py
```diff
@@ -62,6 +62,18 @@ def deleteEvent(eventId):
 
         event.delete_instance(recursive = True, delete_nullable = True)
 
+def deleteAllRecurringEvents(eventId):
```
AndersonStettner: This function name is not entirely accurate, and will not be accurate when the requested changes get implemented. Is there a name the could better fit what the function is doing?

app/controllers/admin/routes.py
sreynit02: fixed

app/logic/events.py
sreynit02: Thanks for the suggestion. I will use recurringSeries as recurringEvent has also been used somewhere else.

app/logic/events.py
sreynit02: Changed to deleteRecurringSeries()

app/templates/admin/createEvent.html
```diff
@@ -227,11 +227,16 @@ <h5 class="modal-title" id="warningModel">Delete Event</h5>
       </div>
       <div class="modal-footer">
         <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Close</button>
+        {% if eventData.recurringId != None %}
```
tylerpar99: This would be better in a positive statement. Instead try {% if eventData.recurringId %} should give  you the same result without the '!='

app/logic/events.py
```diff
@@ -62,6 +62,31 @@ def deleteEvent(eventId):
...
+        if event:
+            if event.recurringId:
+                recurringId = event.recurringId
+                recurringSeries = list(Event.select().where((Event.recurringId == recurringId) & (Event.startDate >= event.startDate)))
+        for seriesEvent in recurringSeries:
+            seriesEvent.delete_instance(recursive = True)
+
+def deleteAllRecurringEvents(eventId):
+        """
+        Deletes all recurring events.
```
AndersonStettner, line 79: You do not have a test for this function.

tests/code/test_events.py
```diff
@@ -479,9 +479,25 @@ def test_deleteEvent():
...
+        event = Event.get_by_id(recurringEvents[1].id)
+        recurringId = event.recurringId
+
+        totalRecurringEvents = len(Event.select().where(Event.recurringId == recurringId))
+        #checks the number of all recurring events that will take place after a recurring event plus the event itself.
+        eventPlusAllRecurringEventsAfter = len(Event.select().where((Event.recurringId == recurringId) & (Event.startDate >= event.startDate)))
+
+        with app.app_context():
+            g.current_user = User.get_by_id("ramsayb2")
+            deletingEvent = deleteEventAndAllFollowing(recurringEvents[1])
```
AndersonStettner: I think that if you are wanting to test the delete current and all following function, it would be better to not delete the first event in the series since that is basically just checking that you are deleting the whole series (which is now its own separate function). I think that it would be better to delete an event in the middle of the series and make sure that the older ones are still there.

AndersonStettner (changes requested):
Make sure to pull in development.

AndersonStettner (changes requested):
Lets also change how the modal looks. To move the Close button over just add me-auto to the class of the button.

Modal for deleting recurring events:
Delete Event

Modal for deleting normal events:
Delete single

tests/code/test_events.py
sreynit02: fixed!

AndersonStettner (changes requested):
Good job on getting the buttons to work. One last note would be to move the buttons so the look like my previous comment: put all of the delete buttons on their own row and move the close button to the left. Make sure to also move the close button to the left on the non-recurring delete modal as well.

sreynit02:
> Good job on getting the buttons to work. One last note would be to move the buttons so the look like my previous comment: put all of the delete buttons on their own row and move the close button to the left. Make sure to also move the close button to the left on the non-recurring delete modal as well.

Edits made!

AndersonStettner (approved):
Good job!

PR #751, opened 2022-09-15
Add new background check type, disable save after press, and allow deletion of BG check

### LINKED ISSUE #749: Add a new Background check category: Student Life. Also give admin the ability to delete a BG check and disable the save button after it has been pressed

AndersonStettner:
Issue: CELTS asked for a new background check type to be added (Student Life BG check). They also had times where the double clicked the save button so they wished for it to be disabled after they clicked it and wanted to be able to delete background checks. 
___________________________________________________________
Did: 
- Added the new background check type called Berea Student Life (BSL). 
- Added a red X button to the right of individual background check histories that, when pressed, deletes that entry from the database. 
- Made it so that the page is reloaded after a background check is submitted and made sure that the accordions were opened and focusing correctly. 
- Made it so that if a user tries to save a background check they will not be able to until both the status and date are filled in. 
- Moved Toast notifications so that they will be able to be used from anywhere in the application.
- Changed getUserBGCheckHistory so that it no longer formats the data. The formatting is now done in the HTML. 
- Refactored test_getUserBGCheckHistory to match with the new way background check history is returned. 

___________________________________________________________
To test:
- Checkout the branch add-and-enhance-background-check
- With a newly reset database, run tests. 
- Using the application from the Admin user perspective:
1. On the Admin User Profile submit a few background checks and make sure they are appearing as they should
2. Try and delete one of the background checks that should now be in the history
3. Make sure the Toast appears in the bottom right confirming that the background check was deleted 
4. Reload the page to make sure the background check stays gone and check the database to make sure it is removed from there as well 
5. Repeat on a normal students User Profile and Student Staff User Profile.


AndersonStettner:
[Add a new Background check category: Student Life. Also give admin the ability to delete a BG check and disable the save button after it has been pressed #749](https://trello.com/c/4U3spXI4/542-add-a-new-background-check-category-student-life-also-give-admin-the-ability-to-delete-a-bg-check-and-disable-the-save-button-af)

app/controllers/admin/volunteers.py
```diff
@@ -121,6 +121,14 @@ def updateBackgroundCheck():
         addUserBackgroundCheck(user, type, bgStatus, dateCompleted)
         return " "
 
+@admin_bp.route('/deleteBackgroundCheck', methods = ['POST'])
+def deleteBackgroundCheck():
+    if g.current_user.isCeltsAdmin:
+        eventData = request.form
+        bgId = BackgroundCheck.get_by_id(eventData['bgID'])
+        BackgroundCheck.delete().where(BackgroundCheck.id == bgId).execute()
```
BrianRamsay: `bgId` isn't an id, it's a `BackgroundCheck` object

app/static/js/userProfile.js
```diff
@@ -84,17 +84,25 @@ $(document).ready(function(){
...
-    if (bgStatus != '' && bgDate == '' ) {
-        displayMessage("Date<br>Empty!", "danger")
-        return
+
+    if (bgDate == ''){
+      $("#" + bgCheckType + "_date").addClass("invalid");
+      window.setTimeout(() => $("#" + bgCheckType + "_date").removeClass("invalid"), 1000);
+      $("#" + bgCheckType + "_date").focus()
+      $(this).prop("disabled", false);
+      return false
```
BrianRamsay, line 109: In each of these we are selecting the same element multiple times. Save it in a variable at the top so you only have to get it once.

app/templates/main/userProfile.html
```diff
@@ -246,7 +246,7 @@ <h3 class="accordion-header" id="headingFive">
       <div id="collapseFive" class="accordion-collapse collapse {{show}}" aria-labelledby="headingFive" data-bs-parent="#userProfile">
         <div class="accordion-body">
           <div class="table-responsive">
-            <table class="table">
+            <table class="table" id="bgHistoryTable" name="bgHistoryTable">
```
BrianRamsay: The `name` attribute is meaningless for non-form elements, as far as I know.

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #759, opened 2022-09-16
Fixing timepicker when creating an event

### LINKED ISSUE #729: Validate time entry in Firefox and iOS. #729
Firefox doesn't use the Chrome time picker so we need to validate the text that users enter. Solution must work on iOS as well

Karina-Agliullova:
Issue #729
**Issue:** Timepicker doesn't work in anything but Chrome. The time picker icon works in Chrome, but not in other browsers (Firefox)
**Solution:** Added CSS for jquery to fix the import issue. The time picker icon issue is still in progress.
**Testing:** When creating an event, tested the time picker in Chrome, and Firefox.


sreynit02:
Paw & Sreynit: We think that it would be better if the time options show when the time picker bar is clicked rather than just having the icon clickable.

martinj2-dot:
> Paw & Sreynit: We think that it would be better if the time options show when the time picker bar is clicked rather than just having the icon clickable.

So from what I can tell the timepicker is working as intended, by that I mean in chrome it will allow you to either type a custom time OR select a dropdown if you want to pick a time that way. From what I can tell by playing around with it and trying to make the dropdown appear when clicking on the box is that that isn't possible for the chrome library. I would like to hear more thoughts on this because I haven't been able to fix it yet.

AndersonStettner (changes requested):
Good work on getting the drop down to work but we still want to make sure a user cant input random text strings in Firefox:
image

Since we accept text, our application, understandably, breaks when submitted:
image

sreynit02 (changes requested):
Agreed with Anderson's Comment

AndersonStettner (changes requested):
Everything is looking good and we can now type time into the box just fine on Firefox. One last thing to add would be making the button to the right of the input box an actual button so that when it is clicked it focuses the time box.

AndersonStettner (changes requested):
When typing in a time on Firefox, the time gets deleted when the user types the P of PM.

martinj2-dot:
Deprecating PR. Instead we are going to find a new time picker library.

PR #760, opened 2022-09-16
Volunteer duplicates

### LINKED ISSUE #735: Investigate why track Volunteers is showing duplicates
There are some cases where Track Volunteers will show duplicated students. Investigate why and fix.

gastonjarju:
Fixes #735 

Removed the duplicate rows of a volunteer that occurs each time the event is in the past. Instead of having multiple duplicated rows for a single volunteer, we now have a single row.

To test:
Navigate to the Event List: IP_Address/eventsList/1/track_volunteers.  Click on edit event to confirm if event has already occurred. If yes, click on Track Volunteers and for the same volunteer that is on view, it will show only one row per volunteer. Also, you can add a new volunteer for a past event and only one row would be created per volunteer.


app/controllers/admin/volunteers.py
```diff
@@ -42,8 +42,13 @@ def trackVolunteersPage(eventID):
 
     eventRsvpData = list(EventRsvp.select().where(EventRsvp.event==event))
     eventParticipantData = list(EventParticipant.select().where(EventParticipant.event==event))
-    eventVolunteerData = (eventParticipantData + eventRsvpData)
-
+    participantsAndRsvp = (eventParticipantData + eventRsvpData)
+    eventVolunteerData = []
+    volunteerUser = []
```
sreynit02, line 47: What is the purpose of having these two new lists, it seems like the same info is being passed into each list. The only difference is one list is being passed the object and the other is being passed object.user

sreynit02 (changes requested):
The functionality works correctly now (no duplicate volunteer added) and no test failed. Just have one question to address.

app/controllers/admin/volunteers.py
zachneill: The object list includes both EventRsvp and EventParticipant objects. So, they can't be compared. The user list appends the user, which is in both objects. That's how we were able to compare them. Just one list doesn't work for how we're doing the comparison.

AndersonStettner (commented):
Everything looks good. When a volunteer is added to an event after already being in the table the flash message we show is something along the lines of "User has been added successfully" but I think it would be better for it to say that they tried to add a user that has already been added.

AndersonStettner (changes requested):
Another thing, when an event is in the past, and a volunteer has attended the event, their "Total Hours" is grayed out which makes it seem like they have not actually attended the event. Make sure that if a volunteer has attended an event, their hours are displayed correctly on the Event Participant page.

AndersonStettner (approved):
Looks good.

sreynit02 (approved):

PR #762, opened 2022-09-16
Firefox kiosk error

### LINKED ISSUE #753: Make sure kiosk mode works in Firefox
During the first session of the All Volunteer training held by CELTS they used Firefox to run the kiosk. We were told that it stopped working after ~10 students. They switched to Chrome during the second session and there were no issues. Investigate what happened in Firefox and fix.

zachneill:
Fixes #753 

I believe I replicated their issue. If you use the kiosk a little too fast, it throws a string index out of range error in the routes file. I suspect it tried to index an empty bnumber string. After adding a check for the bnumber to not be empty, I haven't been able to replicate the problem. 

To test:

Spam the heck out of the kiosk by any means and see if it stops working/throwing failure flashers no matter what.

### DIFF
--- a/app/controllers/events/routes.py
+++ b/app/controllers/events/routes.py
@@ -48,6 +48,9 @@ def signinEvent():
     bnumber = request.form["bNumber"]
     programid = ProgramEvent.select(ProgramEvent.program).where(ProgramEvent.event == eventid)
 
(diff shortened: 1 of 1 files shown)


sreynit02 (approved):
Kisok is working properly in Firefox now!

Issue #731, opened 2022-09-07
Add the concept of a Bonner cohort


AndersonStettner:
We need to add the concept of Bonner and a Bonner cohort. 

Needs:
- Need a way to identify who is a Bonner student.

- The concept of a Bonner cohort 
1. Each year there will be an incoming freshman Bonner cohort so there will be 4 active cohorts at a time. 
2. Each Bonner should be tied to their cohort so we are able to track all Bonners within a cohorts progress/event attendance. 

Sub-issue of #603

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Add the concept of a Bonner cohort](https://trello.com/c/JYHMLvTD/535-add-the-concept-of-a-bonner-cohort)

Issue #734, opened 2022-09-07
Create a new Bonner Admin Page


AndersonStettner:
Create a new page for the Admin that allows them to: 
- add/remove Bonner students 
- Show them all upcoming Bonner events
- Show them all current Bonner students broken down by the Bonner cohort

Sub-issue of #603

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Create a Bonner Admin page](https://trello.com/c/drkfgJdT/536-create-a-bonner-admin-page)

PR #765, opened 2022-09-21
Bonner Cohort Management

BrianRamsay:
I have added basic data structures and an admin management page for Bonner cohorts.

CELTS admins should be able to:
- view all bonner students in a list within their cohorts
- add students to different Bonner years
- download an excel document with all of the bonner students.
- view a basic list of bonner events for the term with the # of rsvps and participants

Resolves issues #731 and #734


sreynit02 (commented):
Some recommendation for the UI:
-When displaying the cohorts (years &all bonners in each year), we think it might look better if make each year and the corresponding bonners in different column. That way there is less white space to the right. Also, we actually display all the bonners in each cohort, the accordion would not be too long vertically if we display each year into columns. 
-We also think that when displaying all the bonner cohorts, the font of the year should be bigger than the student names.

app/templates/admin/bonnerManagement.html
```diff
@@ -0,0 +1,151 @@
...
+    </div>
+  </div>
+</div>
+
+  <div class="accordion-item">
+    <h3 class="accordion-header" id="headingThree">
+      {% set focus = "open" if visibleAccordion == "requirements" else "collapsed" %}
+      <button class="accordion-button {{focus}} " type="button" data-bs-toggle="collapse" data-bs-target="#collapseThree" aria-expanded="false" aria-controls="collapseThree"> Requirements </button>
+    </h3>
+    {% set show = "show" if visibleAccordion == "requirements" else "" %}
```
AndersonStettner, line 127: Your accordion focus/opening is off:
image

AndersonStettner:
Not necessarily related to this PR just something I noticed: The description for Bonners on the Service Transcript is the Habitat for Humanity description.

AndersonStettner:
Another thing not necessarily related to this PR: It has been mentioned briefly in the past that we will probably need different behaviors for when a student leaves Bonners voluntarily vs. when they leave involuntarily.  

Could be as simple as: Student leaves voluntarily we could just remove them from cohort and when a student leaves involuntarily we could remove them from the cohort AND ban them from Bonner events.

app/templates/admin/bonnerManagement.html
```diff
@@ -0,0 +1,151 @@
...
+                        <td><a href= '/eventsList/{{event.id}}/view' class="link-primary">{{event.name}}</a></td>
+                        <td nowrap>{{event.startDate.strftime('%m/%d/%Y')}}</td>
+                        <td nowrap>{{event.timeStart.strftime('%-I:%M %p')}}</td>
+                        <td>{{event.location}}</td>
+                        <td>{{event.rsvps|count}}</td>
+                        <td>{{event.participants|count}}</td>
+                      </tr>
+                  {% endfor %}
+                {% else %}
+                  <h5>You have no Bonner events.</h5>
```
AndersonStettner: Personally, I don't think we should use personal wording here. The page is, in my head, a general info page that the admin will use for Bonner so, it would be better to say something like "There are no Bonner events."

app/models/bonnerYear.py
```diff
@@ -0,0 +1,9 @@
+from app.models import *
+from app.models.user import User
+
+class BonnerYear(baseModel):
```
AndersonStettner: Personally, I do not like the name BonnerYear and I think it is a bit misleading. Just by going off of the name it seems like it is just keeping track of cohort years ( my first impression when I saw it was that it was keeping track of which years were which cohort id) but it is actually the Bonner student with their cohort year. I think we might want to rename it to better match what it actually is.

AndersonStettner (changes requested):

BrianRamsay:
> Some recommendation for the UI: -When displaying the cohorts (years &all bonners in each year), we think it might look better if make each year and the corresponding bonners in different column. That way there is less white space to the right. Also, we actually display all the bonners in each cohort, the accordion would not be too long vertically if we display each year into columns. 

@sreynit02 Do you think the contents of the accordion should scroll right and left to see all of the columns or reflow depending on the width of the page? Or, do we even need 'All', since they have the option to download the whole list?

AndersonStettner:
Lets add a few students as Bonners to base data.

sreynit02:
> 

I think scrolling left-right would be worse than up-down.  I think we don't need 'All' at all because we already displayed them under the search box for every year. If we got rid of "All",  we should add the volunteer profile link when we display their names for each year. 

A recommendation to remove the white space to the right of the Cohort Management accordion would be to center the content and make the year tabs' length the same length as the space bar column.

sreynit02:
> > Some recommendation for the UI: -When displaying the cohorts (years &all bonners in each year), we think it might look better if make each year and the corresponding bonners in different column. That way there is less white space to the right. Also, we actually display all the bonners in each cohort, the accordion would not be too long vertically if we display each year into columns.
> 
> @sreynit02 Do you think the contents of the accordion should scroll right and left to see all of the columns or reflow depending on the width of the page? Or, do we even need 'All', since they have the option to download the whole list?

I think scrolling left-right would be worse than up-down. I think we don't need 'All' at all because we already displayed them under the search box for every year. If we got rid of "All", we should add the volunteer profile link when we display their names for each year.

A recommendation to remove the white space to the right of the Cohort Management accordion would be to center the content and make the year tabs' length the same length as the space bar column.

BrianRamsay:
I removed 'All' and centered the columns. Do we like it centered or left-justified better?

sreynit02:
> I removed 'All' and centered the columns. Do we like it centered or left-justified better?

Centered please :)

AndersonStettner (approved):
We probably want to add one or two people as Bonners in base data and fix the tests if they break due to those additions. Aside from that, I think everything is looking good.

PR #770, opened 2022-09-30
Allow events to be created on past days

AndersonStettner:
Added a new option on the Create Event page: Allow start date to be in the past. When checked it will allow the user to create an event before the current day (in the past).

### DIFF
--- a/app/static/js/createEvents.js
+++ b/app/static/js/createEvents.js
@@ -56,13 +56,11 @@ $(document).ready(function() {
   $("#allowPastStart").click(function() {
     var bloo = $("#allowPastStart:checked").val()
     if (bloo == 'on') {
-      // can time travel into the past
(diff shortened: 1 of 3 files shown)


app/static/js/createEvents.js
```diff
@@ -53,6 +53,20 @@ $(document).ready(function() {
       $("#endDatePicker").prop('required', false);
     }
   });
+  $("#allowPastStart").click(function() {
+    var bloo = $("#allowPastStart:checked").val()
```
sreynit02: I think variable name like 'allowPast' would better represent its function.

sreynit02 (changes requested):
The functionality of this works good! 

I think it the button "Allow start date to be in the past." should be removed from the edit event page. 
image

sreynit02 (approved):
looks good!

PR #773, opened 2022-10-03
Increase word limit 747

### LINKED ISSUE #747: Increase the word limit in the event description and tell the user what that limit is.
The current the word limit is 255 (the default limit) and we want it increased to 350. We should also tell the user what the word limit is somewhere.

thawpt:
Did:
-Added text near "Description" to let user know maximum amount of words they can input 
-Changed description to TextField in event.py model to allow more words to be stored in database
-Added an event handler  to prevent user from inputting more than 350 words in the text box 
-Also changed serviceLearningDesignatedSections (course.py )and questionContent(courseQuestion.py) to textfield in model

Test: 
-Create an event, fill in all the required  fields, in the description text area you can insert up to 350 words (will not let you type in more)
-Save the event


martinj2-dot (changes requested):
image

This is the error Flash Message I get when trying to add a description with more than 350 words. I think it would be best to add a better notification to notify them that they put too many words.

Andersonstettner:
> 

This is not due to going over word limit but because we are not doing our encoding correctly so it will be a separate issue.

AndersonStettner:
> image
> 
> This is the error Flash Message I get when trying to add a description with more than 350 words. I think it would be best to add a better notification to notify them that they put too many words.

This is not due to going over word limit but because we are not doing our encoding correctly so it will be a separate issue.

AndersonStettner:
[Increase the word limit in the event description and tell the user what that limit is. #747](https://trello.com/c/XoPaUFXD/539-increase-the-word-limit-in-the-event-description-and-tell-the-user-what-that-limit-is-747)

tylerpar99 (commented):
Works great when typing, one issue is it doesn't count pasting. The function seems like it counts as you type however because pasting doesn't require typing I was able to pass in 800 words. Now when you send in the data it will cut it off on the SQL end but we should alert our user to this. My recommendation would be to count the length of the data before sending it to the backend and if it is over 350 words let the user know it will be cut off or limited in someway.

thawpt:
fixed problem (I think). now it should automatically cut off words beyond 350 when pasted in the text area.

AndersonStettner (changes requested):
It's looking good. One last thing, if the person is pasting something into the description after they reached the limit the first word of the text that is pasted in stays despite the being over the limit. 
image

Also, if the user goes to type after hitting the word limit they are able to add as much as they want as long as they don't space the words out.
image
image

AndersonStettner (changes requested):
Also, it would be nice nice for the function that limits the word limit to be generalized and put into base.js so it can be called from anywhere like msgToast and msgFlash. Could be something like wordLimit(350). We also need to display to the user how many words/characters they have remaining until they reach the limit.

AndersonStettner (changes requested):
Looks good, only thing we can see is that when a user goes to look at a saved description the character count is not updated until after they make an action in the box. We want the character count to accurately represent the amount of characters in the description at all times.

AndersonStettner (approved):
Looks good!

martinj2-dot (approved):
Great job, looks great.

Karina-Agliullova (approved):
Good work!

Issue #721, opened 2022-08-31
All Volunteer trainings should show up as required trainings for all programs.


AndersonStettner:
For now, we want All volunteer training to be required training for all programs but when you hover over the "view" training for programs it is not on all of them.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [All Volunteer trainings are not being added to all the programs correctly #721](https://trello.com/c/et6fKd8V/518-all-volunteer-trainings-are-not-being-added-to-all-the-programs-correctly-721)

Lyisus:
Working on this issue alongside issue #691

PR #775, opened 2022-10-06
Added a program column in event table and a CELTS-sponsored event program

gahimbaref:
#721 
- Added a column to the events table called program to get rid of the many-to-many relationship of the program event table
- Added another program called CELTS-Sponsored events to represent events that belong to no program or all programs
- Updated test data accordingly to accomodate for the new column
- Used the new Program column in the Event table to make it so that all trainings relevant to a program show up when hovering over the "view" link in the Program Table in the Volunteer Profile page

Files Changed: 
- database/test_data.py
- app/controllers/admin/routes.py
- app/logic/participants.py
- app/models/event.py
- tests/code/test_participants.py


app/logic/participants.py
```diff
@@ -115,23 +117,25 @@ def getEventParticipants(event):
...
     academicYear = currentTerm.academicYear
+    otherCeltsEventsProgram = Program.get(11) # This is the ID for the other celts sponsored events
+                                              # TODO: make a query that gets the program data without a need for the ID, in case the ID changes
+
+    programTrainings = (Event.select(Event, EventParticipant)
+                                .join(EventParticipant, JOIN.LEFT_OUTER)
+                                .where((Event.isTraining),
+                                        (Event.program == program) | (Event.program == otherCeltsEventsProgram),
+                                        Event.term == currentTerm,
+                                        (EventParticipant.user == user) | (EventParticipant.user == None)))
```
AndersonStettner: Why are you doing a .get and checking specifically in the where for the CELTS program? If you actually made it a program this should not be necessary since it will be included with the other programs.

app/logic/participants.py
```diff
@@ -115,23 +117,25 @@ def getEventParticipants(event):
...
-                               .join(EventParticipant, JOIN.LEFT_OUTER).switch()
-                               .join(ProgramEvent).switch()
-                               .join(Term)
-                               .where((Event.isTraining | Event.isAllVolunteerTraining),
-                                      ProgramEvent.program == program,
-                                      Event.term.academicYear == academicYear,
-                                      EventParticipant.user.is_null(True) | (EventParticipant.user == user)))
 
     userParticipatedEvents = {}
     for training in programTrainings.objects():
```
AndersonStettner, line 136: I am aware yall are not the ones who wrote the logic below but what is the point of it? If we are checking that the user is in the participants table in the initial query then it is implied they have already participated. You cannot be in the participants table without being entered manually, by and admin or program manager, or by swiping your ID at the kiosk when you attend events.

app/models/event.py
```diff
@@ -20,6 +20,7 @@ class Event(baseModel):
     recurringId = IntegerField(null=True)
     contactEmail = CharField(null=True)
     contactName = CharField(null=True)
+    program = ForeignKeyField(Program, null=True)
```
AndersonStettner, line 23: Are you adding this just to do Event.program? If so, why are you doing it this way instead of using ProgramEvent like we do everywhere else?

AndersonStettner (changes requested):
Remove the first "CELTS-Sponsored Event" on the create event page. It is still treated as a non-program event when you create events for it and we don't want it on there twice. Also, it might be better to rename it to something like "CELTS-Sponsored" or "CELTS-Sponsored Program" since it is now a program.

app/models/event.py
Lyisus: This is part of a refactor that Fleur and Luis are working on to get rid of the many-to-many relationship in the programEvent table for a many-to-one relationship using a Program column in the Event table. Issue #792 was made to keep track of the refactoring's progress. This should make it easier to figure out what event belongs to what program and "Celts-Sponsored Events" would work as the program events that do not belong to other programs belong to.

app/logic/participants.py
Lyisus: The .get is to get the program the All-Volunteer Trainings belong to. Then, the following query gets all events the user has participated in that belong to any given program OR are an All-Volunteer Training. Though, now I see we have a column that keeps track of which events are All-Volunteer Trainings. So, this query could be a bit clearer.

app/logic/participants.py
Lyisus: The purpose of the getUserParticipatedTrainings is to get a list of all trainings and figure out which of those trainings the inputted user has participated in. Specifically for the tool tip in the user profile that shows all the trainings necessary for a program. Before it failed to find All-Volunteer Trainings, but now it should be working.

app/logic/participants.py
sreynit02: on the .get I would not recommend hard-coding the id. Maybe use the program name instead.

AndersonStettner: This is still not getting the All-Volunteer trainings but it will get everything in the specific program and the trainings for the CELTS sponsored events program will appear.

AndersonStettner (changes requested):

app/logic/participants.py
```diff
@@ -115,31 +117,29 @@ def getEventParticipants(event):
...
-                                      ProgramEvent.program == program,
-                                      Event.term.academicYear == academicYear,
-                                      EventParticipant.user.is_null(True) | (EventParticipant.user == user)))
+    programTrainings = (Event.select(Event, EventParticipant)
+                                .join(EventParticipant, JOIN.LEFT_OUTER)
+                                .where((Event.isTraining),
+                                        (Event.program == program) | (Event.isAllVolunteerTraining),
+                                        Event.term == currentTerm))
 
     userParticipatedEvents = {}
```
AndersonStettner: I think we should change the name of this dict since it is the participated trainings.

AndersonStettner (changes requested):

tylerpar99:
Not sure if this is still active but there are currently merge conflicts with development.

PR #777, opened 2022-10-07
Change sorting order of events on Events List page.

AndersonStettner:
Requested change: sort events on the Events List page chronologically instead of alphabetically. 

Did: 
- Changed it so that events on the Events List page are sorted chronologically (by start date and start time).
- Moved the "Event Program" column to the first column and renamed it to just "Program".
- Added the order_by of the start time to the Upcoming Events section to a users profile.
- Added more detail on the RSVP'd column of a users Upcoming Events.
- Fixed the test that broke with the new chronological ordering.

To test: 
- With a freshly reset database run the tests.
- After starting the application create multiple events that are apart of the same program and make sure they are sorted by both start date and start time.
- Create a events that require the user to RSVP and check that the RSVP status on the User Profile matches the users RSVP'd status.


app/templates/events/event_list.html
```diff
@@ -57,34 +57,34 @@ <h1 class="text-center">Events List for {{selectedTerm.description}}</h1>
   <table class="table table-striped">
     <thead>
       <tr>
+        {% if type != "bonner" %}
+        <th>Program</th>
+        {% endif %}
         <th scope="col">Event Name</th>
         <th scope="col">Date</th>
         <th scope="col">Time</th>
```
sreynit02, line 65: Maybe consider changing this to Start Time instead of just Time?

sreynit02 (commented):
This PR looks good. I like how you can specified the different type of RSVP status. The events are ordered correctly on time and date.

app/templates/events/event_list.html
AndersonStettner: Since that column has both start time and end time I think it would be fine to keep it as it is.

sreynit02: that's true. I meant to request this change for the userProfile page since we only have the start time there.

BrianRamsay (approved):

PR #778, opened 2022-10-07
Add Notes to User Profile

### LINKED ISSUE #665: Add Notes on a users Volunteer Profile
Add a notes section to students profiles. 

They should have different levels of permissions when created:
- Only Admin can view
- Admin and Student Staff can view
- Everyone can view (that student, Admin, and Student Staff) 
- We also need a Bonner specific option

zachneill:
This fixes #665.

There is a notes section in users' profiles. There are 3 tiers of visibility:
* 1 --> Everyone
* 2 --> Student Staff and Admins
* 3 --> Admins

Additionally, there is a bonner scholars option. You cannot see them if you are not a bonner scholar. You can create and delete them with the ProfileNote table. For this PR to work you'll need the new table by resetting the database. There is also a noteType column in the Note table. 

To test:
Go to a user's profile. 
There is a notes tab and adding and deleting buttons.


AndersonStettner (changes requested):
The "Add Note" button is not working and there is not delete button for the notes. Also, if we are going to have all of the notes in the same table in the database then I think we might want to add one or two new columns to differentiate between ban/unban notes and the user profile notes. There is also a merge conflict with development and test_bonner.py is now failing.

AndersonStettner:
[Add Notes on a users Volunteer Profile #665](https://trello.com/c/3oKOOyxE/464-add-notes-on-a-users-volunteer-profile-665)

app/templates/main/userProfile.html
```diff
@@ -294,6 +295,112 @@ <h6>{{bgType.description}}</h6>
     {% endif %}
     <!-- End Background Check Table -->
 </div>
+<!-- Start Notes section -->
+  <div class="accordion-item">
```
AndersonStettner: This accordion item is not inside the main User Profile accordion so there is a small separation between the the Notes tab and the Background Check tab.

app/static/js/userProfile.js
```diff
@@ -61,6 +62,23 @@ $(document).ready(function(){
...
+
+  $(".deleteNoteButton").click(function() {
+    let username = $(this).data('username')
+    let noteid = $(this).data('noteid')
+    $.ajax({
+      method: "POST",
+      url:  "/" + username + "/deleteNote",
+      data: {"id": noteid},
+      success: function(response) {
+        location.reload();
```
AndersonStettner: This will reload the page and take the user off of the notes section. We want the user to stay in the notes section after the reload.

app/templates/main/userProfile.html
```diff
@@ -294,6 +295,112 @@ <h6>{{bgType.description}}</h6>
...
+    <h3 class="accordion-header" id="headingSix">
+      {% set focus = "open" if visibleAccordion == "notes" else "collapsed" %}
+      <button class="accordion-button {{focus}}" type="button" data-bs-toggle="collapse" data-bs-target="#collapseSix" aria-expanded="false" aria-controls="collapseSix">
+          Notes
+      </button>
+    </h3>
+    {% set show = "show" if visibleAccordion == "notes" else "" %}
+    <div id="collapseSix" class="accordion-collapse collapse {{show}}" aria-labelledby="headingSix" data-bs-parent="#userProfile">
```
AndersonStettner, line 312: Once you fix the page reload in the js, so that it reloads on the accordion, you will need to fix the focusing and opening of the accordions.
image

AndersonStettner (changes requested):

app/templates/main/userProfile.html
zachneill: fixed

app/static/js/userProfile.js
zachneill: fixed

app/templates/main/userProfile.html
zachneill: Fixed

app/static/js/userProfile.js
zachneill: Fixed

app/static/js/userProfile.js
```diff
@@ -61,6 +62,47 @@ $(document).ready(function(){
...
 
+  $("#addNoteButton").click(function() {
+    $("#noteModal").modal("toggle");
+  });
+
+  $("#bonnerInput").click(function() {
+    $("#bonnerInput").val(
+      $("#bonnerInput").val()== "on" ? "off" : "on"
+    )
+    console.log($("#bonnerInput").val())
```
bledsoef: Remove console.log()

app/static/js/userProfile.js
```diff
@@ -61,6 +62,47 @@ $(document).ready(function(){
...
+  });
+
+// action="/{{volunteer.username}}/addNote"  method="post"
```
bledsoef: Remove commented out code if its unnecessary.

bledsoef (dismissed):
It looks pretty good, besides the comments here and there, but we also want to make sure that when a user adds a Bonner Note and they are not a bonner, they can access it. We don't want them creating a note and then wanting to view or delete it and not being able to find it anywhere.

Try harder next time.

AndersonStettner (changes requested):
The Notes accordion is not behaving as it should after the note is submitted and the page is reloaded.

zachneill:
> The Notes accordion is not behaving as it should after the note is submitted and the page is reloaded.

Could you elaborate on this?

AndersonStettner:
> > The Notes accordion is not behaving as it should after the note is submitted and the page is reloaded.
> 
> Could you elaborate on this?

Yes, currently, after a note is submitted and the page is reloaded it looks like this:
image

The "Upcoming Events" accordion is the one that is focused and opened but it should be closed and the "Notes" accordion is the one that should be focused and open like this: 
image

AndersonStettner:
Another thing I noticed while getting those screenshots: I think all of the accordions are broken. When you indicate an interest on the "Programs" accordion it should be reloaded and have that accordion open but nothing happens, the page isn't reloaded.

AndersonStettner (changes requested):
It's looking good but a few things:
- The accordion is still a little messed up after reloading from notes 
- The "Program" table is not taking in a users interest 
- A normal user is not able to add any notes to their own profile
- Also as a suggestion, if there are no notes yet why don't we just show the user a sentence like "There are no notes yet"

AndersonStettner (changes requested):
It looks good and everything seems to be working well, good job. Though, Brian and I are not a big fan of how the Notes accordion broken down into different sections. We think it would be better to do it like we have it in LSF, just show all the notes together in a list, say who made them, and tell who can view it/why it was made. 

Also, while only Bonners will be able to see the Bonner notes the check to indicate that it is a Bonner note, I think that it might read better if it is labeled something like "Bonner Info" or "For Bonners".

app/controllers/main/routes.py
```diff
@@ -125,11 +127,46 @@ def viewUsersProfile(username):
...
             )
     abort(403)
 
+@main_bp.route('/profile/addNote', methods=['POST'])
+def addNote():
+    """
+    This function adds a note to the user's profile.
+    """
+    postData = request.form
+    try:
```
Andersonstettner, line 141: Can replace the whole first try and except here with a check for Bonner in postData since the value only ever exists when it is "on".

Andersonstettner (changes requested):

AndersonStettner (changes requested):
Also, the accordions are still a little messed up. 

The "Upcoming Events" accordion should be closed after a note is submitted and the page is reloaded.
image

AndersonStettner (changes requested):
The check "For Bonners" should only appear on profiles of Bonner students, not on everyone.

Karina-Agliullova:
image
When adding a note for a student, the student is not able to delete their own note.

Karina-Agliullova (approved):
Good job! All the comments were addressed!

AndersonStettner (changes requested):
Looks good, one last thing, the "For Bonner" check should only appear on the profiles of Bonners, so if they are not Bonners students they should not have the option to make it for a Bonner.

zachneill:
> Looks good, one last thing, the "For Bonner" check should only appear on the profiles of Bonners, so if they are not Bonners students they should not have the option to make it for a Bonner.

I thought we'd fixed that in the recent changes. In the jinja we've got: 
if g.current_user.isBonnerScholar and volunteer.isBonnerScholar 
I believe this works

AndersonStettner:
> > Looks good, one last thing, the "For Bonner" check should only appear on the profiles of Bonners, so if they are not Bonners students they should not have the option to make it for a Bonner.
> 
> I thought we'd fixed that in the recent changes. In the jinja we've got: if g.current_user.isBonnerScholar and volunteer.isBonnerScholar I believe this works

Ah, you are right, my bad, I did not realize Brian was in as a Bonner.

AndersonStettner (approved):

tests/code/test_bonner.py
```diff
@@ -24,22 +24,22 @@ def test_getBonnerCohorts():
         BonnerCohort.create(user="lamichhanes2", year=currentYear-6)
         cohorts = getBonnerCohorts()
         assert len(cohorts) == 7
-        assert len(cohorts[currentYear]) == 0
+        assert len(cohorts[currentYear]) == 2
```
bledsoef, line 27: This test fails.

bledsoef (changes requested):

app/static/js/userProfile.js
```diff
@@ -61,6 +61,45 @@ $(document).ready(function(){
...
+  $('#addNoteForm').submit(function(event) {
+    event.preventDefault()
+    let username = $("#notesSaveButton").data('username')
+    $.ajax({
+      method: "POST",
+      url:  "/profile/addNote",
+      data: {"username": username,
+             "visibility": $("#noteDropdown").val(),
+             "noteTextbox": $("#addNoteTextArea").val(),
+             "bonner": $("#bonnerInput").val()=="on"? "on":"off"},
```
BrianRamsay, line 83: Is there a reason to keep using 'on' and 'off' instead of just passing true or false to the controller?

tests/code/test_users.py
```diff
@@ -103,9 +104,37 @@ def test_removeUserInterestt():
...
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_deleteUserProfileNote():
     with mainDB.atomic() as transaction:
+        with app.app_context():
+            g.current_user = "ramsayb2"
+
+            addedNote = addProfileNote(1, True, "Test profile note", "neillz")
+            profileNote = deleteProfileNote(addedNote)
```
BrianRamsay, line 124: We should test once between these to make sure that the note actually exists in the db with the content we expect.

BrianRamsay (changes requested):
UI Changes:
The Bonner flag should come first, as the visibility options are slightly different depending on this value (see below).

For student leaving a note on their profile, the Visibility heading and dropdown should be replaced with a notice that "This note will only be visible to you, CELTS Student Staff, and Administrators." (or "you and Bonner administrators", maybe, for a Bonner note) (done)

"For Bonners" should be changed to something like "Related to Bonner Scholars Program" (done)

Student Staff don't need to see the Bonner checkbox option at all. (done? the only reason the Student Staff is able to see Bonner checkbox is because they are in as a Bonner by default) 

"Visible By" in the table should be "Visible To" (done)

The table and dropdown are not consistent with their role separators. Pick "," or "/". (done)

Something weird is going on with the Bonner flag and visibility. I don't think there's a reason to change the visibility text, when they are effectively the same visibility levels. What are "Bonner Admins"? Is there a reason we don't keep the visibility text the same, like "Admins" or "Admins/Student Staff" or "Everyone"? That being said, bonner notes aren't visible to student staff, so there is a 4th option that we need that is Admins/Bonner Scholar. That option would only be available to administrators leaving a bonner note, as they could choose between "Admins Only" and "Admins/Bonner Scholar".

app/templates/main/userProfile.html
```diff
@@ -237,102 +238,215 @@ <h5>{{volunteer.firstName}} {{volunteer.lastName}} is the manager of:</h5>
...
-      <div class="modal-content">
-        <div class="modal-header">
-          <h5 class="modal-title">Ban Volunteer</h5>
-          <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+  </div>
+  {% endif %}
+  <!-- End Background Check Table -->
+  <!-- Start Notes section -->
+  <div class="accordion-item" id="accordion5">
+    <h3 class="accordion-header" id="headingSix">
```
AndersonStettner, line 305: Why are you giving it the div id accordion5 and the accordion header id headingSix? It is the sixth accordion on the page.

AndersonStettner:

BrianRamsay:
Merged with mega-branch

PR #779, opened 2022-10-10
Food check box

### LINKED ISSUE #757: Add a new check box when creating an event to say that food will be provided during the event.

thawpt:
Did:
-added a "Food is Provided" switch in the create event page that would allow users to switch it if food is provided.
-added foodProvided to  events.py and the event modal so that it  saves to database
-added conditional in eventView.html to show user this event provided food

Test:
-create an event, fill in necessary information, switch the "Food is provided" switch
- Save the event, go to "View Event"

Fixes issue #757


AndersonStettner:
[Add a new check box when creating an event to say that food will be provided during the event. #757](https://trello.com/c/smsFmGkY/548-add-a-new-check-box-when-creating-an-event-to-say-that-food-will-be-provided-during-the-event-757)

AndersonStettner (changes requested):
Revert the changes to userProfile.html.

app/templates/admin/createEvent.html
```diff
@@ -161,18 +161,34 @@ <h1 id="title">{{page_title}}</h1>
...
-                <input class="form-check-input" type="checkbox" id="rsvp" name="isRsvpRequired" {{"checked" if eventData.isRsvpRequired}}/>
-              </div>
-              <div class="form-check form-switch mb-4">
-                <label class="custom-control-label" for='earnServiceHours'><strong>This event earns service hours.</strong></label>
-                <input class="form-check-input" type="checkbox" id="earnServiceHours" name="isService" {{"checked" if eventData.isService}}/>
+            <div class="pt-4 mb-4">
+              <div class="row col-md-12">
+                <div class="form-group col-md-6">
+                  <div class="form-check form-switch mb-2">
+                    <label class="custom-control-label" for="food"><strong>Food is provided.</strong></label>
```
AndersonStettner: I think it would be nice if we could try and have this label match the others. Maybe something like "This event will have food provided." or "This event will provide food."

app/templates/eventView.html
```diff
@@ -37,6 +37,9 @@ <h1 id="title">{{page_title}}</h1>
             <div class="row">
                 <div class="col-md-6">
                     <h3><br>Description</h3><p>{{eventData.description|urlize}}</p><br>
+                    {% if eventData.foodProvided %}
+                        <p>This event provides food.</p>
```
AndersonStettner: Opposite of my comment above, I think it might look better if we change the wording here so it is not much like the other statements. Maybe something like "Food will be provided at this event." or "There will be food provided at this event."

AndersonStettner (approved):

PR #782, opened 2022-10-12
Banned permissions

### LINKED ISSUE #657: Banned Volunteer permissions
Banned volunteers should:

- NOT see upcoming events when they are banned from that program. They will still be able to see the program events on the Events List page but they will not appear as upcoming events on their Student Profile. 
- NOT be able to RSVP (pretty sure this is already done)

AndersonStettner:
Issue: If a user is banned from a program do not show events from that program in their Upcoming Events on their User Profile. 

Did: Added the bannedEvents query to getUpcomingEventsForUser. Compared the list that is gotten from the new query to events_list and removed all events from events_list that also appeared in the bannedEvents list. Added the new test cases to the upcoming events test.

To test:
- Run the test suite.
- Create events under different programs that will happen within the day, then, go look at the user profile.
- Mark an interest in the programs you created events in and make sure they appear in the upcoming events section.
- Ban the person you are looking at from the programs you have created events under and make sure those events no longer appear in their upcoming events section.
- Unban the person you are looking at and make sure the events return to their upcoming events section.

### DIFF
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -251,10 +251,11 @@ def getUpcomingEventsForUser(user, asOf=datetime.datetime.now()):
                        .join(EventRsvp, JOIN.LEFT_OUTER, on=(Event.id == EventRsvp.event))
                        .where(Event.startDate >= asOf,
                               (Interest.user == user) | (EventRsvp.user == user),
-                              ProgramBan.user.is_null(True) | (ProgramBan.endDate.is_null(False) & (ProgramBan.endDate < asOf)))
+                              ProgramBan.user.is_null(True) | (ProgramBan.endDate < asOf))
                        .distinct() # necessary because of multiple programs
                        .order_by(Event.startDate, Event.name)
                      )
+
     events_list = []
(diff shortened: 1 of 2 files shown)


AndersonStettner:
[Banned Volunteer permissions #657](https://trello.com/c/Sv7YE0f9/456-banned-volunteer-permissions-657)

app/logic/events.py
```diff
@@ -253,6 +254,14 @@ def getUpcomingEventsForUser(user, asOf=datetime.datetime.now()):
                     .order_by(Event.startDate, Event.name).execute() # keeps the order of events the same when the dates are the same
                     )
 
+    bannedEvents = list(Event.select()
```
bledsoef: I think we could benefit from putting this in its own logic function and testing separately as I can think of times in the future where this may be handy to have as a logic function.

bledsoef (changes requested):

bledsoef (dismissed):

BrianRamsay (approved):

Issue #755, opened 2022-09-16
Make the location of Events auto populate the same way Event contact info does. 


AndersonStettner:

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Make the location of Events auto populate the same way Event contact info does. #755](https://trello.com/c/Udq147N1/546-make-the-location-of-events-auto-populate-the-same-way-event-contact-info-does-755)

PR #783, opened 2022-10-14
Auto Populate Event Location 

sreynit02:
fixes PR #755 

Did: 
- added a column called 'programLocation' to Program. 
- On program management page, we added a 'location' column so that admin can update the location of a program. 
- Populate the location for an event of the event creation page to be the same as the location set for a program
- fixed tests

Test: 
- reset database and run tests. no test should fail
- go to Admin>Settings>Program Managment: Set a location for a program
- create event for that program from above. On the create event page you should see the location has been populated with what you had entered for the program location.


app/logic/userManagement.py
```diff
@@ -70,12 +70,12 @@ def addNextTerm():
...
 
-def changeProgramInfo(newProgramName, newContactEmail, newContactName, programId):
+def changeProgramInfo(newProgramName, newContactEmail, newContactName, newLocation, programId):
     """Updates the program info with a new sender and email."""
     program = Program.get_by_id(programId)
-    updatedProgram = Program.update({Program.programName:newProgramName,Program.contactEmail: newContactEmail, Program.contactName:newContactName}).where(Program.id==programId)
+    updatedProgram = Program.update({Program.programName:newProgramName,Program.contactEmail: newContactEmail, Program.contactName:newContactName, Program.programLocation:newLocation}).where(Program.id==programId)
     updatedProgram.execute()
-    createLog(f"{program.programName}'s settings changed to: Name: {newProgramName}; Reply-to-email: {newContactEmail}; Sender name: {newContactName}.")
+    createLog(f"{program.programName}'s settings changed to: Name: {newProgramName}; Reply-to-email: {newContactEmail}; Sender name: {newContactName}; location: {newLocation}")
```
AndersonStettner: I am not a big fan of how this looks on the page. I think it might be better to make it so that the admin log only logs the fields that were changed so if the email was changed the log would read something like program.programName's email was change to: hellothere@gmail.com.

app/controllers/admin/routes.py
```diff
@@ -91,9 +91,10 @@ def createEvent(templateid, programid=None):
         eventData['contactName'] = "CELTS Admin"
         eventData['contactEmail'] = app.config['celts_admin_contact']
         if program:
-            if program.contactName and program.contactEmail:
+            if program.contactName or program.contactEmail or program.programLocation:
```
bledsoef: The issue with doing "or" like this is that if they have a location, but no contactEmail or contactName, it will default to an empty string, this is a problem because just above that line we want to make sure if there is no contactEmail or contactName it defaults to "CELTS Admin" and the default CELTS contact.

bledsoef (changes requested):
The PR looks good besides the change I requested, the only other comment is a suggestion to maybe change program.programLocation to program.location, only because it feels a little redundant and matches formatting of program.contactEmail and program.contactName.

app/controllers/admin/routes.py
sreynit02: > 

Made another if statement that detects if program.programLocation is changed. This way contactEmail and contactName will not be empty strings when only program.programLocation is changed.

tests/code/test_userManagement.py
```diff
@@ -67,21 +67,24 @@ def test_changeProgramInfo():
         eventName = "Test Event Name"
         contactName = "New Test Name"
         contactEmail = 'newtest@email'
+        location = "Danforth Tech"
         currentProgramInfo = Program.get_by_id(programId)
 
         assert currentProgramInfo.programName == "Adopt-a-Grandparent"
         assert currentProgramInfo.contactName == ""
         assert currentProgramInfo.contactEmail == ""
+        assert currentProgramInfo.contactEmail == ""
```
AndersonStettner: Is this supposed to be .programLocation?

AndersonStettner (changes requested):
It looks good. I still think we might want to change how the admin logs look when we update something in the program management section.

app/logic/userManagement.py
sreynit02: So if all the fields were changed, you would prefer the fields to be logged separately or just one?

tests/code/test_userManagement.py
sreynit02: fixed

app/logic/userManagement.py
AndersonStettner: If they are all changed at the same time, I think it would be fine to show them all together but I think we would still want to change the wording in that case.

AndersonStettner (dismissed):
Everything looks good, good job.

martinj2-dot (dismissed):
Looks good!

Karina-Agliullova (dismissed):
Good work!

app/models/program.py
```diff
@@ -10,6 +10,7 @@ class Program(baseModel):
     isBonnerScholars = BooleanField(default=False)
     contactName = CharField(null=True,default='')
     contactEmail = CharField(null=True,default='')
+    programLocation = CharField(null=True,default='')
```
BrianRamsay: I think this should be `defaultLocation` instead. There isn't really a location for each program, there is a default location that events can override.

BrianRamsay (changes requested):

app/models/program.py
sreynit02: fixed!

qasema:
Looks Great

qasema (approved):

bledsoef (approved):
Looks great!

BrianRamsay (approved):

PR #793, opened 2022-10-24
Can now create nonProgram events

AndersonStettner:

Issue:
When a user attempted to create a non-program event they were thrown an error from the dict object since it was looking for program in the event data but since there is no program in non-program events the didn't know what to do. 

Did:
Added a new conditional to check whether or not there was a program in the event data. If there was a program then continue on to the existing logic and if there was no program then it would just skip it. 

To test:
- Run the test suite with a freshly reset database and make sure all tests pass (things were only changed in the HTML so the  had better pass) 
- Run the application and make sure that you are able to create non-program events
- With the application running verify that the requirements dropdown appears as it should on the Bonner event creation page but not on any of the other pages.

### DIFF
--- a/app/templates/admin/createEvent.html
+++ b/app/templates/admin/createEvent.html
@@ -158,15 +158,17 @@
         <div class="mb-4">
           <div class="row col-md-12" >
             <div class="form-group">
-            {% if eventData['program'].isBonnerScholars %}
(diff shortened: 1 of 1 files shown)


AndersonStettner:
[Can now create nonProgram events](https://trello.com/c/KXbYNJkt/572-can-now-create-nonprogram-events)

BrianRamsay (approved):

PR #796, opened 2022-11-09
The time when a user RSVPd to an event is now shown on the participants table.

AndersonStettner:

Issue: Since there are events that are first come first serve CELTS needed a way to see when users had RSVPd. 

Did: Added a new column to the EventRsvp model and added the RSVP Time column to the participants table which displays the volunteers in the order they RSVPd for that event. 

To test:
- With a fresh database run the tests and verify they all pass.
- Run the application and create a new event in the future that requires an RSVP.
- Add users to the participants table and make sure they are appearing in the order they are added. 
- Wait a minute or two and add another user to the participant table and verify they are added at the bottom of the table.

### DIFF
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -41,7 +41,7 @@ def trackVolunteersPage(eventID):
     if not (g.current_user.isCeltsAdmin or (g.current_user.isCeltsStudentStaff and isProgramManager)):
         abort(403)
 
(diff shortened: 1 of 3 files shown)


AndersonStettner:
[The time when a user RSVPd to an event is now shown on the participants table. #796](https://trello.com/c/GtUL3qdJ/573-the-time-when-a-user-rsvpd-to-an-event-is-now-shown-on-the-participants-table-796)

sreynit02 (approved):
Looks good to me!

BrianRamsay (approved):

PR #798, opened 2022-11-17
New CELTS Landing Page

### LINKED ISSUE #756: Create a new home/landing page
Create a new home/landing page that will have all of the programs on them with short blurbs/descriptions. The name of the Program on the landing page should link to the Events List page. Tell who the Program Manager of each program.

zachneill:
This fixes #756 which was to create a new landing page. It has each program, their program manager, a photo, and the description. The photos are in static/images/landingPage. The descriptions are a property in the program model. You can click "Read more" for a modal with the full description. The "View Events" button takes the user to the tab "Student Led Service," "Bonner Scholars," or "Other Events."

To test: 
- go to http://YOUR.IP.ADDRESS:8080
- Test the buttons and new functionality.

Note: You may have to do source setup.sh and ./reset database test to get the new changes.

### DIFF
--- a/app/templates/main/landingPage.html
+++ b/app/templates/main/landingPage.html
@@ -61,7 +61,7 @@
           <img id="modalImage" src="" class="img-fluid rounded" alt="program-image">
         </div>
         <p class="mt-3" id="modalDescription"></p>
-        <a type="button" id="modalLearnMore" class="float-end link-primary" href="">Learn more</a>
+        <a type="button" id="modalLearnMore" class="float-end link-primary" target="_blank" href="">Learn more</a>
       </div>
       <div class="modal-footer justify-content-between">
         <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Close</button>


AndersonStettner (changes requested):
When a user clicks "Read More" I think we might want to move the "Close" button to the left and put a "Go" button on the right where the close button is currently. Also, I think it would be nice if, when a user clicks "Go", and the program events are inside an accordion, we show that page with the accordion open. I also feel like we would want to have the cover photos be the same size so we don't have such a difference in the sizes of the program cards. 

Also, not a needed change but I think it might be nice if we could incorporate the CELTS logo in some way on the landing page and also possibly Berea's logo. As long as we can make them not look out of place.

qasema (changes requested):
I think we still need to include the "Go" button in the Modal. we should also display the "read more" button only when the description text is over a certain word limit. The way the "read more" button works right now is that it is displayed even if the text is one short sentence.

AndersonStettner (commented):
- “Go” to “View Events”

- Standardize the picture size/crop the pictures that are uploaded so they are all the same size.

- Make sure the picture is centered when it is in the modal. 

- Possibly link to the Berea CELTS website for more information? Add it at the end of the text in the modal something like “Learn more!”.

- Turn the name of the program into a link and have it act like a second “View Events” button. 

- Add a footer to the cards and make the “View Events” button centered at the bottom of the card

- Make where “Read more” appears the same across all events. 

- If there are no events scheduled for a program in that semester, we should either not allow the user to press the “View Events” button or, have all the program accordions always be shown in the events list so the button redirects to an empty accordion that tells the user there are no events scheduled yet. 

- Remove "Link" from title of the page.

- Also, something we need clarification on: Do they want us to put Bonner on the landing page and if they do, we will probably want to remove the "View Events" button on that card since most users can't participate in Bonner events.

AndersonStettner (commented):
The page is looking really good, two UI things:
1. I think we should change the wording when there is no event created yet for the program "No Events This Term" seems definitive and might make some users think there will be no events through the term for the rest of the semester, something like "Currently, there are no events planned". 
2. If a student is a Bonners then we should put the view events button in the card. When I am go to the landing page as a Bonners student the card looks no different than if I were to look at it as a normal student.

AndersonStettner (commented):
Personally, I would say to either make it "Currently, no events are planned." or "No events planned currently."

zachneill:
> Personally, I would say to either make it "Currently, no events are planned." or "No events planned currently."

changed to No Events Planned Currently

AndersonStettner (changes requested):
When the user clicks "Read more" and is taken to the CELTS website we should open a new tab and open it there instead of doing it in the same tab.

AndersonStettner:

AndersonStettner (changes requested):
If the user is not a Bonner or an Admin then we should not show the Bonners Scholars card at all.

BrianRamsay:
Merged with mega-branch

Issue #758, opened 2022-09-16
Add the ability to for a user to put in their dietary preferences and prompt them for it on some cases 


AndersonStettner:
Allow a user to input their dietary preferences somewhere on their User Profile. 
In the case that a User is RSVPing for an event that has had the "food will be provided" option checked and they have not given dietary preferences prompt them to enter them.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Add the ability to for a user to put in their dietary preferences and prompt them for it on some cases #758](https://trello.com/c/aqU7cbe0/549-add-the-ability-to-for-a-user-to-put-in-their-dietary-preferences-and-prompt-them-for-it-on-some-cases-758)

PR #802, opened 2022-11-29
User diet

thawpt:
Addresses Issue #758 
Did:
-Added a new table to the database (dietaryrestriction)
- Added  "Dietary Restriction" to the the modal in My Profile
- Added input field where user is able to type in their dietary restriction as well as update it as needed.

Test:
- Go to My Profile
- Click on Dietary Restriction
- Type in your restrictions, then hit save
- Run test


app/static/js/userProfile.js
```diff
@@ -158,7 +158,29 @@ $(document).ready(function(){
         }
     });
 
-    setupPhoneNumber("#updatePhone", "#phoneInput")
+  $("#updatePhone").on('click', function() {
+    var username = $(this).data("username")
+    validatePhoneNumber(this, "#phoneInput", username)
+  });
```
AndersonStettner: This is old code that was carried over. It breaks the phone number handler.

app/models/dietaryRestriction.py
```diff
@@ -0,0 +1,6 @@
+from app.models import *
+from app.models.user import User
+
+class DietaryRestriction(baseModel):
+    user = ForeignKeyField(User)
+    dietRestriction = CharField()
```
AndersonStettner: Make this a text field

AndersonStettner (changes requested):

app/static/js/userProfile.js
```diff
@@ -158,7 +158,29 @@ $(document).ready(function(){
...
+    let data = {
+      dietInfo: $("#diet").val(),
+      user: $(this).data("user")
+    }
+    $.ajax({
+      type: "POST",
+      url: "/updateDietInformation",
+      data: data,
+      success: function(s){
+        location.reload();
```
AndersonStettner: use reloadWithAccordion() instead of location.reload()

AndersonStettner (changes requested):
The accordion focusing is messed up. With accordion=dietartyInformation the "Dietary Restrictions" accordion is focused and closed while the "Upcoming Events" accordion is focused and open. The "Upcoming Events" accordion should be closed and not focused and the "Dietary Restrictions" accordion should be focused and open. 
image

app/models/dietaryRestriction.py
sreynit02: fixed

app/static/js/userProfile.js
sreynit02: fixed

app/static/js/userProfile.js
sreynit02: deleted,

AndersonStettner (changes requested):
Dietary restrictions does not need its won model, we can just add it as a field on the user since it is a one to one mapping.

sreynit02:
> Dietary restrictions does not need its won model, we can just add it as a field on the user since it is a one to one mapping.

Deleted UserDiet table and added a field to User table for diet information

AndersonStettner:

sreynit02:
> Any further work of this issue/pull request should be done on PR #819

Nope

solijonovam (changes requested):
test_event_list.py  is failing.

BrianRamsay:
Merged with mega-branch

Issue #768, opened 2022-09-26
End Time is not checked when creating recurring events


sreynit02:
At the moment, when creating a recurring event, if the end time of the event is the same as the start time, the application does not ask the user to fill in a different time.

Andersonstettner:
Example:
![image](https://user-images.githubusercontent.com/26262045/193613467-ef0127c5-4db3-4445-8684-e005dd11c3a0.png)
![image](https://user-images.githubusercontent.com/26262045/193613541-02b44875-9623-4974-8bd6-592b7b63856f.png)

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [End Time is not checked when creating recurring events #768](https://trello.com/c/0s7oG47P/586-end-time-is-not-checked-when-creating-recurring-events-768)

PR #803, opened 2022-11-29
Time check

thawpt:
Addresses issue #768 
Did:
- Check to make sure start time is not after end time
- Made sure start time is not the same as the end time
- Added a clearer error message
Test:
- Create an event- make it recurring, make end time a time before the start time
- Once you try to create event, the error message "Event end time must be after start time." should pop up 
- Create an event- make it recurring, make end time the same as start time
- Once you try to create event, the error message "Event end time must be after start time." should pop up


AndersonStettner (changes requested):
Make sure that if the event is not created due to one of the blockers (ie: the start and end time are the same) when the page is reloaded to show the error message the event is still marked as a recurring event.

AndersonStettner:
Almost, after reload the recurring switch is still checked but the "End Date" box does not appear.
image

app/templates/admin/createEvent.html
```diff
@@ -114,7 +115,8 @@ <h1 id="title">{{page_title}}</h1>
                     </div>
                 </div>
             </div>
-            <div class="form-group col-md-6 d-none" id ="endDateStyle">
+            {%if eventData.isRecurring == True and isNewEvent%}
```
AndersonStettner: Set a variable to that will add or remove d-none depending on the condition so we don't have to repeat everything kind of like how we show and focus accordions.

AndersonStettner (changes requested):

app/templates/admin/createEvent.html
sreynit02: fixed!

AndersonStettner:
[End Time is not checked when creating recurring events #768](https://trello.com/c/0s7oG47P/586-end-time-is-not-checked-when-creating-recurring-events-768)

app/static/js/createEvents.js
```diff
@@ -34,16 +34,48 @@ function format24to12HourTime(timeStr){
...
+    calculateRecurringEventFrequency();
+  }
+
     $("#attachmentObject").fileinput({
         allowedFileExtensions:["pdf","jpg","png","gif", "csv", "docx", "jpg", "jpeg", "jfif"]
     })
   // Disable button when we are ready to submit
   $("#saveEvent").on('submit',function(event) {
-      $(this).find("input[type=submit]").prop("disabled", true)
+      debugger;
```
AndersonStettner: Remove debug code

AndersonStettner (changes requested):

app/static/js/createEvents.js
qasema: @Andersonstettner removed

app/static/js/createEvents.js
```diff
@@ -34,16 +34,47 @@ function format24to12HourTime(timeStr){
...
+      });
+
+  }
 /*
  * Run when the webpage is ready for javascript
  */
 $(document).ready(function() {
+  if ( $("#startDatePicker").val() && $("#endDatePicker").val()){
+    calculateRecurringEventFrequency();
+  }
```
AndersonStettner, line 70: Remove this check, it does not have any actual use and it breaks the code.

AndersonStettner (changes requested):

app/static/js/createEvents.js
AndersonStettner: Keep check but change it to check if the start and end dates are different instead

app/templates/admin/createEvent.html
```diff
@@ -114,7 +115,14 @@ <h1 id="title">{{page_title}}</h1>
                     </div>
                 </div>
             </div>
-            <div class="form-group col-md-6 d-none" id ="endDateStyle">
+
+            {%if eventData.isRecurring == True and isNewEvent%}
```
Karina-Agliullova: Add space after and before the "%" sign when using jinja in all jinja code

Karina-Agliullova (commented):
Everything is fine. It just looks messy when spaces are not used in jinja code. so please include them. Other than that, good job!

Karina-Agliullova:
Screenshot (1286)_LI
Screenshot (1287)_LI
When creating an event and setting it as recurring, after clicking save and saving it successfully, the end date option disappears. Also, if you check the saved recurring event on the events list page and try to edit one of the upcoming events, the end date option is also missing on the edit event page. Investigate why it occurs and fix the issue

Karina-Agliullova (approved):
Good work!

AndersonStettner (approved):
Looks good, good job!

PR #822, opened 2023-01-09
Rework SLC Proposal Forms

### LINKED ISSUE #820: Refactor Service-Learning forms
- Make our version of the Service-Learning proposal match the original proposal form on the Berea website. This will include rewording and adding in new questions. 

- Make sure to add the text blurbs/descriptions that are at the start of each section.  

- On the middle page of the proposal (where the user puts in the instructor, course name, etc.) add a place for the instructor to upload a copy of the syllabus for the course. 

- Before the user begins working on a proposal we should make sure they know that they are able to save and come back to the form (as long as they have a course instructor and course name filled in) and we should also give them an option to look over the entire form to read through it (link out to it on the Berea web page?).

- Add a download option to the dropdown menu.

- If a user attempts to submit a proposal without completing everything inform the user of what they are missing.

### LINKED ISSUE #832: Change wording/messaging for SLC statuses.
Some of the feedback given is that our messaging about a courses status is confusing, especially the "Incomplete" status. Change "Incomplete" to "In Progress" and think about other possible changes that could be made.

### LINKED ISSUE #843: Cannot exit SLC Proposal form from the last page when viewing
There is no exit button on the last page of the SLC Proposal form when viewing. An exit button should be added so you don't have to click back through to the beginning.

AndersonStettner:
Fixes issues #843, #832 and #820 
## Changes
- Reworked FileHandler so that it works with events and courses
- Improved the middle page of the SLC Proposal/made it better match the original document. This includes:
1. changed how isRegularlyOccurring works and added isPreviouslyApproved, previouslyApprovedDescription, and hasSlcComponent fields to the Course model.
2. Adding a place for the user to upload a copy of the course syllabus 
3. Adding a new "Save and Exit" button on the middle page of the SLC proposal and renamed "Save for Later" to "Save and Exit" on the last page
4. Added a "Exit" button on the last page of the SLC proposal when a used the "View" action
- Spaced the SLC questions out from each other on the last page of the proposal
- Added missing text to both the first and last pages of the SLC proposal
- Added a new "Print" option to the action dropdown to let a user download/print their proposal (this option was also added to the Admin "Course Management" page
- Added a text blurb to the Faculty "Course Proposal" page that, among other things, lets the user download a copy of the official proposal form from CELTS
- Combined "courseManagement.js" and "slcManagement.js" files
- Changed the course status message from "INCOMPLETE" to "IN_PROGRESS" and added a hover to them when they appear in the table (the "Status" section) to let the user know what they mean
- Rewrote test to match all the changes we made
- Made user that if a course is renewed through the "Renew" action then "isPreviouslyApproved" is checked by default
- Added test data to make sure all fields on the SLC questionnaire can be edited.

## Testing
- Pull the branch and make sure it is up to date
- Reset your database and run the tests to verify they all pass
- Run the application with "flask run" 
- Verify you are still able to attach a file to an event and that you are also able to delete an event attachment
- Navigate to the "Course Proposals" page
- Verify that you are able to download a copy of the SLC proposal form out of the "here" hyper link in the text blurb
- Hover over the Status of any of the existing proposals and make sure a hover pops up and that its contents match what it is attached to
-  Click the "Create Course Proposal" button and go through the proposal process. 
1. On the middle page of the proposal verify all of the new options work and save to the database as expected
2. Upload a file to the syllabus section and make sure it is actually uploaded
3. Delete the file you have uploaded and make sure it is actually deleted
5. On the middle page use the "Save and Exit" button and verify it has saved the changes made to that middle page when you return to it 
6. Once you finish the proposal test both the "Save and Exit" and "Submit" buttons on the last page and make sure any changes are saved and the status that is displayed on the table matches the action you took (Save for Later the status should be "In Progress" and Submit should have a status of "Submitted")
7. Using the Action drop-down on the course table verify that when you use the "Print" action, a new window is opened with a print screen (don't actually print the form, set the "Destination" as "Save as PDF") and after click "Print" or "Close" inside the print screen that new window is closed and that file is downloaded
8. Using the "View" action from the drop-down, go through all the pages and make sure you are not able to edit any fields and that all the buttons correspond to what they should when a user is viewing a proposal

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -9,6 +9,8 @@ class User(baseModel):
     firstName = CharField()
     lastName  = CharField()
     isStudent = BooleanField(default = False)
+    major = CharField(null = True)
+    classLevel = CharField(null = True)
     isFaculty = BooleanField(default = False)
     isStaff = BooleanField(default = False)
     isCeltsAdmin = BooleanField(default  =False)
--- a/database/import_users.py
+++ b/database/import_users.py
@@ -34,7 +34,7 @@ def addToDb(userList):
 
         except peewee.IntegrityError as e:
             if user['username']:
-                (User.update(firstName = user['firstName'], lastName = user['lastName'], email = user['email'])
+                (User.update(firstName = user['firstName'], lastName = user['lastName'], email = user['email'], major = user['major'], classLevel = user['classLevel'])
                      .where(user['bnumber'] == User.bnumber)).execute()
             else:
                 print(f"No username for {user['bnumber']}!", user)
@@ -58,6 +58,8 @@ def getFacultyStaffData():
             "isStudent": False,
             "isFaculty": True,
             "isStaff": False,
+            "major": None,
+            "classLevel": None,
           }
         for row in c.execute('select * from STUSTAFF')
     ]
@@ -76,6 +78,8 @@ def getStudentData():
             "firstName": row[2].strip(),
             "lastName": row[3].strip(),
             "isStudent": True,
+            "major": row[6].strip(),
+            "classLevel": row[4].strip()
           }
         for row in c.execute('select * from STUDATA')
     ]
--- a/database/test_data.py
+++ b/database/test_data.py
@@ -45,7 +45,10 @@ users = [
         "isStaff": True,
         "isCeltsAdmin": True,
         "isCeltsStudentStaff": False,
-        "dietRestriction": "Diary"
+        "dietRestriction": "Diary",
+        "major": None,
+        "classLevel": None,
+
     },
     {
         "username" : "khatts",
@@ -58,6 +61,8 @@ users = [
         "isFaculty": False,
         "isCeltsAdmin": True,
         "isCeltsStudentStaff": False,
+        "major": None,
+        "classLevel": None,
     },
     {
         "username": "neillz",
@@ -70,6 +75,8 @@ users = [
         "isFaculty": False,
         "isCeltsAdmin": False,
         "isCeltsStudentStaff": True,
+        "major": None,
+        "classLevel": None,
     },
 
     {
@@ -83,6 +90,8 @@ users = [
         "isFaculty": True,
         "isCeltsAdmin": False,
         "isCeltsStudentStaff": False,
+        "major": None,
+        "classLevel": None,
     },
     {
         "username" : "ayisie",
@@ -95,6 +104,8 @@ users = [
         "isFaculty": False,
         "isCeltsAdmin": False,
         "isCeltsStudentStaff": False,
+        "major": None,
+        "classLevel": None,
 
     },
     {
@@ -108,6 +119,8 @@ users = [
         "isFaculty": False,
         "isCeltsAdmin": False,
         "isCeltsStudentStaff": False,
+        "major": None,
+        "classLevel": None,
     },
     {
         "username": "bryanta",
@@ -117,6 +130,8 @@ users = [
         "firstName": "Alex",
         "lastName": "Bryant",
         "isStudent": True,
+        "major": None,
+        "classLevel": None,
     },
     {
         "username": "partont",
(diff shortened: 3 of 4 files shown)


qasema (changes requested):
when Adding a file. If the path of that attachment is deleted or the attachment itself is missing, the file cannot be deleted anymore from the event interface. I think the file should be displayed only if it exists.

qasema (changes requested):
when the event is deleted, the path of that attachments and the attachments of that event are not deleted from the file directory

qasema (changes requested):
When printing the SLC proposal, It prints an extra page. Refer to this PR #438

qasema (changes requested):
When viewing the SLC proposal, you should disable the upload button.

qasema (changes requested):
The same thing with the course syllabus. I when course is deleted the attachment remains in the file directory.

qasema (changes requested):
When uploading the course syllabus, the page refreshes, and all the information on that first page gets deleted when you create the course for the first time.

Karina-Agliullova (dismissed):
Eveything works fine. Good job!

sreynit02 (commented):
When I view a SLC proposal, I was able to delete the uploaded syllabus. Shouldn't we disable the user from doing that in view mode?

app/controllers/serviceLearning/routes.py
```diff
@@ -191,6 +210,55 @@ def renewCourse(courseID, termID):
...
 
+@serviceLearning_bp.route('/serviceLearning/print/<courseID>', methods=['GET'])
+def printCourse(courseID):
+    """
+    This function will print a PDF of an SLC proposal.
+    """
+    instructors = CourseInstructor.select().where(CourseInstructor.course==courseID)
+    courseInstructors = [instructor.user for instructor in instructors]
+    theCreator = Course.select().join(User).where(Course.createdBy.username, Course.id==courseID)
+    if g.current_user.isCeltsAdmin or g.current_user in courseInstructors or g.current_user in theCreator:
```
BrianRamsay: theCreator should be a list of users rather than courses. Or don't use a list check at all and use .exists() on the query

app/logic/courseManagement.py
```diff
@@ -44,41 +45,52 @@ def approvedCourses(termId):
...
+        for instructor in instructorList:
+            CourseInstructor.create(course=course, user=instructor)
+        createLog(f"Saved SLC proposal: {courseData['courseName']}")
+        if attachment:
+            addfile= FileHandler(attachment, courseId=course.id)
+            addfile.saveFiles()
+        return Course.get_by_id(course.id)
+    except Exception as e:
+        print(e)
+        return False, e
```
BrianRamsay: You add a blanket handler and change the return value, but nothing that calls `updateCourse` checks the return value for failure.

app/logic/serviceLearningCoursesData.py
```diff
@@ -39,6 +41,15 @@ def withdrawProposal(courseID):
     """Withdraws proposal of ID passed in. Removes foreign keys first.
     Key Dependencies: QuestionNote, CourseQuestion, CourseParticipant,
     CourseInstructor, Note"""
+
+    # delete syllabus
+    try:
+        syllabus = AttachmentUpload.select().where(AttachmentUpload.course==courseID).get()
+        FileHandler(courseId = courseID).deleteFile(syllabus.id)
+    except:
```
BrianRamsay: We should only pass exceptions that are expected (like the file or attachment not being there), but this will pass ANY exception, including things like syntax errors

app/static/js/slcNewProposal.js
```diff
@@ -3,7 +3,38 @@ import searchUser from './searchUser.js'
...
+      let fileId=  $(this).data("id")
+      let fileData = {fileId : fileId,
+                        courseId:this.id}
+      $.ajax({
+        type:"POST",
+        url: "/deleteCourseFile",
+        data: fileData, //get the startDate, endDate and name as a dictionary
+        success: function(){
+            $("#attachment_"+fileId).remove()
+            $("#attachment_"+fileId).remove()
```
BrianRamsay: Do we really need to remove it twice?

BrianRamsay (changes requested):

Issue #799, opened 2022-11-17
Upgrade Flask


martinj2-dot:
The current version of Flask running on CELTS is 2.0.1. 

The most up to date version is 2.2.2.

https://pypi.org/project/Flask/

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Upgrade Flask #799](https://trello.com/c/ghya89sX/587-upgrade-flask-799)

PR #824, opened 2023-01-25
upgrade flask to 2.2.2

qasema:
Fix issue #799  upgrade flask to 2.2.2 which depends on werkzeug 2.2.2, and werkzeug 2.2.2 depends on MarkupSafe==2.1. So I upgraded only these three libraries


sreynit02 (approved):
Flask upgraded successfully and all tests passed!

BrianRamsay (approved):

Issue #652, opened 2022-08-04
SLC Proposal input boxes should accept links


AndersonStettner:
Make links in course description and proposal questions clickable (|urlize)

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [SLC Proposal input boxes should accept links #652 Open](https://trello.com/c/YX0igiEe/465-slc-proposal-input-boxes-should-accept-links-652-open)

PR #825, opened 2023-01-30
made area for responses in slc course proposal accept links

gahimbaref:
#652 

Did:
- Made the textarea in slc proposal responses turn text into links 
- Bolded the questions
- indented the responses

Test:
- Go to course proposal and create a new proposal
- Enter the information and press continue
- When you get to the 6 questions, add a link in the response such as "www.google.com"  and click submit
- Go back to the course proposal page and select "view" on the created proposal, the response should be indented and any links in the text should be linkified.


tylerpar99 (changes requested):
Smaller links work as expected. However, larger links will break the entire page.

Screenshot (53)

tylerpar99 (approved):
Looks good!

PR #826, opened 2023-01-30
Slc proposal edit permissions Issue#804

### LINKED ISSUE #804: Fix and check on SLC Proposal  edit permissions
Make sure a proposal is no longer editable after it has been approved (URL hack) and make sure that a proposal is not editable by anyone other than admin, the creator, and any faculty that may be assigned as the instructor.

Karina-Agliullova:
**Issue:** The proposal should not be editable after it has been approved. The proposal should not be editable by anyone other than the admin, the creator, and any faculty that may be assigned as the instructor.

**Solution:** Created a new function called ViewProposal to be able to redirect the user to that page when trying to edit an approved proposal. Made a conditional for the user to be an admin, creator, or faculty to be able to edit proposals.

**Test:** On the Course Proposals page, try to edit an approved proposal, and it will redirect you to viewing that proposal and flash a warning message. If you try to edit any other proposals, you should be able to do so. If you try to view the approved or any other proposals, you should be able to do so.


app/controllers/serviceLearning/routes.py
```diff
@@ -62,6 +61,41 @@ def slcEditProposal(courseID):
...
+            return render_template('serviceLearning/slcNewProposal.html',
+                                        course = course,
+                                        questionanswers = questionanswers,
+                                        terms = terms,
+                                        courseInstructor = courseInstructor,
+                                        isAllSectionsServiceLearning = isAllSectionsServiceLearning,
+                                        isPermanentlyDesignated = isPermanentlyDesignated,
+                                        redirectTarget=getRedirectTarget())
+        else:
+            flash("You cannot edit an approved SLC proposal. Contact ", 'warning')
```
tylerpar99: Contact who?

tylerpar99 (changes requested):

app/controllers/serviceLearning/routes.py
Karina-Agliullova: Fixed it

solijonovam (commented):
looks all good!

AndersonStettner (changes requested):
Students are able to go to the view proposal link if they type in the url, this should not be allowed. We want to keep the same visibility restrictions as with editing proposal.

solijonovam (changes requested):
There is an inconsistency with telling the user that a page is inaccessible. instead of doing a flasher, we should do this: 
image
if a user should not be able to reach any URL then we need to give 'acces denied' page, but if we are only redirecting to a specific page after access denied than we should do a flasher.

app/controllers/serviceLearning/routes.py
```diff
@@ -34,33 +34,75 @@ def serviceCourseManagement(username=None):
         return redirect(url_for('main.events', selectedTerm=g.current_term))
 
 @serviceLearning_bp.route('/serviceLearning/viewProposal/<courseID>', methods=['GET'])
+def slcViewProposal(courseID):
```
sreynit02: This function is not necessary because the view proposal functionality has already been handled by 'slcNewProposal.js' and 'slcQuestionnaire.html'

To direct the page to viewProposal when s1 click edit on approved proposal, you could change a condition in the 'changAction' function in the 'slcNewProposal.js'. 

Please add 
if g.current_user.isStudent:
        abort(403)

in the `slcEditProposal` to make sure students cannot access the view and edit proposal page.

sreynit02 (changes requested):

app/controllers/serviceLearning/routes.py
```diff
@@ -40,27 +40,36 @@ def slcEditProposal(courseID):
...
-                                terms = terms,
-                                courseInstructor = courseInstructor,
-                                isAllSectionsServiceLearning = isAllSectionsServiceLearning,
-                                isPermanentlyDesignated = isPermanentlyDesignated,
-                                redirectTarget=getRedirectTarget())
+    if g.current_user.isStudent:
+        abort(403)
+    if g.current_user.isCeltsAdmin or g.current_user.isFaculty:
+        course = Course.get_by_id(courseID)
+        coursestatus = CourseStatus.get_by_id(course.status)
```
tylerpar99: Is this necessary? The `course` object we are getting has `course.status` which will provide the same object you are re-grabbing for `coursestatus`.

app/static/js/slcManagement.js
```diff
@@ -10,6 +10,7 @@ $(document).ready(function() {
 
 function changeAction(action){
   courseID = action.id;
+  var courseSt = $("#courseStatus-" + courseID).html();
```
tylerpar99: We should just spell out the full variable name to make it more clear. (courseStatus)

tylerpar99 (changes requested):

app/templates/serviceLearning/slcNewProposal.html
```diff
@@ -12,7 +12,7 @@
 {% endblock %}
 
 {% block app_content %}
-<form id="slcNewProposal" method="post" action="/serviceLearning/newProposal">
+<form id="slcNewProposal" class="{{coursestatus}}" method="post" action="/serviceLearning/newProposal">
```
tylerpar99: Doesn't seem like this should be here.

app/templates/serviceLearning/slcManagement.html
```diff
@@ -32,7 +32,7 @@ <h1 class="text-center mb-2">{{user.firstName}} {{user.lastName}}'s SLC Proposal
           <td>{{courseDict[course]['creator']}}</td>
           <td>{{courseDict[course]['faculty']|join(", ")}}</td>
           <td>{{courseDict[course]['term'].description}}</td>
-          <td>{{courseDict[course]['status']}}</td>
+          <td id="courseStatus-{{courseDict[course]['id']}}">{{courseDict[course]['status']}}</td>
```
tylerpar99, line 35: Not sure what function the `courseStatus-` serves.

tylerpar99 (changes requested):

app/templates/serviceLearning/slcManagement.html
AndersonStettner: I think it would be to make the id name more descriptive. With the without `courseStatus-` then the id will just be an int, this way we know what that int is for without having to trace it back through the code.

app/templates/serviceLearning/slcNewProposal.html
Karina-Agliullova: fixed

app/static/js/slcManagement.js
Karina-Agliullova: fixed

app/controllers/serviceLearning/routes.py
Karina-Agliullova: fixed

AndersonStettner (changes requested):
One of the cases mentioned in the issue is "make sure that a proposal is not editable by anyone other than admin, the creator, and any faculty that may be assigned as the instructor." but that is not being done currently. 

Example: 
- Submit the Databases course which is created by "Brian Ramsay" and an instructor "Brian Ramsay" 
- Switch your user to "Scott Heggen" 
- You are able to, as Scott,  input the url (/serviceLearning/editProposal/1) to the databases course for both editing and viewing

A user should not be able view a course page unless they are a CELTS admin, the creator of the course, or an instructor of the course. Also, in the above example Scott is still edit the Database course despite it being approved.

app/controllers/serviceLearning/routes.py
```diff
@@ -40,27 +40,36 @@ def slcEditProposal(courseID):
...
+        return render_template('serviceLearning/slcNewProposal.html',
+                                    course = course,
+                                    questionanswers = questionanswers,
+                                    terms = terms,
+                                    courseInstructor = courseInstructor,
+                                    isAllSectionsServiceLearning = isAllSectionsServiceLearning,
+                                    isPermanentlyDesignated = isPermanentlyDesignated,
+                                    redirectTarget=getRedirectTarget())
+    else:
+        flash("Unauthorized to view page", 'warning')
```
AndersonStettner: Instead of having the flash and redirect we can just shoot a 403 which will match how we handle things like this throughout the application.

AndersonStettner (changes requested):
Nice job, everything is looking good except that approved courses are still editable if hey change the viewProposal to editProposal in the URL and the one comment I left.

tylerpar99:
When I attempt to edit an approved course proposal it throws the following error:
image

tylerpar99:
Also the terminal prints hello on a route visit through your function

tylerpar99:
Looks good!

AndersonStettner (dismissed):
Looks good!

tylerpar99 (dismissed):

solijonovam (approved):

sreynit02 (approved):
looks good!

PR #834, opened 2023-02-03
change ID names and added labels back in.

AndersonStettner:
*There is no GitHub issue attached to this PR

I merged a PR and then noticed that the IDs were very general so I made them more descriptive and I added the label tag back onto the question prompts for screen readers.


tylerpar99 (approved):
Tested. Everything still works and Id names are much better.

Issue #809, opened 2022-12-09
Change the banned indicator of volunteers on the Track Volunteers page.


AndersonStettner:
Currently, when a student who is banned from an event appears on the Track Volunteers page of an event we indicate they are banned by putting a checkmark in the "Banned" column and and "X" if they are not banned. Either swap the symbols or use a different symbol to better indicate a user is banned. Could possibly use the same symbol that shows up next to a name when you are searching from them in the "Add Volunteer" modal.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Change the banned indicator of volunteers on the Track Volunteers page. #809](https://trello.com/c/E18cA6cz/580-change-the-banned-indicator-of-volunteers-on-the-track-volunteers-page-809)

PR #836, opened 2023-02-07
Removed the banned column on the Volunteer Tracker for event

bledsoef:
Fixes issue #809 
## Changes
- Removed the banned column
- Substituted the red slash icon with a red x instead
- If a volunteer is banned, put x next to name to indicate ban.


sreynit02 (approved):
Looks good!

PR #847, opened 2023-03-03
Food provided is now checked when "This event requires RSVP" is checked.

### LINKED ISSUE #830: Make it so that if the "Food will be provided" is checked the automatically check the "Requires RSVP" as well.
So if food is provided at an event it will require RSVP but if an event requires RSVP that does not necessarily mean it will have food provided.

bledsoef:
Fixes issue #830 

## Changes
- Added Javascript function that updates the food is provided field when the event requires RSVP.
- Removed redundant function in JS that wasn't necessary.

### DIFF
--- a/app/static/js/createEvents.js
+++ b/app/static/js/createEvents.js
@@ -157,14 +157,15 @@ $(document).ready(function() {
     }
   });
 
-  $("#checkIsTraining").click(function(){
-    if ($("input[name='isTraining']:checked").val() == 'on'){
-      $("#checkIsRequired").prop('checked', true);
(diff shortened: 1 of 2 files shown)


AndersonStettner:
[Make it so that if the "Food will be provided" is checked the automatically check the "Requires RSVP" as well. #830](https://trello.com/c/LOnopg0t/595-make-it-so-that-if-the-food-will-be-provided-is-checked-the-automatically-check-the-requires-rsvp-as-well-830)

tylerpar99:
Terriblely wrote PR, but the code is good.

tylerpar99 (approved):

Issue #748, opened 2022-09-14
Add the ability to limit the number of volunteers that can RSVP to an event. 


AndersonStettner:
Also, add a waitlist after RSVP limit has been reached.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Add the ability to limit the number of volunteers to an event. #748](https://trello.com/c/HsFbG9aG/541-add-the-ability-to-limit-the-number-of-volunteers-to-an-event-748)

Andersonstettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Add a new Background check category: Student Life #749](https://trello.com/c/4U3spXI4/542-add-a-new-background-check-category-student-life-749)

PR #849, opened 2023-03-08
Add ability to limit event RSVP and add a RSVP waitlist

AndersonStettner:
## This fixes issue #748 : Combined with another issue, now in PR #879
CELTS would like to be able to limit the amount of people that can RSVP to an event. 

## Did: 
Added a new column to the Event table `rsvpLimit` and created a function `getCurrentRsvpAmmount()` in events.py to get the amount of people that have RSVPd to an event. The RSVP button is enabled/disabled based off the current RSVP amount. The wording on the Event View page now changes based off of whether or not RSVP is required, if there is a limit or not, and how many spots are left if there is a limit. 

## To Test:
- Checkout the branch `limitRsvp`
- Reset your database with `database/reset_database.sh test`
- Run the test suite with `tests/run_tests.sh ` and verify that all tests pass
- Run the application with `flask run` 
- Create multiple events, some that require RSVP but have no limit and some that require RSVP with a limit
- Verify that all the behavior that you encounter is as you expect


AndersonStettner:
[Add the ability to limit the number of volunteers that can RSVP to an event. #748](https://trello.com/c/HsFbG9aG/541-add-the-ability-to-limit-the-number-of-volunteers-that-can-rsvp-to-an-event-748)

tylerpar99 (changes requested):
I think if we are going to disable the button we should tell the user why, if not they might think the site is broken. Also, if I inspect the page and delete the `disabled` tag on the HTML button I can just RSVP anyway so we need to block them in `.js` or `.py`.

tylerpar99:
This PR is good, just waiting on a companion PR that will fix some vulnerabilities. This PR should be left as is and will be merged with the branched PR.

qasema:
looks good, and agree with Tyler regarding the button being disabled

Karina-Agliullova:
While testing the PR, I noticed that I could select a negative number of people attending the event. I don't think it makes sense to go negative when setting the attendance limit

tests/code/test_events.py
```diff
@@ -441,6 +435,7 @@ def test_saveEventToDb_update():
                         'recurringId': 3,
                         'isTraining': True,
                         'isRsvpRequired': False,
+                        'rsvpLimit': None,
```
Karina-Agliullova: Why is newEventData{} implemented twice?

Issue #831, opened 2023-02-01
Have an email sent to Ashley Cochrane when an SLC proposal is submitted.


AndersonStettner:
When a person submits an SLC proposal we should an email to Ashley with the name of the course instructor, the name of the person who created the from (possibly only include this if the creator is different than the instructor), and the course name.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Have an email sent to Ashley Cochrane when an SLC proposal is submitted. #831](https://trello.com/c/aBqEdu9K/596-have-an-email-sent-to-ashley-cochrane-when-an-slc-proposal-is-submitted-831)

PR #850, opened 2023-03-08
Email Ashley SLC submitted courses Issue#831

Karina-Agliullova:
**NOT FINISHED!**
Issue: https://github.com/BCStudentSoftwareDevTeam/celts/issues/831

**Problem:** When a person submits an SLC proposal we should email Ashley with the name of the course instructor, the name of the person who created the form (possibly only include this if the creator is different than the instructor), and the course name.

**Solution:** Using an email template to build a template for the email that needs to be sent to Ashley. The email template is not working as of now.

**Testing:** As of now, it cannot be tested. Ideally, you would need to create an SLC course and submit it while logged in as an instructor. It should automatically send an email to Ashley and notify you with the flash message


AndersonStettner:
[Have an email sent to Ashley Cochrane when an SLC proposal is submitted. #831](https://trello.com/c/aBqEdu9K/596-have-an-email-sent-to-ashley-cochrane-when-an-slc-proposal-is-submitted-831)

Issue #854, opened 2023-03-24
Add a students major and class level to the User table


AndersonStettner:
Sub-issue of #857 

In the User table add the class level and major to the students. There is already a script in LSF to pull this info from Tracy, it will need to be modified a bit to be used in CELTS and to get the exact info we want.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Add a students major and class level to the User table #854](https://trello.com/c/q0Cdwe51/602-add-a-students-major-and-class-level-to-the-user-table-854)

PR #855, opened 2023-03-27
Add a student's major and class level to the User table

sreynit02:
Fixes issue #854 
Did: 
- added "major" and "classLevel" to User table
- configure import_users.py so it can pull these two information from Tracy to our database

Test: 
- reset database
- update the Tracy password in config/default.yml (ask Brian or Sreynit or Fleur for password)
- In database run 'python import_users.py'
- Check the User table. "major and classLevel" for students should be populated with their major and class level.

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -9,6 +9,8 @@ class User(baseModel):
     firstName = CharField()
     lastName  = CharField()
     isStudent = BooleanField(default = False)
+    major = CharField(null = True)
+    classLevel = CharField(null = True)
     isFaculty = BooleanField(default = False)
     isStaff = BooleanField(default = False)
     isCeltsAdmin = BooleanField(default  =False)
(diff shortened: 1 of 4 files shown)


AndersonStettner (changes requested):
add pyodbc to requirements.txt and the students who are in base data are not being updated correctly. 
EX: I changed Sreynit's name to just Sr, re-ran import_users and it fixed the name but still did not add a class level or major, it is like that for all students in base data.

sreynit02:
> add pyodbc to requirements.txt and the students who are in base data are not being updated correctly. EX: I changed Sreynit's name to just Sr, re-ran import_users and it fixed the name but still did not add a class level or major, it is like that for all students in base data.

fixed!

AndersonStettner (changes requested):
Nevermind, I went ahead and made the change since it was so small, everything looks good!

AndersonStettner (approved):

Issue #856, opened 2023-03-29
Add a view counter to event pages


AndersonStettner:
Sub-issue of #857 

CELTS would like to see how many times event pages are viewed in their reports. Add a "Views" counter to the event pages that counts every time someone opens the event view. 

Presumably, this is only added to when they are on the `viewEvent` page and not the `trackVolunteers` or `editEvent` page. Also, I do not think this should be a count that is visible to anyone looking at the event page itself, it is a value tracked in the backend to be displayed elsewhere.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Add a view counter to event pages #856](https://trello.com/c/HdkV5aat/603-add-a-view-counter-to-event-pages-856)

PR #865, opened 2023-04-19
Views count

qasema:
fixed issue #856, add a view counter on the event view page to keep track of the people who view an event. 

## Testing
-  checkout branch viewsCount
-  reset the database 
-  make sure tests are not failing 
-  try to view an event (not admin)
-  switch the current user to admin and see if the view is counted 
-  only the admin can see the views, admin views do not count, and no duplicate views 
-  check if the EventView table is updated


solijonovam (commented):
looks good to me!

solijonovam (commented):
tests all pass! good job ala

app/logic/events.py
```diff
@@ -450,3 +451,11 @@ def getTomorrowsEvents():
     tomorrowDate = date.today() + timedelta(days=1)
     events = list(Event.select().where(Event.startDate==tomorrowDate))
     return events
+def eventViewCount(viewer,event):
```
AndersonStettner: We add a test for new logic functions.

AndersonStettner (changes requested):

app/controllers/admin/routes.py
```diff
@@ -147,7 +148,12 @@ def createEvent(templateid, programid=None):
 @admin_bp.route('/eventsList/<eventId>/view', methods=['GET'])
 @admin_bp.route('/eventsList/<eventId>/edit', methods=['GET','POST'])
 def eventDisplay(eventId):
-    # Validate given URL
```
AndersonStettner, line 150: Why did you remove this comment and not move it down with the content it was originally with?

app/controllers/admin/routes.py
```diff
@@ -147,7 +148,12 @@ def createEvent(templateid, programid=None):
...
+    viewCount = EventView.select().where(EventView.event == eventId).count() # count the event views
```
AndersonStettner: Instead of having a trailing comment why not just change the name of the variable so it is clear by just reading its name.

app/controllers/admin/routes.py
```diff
@@ -147,7 +148,12 @@ def createEvent(templateid, programid=None):
 @admin_bp.route('/eventsList/<eventId>/view', methods=['GET'])
 @admin_bp.route('/eventsList/<eventId>/edit', methods=['GET','POST'])
 def eventDisplay(eventId):
-    # Validate given URL
+    viewCount = EventView.select().where(EventView.event == eventId).count() # count the event views 
+    if not g.current_user.isCeltsAdmin and request.method == 'GET' and request.path == f'/eventsList/{eventId}/view':
+        viewer = User.get_by_id(g.current_user)
+        event = Event.get_by_id(eventId)
+        eventViewCount(viewer,event) #insert into the EventView table
```
AndersonStettner: There really is not a reason for this comment.

app/logic/events.py
```diff
@@ -450,3 +451,11 @@ def getTomorrowsEvents():
...
+    """This checks if the current user already viewed the event. If not, insert a recored to EventView table"""
+    event_view_exists = EventView.select().where(EventView.user == viewer, EventView.event == event).exists()
```
AndersonStettner: We use camleCase for variable naming/declaration across our logic files, please stay consistent.

app/logic/events.py
```diff
@@ -450,3 +451,11 @@ def getTomorrowsEvents():
...
     events = list(Event.select().where(Event.startDate==tomorrowDate))
     return events
+def eventViewCount(viewer,event):
+    """This checks if the current user already viewed the event. If not, insert a recored to EventView table"""
+    event_view_exists = EventView.select().where(EventView.user == viewer, EventView.event == event).exists()     
+    if not event_view_exists:
+        # Insert new EventView record for the user and event
+        date = datetime.datetime.now()
+        new_event_view = EventView.create(user=viewer, event=event, viewedOn=date)
+        new_event_view.save()
```
AndersonStettner: camelCase

app/templates/eventView.html
```diff
@@ -123,5 +123,12 @@ <h3><br>Program Trainings</h3>
                     <button type="submit" class="btn btn-primary" value="{{eventData.id}}" id="rsvpBtn">RSVP</button>
                  {% endif %}
               {% endif %}
+              {% if g.current_user.isCeltsAdmin or isProgramManager %}
+            <div class="d-flex align-items-center">
+                <i class="bi bi-eye-fill me-2"></i>
+                <span class="me-2">Views:</span>
+                <span id="view-count">{{viewCount}}</span>
```
AndersonStettner: We have been using camelCase for id names everywhere else, please keep it contestant.

app/templates/eventView.html
```diff
@@ -123,5 +123,12 @@ <h3><br>Program Trainings</h3>
                     <button type="submit" class="btn btn-primary" value="{{eventData.id}}" id="rsvpBtn">RSVP</button>
                  {% endif %}
               {% endif %}
+              {% if g.current_user.isCeltsAdmin or isProgramManager %}
+            <div class="d-flex align-items-center">
+                <i class="bi bi-eye-fill me-2"></i>
+                <span class="me-2">Views:</span>
```
AndersonStettner: Personally I would prefer it to be something like "Unique Page Views" or at least "Page Views".

AndersonStettner (changes requested):

app/templates/eventView.html
qasema: agree

qasema:
[Event Page views count](https://trello.com/c/mhqN5wfW/609-event-page-views-count)

qasema:
[Add a view counter to event pages #856](https://trello.com/c/HdkV5aat/603-add-a-view-counter-to-event-pages-856)

sreynit02 (dismissed):
Looks good!

app/controllers/admin/routes.py
```diff
@@ -147,6 +148,11 @@ def createEvent(templateid, programid=None):
 @admin_bp.route('/eventsList/<eventId>/view', methods=['GET'])
 @admin_bp.route('/eventsList/<eventId>/edit', methods=['GET','POST'])
 def eventDisplay(eventId):
+    pageViewsCount = EventView.select().where(EventView.event == eventId).count() # count the event views
```
AndersonStettner: Removed comment, the variable name now gives accurate description of what it is doing so the comment is no longer needed.

app/controllers/admin/routes.py
```diff
@@ -147,6 +148,11 @@ def createEvent(templateid, programid=None):
 @admin_bp.route('/eventsList/<eventId>/view', methods=['GET'])
 @admin_bp.route('/eventsList/<eventId>/edit', methods=['GET','POST'])
 def eventDisplay(eventId):
+    pageViewsCount = EventView.select().where(EventView.event == eventId).count() # count the event views 
+    if not g.current_user.isCeltsAdmin and request.method == 'GET' and request.path == f'/eventsList/{eventId}/view':
+        viewer = User.get_by_id(g.current_user)
```
AndersonStettner: `g.current_user` is the user so, you already have the user, making the `.get_by_id` unnecessary

app/logic/events.py
AndersonStettner: `eventViewCount` is a bit misleading of a name, something like `recordEventView`, `logEventView`, or `addEventView` would make more sense since that is what the function is doing.

app/logic/events.py
```diff
@@ -450,3 +451,11 @@ def getTomorrowsEvents():
     tomorrowDate = date.today() + timedelta(days=1)
     events = list(Event.select().where(Event.startDate==tomorrowDate))
     return events
+def eventViewCount(viewer,event):
+    """This checks if the current user already viewed the event. If not, insert a recored to EventView table"""
+    eventViewExists = EventView.select().where(EventView.user == viewer, EventView.event == event).exists()     
+    if not eventViewExists:
+        # Insert new EventView record for the user and event
+        date = datetime.datetime.now()
```
AndersonStettner: Use `.get_or_create()` instead of `.exists()` and ignore viewed on since you are adding default to model. Also, add the logic to exclude Admin into the function and add it to the tests.

tests/code/test_event_list.py
```diff
@@ -192,3 +193,32 @@ def test_getOtherEvents(special_otherEvents):
...
+                                    bnumber = "B000000000",
+                                    email = "test@test.com",
+                                    phoneNumber = "Null",
+                                    firstName = "TestFirst",
+                                    lastName = "TestLast",
+                                    isStudent = False,
+                                    isFaculty = True,
+                                    isStaff = False,
+                                    isCeltsAdmin = False,
+                                    isCeltsStudentStaff = False)
```
AndersonStettner: Indent `viewer `and `testEvent `properly.

tests/code/test_event_list.py
```diff
@@ -192,3 +193,32 @@ def test_getOtherEvents(special_otherEvents):
...
+                            timeStart = "18:00:00",
+                            timeEnd = "21:00:00",
+                            location = "basement",
+                            isTraining = True,
+                            startDate = "2021-12-12",
+                            endDate = "2021-12-13")
+        viewer = User.create(username = "eventViewer",
+                                    bnumber = "B000000000",
+                                    email = "test@test.com",
+                                    phoneNumber = "Null",
```
AndersonStettner: Give it a fake phone number. Just do something like 000-000-0000

app/models/eventViews.py
```diff
@@ -0,0 +1,8 @@
+from app.models import*
+from app.models.user import User
+from app.models.event import Event
+
+class EventView(baseModel):
+    user = ForeignKeyField(User)
+    event = ForeignKeyField(Event)
+    viewedOn = DateTimeField()
```
AndersonStettner: We should make `viewedOn` have a default of now.

app/templates/eventView.html
```diff
@@ -123,5 +123,12 @@ <h3><br>Program Trainings</h3>
                     <button type="submit" class="btn btn-primary" value="{{eventData.id}}" id="rsvpBtn">RSVP</button>
                  {% endif %}
               {% endif %}
+              {% if g.current_user.isCeltsAdmin or isProgramManager %}
+            <div class="d-flex align-items-center">
+                <i class="bi bi-eye-fill me-2"></i>
+                <span class="me-2">Page Views:</span>
```
AndersonStettner: Also, after thinking about it more, "Unique Viewers" would probably be the best descriptor.

AndersonStettner (changes requested):

app/logic/events.py
qasema: I think that exclusion is happening in the controller. shall we add that extra layer in logic

app/templates/eventView.html
qasema: how about we make it like "Single User Views" because Unique Viewers could be a bit misleading

tests/code/test_event_list.py
qasema: should that necessarily make the test fail?

app/templates/eventView.html
AndersonStettner: How is it misleading? Unique Viewers seems to be a very accurate descriptor to me. It is a count of all unique users that have viewed the page.

tests/code/test_event_list.py
AndersonStettner: It is unclear right now whether you want to be empty or Null. Either way it is incorrect because this is not how you make a variable null in python.

app/logic/events.py
AndersonStettner: Move the check from the controller and put it into the logic and test it.

app/logic/events.py
```diff
@@ -450,3 +451,6 @@ def getTomorrowsEvents():
     tomorrowDate = date.today() + timedelta(days=1)
     events = list(Event.select().where(Event.startDate==tomorrowDate))
     return events
+def addEventView(viewer,event):
+    """This checks if the current user already viewed the event. If not, insert a recored to EventView table"""
+    eventViewExists = EventView.get_or_create(user = viewer, event = event)
```
AndersonStettner: The variable is irrelevant here.

AndersonStettner (changes requested):

tests/code/test_event_list.py
qasema: but most of the previous test cases used are using the same thing. and I just used it.

app/templates/eventView.html
qasema: haha, I meant that Unique Viewers could refer to certain users and we refer to them as unique

tests/code/test_event_list.py
AndersonStettner: That is not an excuse, when things like this are noticed the person who is working on the pr/issue goes and fixes them.

app/templates/eventView.html
AndersonStettner: Brian and I agree that Unique Viewers is the best description here.

AndersonStettner (dismissed):

BrianRamsay (approved):

PR #877, opened 2023-05-01
Spreadsheet

AndersonStettner:


app/logic/spreadsheet.py
```diff
@@ -0,0 +1,515 @@
...
+def volunteersMajors():
+    # Majors represented in volunteering
+    majorQuery = (User.select(User.major, fn.COUNT(fn.DISTINCT(EventParticipant.user_id)).alias('count'))
+                      .join(EventParticipant, on=(User.username == EventParticipant.user_id))
+                      .group_by(User.major))
+    major_query_dict = {row.major: float(row.count) for row in majorQuery}
+
+    return major_query_dict
+
+def classLevelsInVolunteering():
```
AndersonStettner: Combine class level and major queries into one function where you pass through which one you want to get.

app/logic/spreadsheet.py
```diff
@@ -0,0 +1,515 @@
...
+def repeatVolunteersAllPrograms():
+    # Get people who participated in events more than once (all programs)
+    repeatAllProgramQuery = (EventParticipant.select(EventParticipant.user_id, fn.COUNT(EventParticipant.user_id).alias('count'))
+                                             .group_by(EventParticipant.user_id)
+                                             .having(fn.COUNT(EventParticipant.user_id) > 1))
+    return repeatAllProgramQuery
+
+def retentionRate():
+    #fall participation
+    fallParticipationQuery=(ProgramEvent.select(ProgramEvent.program_id, EventParticipant.user_id.alias('participants'), Program.programName.alias("progName"))
```
AndersonStettner: Combine Fall and Spring into the same query and pass through which term you want retention rates on.

app/logic/spreadsheet.py
```diff
@@ -0,0 +1,515 @@
...
+                                        .join(EventParticipant, on=(ProgramEvent.event == EventParticipant.event))
+                                        .join(Program, on=(Program.id == ProgramEvent.program_id))
+                                        .join(Event, on=(EventParticipant.event_id == Event.id))
+                                        .join(Term, on=(Event.term_id == Term.id) )
+                                        .where(Term.description == "Fall 2022"))
+                                  
+
+    fallParticipationDict = {}
+    for result in fallParticipationQuery.dicts():
```
AndersonStettner: Since the fall and spring are becoming one query then we only need one of these. Do it with a defaultdict instead of a normal dict to simplify the for loop.

app/logic/spreadsheet.py
```diff
@@ -0,0 +1,515 @@
...
+from app.models.programEvent import ProgramEvent
+from app.models.program import Program
+from app.models.event import Event
+from app.models.term import Term
+from peewee import *
+import openpyxl
+import pandas as pd
+from openpyxl.utils import get_column_letter
+from openpyxl.worksheet.table import Table, TableStyleInfo
+
```
AndersonStettner, line 14: Add new things to requirements

BrianRamsay (approved):

PR #912, opened 2023-06-12
Stevenson and Lawrence are on the scene!

hoerstl:
Added our names and roles to the default.yml file.


AndersonStettner (approved):

BrianRamsay (commented):
done!

PR #928, opened 2023-06-13
Liliane and Bineta changed the Kiosk route

### LINKED ISSUE #678: Change the Event Kiosk route to match the other event routes
The kiosk rout should be /event/< eventid >/kiosk. Change it and modify whatever else is need so that it works correctly.

azabeli:


app/templates/events/trackVolunteers.html
```diff
@@ -2,57 +2,59 @@
 {% extends "base.html"%}
 
 {% block scripts %}
-  {{super()}}
-  <script type="module" src="/static/js/trackVolunteers.js"></script>
-  <script type="text/javascript" src="https://cdn.datatables.net/v/dt/dt-1.11.3/r-2.2.9/datatables.min.js"></script>
+{{super()}}
+<script type="module" src="/static/js/trackVolunteers.js"></script>
+<script type="text/javascript" src="https://cdn.datatables.net/v/dt/dt-1.11.3/r-2.2.9/datatables.min.js"></script>
```
BrianRamsay, line 7: Somehow a lot the indentation on this page was messed up. Not sure how. We need to reset that back to the way it was

BrianRamsay (changes requested):

PR #934, opened 2023-06-15
Restricted permissions to the course management page

### LINKED ISSUE #929: Individual course management page should only be visible to individual and admin
If a username is given to the individual course management page (`/serviceLearning/courseManagement`), only the provided user or an admin should be able to see it. Other users should get a 403 Forbidden

hoerstl:
We've changed the permissions for the course management pages for users to be restricted to all administrators and faculty that are attempting to view their own page. Faculty cannot view other faculty's pages and students should not have one so they cannot either.

Fixes issue #929

### DIFF
--- a/app/controllers/serviceLearning/routes.py
+++ b/app/controllers/serviceLearning/routes.py
@@ -26,10 +26,10 @@ def serviceCourseManagement(username=None):
         user = User.get(User.username==username) if username else g.current_user
     except DoesNotExist:
         abort(404)
-
+    
     isRequestingForSelf = g.current_user == user 
     if g.current_user.isCeltsAdmin or (g.current_user.isFaculty and isRequestingForSelf):
-        setRedirectTarget("/serviceLearning/courseManagement")
+        setRedirectTarget(request.full_path)
         courseDict = getServiceLearningCoursesData(user)
         termList = selectSurroundingTerms(g.current_term, prevTerms=0)
         return render_template('serviceLearning/slcManagement.html',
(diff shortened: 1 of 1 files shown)


app/controllers/serviceLearning/routes.py
```diff
@@ -22,20 +22,23 @@
...
-    if g.current_user.isCeltsAdmin or g.current_user.isFaculty:
-        setRedirectTarget("/serviceLearning/courseManagement")
+    try:
         user = User.get(User.username==username) if username else g.current_user
+    except DoesNotExist:
+        abort(404)
+
+    isRequestingForSelf = g.current_user == user 
+    if g.current_user.isCeltsAdmin or (g.current_user.isFaculty and isRequestingForSelf):
+        setRedirectTarget("/serviceLearning/courseManagement")
```
BrianRamsay: Instead of always coming back to `/serviceLearning/courseManagement` in `setRedirectTarget`, we should come back to the original route they came in on (including the username if necessary). Check out the flask `request` object

BrianRamsay (changes requested):

app/controllers/serviceLearning/routes.py
hoerstl: On it!

hoerstl (commented):
Still working on the newly found bug

hoerstl:
We have fixed the redirection issue! May we merge now?

BrianRamsay (approved):

PR #938, opened 2023-06-16
896 dietary restrictions page

### LINKED ISSUE #896: Show dietary restrictions of users who have RSVPd for an event
When an event requires RSVP we also automatically check that food will be provided at the event so, we should show the dietary restrictions of all the users that have RSVPd to that event somewhere on the Track Volunteers page.

andersoncedu:
Issue:
Needed page to show dietary restrictions for people attending an event that provides food.

Solution:
We implemented a new dietaryRestrictions.html file which adds a new tab to the event view.
The tab shows dietary restrictions enabling only when food is to be provided for the event.

Test:
Create student user accounts that have dietary restrictions for an event that provides food and add these users to the event. Ensure proper display of the dietary restrictions of the users.


AndersonStettner (changes requested):
The following error happens when there is a user in the waitlist: 
image

This error occurs whether or not the RSVPd user or the wait-listed user have dietary restrictions.

AndersonStettner (changes requested):
I got the following error when I tried to go to the dietary restrictions page:

image


image

app/templates/events/dietary_restrictions.html
```diff
@@ -0,0 +1,97 @@
...
+      {% for participant in volunteerUser %}
+        <tr>
+          <td class="col-auto" nowrap>{{participant.firstName}} {{participant.lastName}}</td>
+          <td class="col">{{participant.dietRestriction}}</td>
+        </tr>
+      {% endfor %}
+      {% if event.isPast %}
+        {% for participant in waitlistUser %}
+          <tr>
+            <td class="col-auto" nowrap>>{{participant.firstName}} {{participant.lastName}}</td>
```
AndersonStettner: Is the extra > on purpose? If so what is it for?

AndersonStettner (changes requested):

app/controllers/admin/volunteers.py
```diff
@@ -89,6 +89,34 @@ def trackVolunteersPage(eventID):
 
 
 
+@admin_bp.route('/event/<eventID>/dietary_restrictions', methods=['GET'])
+def dietaryRestrictionsPage(eventID):
```
AndersonStettner, line 93: Access should be restricted to admin and student staff

AndersonStettner (changes requested):

app/templates/events/dietary_restrictions.html
andersoncedu: It was a typo.

AndersonStettner (changes requested):
There should not be so much space between the name and the dietary restriction:  

image

app/templates/events/dietaryRestrictions.html
```diff
@@ -0,0 +1,61 @@
...
+          </tr>
+        {% endfor %}
+      {% endif %}
+    </tbody>
+  </table>
+{% endif %}
+
+{% if waitlistUser and not event.isPast %}
+  <br>
+  <h3>Wait List</h3>
```
AndersonStettner: Wait List should be Waitlist.

AndersonStettner (changes requested):
Since we do not show the Waitlist after an event is in the past, we should not show those on the Waitlist in the Dietary Restrictions page. It should only have event participants at that point.

AndersonStettner (approved):

Issue #564, opened 2022-07-22
Improve Id Usage


zachneill:
There is some Bootstrap 5 default code in development. E.g. id="myTable" in reviewproposal.html and manageServiceLearningFaculty.html, and id="exampleModalLabel" in trackVolunteers.html. I'd find new names for these and any other code that was ctrl+C-->ctrl+V'd from the [BS5 docs](https://getbootstrap.com/docs/5.0/getting-started/introduction/). 

Additionally, id="eventLength" and id="eventID" are used twice in trackVolunteers.html. ids should always be used once, so I'd refactor one or the other. 

Closed PR #490 may help.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Improve Id Usage #564](https://trello.com/c/cDs9GSJ6/430-improve-id-usage-564)

Mercy-Eze:
Issue closed on #939

PR #939, opened 2023-06-16
improved the ids

azabeli:


AndersonStettner (changes requested):
These are a good start but I think it would be good to be a bit more descriptive in the id names. For something like renaming a table it might be good to name the id based off of whatever information that table is showing.

app/templates/events/trackVolunteers.html
```diff
@@ -236,7 +236,7 @@ <h4>RSVP Waitlist</h4>
     <div class="modal-content">
       <div class="modal-header">
         {% set modalTitle = "Add Volunteer to Waitlist" if rsvpFull else("Add Volunteer RSVP" if rsvpNeeded else "Add Volunteer") %}
-        <h5 class="modal-title" id="exampleModalLabel">{{modalTitle}}</h5>
+        <h5 class="modal-title" id="ModalLabel">{{modalTitle}}</h5>
```
BrianRamsay: If the id is never used, it doesn't need to be there at all.

BrianRamsay (changes requested):
We should review more html than just the examples that Zach provided.

AndersonStettner:
[Improve Id Usage #564](https://trello.com/c/cDs9GSJ6/430-improve-id-usage-564)

AndersonStettner (changes requested):
Resolve the merge conflicts.

BrianRamsay (dismissed):

PR #946, opened 2023-06-16
Added conditional to route to search for userinput before redirecting…

### LINKED ISSUE #941: Fix `Page Not Found` when an admin searches for no student on the `/student_search` page
We do not check to see if there is anything in the search bar when an Admin tries to search for a student on the `/student_search` page. This means that if they hit enter with nothing in the search bar the get a `Page Not Found` error page. 

In cases like this, we should not go through with the search/redirect that happens and instead let the user know they have not entered anything and keep them on the same page.

andersoncedu:
Added condition in routes to test user input for emptystring and route back to the page with an error message instead of routing to search results, which would break.

Resolves #941

### DIFF
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -192,8 +192,9 @@ def eventDisplay(eventId):
     if eventData['program'] and eventData['program'].isBonnerScholars:
         requirements = getCertRequirements(Certification.BONNER)
         bonnerCohorts = getBonnerCohorts(limit=5)
-
+    
     rule = request.url_rule
+
     # Event Edit
     if 'edit' in rule.rule:
         return render_template("admin/createEvent.html",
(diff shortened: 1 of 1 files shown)


app/controllers/admin/routes.py
```diff
@@ -272,9 +273,13 @@ def addRecurringEvents():
...
     volunteerName= request.form.copy()
-    username = volunteerName['searchStudentsInput'].strip("()")
-    user=username.split('(')[-1]
-    return redirect(url_for('main.viewUsersProfile', username=user))
+    if volunteerName['searchStudentsInput']:
+        username = volunteerName['searchStudentsInput'].strip("()")
+        user=username.split('(')[-1]
+        return redirect(url_for('main.viewUsersProfile', username=user))
+    else:
+        flash(f"You have not entered anything.", category='danger')
```
AndersonStettner: I think it would be better to have an error message that is a bit more descriptive. Maybe something along the lines of "Please enter a students first name or username to search."

AndersonStettner (changes requested):

AndersonStettner (approved):

PR #947, opened 2023-06-20
Combined tables as requested in the issue on Github.

### LINKED ISSUE #907: Combine the Add Volunteer and the Add Past Volunteers modals
In `trackVolunteers.html`

Currently we are creating separate modals for before and after an event has occurred. This should be done with conditionals in the modal to change button names and/or what is displayed instead of essentially copying and pasting.

andersoncedu:
Combined tables for past volunteers and volunteers.

Resolves #907

### DIFF
--- a/app/templates/events/trackVolunteers.html
+++ b/app/templates/events/trackVolunteers.html
@@ -234,12 +234,18 @@
 <div class="modal fade float-start" id="addVolunteerModal" tabindex="-1" aria-labelledby="addVolunteerModalLabel" aria-hidden="true">
   <div class="modal-dialog">
     <div class="modal-content">
+      <form action="/addVolunteersToEvent/{{event.id}}" method="POST">
       <div class="modal-header">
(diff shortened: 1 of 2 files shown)


app/templates/events/trackVolunteers.html
```diff
@@ -234,57 +234,55 @@ <h4>RSVP Waitlist</h4>
...
+      <div class="modal-body"> 
         <div class="input-group">
           <div class="form-group input-group-x form-outline ui-widget" style="width:80%;">
             <input type="input" id="addVolunteerInput" class="form-control" placeholder="Search" autocomplete="off"/>
           </div>
           <button type="button" class="btn btn-primary btn-sm">
             <i class="bi bi-search ms-0"></i>
           </button>
+          <div class="modal-body pl-5">
+            {%for participant in recurringVolunteers%}
```
AndersonStettner, line 257: The way this is right now makes all previous volunteers show up in past events no matter what. What we want to have happen is for the previous volunteers to show up when the event is a recurring event and it should only be the volunteers from the previous events in the series.

AndersonStettner (changes requested):

PR #954, opened 2023-06-21
Merged Previous and Current Volunteer Modals

### LINKED ISSUE #907: Combine the Add Volunteer and the Add Past Volunteers modals

hoerstl:
There used to be two modals that allowed you to add volunteers to an event. One for new volunteers and one for volunteers that had volunteered in the past. Now, we have combined **and** improved their functionality into the same modal. Since we have combined them, we have added a hoverable indicator to mark the previous volunteers from the rest.

Resolves #907


app/static/js/trackVolunteers.js
```diff
@@ -2,7 +2,11 @@ import searchUser from './searchUser.js'
 
 $(document).ready(function() {
   $('[data-toggle="tooltip"]').tooltip();
+  // var recurringVolunteers = $("#recurringVolunteers").val()
```
AndersonStettner: Remove commented out code if it is not going to be used.

app/static/js/trackVolunteers.js
```diff
@@ -21,6 +25,23 @@ $(document).ready(function() {
       });
     });
 
+    function updateSelectVolunteer(){
+      let checkboxlist = $("#addVolunteerModal input[type=checkbox]")
+      var shouldEnableButton = false
+      console.log("updating button")
```
AndersonStettner: Remove debug output.

app/static/js/trackVolunteers.js
```diff
@@ -47,24 +68,29 @@ $(document).ready(function() {
       })
     })
 
-    var userlist = []
+    var userlist = $(".recurringVolunteer").map(function(){
+      return $(this).val()
+    }).get()
+    console.log(userlist);
```
AndersonStettner: Remove debug output.

app/static/js/trackVolunteers.js
```diff
@@ -47,24 +68,29 @@ $(document).ready(function() {
...
     function callback(selected) {
-      $("#selectVolunteerButton").prop('disabled', false);
       let user = $("#addVolunteerInput").val()
+      console.log(user)
```
AndersonStettner: Remove debug output.

app/static/js/trackVolunteers.js
```diff
@@ -80,8 +106,9 @@ $(document).ready(function() {
     $.ajax({
       url: `/removeVolunteerFromEvent/${username}/${eventId}`,
       type: "POST",
-      success: function(s) {
-         location.reload();
+      success: function(response) {
+        console.log(response);
```
AndersonStettner: Remove debug output.

AndersonStettner (changes requested):

app/templates/events/trackVolunteers.html
```diff
@@ -248,43 +246,27 @@ <h5 class="modal-title" id="exampleModalLabel">{{modalTitle}}</h5>
...
+          {% endif %}
         </ul>
       </div>
       <div class="modal-footer">
         <button type="button" class="btn btn-secondary me-auto" data-bs-dismiss="modal">Close</button>
-        {% set buttonName = "Add to Waitlist" if rsvpFull else ("Add RSVP" if rsvpNeeded else "Add Participant") %}
-        <button id="selectVolunteerButton" type="button" class="btn btn-primary" disabled>{{buttonName}}</button>
+        <button type="button" id = "selectAllVolunteers" class="btn btn-secondary" >Select All</button>
+        {% set buttonName = "Add to Waitlist" if rsvpFull else ("Add to RSVP" if rsvpNeeded else "Add Participant(s)") %}
+        <button id="addVolunteersButton" type="button" class="btn btn-primary" disabled>{{buttonName}}</button>
```
AndersonStettner: Lets change the `Select All` button color to blue and the `{{buttonName}}` button to green.

AndersonStettner (changes requested):

app/templates/events/trackVolunteers.html
```diff
@@ -248,43 +246,27 @@ <h5 class="modal-title" id="exampleModalLabel">{{modalTitle}}</h5>
...
+                  <label class="text-dark" data-toggle="previousVolunteerHover">*</label>
+                </li>
+              {% endfor %}
+            {% endif %}
+          {% endif %}
         </ul>
       </div>
       <div class="modal-footer">
         <button type="button" class="btn btn-secondary me-auto" data-bs-dismiss="modal">Close</button>
+        <button type="button" id = "selectAllVolunteers" class="btn btn-primary" >Select All</button>
```
AndersonStettner, line 247: Everything is looking good. One final thing is, lets only show the `Select All` button when it can be used. This means it should only appear when it is a recurring event and there are previous volunteers to add.

AndersonStettner (changes requested):

app/static/js/trackVolunteers.js
```diff
@@ -21,10 +30,26 @@ $(document).ready(function() {
...
+          if(checkbox["checked"] == true){
+            shouldEnableButton = true
+          }
+      })
+      if (shouldEnableButton){
+        $("#addVolunteersButton").prop("disabled", false)
+      }
+      else{
+        $("#addVolunteersButton").prop("disabled", true)
+      }
```
BrianRamsay: I would combine all of this into the loop. Instead of using `shouldEnableButton` you could just do the actual property setting in place of it.

BrianRamsay: If you wanted to get really fancy, this whole bit could be two pieces, a reduce to get a single boolean value, and a property setting based on that boolean. This isn't necessarily a recommendation, but might be fun.

app/templates/events/trackVolunteers.html
```diff
@@ -248,43 +246,29 @@ <h5 class="modal-title" id="exampleModalLabel">{{modalTitle}}</h5>
...
+                  <label class="text-dark" data-toggle="previousVolunteerHover">*</label>
+                </li>
+              {% endfor %}
+            {% endif %}
+          {% endif %}
         </ul>
       </div>
       <div class="modal-footer">
         <button type="button" class="btn btn-secondary me-auto" data-bs-dismiss="modal">Close</button>
+        {% if event.isRecurring and (recurringVolunteers | length != 0) and not event.isFirstRecurringEvent  %}
```
BrianRamsay: You can probably use `recurringVolunteers` on its own rather than using length and comparing to 0, as an empty string evaluates to False. I'd check that in jinja first. And if not, adding `| length` without the comparison should work.

Also, `not event.isFirstRecurringEvent` is kind of redundant, as `recurringVolunteers` will always be empty in that case, so we'll shortcut the expression before we get to the end.

BrianRamsay (changes requested):

app/templates/events/trackVolunteers.html
hoerstl: Thanks Bryon for the awesome conditional tip!

About the redundancy in the `not event.isFirstRecurringEvent`, that part of the conditional is there to make sure that volunteers that have been added to the first of a series of events are not suggested as recurring volunteers for that same instance of the event (that is, week 1). I believe that we should leave this one in.

AndersonStettner (changes requested):
Resolve the merge conflict.

hoerstl (commented):
Successfully merged changes from development

AndersonStettner (approved):

PR #955, opened 2023-06-22
943 Navigation Tab for Event Page

### LINKED ISSUE #943: Refactor Event page navigation tabs
The navigation tabs on the event page (/eventsList/12/view) are currently duplicated across different templates for each tab. We should create a template for the event page that includes the title and the navigation and then includes the appropriate template for the active tab.

andersoncedu:
We added a new event_nav page to files and used it as a template for the event view pages.

Resolves #943


AndersonStettner (changes requested):
I get the following error when trying to create a new event:

image

app/templates/eventView.html
```diff
@@ -1,40 +1,23 @@
 {% set title = eventData.name %}
-{% extends "base.html" %}
+
+{% set eventName = eventData.name %}
+{% set eventProgram = eventData.program %}
+{% set eventProgramName = eventData["program"].programName %}
+{% set eventPast = event.isPast %}
+{% set eventID = eventData.id %}
```
BrianRamsay: Let's clean up the data we are sending in, since we are having to specify it multiple times. There are enough different values here that we should consider just using eventData and eventPast in events_nav.html

app/templates/eventView.html
```diff
@@ -1,40 +1,23 @@
...
+
+{% set tabName = 'view' %}
+
```
BrianRamsay: I'd remove the blank line between these statements to show that `tabName` is connected to the extends

BrianRamsay (changes requested):

hoerstl:
Updated the branch and fixed the crashing error. The event title now displays correctly.

app/templates/events_nav.html
```diff
@@ -0,0 +1,38 @@
...
+
+
+{% block app_content %}
+    {% block content_pageHeading %}
+    <div class="text-center">
+        <h1 id="title">{{page_title}}</h1>
+        {% if eventData.name != 'All Volunteer Training' and eventData['program'] == None %}
+            <h3>CELTS-Sponsored Event</h3>
+        {% elif eventData['program'] %}
+            <h3>{{eventData["program"].programName}} Event</h3>
```
AndersonStettner: This should not appear when creating an event:

image

AndersonStettner (changes requested):
We should be consistent with our file naming. Instead of `events_nav.html` it should be `eventsNav.html`.

hoerstl:
We've removed the header and renamed the file to your liking. Thanks be to Andonson.

andersoncedu:
Ready for review.

AndersonStettner (approved):

PR #967, opened 2023-06-27
Data has been removed from URL of the Ajax request and added restriction in route

### LINKED ISSUE #953: Remove the data from the url of the Ajax request into post data and add some restrictions to the route that it is calling.
In `trackVolunteers.js` an AJAX request for removing volunteers from an event passes username and eventID through the URL. The should be moved to POST data and handled in the controller as request data. 

Also, make it so the controller only allows Admin and Student Staff  to hit the route.

Files: `trackVolunteers.js`, `volunteers.py`

Mercy-Eze:
Resolves #953

### DIFF
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -124,11 +124,14 @@ def rsvpFromWaitlist(username, eventId):
 def isVolunteerBanned(username, eventId):
     return {"banned":1} if isBannedFromEvent(username, eventId) else {"banned":0}
 
-@admin_bp.route('/removeVolunteerFromEvent/<user>/<eventID>', methods = ['POST'])
-def removeVolunteerFromEvent(user, eventID):
-    (EventParticipant.delete().where(EventParticipant.user==user, EventParticipant.event==eventID)).execute()
-    (EventRsvp.delete().where(EventRsvp.user==user, EventRsvp.event==eventID)).execute()
-    flash("Volunteer successfully removed", "success")
+@admin_bp.route('/removeVolunteerFromEvent', methods = ['POST'])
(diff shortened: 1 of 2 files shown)


app/controllers/admin/volunteers.py
```diff
@@ -124,11 +124,14 @@ def rsvpFromWaitlist(username, eventId):
...
-    (EventParticipant.delete().where(EventParticipant.user==user, EventParticipant.event==eventID)).execute()
-    (EventRsvp.delete().where(EventRsvp.user==user, EventRsvp.event==eventID)).execute()
-    flash("Volunteer successfully removed", "success")
+@admin_bp.route('/removeVolunteerFromEvent', methods = ['POST'])
+def removeVolunteerFromEvent():
+    user = request.form.get('username')
+    eventID = request.form.get('eventId')
+    if g.current_user.isAdmin: 
+        (EventParticipant.delete().where(EventParticipant.user==user, EventParticipant.event==eventID)).execute()
+        (EventRsvp.delete().where(EventRsvp.user==user, EventRsvp.event==eventID)).execute()
```
BrianRamsay, line 165: Almost there! I think we want to correct the old code so that if the event is in the past we remove from EventParticipant, but if it is in the future we remove from EventRsvp

BrianRamsay (changes requested):

app/controllers/admin/volunteers.py
BrianRamsay: we will address this in a future issue, as it causes some unintentional UI weirdness

BrianRamsay (dismissed):

BrianRamsay (approved):

PR #968, opened 2023-06-28
792 refactor program events table

### LINKED ISSUE #792: Refactor programEvent table
Get rid of the programEvent table by replacing it with a program column in the event table
Change all the files that use the programEvent table to use the program column instead.

Once the refactoring has been completed, create a script that will change the database structure and all its contents from the one using the programEvent table to the one using the program column in the event table without losing any data. This script is to use in production to correctly update production's database without losing any data.

michelmonc:
Did:
-  Created a new column in the Events table called program_id
- Updated the test in the database with the program Events that were already created in the new Event.programEvents
- Reset the database and the new tests appeared in the database
-  Looked for every instance of the programEvent in the environment and redirect it to Event.program_id
- Deleted the programEventes Table
-  Updated all queries that were joining or using the programEvents Table to just use the Event Table.
Test:
- Reset Database
- Check if the right program_Id is inserted in the right Event Id
- Create a new Event and then Create a new Recursive Events
- Request Transcript of the new Event
- Check the program_id of your own profiles by checking the events you have attended.


app/controllers/events/routes.py
```diff
@@ -46,7 +45,9 @@ def signinEvent():
     """Utilizes form data and sign in function. Returns correct flasher message."""
     eventid = request.form["eventid"]
     bnumber = request.form["bNumber"]
-    programid = ProgramEvent.select(ProgramEvent.program).where(ProgramEvent.event == eventid)
+    
+    # Correct query
+    programid= Event.select(Event.program_id). where(Event.id == eventid)
```
BrianRamsay: I don't think you tested this. A select query returns a ModelSelect object, not a program id.

app/controllers/main/routes.py
```diff
@@ -42,8 +41,11 @@ def redirectToLogout():
 @main_bp.route('/', methods=['GET'])
 def landingPage():
     managerProgramDict = getManagerProgramDict(g.current_user)
-    programsWithEvents = list(ProgramEvent.select(ProgramEvent.program).where(ProgramEvent.event.term == g.current_term).join(Event).distinct())
-    programsWithEventsList = [program.program.id for program in programsWithEvents]
+
+    
+    programsWithEvents = list(Event.select(Event.program_id).where(Event.term == g.current_term).distinct()) # check before submitting with Anderson
```
BrianRamsay: That comment should be removed.

app/logic/emailHandler.py
```diff
@@ -71,8 +70,11 @@ def fetch_event_programs(self, program_id):
...
         if program_id == 'Unknown' or program_id is None:
-            programEvents = ProgramEvent.select(ProgramEvent.program).where(ProgramEvent.event==self.event.id)
-            return [program.program for program in programEvents.objects()]
+            
+            programEvents = Event.select(Event.program_id).where(Event.id==self.event.id)
+            
+            
+            return [program.program_id for program in programEvents.objects()] 
         else:
             return [Program.get_by_id(program_id)]
```
BrianRamsay: This whole function is irrelevant now. It is getting either a program or a list of programs for an event, but events now only have one program (or none). So it should return the Program for the passed in `program_id`, or `event.program`

app/logic/events.py
```diff
@@ -27,8 +25,8 @@ def getEvents(program_id=None):
 
     if program_id:
         Program.get_by_id(program_id) # raises an exception if program doesn't exist
-        return (Event.select(Event).join(ProgramEvent)
-                    .where(ProgramEvent.program == program_id).distinct())
+        return (Event.select(Event)
+                     .where(Event.program_id == program_id).distinct())
```
BrianRamsay: No `.distinct()` is needed, as you aren't doing any joins. There aren't duplicate events in the `event` table. Also, you don't need to explicitly select `Event`  anymore.

app/logic/events.py
```diff
@@ -164,8 +162,6 @@ def saveEventToDb(newEventData):
             # Create or update the event
             if isNewEvent:
                 eventRecord = Event.create(**eventData)
-                if 'program' in newEventData:
-                    ProgramEvent.create(program=newEventData['program'], event=eventRecord)
```
BrianRamsay, line 169: You removed how we used to set program, but I don't see where you save the new location of program

app/models/event.py
```diff
@@ -21,6 +21,7 @@ class Event(baseModel):
     recurringId = IntegerField(null=True)
     contactEmail = CharField(null=True)
     contactName = CharField(null=True)
+    program_id = ForeignKeyField(Program, null= True)
```
BrianRamsay: Look at how we name all of our other foreign keys. We don't use `_id` in the peewee model, so that we can reference our program object with `event.program`. Using `event.program_id` is confusing because it's not an ID, it's a whole object in peewee.

app/models/event.py
```diff
@@ -29,20 +30,25 @@ def __str__(self):
 
     @property
     def noProgram(self):
-        return not self.programEvents.exists()
+        #Ask Anderson 
+        return not self.program_id.exists()
```
BrianRamsay: you don't need .exists(), because `self.program` isn't a query, it's a value. Just do `not self.program`

BrianRamsay (changes requested):
I stopped reviewing because it will be easier to see what needs to be changed once `Event.program_id` is changed to `Event.program`

app/controllers/events/routes.py
ojmakinde: Thank you for the correction, Brian. Wasn't the previous code also returning a ModelSelect object? Should we use a .get() instead?

BrianRamsay: `program = Event.get_by_id(eventid).program`

app/logic/emailHandler.py
ojmakinde: Thank you for your observation. In an attempt to resolve this, we have resorted to returning "self.event.program", and we have scraped the query.

app/controllers/events/routes.py
```diff
@@ -46,7 +45,9 @@ def kioskSignin():
     """Utilizes form data and sign in function. Returns correct flasher message."""
     eventid = request.form["eventid"]
     bnumber = request.form["bNumber"]
-    programid = ProgramEvent.select(ProgramEvent.program).where(ProgramEvent.event == eventid)
+    
+    # Correct query
```
BrianRamsay: unnecessary comment. Why would it be incorrect?

app/controllers/events/routes.py
```diff
@@ -46,7 +45,9 @@ def kioskSignin():
...
+    programid= Event.select(Event.program). where(Event.id == eventid)
```
BrianRamsay: `programid` isn't used in this function anymore

app/controllers/main/routes.py
```diff
@@ -43,8 +42,10 @@ def redirectToLogout():
 @main_bp.route('/', methods=['GET'])
 def landingPage():
     managerProgramDict = getManagerProgramDict(g.current_user)
-    programsWithEvents = list(ProgramEvent.select(ProgramEvent.program).where(ProgramEvent.event.term == g.current_term).join(Event).distinct())
-    programsWithEventsList = [program.program.id for program in programsWithEvents]
+
+
+    programsWithEvents = list(Event.select(Event.program).where(Event.term == g.current_term).distinct())
```
BrianRamsay: don't limit the select (don't pass any arguments), and you don't need `.distinct()` anymore.

BrianRamsay: The variable also isn't `programsWithEvents`, it's `eventsInTerm` or something.

app/logic/emailHandler.py
```diff
@@ -71,11 +70,17 @@ def fetch_event_programs(self, program_id):
...
 
+            # so this should essentially return
+            # var = Program.get_by_id(program)
+            # return var.programName ???
+
+
+        # The function above should essentially return just the program name, since an event can only belong to one program.
+
```
BrianRamsay, line 76: Remove extra comments

app/logic/emailHandler.py
```diff
@@ -71,11 +70,17 @@ def fetch_event_programs(self, program_id):
         """ Fetches all the programs of a particular event """
         # Non-student-led programs have "Unknown" as their id
         if program_id == 'Unknown' or program_id is None:
-            programEvents = ProgramEvent.select(ProgramEvent.program).where(ProgramEvent.event==self.event.id)
-            return [program.program for program in programEvents.objects()]
+            return self.event.program
         else:
             return [Program.get_by_id(program_id)]
```
BrianRamsay: This function should return the same data type. Right now it's returning a list of program objects in one case and a single program object in another.

app/logic/events.py
```diff
@@ -124,12 +112,14 @@ def attemptSaveEvent(eventData, attachmentFiles = None):
...
 
 def saveEventToDb(newEventData):
-
     if not newEventData.get('valid', False):
         raise Exception("Unvalidated data passed to saveEventToDb")
 
     isNewEvent = ('id' not in newEventData)
 
+    if not isNewEvent:
+        newEventData['program'] = (Event.get_by_id(newEventData['id'])).program
```
BrianRamsay: You don't need this anymore, since you are only using program for a new event.

app/logic/spreadsheet.py
```diff
@@ -7,16 +7,14 @@
...
 from app.models.term import Term
 
 
 def volunteerHoursByProgram():
     query = ((Program.select(Program.programName, fn.SUM(EventParticipant.hoursEarned).alias('sum'))
-                     .join(ProgramEvent)
-                     .join(EventParticipant, on=(ProgramEvent.event == EventParticipant.event))
+                     .join(EventParticipant, on=(Event.program== EventParticipant.event))
                      .group_by(Program.programName)
                      .order_by(Program.programName)))
```
BrianRamsay, line 19: This query doesn't work anymore. You can't use `Event.program` in the join condition because Event is not included in the query

app/logic/spreadsheet.py
```diff
@@ -82,10 +80,10 @@ def getRetentionRate():
...
-                                      .join(EventParticipant, JOIN.LEFT_OUTER, on=(ProgramEvent.event == EventParticipant.event))
-                                      .join(Program, on=(Program.id == ProgramEvent.program_id))
-                                      .join(Event, on=(ProgramEvent.event_id == Event.id))
+    participationQuery = (Event.select(Event.program, EventParticipant.user_id.alias('participant'), Program.programName.alias("progName"))
+                                      .join(EventParticipant, JOIN.LEFT_OUTER, on=(Event.id == EventParticipant.event))
+                                      .join(Program, on=(Program.id == Event.program))
+                                      .join(Event, on=(Event.event_id == Event.id))
                                       .join(Term, on=(Event.term_id == Term.id) )
                                       .where(Term.description == termDescription ))
     print(participationQuery)
```
BrianRamsay: This wasn't you, but let's get rid of this debug output.

app/logic/spreadsheet.py
```diff
@@ -82,10 +80,10 @@ def getRetentionRate():
...
 
 def termParticipation(termDescription):
-    participationQuery = (ProgramEvent.select(ProgramEvent.program_id, EventParticipant.user_id.alias('participant'), Program.programName.alias("progName"))
-                                      .join(EventParticipant, JOIN.LEFT_OUTER, on=(ProgramEvent.event == EventParticipant.event))
-                                      .join(Program, on=(Program.id == ProgramEvent.program_id))
-                                      .join(Event, on=(ProgramEvent.event_id == Event.id))
+    participationQuery = (Event.select(Event.program, EventParticipant.user_id.alias('participant'), Program.programName.alias("progName"))
+                                      .join(EventParticipant, JOIN.LEFT_OUTER, on=(Event.id == EventParticipant.event))
+                                      .join(Program, on=(Program.id == Event.program))
+                                      .join(Event, on=(Event.event_id == Event.id))
```
BrianRamsay: You are already selecting from Event, so joining Event in again doesn't add anything.

app/logic/transcript.py
```diff
@@ -34,22 +32,20 @@ def getProgramTranscript(username):
...
         .order_by(Event.term)
         .having(fn.SUM(EventParticipant.hoursEarned > 0)))
     transcriptData = {}
     for program in programData:
         if program.program in transcriptData:
-            transcriptData[program.program].append([program.event.term.description, program.hoursEarned])
+            transcriptData[program.program].append([program.term.description, program.hoursEarned])
         else:
-            transcriptData[program.program] = [[program.event.term.description, program.hoursEarned]]
+            transcriptData[program.program] = [[program.term.description, program.hoursEarned]]
```
BrianRamsay: They were using the term `program` a lot of places here, but you are now selecting from events. This query should probably be selecting from program, not event

app/models/event.py
```diff
@@ -29,20 +30,25 @@ def __str__(self):
...
+    #         print("#################")
+    #         print(countPE)
+    #         print("###################")
+    #         if len(countPE) == 1:
+    #             self._spCache = countPE[0].program
+    #         else:
+    #             self._spCache = None
 
-        return self._spCache
+    #     return self._spCache
```
BrianRamsay: Get rid of the function, don't just comment it out. If we need it back we can get it from git.

tests/code/test_events.py
```diff
@@ -30,54 +29,26 @@
...
-
-@pytest.mark.integration
-def test_getEventsInvalidProgram():
-
-    # Invalid program
-    with pytest.raises(DoesNotExist):
-        getEvents(program_id="asdf")
+#     # program/event passed
+#     event = Event.get_by_id(11)
+#     assert event.isPast
```
BrianRamsay: Lots of deletions and comments here. Looks like there are some legit tests that are commented out.

BrianRamsay (changes requested):

app/logic/emailHandler.py
BrianRamsay: This still isn't resolved. The function should not return a list in one case and an object in another.

app/controllers/events/routes.py
BrianRamsay: not resolved

app/logic/emailHandler.py
```diff
@@ -206,10 +204,10 @@ def send_email(self):
         template_id, subject, body = self.build_email()
 
         if len(self.program_ids) == 1:
-            if self.program_ids[0].contactEmail:
-                defaultEmailInfo["replyTo"] = self.program_ids[0].contactEmail
-            if self.program_ids[0].contactName:
-                defaultEmailInfo["senderName"] = self.program_ids[0].contactName
+            if self.program_ids.contactEmail:
+                defaultEmailInfo["replyTo"] = self.program_ids.contactEmail
+            if self.program_ids.contactName:
```
BrianRamsay: `self.program_ids` is an inaccurate name now. It should be singular

app/models/event.py
BrianRamsay: not addressed

BrianRamsay (changes requested):
Did you guys forget to push some changes? There are a few things here that I thought were resolved.

AndersonStettner:
[Refactor programEvent table #792](https://trello.com/c/7sFohKiP/571-refactor-programevent-table-792)

BrianRamsay (approved):

PR #970, opened 2023-06-30
Fixed the file upload box

### LINKED ISSUE #898: Adjust the size of the file upload box on `Create Event` and `Edit Event` pages
The upload box takes up way too much room currently, look into resizing it or possibly using a different tool for our file upload.

hoerstl:
Issue: The file upload box was a krajee plugin that we were using to take care of the file uploads. Unfortunately, this box was very big and took up too much space on the page. Double unfortunate was that there was no in-house way to resize that file-upload input tag. Therefore, we needed to replace it.

Fix: Chris and I overhauled the previous file upload box and replaced it with our own, superior version. Ours doesn't have the big drag-and-drop bar but still has the functionality. When you add a file, the file appears below the select file bar. The file input doesn't allow duplicate filenames to be uploaded. The file upload box also works on Firefox.

To Test: Create an event and add some attachments of varying file types to verify that the correct file icons appear for different files. Ensure that the files save correctly.

### DIFF
--- a/app/static/js/createEvents.js
+++ b/app/static/js/createEvents.js
@@ -72,11 +72,8 @@ function getSelectedFiles(){
 
 
 function hasUniqueFileName(fileName){
-  let nameHolders = $(".fileName").toArray();
-  for (let i = 0; i < nameHolders.length; i++){
-    if (fileName == $(nameHolders[i]).data("filename")) return false;
-  }
-  return true;
+  if ($(".fileName[data-filename='" + fileName + "']").length == 0) { return true }
+  return false;
 }
 
 
@@ -116,27 +113,28 @@ $(document).ready(function() {
             default:
               iconClass = 'bi-file-earmark-arrow-up';
           }
-          $("#attachedObjectContainer").append("<div class='border row p-0 m-0' id='attachedFilesRow" +fileNum+"'></div>")
-          $("#attachedFilesRow"+fileNum).append("<i class='col-auto fs-3 px-3 bi " + iconClass + "'></i>")
-          $("#attachedFilesRow"+fileNum).append("<div id='attachedFile" + fileNum + "' data-filename='" + file.name + "' class='fileName col-auto pt-2'>" + fileName + "</div>");
-          $("#attachedFilesRow"+fileNum).append("<div class='col' style='text-align:right'> \
-                                                  <div class='btn btn-danger fileHolder p-1 my-1 mx-1' id='trash" + fileNum + "'>\
-                                                    <span class='bi bi-trash fs-6'></span>\
-                                                  </div>\
+          $("#attachedObjectContainer").append("<div class='border row p-0 m-0' id='attachedFilesRow" +fileNum+"'> \
+                                                  <i class='col-auto fs-3 px-3 bi " + iconClass + "'></i> \
+                                                  <div id='attachedFile" + fileNum + "' data-filename='" + file.name + "' class='fileName col-auto pt-2'>" + fileName + "</div> \
+                                                  <div class='col' style='text-align:right'> \
+                                                    <div class='btn btn-danger fileHolder p-1 my-1 mx-1' id='trash" + fileNum + "' data-filenum='" + fileNum + "'>\
+                                                      <span class='bi bi-trash fs-6'></span>\
+                                                    </div>\
+                                                  </div> \
                                                 </div>")
           $("#trash"+fileNum).data("file", file);
           $("#trash"+fileNum).on("click", function() {
-            let elementFileNum = $(this).prop("id").split("trash")[1];
+            let elementFileNum = $(this).data('filenum');
             $("#attachedFilesRow" + elementFileNum).remove();
-            $("#attachmentObject").prop('files', getSelectedFiles);
+            $("#attachmentObject").prop('files', getSelectedFiles());
           })
           fileNum++;
         }
         else{
-          msgFlash("File with filename '" + file.name + "' has already been added to this event")
+          msgToast("File with filename '" + file.name + "' has already been added to this event")
         }
       }
-      $("#attachmentObject").prop('files', getSelectedFiles);
+      $("#attachmentObject").prop('files', getSelectedFiles());
     });
 
 
--- a/app/static/js/createEvents.js
+++ b/app/static/js/createEvents.js
@@ -72,8 +72,7 @@ function getSelectedFiles(){
 
 
 function hasUniqueFileName(fileName){
-  if ($(".fileName[data-filename='" + fileName + "']").length == 0) { return true }
-  return false;
+  return $(".fileName[data-filename='" + fileName + "']").length == 0;
 }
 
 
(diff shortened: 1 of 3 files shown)


app/static/js/createEvents.js
```diff
@@ -60,18 +60,86 @@ function format24to12HourTime(timeStr){
...
+}
+
+
+function hasUniqueFileName(fileName){
+  let nameHolders = $(".fileName").toArray();
+  for (let i = 0; i < nameHolders.length; i++){
+    if (fileName == $(nameHolders[i]).data("filename")) return false;
+  }
+  return true;
+}
```
BrianRamsay, line 76: it might be simpler for this function to use something like the `.filter` method on `$('.fileName')` to get the # of filenames that match and then return whether that result is 0 length or not

app/static/js/createEvents.js
```diff
@@ -60,18 +60,86 @@ function format24to12HourTime(timeStr){
...
+              iconClass = 'bi-file-earmark-arrow-up';
+          }
+          $("#attachedObjectContainer").append("<div class='border row p-0 m-0' id='attachedFilesRow" +fileNum+"'></div>")
+          $("#attachedFilesRow"+fileNum).append("<i class='col-auto fs-3 px-3 bi " + iconClass + "'></i>")
+          $("#attachedFilesRow"+fileNum).append("<div id='attachedFile" + fileNum + "' data-filename='" + file.name + "' class='fileName col-auto pt-2'>" + fileName + "</div>");
+          $("#attachedFilesRow"+fileNum).append("<div class='col' style='text-align:right'> \
+                                                  <div class='btn btn-danger fileHolder p-1 my-1 mx-1' id='trash" + fileNum + "'>\
+                                                    <span class='bi bi-trash fs-6'></span>\
+                                                  </div>\
+                                                </div>")
```
BrianRamsay, line 123: Since you are already doing a big block of html, you could combine these append statements into one html block

BrianRamsay: I'd also suggest creating a variable for the id so you don't have to repeat `"#attachedFilesRow"+fileNum` in so many places.

app/static/js/createEvents.js
```diff
@@ -60,18 +60,86 @@ function format24to12HourTime(timeStr){
...
+          $("#trash"+fileNum).data("file", file);
+          $("#trash"+fileNum).on("click", function() {
+            let elementFileNum = $(this).prop("id").split("trash")[1];
+            $("#attachedFilesRow" + elementFileNum).remove();
+            $("#attachmentObject").prop('files', getSelectedFiles);
+          })
+          fileNum++;
+        }
+        else{
+          msgFlash("File with filename '" + file.name + "' has already been added to this event")
```
BrianRamsay: let's use our toast messages for this instead of a flash. `msgToast`? I forget.

app/static/js/createEvents.js
```diff
@@ -60,18 +60,86 @@ function format24to12HourTime(timeStr){
...
+          $("#attachedFilesRow"+fileNum).append("<i class='col-auto fs-3 px-3 bi " + iconClass + "'></i>")
+          $("#attachedFilesRow"+fileNum).append("<div id='attachedFile" + fileNum + "' data-filename='" + file.name + "' class='fileName col-auto pt-2'>" + fileName + "</div>");
+          $("#attachedFilesRow"+fileNum).append("<div class='col' style='text-align:right'> \
+                                                  <div class='btn btn-danger fileHolder p-1 my-1 mx-1' id='trash" + fileNum + "'>\
+                                                    <span class='bi bi-trash fs-6'></span>\
+                                                  </div>\
+                                                </div>")
+          $("#trash"+fileNum).data("file", file);
+          $("#trash"+fileNum).on("click", function() {
+            let elementFileNum = $(this).prop("id").split("trash")[1];
```
BrianRamsay: You don't need to get the file num out of the id here, because you already know it. You've just referenced it - `fileNum`.

app/static/js/createEvents.js
```diff
@@ -60,18 +60,86 @@ function format24to12HourTime(timeStr){
...
+            $("#attachedFilesRow" + elementFileNum).remove();
+            $("#attachmentObject").prop('files', getSelectedFiles);
```
BrianRamsay: I see this is adding a function to the files property. But later the function is not executed when it is retrieved. How does this work?

BrianRamsay (changes requested):

app/static/js/createEvents.js
hoerstl: Since we've taken your initial suggestion, we no longer have the need to create a variable for the id! All append statements have been collapsed into a single append statement with a multiline string.

app/static/js/createEvents.js
hoerstl: Thanks for the suggestion Bin, while researching that though, we thought of an even better method to do that. Turns out, we can manage the conditional that we wanted to check all within the initial jQuery selection so we can simply use that to get our list before checking its length.

app/static/js/createEvents.js
hoerstl: Yep! You are correct, msgToast incoming!

app/static/js/createEvents.js
hoerstl: Oh! Actually, that variable is used to get the value from within a callback function so we don't have the luxury of using fileNum here. We have refactored that line though to make it more readable via an extra data attribute.

app/static/js/createEvents.js
hoerstl: OOPS! Thanks for the catch Bry, we actually meant to call that function. We noticed the same error in another place too. Not sure how that was working before. The real intention was to set the `files` property to the actual list of files. Those work better now.

app/static/js/createEvents.js
BrianRamsay: Even better, you don't need the `if` and separate return statements. Just return that conditional directly.

app/static/js/createEvents.js
```diff
@@ -60,18 +60,84 @@ function format24to12HourTime(timeStr){
...
+                                                  <div class='col' style='text-align:right'> \
+                                                    <div class='btn btn-danger fileHolder p-1 my-1 mx-1' id='trash" + fileNum + "' data-filenum='" + fileNum + "'>\
+                                                      <span class='bi bi-trash fs-6'></span>\
+                                                    </div>\
+                                                  </div> \
+                                                </div>")
+          $("#trash"+fileNum).data("file", file);
+          $("#trash"+fileNum).on("click", function() {
+            let elementFileNum = $(this).data('filenum');
+            $("#attachedFilesRow" + elementFileNum).remove();
```
BrianRamsay, line 127: `elementFileNum` isn't necessary, because `fileNum` is still in scope and is the same value (and will always be the same).

BrianRamsay (changes requested):

app/static/js/createEvents.js
BrianRamsay: I was wrong.

BrianRamsay (approved):

PR #971, opened 2023-07-03
845 file attachment fix

### LINKED ISSUE #845: Fix file attachment on recurring events.
Currently, if a user attaches a file to an event while creating a recurring event the file is only attached to that first event. All the other events in the recurring series will still have that file in their "attachments" section when viewing the event but the user will be given an error if they try and view them. 

ToDo:
- Make sure that if an attachment is added to an event when creating a recurring event the attachment is actually added to all events that are then made for that recurring series. 
- Make sure that if there is an attachment on a file that is being deleted, that attachment is also deleted in the database.

ojmakinde:
This issue was addressed by ensuring that all recurring events point to the file stored in the directory of the initial recurring event. Consequently, we have made adjustments to the test files to accommodate the modifications made, allowing them to function as intended.

Changes can be tested by instantiating recurring events in the test directory and using the "retrievePath" function to get their paths. All recurring events will point to the same file in the file system. Additionally, recurring events can be instantiated on the web browser, and users can observe that recurring event files all have the same directory in the address bar.

Resolves #845


app/controllers/admin/routes.py
```diff
@@ -306,6 +306,9 @@ def adminLogs():
 def deleteEventFile():
     fileData= request.form
     eventfile=FileHandler(eventId=fileData["eventId"])
+    print("#####################")
+    print(fileData)
+    print("#####################")
```
BrianRamsay: Remove debug output

app/logic/fileHandler.py
```diff
@@ -12,41 +12,55 @@ def __init__(self,files=None, courseId=None, eventId=None):
...
         This creates the directory/path for the object from the "Choose File" input in the create event and edit event.
         :returns: directory path for attachment
         """
+
+        # Added the eventID of the first recurring event to track the file path for subsequent recurring events.
         try:
             # tries to create the full path of the files location and passes if
             # the directories already exist or there is no attachment
-            filePath=(os.path.join(self.path, newfile.filename))
+            filePath=(os.path.join(self.path, "".join(newfilename)))  # This line adds the eventID of the first recurring event.
```
BrianRamsay: The comment doesn't make much sense here. This function doesn't know anything about event IDs.

app/logic/fileHandler.py
```diff
@@ -12,41 +12,55 @@ def __init__(self,files=None, courseId=None, eventId=None):
...
+                        isFileInEvent = AttachmentUpload.select().where(AttachmentUpload.event == self.eventId, AttachmentUpload.fileName == file.filename).exists()
+                        if not isFileInEvent:
+                            AttachmentUpload.create(event = self.eventId, fileName = str(saveOriginalFile.id) + "/" + file.filename)
+                            # saves attachment in the same directory using the eventID of the first recurring event.
+                            file.save(self.getFileFullPath(newfilename = str(saveOriginalFile.id) + "/" + file.filename))
+                            # We need to add the event id to the file path when we call it in file.save.
+                    else:
+                        isFileInEvent = AttachmentUpload.select().where(AttachmentUpload.event == self.eventId, AttachmentUpload.fileName == file.filename).exists()
+                        if not isFileInEvent:
+                            AttachmentUpload.create(event = self.eventId, fileName = str(saveOriginalFile.id) + "/" + file.filename)
```
BrianRamsay: We should be able to refactor and simplify this. Right now we duplicate the `select()` and `create()` calls.

app/logic/fileHandler.py
```diff
@@ -61,10 +75,13 @@ def deleteFile(self, fileId):
...
-        try:
-            file = AttachmentUpload.get_by_id(fileId)
+        file = AttachmentUpload.get_by_id(fileId)
+        filename = file.fileName
+        file.delete_instance()
+
+        # checks if there are other instances with the same filename in the AttachmentUpload table
+        if AttachmentUpload.select().where(AttachmentUpload.fileName == file.fileName).exists():
+            pass
+        else:
```
BrianRamsay: Instead of having nothing in the positive case and the actual functionality in the `else`, just change the conditional so that the `else` is unnecessary and the code is in the `if` block.

app/models/event.py
```diff
@@ -5,6 +5,7 @@
 
 class Event(baseModel):
     name = CharField()
+    program = ForeignKeyField(Program, null=True)
```
BrianRamsay: This change shouldn't be in here

tests/code/test_events.py
```diff
@@ -315,6 +315,8 @@ def test_attemptSaveEvent():
         with app.app_context():
             g.current_user = User.get_by_id("ramsayb2")
             success, errorMessage = attemptSaveEvent(eventInfo)
+            bla = Event.select().where(Event.name == "Attempt Save Test")
+            print([bloo.id for bloo in bla])
```
BrianRamsay: Remove Anderson-isms and debug output

tests/code/test_fileHandler.py
```diff
@@ -2,69 +2,130 @@
 import os
 from werkzeug.datastructures import FileStorage
 
+
+
```
BrianRamsay: This extra whitespace is unnecessary

tests/code/test_fileHandler.py
```diff
@@ -2,69 +2,130 @@
...
 @pytest.mark.integration
 def test_getFileFullPath():
     # test event
-    filePath = handledEventFile.getFileFullPath(eventFileStorageObject[0])
+    filePath = handledEventFile.getFileFullPath("15/" + "".join(eventFileStorageObject[0].filename))
     assert filePath == 'app/static/files/eventattachments/15/eventfile.pdf'
     
     # test course
-    filePath = handledCourseFile.getFileFullPath(courseFileStorageObject[0])
+    filePath = handledCourseFile.getFileFullPath("".join(courseFileStorageObject[0].filename))
```
BrianRamsay: Why are we using `join` here? We don't have multiple filenames, it's just one value.

tests/code/test_fileHandler.py
```diff
@@ -2,69 +2,130 @@
...
+        handledEventFile.deleteFile(secondevent.id)
+        assert os.path.exists(pathDictTest)==False
+        
+
+        # test course
+        coursefiles= AttachmentUpload.select().where(AttachmentUpload.course == 1)
+        pathDictionary = handledCourseFile.retrievePath(coursefiles)
+        fileId = pathDictionary['coursefile.pdf'][1]
+        path = pathDictionary['coursefile.pdf'][0]
+        assert os.path.exists(path)==False
```
BrianRamsay, line 130: You lost the actual `deleteFile` call from the course file test

BrianRamsay (dismissed):

app/logic/events.py
```diff
@@ -114,10 +114,14 @@ def attemptSaveEvent(eventData, attachmentFiles = None):
     try:
         events = saveEventToDb(newEventData)
         if attachmentFiles:
+            # print("################################")
+            # print([e for e in events])
+            # print("################################")
```
BrianRamsay: Remove the comments. These aren't useful for understanding the code, so they shouldn't be in your final version.

app/logic/fileHandler.py
```diff
@@ -12,41 +12,54 @@ def __init__(self,files=None, courseId=None, eventId=None):
         if courseId:
             self.path = os.path.join(self.path, app.config['files']['course_attachment_path'], str(courseId))
         elif eventId:
-            self.path = os.path.join(self.path, app.config['files']['event_attachment_path'], str(eventId))
+            self.path = os.path.join(self.path, app.config['files']['event_attachment_path'])       #Removed event ID from how the file path is being saved.
```
BrianRamsay: Instead of the comment at the end of the line it should be above the line. It makes the line shorter, and you will read it before the line itself. I would also re-word it to say what is happening or the reason, rather than what changed. Like, "Event ID is included in the filename, not in the path"

app/logic/fileHandler.py
```diff
@@ -12,41 +12,54 @@ def __init__(self,files=None, courseId=None, eventId=None):
...
+        
         try:
-            os.makedirs(self.path)
+            extraDir = str(self.eventId) if self.eventId else ""
+            os.makedirs(os.path.join(self.path, extraDir))                                          #Changed the path where the directory is created.
```
BrianRamsay: That comment isn't needed. It doesn't really add any information other than that you have made changes. I would either get rid of it or make it "Create the directory for the event or course". And don't put it so far at the end of the line. Lines shouldn't be so long that they wrap. If it doesn't fit, put it before the code it is commenting.

app/logic/fileHandler.py
```diff
@@ -12,41 +12,54 @@ def __init__(self,files=None, courseId=None, eventId=None):
...
         This creates the directory/path for the object from the "Choose File" input in the create event and edit event.
         :returns: directory path for attachment
         """
+
+        # Added the eventID of the first recurring event to track the file path for subsequent recurring events.
         try:
             # tries to create the full path of the files location and passes if
             # the directories already exist or there is no attachment
-            filePath=(os.path.join(self.path, newfile.filename))
+            filePath=(os.path.join(self.path, "".join(newfilename)))
```
BrianRamsay: `newfilename` is one string. Why are you using `"".join()`?

app/logic/fileHandler.py
```diff
@@ -12,41 +12,54 @@ def __init__(self,files=None, courseId=None, eventId=None):
...
+                    isFileInEvent = AttachmentUpload.select().where(AttachmentUpload.event == self.eventId,
+                                                                    AttachmentUpload.fileName == file.filename).exists()
+                    if saveOriginalFile and saveOriginalFile.id == self.eventId:       # Checks if iterant is the first (recurring) event
+                        if not isFileInEvent:
+                            AttachmentUpload.create(event = self.eventId, fileName = str(saveOriginalFile.id) + "/" + file.filename)
+                            # saves attachment in the same directory using the eventID of the first recurring event.
+                            file.save(self.getFileFullPath(newfilename = str(saveOriginalFile.id) + "/" + file.filename))
+                    else:
+                        if not isFileInEvent:
+                            AttachmentUpload.create(event = self.eventId, fileName = str(saveOriginalFile.id) + "/" + file.filename)
```
BrianRamsay: We still have duplicate `create()` statements

app/logic/fileHandler.py
```diff
@@ -61,10 +74,11 @@ def deleteFile(self, fileId):
...
         Deletes attachmant from the app/static/files/eventattachments/ or courseattachments/ directory
         """
-        try:
-            file = AttachmentUpload.get_by_id(fileId)
+        file = AttachmentUpload.get_by_id(fileId)
+        filename = file.fileName
+        file.delete_instance()
+
+        # checks if there are other instances with the same filename in the AttachmentUpload table
+        if not AttachmentUpload.select().where(AttachmentUpload.fileName == file.fileName).exists():
```
BrianRamsay, line 82: You saved `file.fileName` in a variable, but you don't use it here.

tests/code/test_fileHandler.py
```diff
@@ -3,68 +3,124 @@
...
+        # test saving 2nd event in a hypothetical recurring series
+        handledEventFileRecurring.saveFiles(saveOriginalFile = Event.get_by_id(15))
+        assert AttachmentUpload.select().where(AttachmentUpload.event_id == 16, AttachmentUpload.fileName == '15/eventfile.pdf').exists()
+
+        # test course
+        handledCourseFile.saveFiles()
+        assert AttachmentUpload.select().where(AttachmentUpload.fileName == 'coursefile.pdf').exists()
+
+        # removes saved paths from file directory
+        os.remove(handledEventFile.getFileFullPath('15/eventfile.pdf'))
```
BrianRamsay, line 60: You should also remove the course file that is created.

tests/code/test_fileHandler.py
```diff
@@ -3,68 +3,124 @@
...
+        # creates a second file to simulate recurring events
+        handledEventFileRecurring.saveFiles(saveOriginalFile = Event.get_by_id(15))
+        
+        # creates a course file for deletion
+        handledCourseFile.saveFiles()
+
+        # delete events
+        path = handledEventFile.getFileFullPath("15/eventfile.pdf")
+
+        firstevent = AttachmentUpload.get(AttachmentUpload.event == 15)
```
BrianRamsay, line 115: I'd put a comment before this piece of the test clarifying why the path should still exist after the `deleteFile` call

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #975, opened 2023-07-07
Creating a new table for RSVP'd participants that did not attend

### LINKED ISSUE #969: Refactor delete Event Participant and create a new table on Track Volunteers page for people who RSVPd for the event but did not attend.
Currently, both RSVP and Participant records show up in the same table for a past event. We should separate them, so that if a user is RSVPd for the event but did not attend (no EventParticipant record), they show up in a separate table below the participant list (named something like 'RSVPd Volunteers Who Did Not Attend'). There should be a delete button for each RSVP entry, just like in the main table.

Currently, when someone clicks the delete button on a participant, we delete both the RSVP record and the Event Participant record whether or not they both exist. We should refactor the delete method so that only EventParticipant is removed when deleting from the participant table, and EventRsvp is removed when deleting from a future event or from the new 'Did Not Attend' table.

Desired changes: 
- Create a new "RSVPd Volunteers Who Did Not Attend" table below the Event Participant table on the Track Volunteers page.
- Remove volunteers who only have an EventRsvp record from the main table for past events
- Modify the participant delete logic so that participant and RSVP records are not deleted together, and to allow deletion from the new table.

hoerstl:
Issue: This PR is in direct response to issue #969. The issue asks us to separate attended and non-attended RSVP'd volunteers into different tables once an event has passed. Previously, all volunteers were left in the same table once the event had passed which made it confusing to sort through the people who actually attended the event and add the resulting hours to their accounts. An additional issue with getEventRsvpCountsForTerm was found in the pursuit of our solution as well. 

Solution: We've created another table. The participants are now sorted into the correct tables by whether or not they attended the event. Additionally, these tables have now been placed into an accordion for easy viewing. Another issue that we found while we were working on this one was an incorrect call to getEventRsvpCountsForTerm to get the rsvp count for a single event. The method was being used to get the rsvp count for a specific event rather than for all the events in a term so we created a new method that did the intended purpose of the original use case. The test suit has been altered to test this method.

We also altered the updateEventParticipants method to use lists of username key-value pairs rather than numbering them ('username1', 'username2', etc...). This change meant that we needed to change the datatype that our test functions passed in to match what flask passes (Dictionary -> ImmutableMultiDict). 

To Test: Create an event with an RSVP limit and add plenty of both rsvp'd and waitlisted users. Then go to track_volunteers and verify that both of the types of participants have been added to their respective tables. Ensure that the accordion works correctly. 
Now set the event to the past. This can be done by setting the event to the current date and changing the start and end times to sometime in the AM before the current time of day. Go to track volunteers and mark some of them as attended. Now go and ensure that volunteers that you've marked as attended are separated from volunteers who have not into two different tables.

If the tables seem to have the correct information and you can add participants to and from each table then our fix is working well.

Resolves #969


app/templates/events/trackVolunteers.html
```diff
@@ -105,111 +287,45 @@
...
-                    class='attendanceCheck'
-                    name="checkbox_{{participant.user.username}}"
-                    {% if event.isPast %}
+                  </td>
+                  <td style="text-align:center">
+                    <input
+                      type="checkbox"
+                      class='attendanceCheck'
+                      name="checkbox_{{participant.user.username}}"
+                      {# TODO: Remove below if statement once working #}
```
BrianRamsay: This if statement and block can go now, presumably

app/controllers/admin/volunteers.py
```diff
@@ -55,36 +55,43 @@ def trackVolunteersPage(eventID):
...
     recurringVolunteers = getPreviousRecurringEventData(recurringEventID)
 
-    currentRsvpAmount = getEventRsvpCountsForTerm(g.current_term)
-
+    currentRsvpAmount = getEventRsvpCount(event.id)
     return render_template("/events/trackVolunteers.html",
                             eventData = eventData,
                             eventVolunteerData = eventVolunteerData,
-                            eventParticipants = eventParticipants,
+                            eventParticipants = [],
```
BrianRamsay: This isn't needed anymore, correct?

app/templates/events/trackVolunteers.html
```diff
@@ -44,59 +44,241 @@
...
+              <div class="col"></div>
+              <button type="submit" class="btn btn-primary float-end nowrap col-auto mx-3">Save</button>
+            </div>
+          {% endif %}
+        </div>
+      </div>
+    </div>
+</form>
+    {% if event.isRsvpRequired and event.rsvpLimit and not event.isPast and eventWaitlistData %}
+    <!-- Waitlist Table -->
```
BrianRamsay: It's good to have this comment here, but let's make it even more of a visual break. It's nice to break up these different segments with more of a section header kind of comment.

tests/code/test_volunteers.py
```diff
@@ -72,7 +73,7 @@ def test_updateEventParticipants():
     with pytest.raises(DoesNotExist):
         EventParticipant.get(EventParticipant.user=="partont", EventParticipant.event==3)
 
-    participantData = {'inputHours_partont':100, 'checkbox_partont':"on", 'event':3, 'username1': 'partont'}
+    participantData = ImmutableMultiDict({'inputHours_partont':100, 'checkbox_partont':"on", 'event':3, 'username': 'partont'})
     volunteerTableUpdate = updateEventParticipants(participantData)
     assert volunteerTableUpdate == True
```
BrianRamsay, line 92: We should update this test (or one of the other ones) to test that having multiple usernames in the ImmutableMultiDict works.

BrianRamsay (changes requested):
Nice work! Just a few small suggestions

tests/code/test_volunteers.py
BrianRamsay: What I meant on this one is that we should actually check that it adds/deletes properly when there are multiple usernames sent. That's a non-trivial part of the code.

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #1007, opened 2023-07-17
Saving files for courses and events without duplication

### LINKED ISSUE #972: Clean up empty directories for file attachments on recurring events.
Currently, when handling attachments for recurring events, separate directories are being created for each recurring event, even though all recurring events reference the same directory in memory. To improve efficiency and prevent unnecessary directory creation, modify the file saving process so that directories are not created for each recurring event after the first one. This will help improve file organization.

michelmonc:
Resolve issue #972 
Did:
-  Check where the attachment is being saved for a recursive event
- Create a function makeDirectory to create the directory when being called to avoid duplicate events and course
- Created a condition to call makeDirectory function and to savefiles
- Created a test to make sure that the file is created and is not duplicated
- Ran tests and all passed
- Remove Comments
- Added Comments

### DIFF
--- a/app/logic/fileHandler.py
+++ b/app/logic/fileHandler.py
@@ -25,6 +25,7 @@ class FileHandler:
         except OSError as e:
             if e.errno != 17:
                 print(f'Fail to create directory: {e}')
+                raise e
         
 
     def getFileFullPath(self, newfilename = ''):
--- a/app/logic/fileHandler.py
+++ b/app/logic/fileHandler.py
@@ -16,11 +16,12 @@ class FileHandler:
             self.path = os.path.join(self.path, app.config['files']['event_attachment_path'])
         
     def makeDirectory(self):
-        # Creating the directory when call in saveFiles
+        # This creates a directory. 
+        # Created to remove duplicates when an event is recurring.
         try:
             extraDir = str(self.eventId) if self.eventId else ""
             os.makedirs(os.path.join(self.path, extraDir))
-        # Occurs when we try to create a directory that already exists
+        # Error 17 Occurs when we try to create a directory that already exists
         except OSError as e:
             if e.errno != 17:
                 print(f'Fail to create directory: {e}')
--- a/app/logic/fileHandler.py
+++ b/app/logic/fileHandler.py
@@ -20,11 +20,10 @@ class FileHandler:
         try:
             extraDir = str(self.eventId) if self.eventId else ""
             os.makedirs(os.path.join(self.path, extraDir))
-        # Passing if the error is 17
+        # Occurs when we try to create a directory that already exists
         except OSError as e:
             if e.errno != 17:
                 pass
-            # Perform additional error handling if needed
         
 
     def getFileFullPath(self, newfilename = ''):
@@ -42,7 +41,7 @@ class FileHandler:
             pass
         except FileExistsError:
             pass
-
+  
         return filePath
 
     def saveFiles(self, saveOriginalFile=None):
@@ -69,13 +68,12 @@ class FileHandler:
                         AttachmentUpload.create(course = self.courseId, fileName = file.filename)
                         saveFileToFilesystem = file.filename
                 
-                # Creating directory and saving file for events and courses
+                # Creating directory and save the file to the filesystem if 'saveFileToFilesystem' is True.
                 if saveFileToFilesystem:
                     self.makeDirectory()
                     file.save(self.getFileFullPath(newfilename = saveFileToFilesystem))        
                         
         except AttributeError: # will pass if there is no attachment to save
-            
             pass
 
     def retrievePath(self,files):
--- a/app/logic/fileHandler.py
+++ b/app/logic/fileHandler.py
@@ -22,9 +22,7 @@ class FileHandler:
             os.makedirs(os.path.join(self.path, extraDir))
         # Occurs when we try to create a directory that already exists
         except OSError as e:
-            if e.errno == 17:
-                pass
-            else:
+            if e.errno != 17:
                 print(f'Fail to create directory: {e}')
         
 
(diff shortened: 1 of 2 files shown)


app/logic/fileHandler.py
```diff
@@ -15,11 +15,17 @@ def __init__(self,files=None, courseId=None, eventId=None):
...
             self.path = os.path.join(self.path, app.config['files']['event_attachment_path'])
         
+    def makeDirectory(self):
+        # Creating the directory when call in saveFiles
         try:
             extraDir = str(self.eventId) if self.eventId else ""
             os.makedirs(os.path.join(self.path, extraDir))
-        except:
-            print("Directory exists.")
+        # Passing if the error is 17
```
BrianRamsay: This comment isn't useful. I can tell that you are looking at error 17 from the code. But I can't tell what error 17 is.

app/logic/fileHandler.py
```diff
@@ -15,11 +15,17 @@ def __init__(self,files=None, courseId=None, eventId=None):
...
+        except OSError as e:
+            if e.errno != 17:
+                pass
+            # Perform additional error handling if needed
```
BrianRamsay: Remove this comment

app/logic/fileHandler.py
```diff
@@ -15,11 +15,17 @@ def __init__(self,files=None, courseId=None, eventId=None):
             # eventID is not included in the path, because it is now a part of the attachment filename.
             self.path = os.path.join(self.path, app.config['files']['event_attachment_path'])
         
+    def makeDirectory(self):
+        # Creating the directory when call in saveFiles
```
BrianRamsay: Don't assume where this function will be called.

BrianRamsay: Also there should be a better description that describes what the function does and/or why it exists

app/logic/fileHandler.py
```diff
@@ -43,25 +49,33 @@ def saveFiles(self, saveOriginalFile=None):
...
-                        file.save(self.getFileFullPath(newfilename = file.filename))
+                        saveFileToFilesystem = file.filename
+                
+                # Creating directory and saving file for events and courses
+                if saveFileToFilesystem:
+                    self.makeDirectory()
+                    file.save(self.getFileFullPath(newfilename = saveFileToFilesystem))        
+                        
         except AttributeError: # will pass if there is no attachment to save
+
```
BrianRamsay: remove the newline here. That will connect the pass to the except better, visually

tests/code/test_fileHandler.py
```diff
@@ -35,6 +35,16 @@ def test_getFileFullPath():
     filePath = handledCourseFile.getFileFullPath(courseFileStorageObject[0].filename)
     assert filePath == 'app/static/files/courseattachments/1/coursefile.pdf'
 
+@pytest.mark.integration
+def test_makingdirectory():
+    #Testing that the file is created and it exists
+    with mainDB.atomic() as transaction:
+        
+        handledEventFile.makeDirectory()
+        handledEventFile.makeDirectory()
```
BrianRamsay: You never actually test that the directory was created. Also, you aren't using the database, so you don't need to use a database transaction.

BrianRamsay (changes requested):

AndersonStettner:
[Clean up empty directories for file attachments on recurring events. #972](https://trello.com/c/2MZBaj2L/642-clean-up-empty-directories-for-file-attachments-on-recurring-events-972)

app/logic/fileHandler.py
BrianRamsay: don't say "when called in saveFiles" in the comment. It doesn't matter to the function where it is called.

app/logic/fileHandler.py
```diff
@@ -15,11 +15,16 @@ def __init__(self,files=None, courseId=None, eventId=None):
...
+        # Creating the directory when call in saveFiles
         try:
             extraDir = str(self.eventId) if self.eventId else ""
             os.makedirs(os.path.join(self.path, extraDir))
-        except:
-            print("Directory exists.")
+        # Occurs when we try to create a directory that already exists
+        except OSError as e:
+            if e.errno != 17:
+                pass
```
BrianRamsay: This try/except is not doing what you want. You wanted it to ignore only error 17, but what you are doing is ignoring all `OSError` exceptions, and the conditional doesn't matter. You need to re-raise the exception if you don't want to catch it, not pass it.

app/logic/fileHandler.py
```diff
@@ -36,31 +41,38 @@ def getFileFullPath(self, newfilename = ''):
...
+                            saveFileToFilesystem = attachmentName
  
                 elif self.courseId:
                     isFileInCourse = AttachmentUpload.select().where(AttachmentUpload.course == self.courseId, AttachmentUpload.fileName == file.filename).exists()
                     if not isFileInCourse:
                         AttachmentUpload.create(course = self.courseId, fileName = file.filename)
-                        file.save(self.getFileFullPath(newfilename = file.filename))
+                        saveFileToFilesystem = file.filename
+                
+                # Creating directory and save the file to the filesystem if 'saveFileToFilesystem' is True.
```
BrianRamsay: `saveFileToFilesystem` will never contain True. And the `if` part is obvious from the code, anyway. We don't want things in comments that are obvious from the code (unless they are more like structural comments). Comments are for organization and explaining the not obvious things.

BrianRamsay (changes requested):

tests/code/test_fileHandler.py
```diff
@@ -35,6 +35,15 @@ def test_getFileFullPath():
     filePath = handledCourseFile.getFileFullPath(courseFileStorageObject[0].filename)
     assert filePath == 'app/static/files/courseattachments/1/coursefile.pdf'
 
+@pytest.mark.integration
+def test_makingdirectory():
+    #Testing that the file is created and it exists
+
```
BrianRamsay: We should make sure that the directory doesn't exist before we try making it, because otherwise we could be not creating it and we wouldn't know it.

BrianRamsay (changes requested):

michelmonc:
Resolve #972

BrianRamsay (approved):

PR #1008, opened 2023-07-18
Removed call to program.id which should have been program.

### LINKED ISSUE #1002: Fix `Upcoming Events` button on the landing page
The buttons on the landing page that indicate whether or not there is an upcoming event for a program are broken. Event if there are future events scheduled for a program, the buttons are still grayed out and say there are no upcoming events. Investigate and fix.

andersoncedu:
Issue: The issue was that upcoming events weren't showing up on the landing page. There was an incorrect call to the program.id instead of program.

Fix: Changed call from program.id to program. Additionally added check to see if event was in past in the list comprehension passed to the landing page.

Test: Add new events in the future and ensure that they display appropriately.

Resolves #1002

### DIFF
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -45,7 +45,7 @@ def landingPage():
 
 
     eventsInTerm = list(Event.select().where(Event.term == g.current_term))
-    programsWithEventsList = [program.program for program in eventsInTerm if not program.isPast]
+    programsWithEventsList = [event.program for event in eventsInTerm if not event.isPast]
 
     return render_template("/main/landingPage.html", managerProgramDict = managerProgramDict,
                                                      term = g.current_term,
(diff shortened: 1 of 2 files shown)


andersoncedu:
[Fix Upcoming Events button on the landing page #1002](https://trello.com/c/WhYBBDvE/665-fix-upcoming-events-button-on-the-landing-page-1002)

AndersonStettner (changes requested):
This does let the button work but not as intended. It will be clickable if there are any events, future or past, and we only want it to be clickable if there are future events.

app/controllers/main/routes.py
```diff
@@ -45,7 +45,7 @@ def landingPage():
 
 
     eventsInTerm = list(Event.select().where(Event.term == g.current_term))
-    programsWithEventsList = [program.program for program in eventsInTerm]
+    programsWithEventsList = [program.program for program in eventsInTerm if not program.isPast]
```
BrianRamsay: Even though this wasn't your change, we should fix this variable name to make sense. It is an `event`, not a `program`.

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #1012, opened 2023-07-25
999 training hover

### LINKED ISSUE #999: Add a "View" hover for training to the `Track Volunteer` page
In the `Programs` accordion on `User Profile` there is a `View` hover that lets the user know if they have or have not completed the training events for a program. We want to add that same functionality to the `Track Volunteers` page.

hoerstl:
Issue: Issue #999 asked us to create a view hover to show all relevant trainings for a program and whether a user had been to them next to each user entry in the track volunteers page. There is a matching view hover in each person's profile that should be duplicated into the track volunteer page when an event is in the future.

Solution: The example inside of the user profile page was perfect for getting the trainings for a single user but what we needed was the ability to get them for a whole list of users without making a query for each. Therefore, we had to change what the getUserParticipatedTrainingEvents function did. It now accepts a list of users and where it returned a list before, it now returns a dictionary with a key for each of the users and their corresponding list in a similar fashion. Now that the underlying function has been changed to fit the use case for both locations, we just used it to pass a new variable to the render_template which we use to create the same hover as is used inside of the user profile page. The hover's data-content is now retrieved via a snippet file holding a macro since it's the same in both spots. This filename is trainingsHoverMacro.html.

To Test: Create an event in the future that belongs to a particular program. For that program in the same academic year, create another event in the past and mark it as a training, then, add any volunteer to it. At this point, go back to the first event you created and manually rsvp (add) the person you added to the second event along with someone else. Ensure that the view hover appears for both of them. Now, ensure that one of the view hovers says that they attended the training and the other says that they did not. 
Another place to check would be the user profile page. Go to that page and ensure that the view hovers there are functional as well. Note that the view hovers on this page show all trainings that are offered in the same academic year as the current term that we've set in our cookies whereas the event page shows all trainings that are offered in the same academic year as the event's term so the two may differ in information if your term is set differently than the event.


app/logic/participants.py
```diff
@@ -140,21 +140,29 @@ def getUserParticipatedTrainingEvents(program, user, currentTerm):
...
+        try:
+            if training.isPast:
+                trainingData[training] = trainingData.get(training, []) + [training.eventparticipant.user_id]
+            else:  # The training has yet to happen
+                trainingData[training] = trainingData.get(training, []) + [training.eventrsvp.user_id]
+        except AttributeError:
+            trainingData[training] = trainingData.get(training, [])
+
+    # Dict; key: username; value: tuple consisting of (trainingObj, didAttend/didRsvp, optional date str if event is in future)
+    result = {}  # Beans- rename this dict
```
AndersonStettner: Rename the dict if you still want to or remove this comment.

AndersonStettner (changes requested):

app/controllers/admin/volunteers.py
```diff
@@ -68,11 +68,13 @@ def trackVolunteersPage(eventID):
...
         eventWaitlistData = [volunteer for volunteer in eventParticipantData + eventRsvpData if volunteer.rsvpWaitlist and event.isRsvpRequired]
-        eventVolunteerData = [volunteer for volunteer in eventParticipantData + eventRsvpData if volunteer not in eventWaitlistData]
+        eventVolunteerData = [volunteer for volunteer in eventRsvpData if volunteer not in eventWaitlistData]
         eventNonAttendedData = []
         
-   
-    
+    program = event.program
+
+    allReleventUsers = [participant.user for participant in eventVolunteerData + eventNonAttendedData + eventWaitlistData]
```
BrianRamsay: we should spell 'relevant' correctly

tests/code/test_participants.py
```diff
@@ -383,11 +383,11 @@ def test_getUserParticipatedTrainingEvents():
...
                                       program = Program.get_by_id(8))
 
-        # If the event has not occured yet, assert their participated status for that event == None
-        allProgramTrainings = Event.select().where(Event.isTraining == 1, Event.program == Program.get_by_id(8))
-        listOfProgramTrainings = [programTraining for programTraining in allProgramTrainings]
-        programTrainings = getUserParticipatedTrainingEvents(Program.get_by_id(8), User.get_by_id("ramsayb2"), currentTerm)
-        for training in programTrainings.keys():
-            assert programTrainings[training][0] == None
-            assert programTrainings[training][1] == datetime.strptime(testingEvent.startDate, "%Y-%m-%d").strftime("%m/%d/%Y")
+        # If the event has not occured yet, assert that someone who has rsvp'd is true and someone who hasn't is false
```
zachneill: there's two r's in the word occurred lawrence

tests/code/test_participants.py
```diff
@@ -356,22 +356,22 @@ def test_getUserParticipatedTrainingEvents():
             if (counter % 2) == 0:
                 EventParticipant.create(user = User.get_by_id("ramsayb2"), event = training)
 
-        programTrainings = getUserParticipatedTrainingEvents(Program.get_by_id(2), User.get_by_id("ramsayb2"), currentTerm)
-        for counter, training in enumerate(programTrainings.keys()):
+        programTrainings = getParticipationStatusForTrainings(Program.get_by_id(2), [User.get_by_id("ramsayb2")], currentTerm)
+        for counter, (training, attended) in enumerate(programTrainings['ramsayb2']):
             if (counter % 2) == 0:
```
zachneill, line 419: this is literally python you don't need the parentheses around counter % 2

zachneill (dismissed):
thank you for titlecasing your branch. i really appreciate it

tests/code/test_participants.py
BrianRamsay: It's an order of operations clarification. I'm ok with it.

BrianRamsay (changes requested):
By default, I see myself as not attending an All Celts Training that is tied to Adopt a Grandparent (from the profile View), but that training isn't in the current term. Trainings are only valid for the current term. All Volunteer Trainings are valid for the academic year. I think this may be correct in the View that is tied to an event, but not the View that is on the program in the profile. They should share data and rendering code.

app/templates/events/event_list.html
```diff
@@ -86,10 +86,10 @@ <h1 class="text-center">Events List for {{selectedTerm.description}}</h1>
             <tr class="">
           {% endif %}
             {% if type != "bonner" and type != "studentLed" %}
-              {% if event.program%}
-                <td class="text-dark">{{event.program.programName}}</td>
-              {% else %}
+              {% if event.isAllVolunteerTraining %}
                 <td class="text-dark">This event does not belong to any program.</td>
```
AndersonStettner: All Volunteer Training does belong to a program now that other CELTS is a program. Though, since All Volunteer is a required training for every program I think it might work to make this message something like "This is a required training for all programs."

app/templates/events/trackVolunteers.html
```diff
@@ -134,9 +135,11 @@ <h2 class="accordion-header">
                     {% else %} {# end hiding the hours and attended inputs #}
                       <td style="text-align:center">
                         {% if participant.user.username not in trainedParticipantsList %}
-                          <span class="bi bi-x-lg" aria-label="Did not complete all trainings."></span>
+                          <span class="bi bi-x-lg" aria-label="Did not complete all trainings."></span> &nbsp&nbsp
```
AndersonStettner: Set a variable for the label.

app/templates/main/userProfile.html
```diff
@@ -146,40 +146,22 @@ <h3 class="accordion-header" id="headingThree">
...
                   <td>
                     {% if row.completedTraining or not trainingList %}
                       <i class="bi bi-check-lg"></i> &nbsp&nbsp
-                      <a class="trainingPopover" id="{{row.program.id}}" href="javascript:void(0);" data-toggle="popover" title="Program Trainings" data-content= "{{trainingListView()}}">View</a>
+                      <a class="trainingPopover" id="{{row.program.id}}" href="javascript:void(0);" data-toggle="popover" title="Program Trainings" data-content= "{{trainingsHover(trainingList)}}">View</a>
                     {% else %}
                       <i class="bi bi-x-lg"></i> &nbsp&nbsp
-                      <a class="trainingPopover" id="{{row.program.id}}" href="javascript:void(0);" data-toggle="popover" title="Program Trainings" data-content= "{{trainingListView()}}">View</a>
+                      <a class="trainingPopover" id="{{row.program.id}}" href="javascript:void(0);" data-toggle="popover" title="Program Trainings" data-content= "{{trainingsHover(trainingList)}}">View</a>
```
AndersonStettner: Both of these hovers are exactly the same, what is the point of the condition?

app/templates/trainingsHoverMacro.html
```diff
@@ -0,0 +1,25 @@
...
+                {% set date = training.startDate.strftime("%m/%d/%Y") %}
+                {% if isRsvpd %}
+                    <p><b>{{trainingName}}</b>: <br>Rsvp'd for training on {{date}}</p>
+                {% else %}
+                    <p><b>{{trainingName}}</b>: <br>Not Rsvp'd for training on {{date}}</p>
+                {% endif %}
+            {% endif%}
+        {% endfor %}
+    {% else %}
+    <p>No trainings for this program.</p>
```
AndersonStettner: I think this should probably be something like "There are no trainings required for this program." or "This program has no required trainings."

tests/code/test_userManagement.py
```diff
@@ -119,8 +119,8 @@ def test_getAllowedPrograms():
     with mainDB.atomic() as transaction:
         # checks the length of all programs an admin has access to and compares that to total programs
         allowedPrograms = len(getAllowedPrograms(User.get_by_id("ramsayb2")))
-        totalPrograms = Program.select().count()
-        assert allowedPrograms == totalPrograms
+        NonOtherPrograms = Program.select().count() - 1 # - Other Celts-Sponsored
```
AndersonStettner: I do not like the variable name `NonOtherPrograms`. the first letter should not be capital and the comment suggests that this is for the other CELTS program so why is programs plural in the variable?

AndersonStettner (changes requested):

AndersonStettner (dismissed):
The `X` is not changing to a `check` despite the volunteer having attended the only training required for the event.

hoerstl:
The tests are broken because now that we've restricted what events the trainedParticipants function checks to see if you've attended, khatts has not attended an event for the current term (id=4). The tests in test_participants.py need to be completely rewritten to accommodate for this correct restriction in what that function returns. To recap, it checks to see if you've been to all program trainings offered within the current semester and all All Volunteer Trainings from the term's academic year.

BrianRamsay (approved):

PR #1013, opened 2023-07-27
691 celts sponsored program

### LINKED ISSUE #691: Make CELTS sponsored events into a program
In order to handle permissions for events that don't have a program we need to change the "other" category to a "CELTS-Sponsored" program. This means all of the places that assume there are events with no program need to be altered.

ojmakinde:
Issue #691

Issue: CELTS-Sponsored Events and All Volunteer Trainings used to have a null value in the program_id column.

Solution: We created a program for CELTS-Sponsored Events and All Volunteer Trainings. Additionally, we ensured all test data were correctly working with our added changes. We changed the links on create event to create events that links to the "Other CELTS-Sponsored Program"

To test, go to Create Event and create a CELTS-Sponsored Event and an All Volunteer Training, both in the past. Add Zach Neill as a volunteer to both events. Visit the database and ensure the both events have a program id value of 9. Then, visit Zach Neill's profile and view his transcript and verify the hours he earned from volunteering at the created events appear and are under the title "CELTS Sponsored Events".

Resolves #691


app/logic/userManagement.py
```diff
@@ -53,7 +53,8 @@ def changeProgramInfo(newProgramName, newContactEmail, newContactName, newLocati
 def getAllowedPrograms(currentUser):
     """Returns a list of all visible programs depending on who the current user is."""
     if currentUser.isCeltsAdmin:
-        return Program.select().order_by(Program.programName)
+        # Beans: ask if this where clause is an acceptable way to filter out the celts sponsored event from appearing in the list of programs
+        return Program.select().where(Program.programName != "CELTS-Sponsored Event").order_by(Program.programName)
```
BrianRamsay: We can just change this to use the id 9, right? Also, why are we not letting admins see these?

BrianRamsay (changes requested):
Instead of having the magic number 9 everywhere, let's define a class variable Program.OTHER_CELTS or something that we can use. (see Certification.BONNER)

app/logic/userManagement.py
hoerstl: Thanks for the suggestion about the class variable. We are not letting admins see these because the buttons that now create other celts sponsored events is returned from getAllowedTemplates. If we allowed this to return the celts sponsored events as a program, that button would appear twice on the page. Since this function is only used to, from the docstring, "Returns a list of all visible programs", we didn't see a need to return this as a program. This is the only use of this function.

app/logic/transcript.py
```diff
@@ -11,7 +11,7 @@
 
 def getOtherEventsTranscript(username):
     """
-    Returns a Other Events query object containing all the non-program
+    Returns a Other Events query object containing all the other-program
```
zachneill: while you're at it changing this docstring, it's "Returns **_an_** Other Events query object..."

database/test_data.py
```diff
@@ -450,6 +450,16 @@
         "contactEmail": "",
         "contactName": ""
     },
+    {
+        "id": 9,
+        "programName": "CELTS-Sponsored Event",
+        "programUrl": "https://www.berea.edu/celts/",
```
zachneill: dead link, it's actually https://www.berea.edu/centers/center-for-excellence-in-learning-through-service

you're leading your own poor staff and students astray. thats bad and you should feel bad

zachneill (changes requested):
this is fun

app/models/program.py
```diff
@@ -14,6 +14,8 @@ class Program(baseModel):
     contactEmail = CharField(null=True,default='')
     defaultLocation = CharField(null=True,default='')
 
+    OTHERCELTSSPONSORED = 9
```
AndersonStettner: To keep with how we are already doing this for other things this should be changed to a program field.

AndersonStettner (changes requested):

database/test_data.py
hoerstl: Atoning for sins is a blessing

app/models/program.py
hoerstl: That's not quite a field in the database. That variable is simply meant as a static and constant class variable for the Program class that we can use to refer to the magic number 9. I don't believe we want to make it a program field because that would add it to the database. See line 8 in /app/models/certification.py for an example of how we use it for BONNER inside of Certification.

hoerstl: Okay, i do it

tests/code/test_scripts.py
```diff
@@ -2,6 +2,7 @@
 from datetime import date, datetime, timedelta
 
 from app.models import mainDB
+from app.models.program import Program
```
AndersonStettner: Why did you add this import?

app/models/event.py
```diff
@@ -21,7 +21,7 @@ class Event(baseModel):
     recurringId = IntegerField(null=True)
     contactEmail = CharField(null=True)
     contactName = CharField(null=True)
-    program = ForeignKeyField(Program, null= True)
+    program = ForeignKeyField(Program, default=9)
```
AndersonStettner: We do not need any default. It can just be `ForeignKeyField(Program)`

app/logic/transcript.py
```diff
@@ -36,7 +36,8 @@ def getProgramTranscript(username):
     EventData = (Event
         .select(Event, fn.SUM(EventParticipant.hoursEarned).alias("hoursEarned"))
         .join(EventParticipant)
-        .where(EventParticipant.user == username)
+        .where(EventParticipant.user == username, 
+        Event.program != 9)
```
AndersonStettner: The where clause is short enough to all be on the same line, also we should not be using a magic number.

app/logic/transcript.py
```diff
@@ -20,7 +20,7 @@ def getOtherEventsTranscript(username):
                     .join(Program, JOIN.LEFT_OUTER).switch(Event)
                     .join(Term)
                     .where(EventParticipant.user==username,
-                    Event.program == None)
+                    Event.program == 9)
```
AndersonStettner: This where clause is short enough to all be on the same line, also we should not be using a magic number.

app/logic/events.py
```diff
@@ -227,7 +227,7 @@ def getOtherEvents(term):
                             .where(Event.term == term,
                                    Event.isTraining == False,
                                    Event.isAllVolunteerTraining == False,
-                                   ((Event.program == None) |
+                                   ((Event.program_id == 9) |
```
AndersonStettner: Like with Bonner events above, you can change `Event.program_id == 9` to `Event.isCeltsSponsored`

app/controllers/events/email.py
```diff
@@ -32,7 +32,7 @@ def retrieveSenderList(eventId):
     senderOptions = [["CELTS (celts@berea.edu)", "Celts"]]
 
     event = Event.get_by_id(eventId)
-    if event.program_id:
+    if event.program_id != 9:
```
AndersonStettner: No magic number.

app/controllers/admin/routes.py
```diff
@@ -48,17 +48,19 @@ def switchUser():
...
+                                celtsSponsoredProgram = celtsSponsoredProgram,
                                 templates=visibleTemplates)
     else:
         abort(403)
 
 
 @admin_bp.route('/eventTemplates/<templateid>/create', methods=['GET','POST'])
 @admin_bp.route('/eventTemplates/<templateid>/<programid>/create', methods=['GET','POST'])
-def createEvent(templateid, programid=None):
+def createEvent(templateid, programid=9):
```
AndersonStettner: No magic number.

AndersonStettner (changes requested):

app/templates/events/event_list.html
```diff
@@ -86,10 +86,10 @@ <h1 class="text-center">Events List for {{selectedTerm.description}}</h1>
             <tr class="">
           {% endif %}
             {% if type != "bonner" and type != "studentLed" %}
-              {% if event.program%}
-                <td class="text-dark">{{event.program.programName}}</td>
-              {% else %}
+              {% if event.isAllVolunteerTraining %}
```
BrianRamsay: We don't really need to have this condition there anymore, because it has a program name

app/controllers/admin/routes.py
```diff
@@ -48,17 +48,18 @@ def switchUser():
 def templateSelect():
     if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff:
         allprograms = getAllowedPrograms(g.current_user)
+        celtsSponsoredProgram = Program.get(Program.programName == 'CELTS-Sponsored Event')
```
BrianRamsay: There is a database flag for this now. Don't select by program name

app/controllers/events/email.py
```diff
@@ -32,7 +32,7 @@ def retrieveSenderList(eventId):
     senderOptions = [["CELTS (celts@berea.edu)", "Celts"]]
 
     event = Event.get_by_id(eventId)
-    if event.program_id:
+    if not event.program.isOtherCeltsSponsored:
```
BrianRamsay: Why are we treating Other CELTS Sponsored differently?

app/logic/transcript.py
BrianRamsay: No magic numbers. Use the database field

app/logic/transcript.py
```diff
@@ -33,10 +33,11 @@ def getProgramTranscript(username):
     """
     # Add up hours earned in a term for each program they've participated in
     
+    OTHERCELTSSPONSORED = 9
     EventData = (Event
         .select(Event, fn.SUM(EventParticipant.hoursEarned).alias("hoursEarned"))
         .join(EventParticipant)
-        .where(EventParticipant.user == username)
+        .where(EventParticipant.user == username, Event.program != OTHERCELTSSPONSORED)
```
BrianRamsay: Extra strange, considering we have a database field for this now.

app/logic/userManagement.py
```diff
@@ -53,7 +53,7 @@ def changeProgramInfo(newProgramName, newContactEmail, newContactName, newLocati
 def getAllowedPrograms(currentUser):
     """Returns a list of all visible programs depending on who the current user is."""
     if currentUser.isCeltsAdmin:
-        return Program.select().order_by(Program.programName)
+        return Program.select().where(Program.isOtherCeltsSponsored == False).order_by(Program.programName)
```
BrianRamsay: Why are we excluding these programs? The previous query didn't restrict by program at all.

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #1014, opened 2023-07-27
Implementation of Canceled event Concept

### LINKED ISSUE #985: Add the concept of a `Canceled Event`
Canceled events would preserve their data and be visible to the same admins and staff. Need to evaluate existing queries for inclusion or exclusion of canceled events. Need to evaluate different views of an event and how Canceled is indicated. An automated email should be sent to people who have RSVPd to a canceled event.

Mercy-Eze:
This PR addresses the issue of preserving data for canceled events while ensuring that the relevant information is visible to authorized admins and staff only. It also introduces improvements for handling different views of a canceled event. It includes the following change:

- Modified Event Queries
- Tests for New Function
- Data Preservation for Canceled Events
- Modified Event View
- Administrative Permission only

Resolves issue #985 

P.S: Automated Email Notifications were not Implemented Following New Directives


AndersonStettner (changes requested):
Make sure recurring events are accounting for when an event in that series is canceled.

app/controllers/admin/routes.py
```diff
@@ -262,6 +262,18 @@ def eventDisplay(eventId):
                                 image = image,
                                 pageViewsCount= pageViewsCount)
 
+
+@admin_bp.route('/event/<eventId>/cancel', methods=['POST'])
+def cancelRoute(eventId):
+    try:
+        cancelEvent(eventId)
+        flash("Event successfully canceled.", "success")
+        return redirect(url_for("main.events", selectedTerm=g.current_term))
```
AndersonStettner: Stay on the event page after canceling it, do not go back out to the events list page.

app/templates/admin/createEvent.html
```diff
@@ -252,13 +252,40 @@
...
           </div>
     </div>
 </form>
-<!-- Modal -->
+<!-- Modal for Cancel Event -->
+<div class="modal fade" id="cancelWarning" tabindex="-1" aria-labelledby="warningModel" aria-hidden="true">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title" id="warningModel">Cancel Event</h5>
```
AndersonStettner: We should use a more specific id name

app/templates/admin/createEvent.html
```diff
@@ -252,13 +252,40 @@
...
+        <p> Are you sure you want to cancel this event? <br> By cancelling this event, the information about it will be retained for any future reference.</p>
+      </div>
+      <div class="modal-footer justify-content-between">
+          <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Close</button>
+          <form action="/event/{{eventData.id}}/cancel" method='post'>
+              <button type="submit" class="btn btn-warning">Cancel Event</button>
+          </form>
+      </div>
+      <div>
+        <input type="hidden" value="{{eventData.id}}" name="id">
```
AndersonStettner: Add some more detail to the name of the input. What is it an id of?

app/templates/admin/createEvent.html
```diff
@@ -269,7 +296,7 @@ <h5 class="modal-title" id="warningModel">Delete Event</h5>
         {% if eventData.recurringId%}
         <div class="modal-body">
           <div class= "container-fluid">
-            <p>This action is irreversible.</p>
+            <p>This action is irreversible. <br> By deleting this event, all information regarding this event will be entirely removed from the website. </p>
```
AndersonStettner: Remove the word "entirely".

app/templates/admin/createEvent.html
```diff
@@ -290,7 +317,7 @@ <h5 class="modal-title" id="warningModel">Delete Event</h5>
        </div>
         {% else %}
         <div class="modal-body">
-          <p>This action is irreversible.</p>
+          <p>This action is irreversible. <br> By deleting this event, all information regarding this event will be entirely removed from the website. </p>
```
AndersonStettner: Remove the word "entirely"

app/templates/admin/createEvent.html
```diff
@@ -252,13 +252,40 @@
...
+<!-- Modal for Cancel Event -->
+<div class="modal fade" id="cancelWarning" tabindex="-1" aria-labelledby="warningModel" aria-hidden="true">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title" id="warningModel">Cancel Event</h5>
+        <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+      </div>
+      <div class="modal-body">
+        <p> Are you sure you want to cancel this event? <br> By cancelling this event, the information about it will be retained for any future reference.</p>
```
AndersonStettner: Indicate the difference between a canceled event and a deleted event. I find this message confusing and I am not really sure what it means.

app/templates/events/event_list.html
```diff
@@ -96,12 +96,15 @@ <h1 class="text-center">Events List for {{selectedTerm.description}}</h1>
...
             {% else %}
-              {% set link_class = 'link-secondary' if event.isPast else 'link-primary' %}
-              <td><a href="/event/{{event.id}}/view" class="{{link_class}}">{{event.name}}</a></td>
+              <td><a href="/event/{{event.id}}/view" class="link-primary">{{event.name}}</a></td>
             {% endif %}
-            <td nowrap>{{event.startDate.strftime('%m/%d/%Y')}}</td>
-            <td nowrap>{{event.timeStart.strftime('%-I:%M %p')}} - {{event.timeEnd.strftime('%-I:%M %p')}}</td>
-            <td class="text-dark">{{event.location}}</td>
+              {% if not user.isAdmin and event.isCanceled %}
+                <td colspan = 3 >This Event Has Been Canceled. Check your Email</td>
```
AndersonStettner: Replace "Check your Email" with "Contact {{programManager}} if you have any questions." and in the case there is no program manager just say "Contact CELTS..."

AndersonStettner (changes requested):

app/templates/events/event_list.html
AndersonStettner: Also, we should not be capitalizing every word in the sentence.

AndersonStettner (changes requested):

app/templates/events/event_list.html
```diff
@@ -96,12 +96,15 @@ <h1 class="text-center">Events List for {{selectedTerm.description}}</h1>
...
             {% else %}
-              {% set link_class = 'link-secondary' if event.isPast else 'link-primary' %}
-              <td><a href="/event/{{event.id}}/view" class="{{link_class}}">{{event.name}}</a></td>
+              <td><a href="/event/{{event.id}}/view" class="link-primary">{{event.name}}</a></td>
             {% endif %}
-            <td nowrap>{{event.startDate.strftime('%m/%d/%Y')}}</td>
-            <td nowrap>{{event.timeStart.strftime('%-I:%M %p')}} - {{event.timeEnd.strftime('%-I:%M %p')}}</td>
-            <td class="text-dark">{{event.location}}</td>
+              {% if not user.isAdmin and event.isCanceled %}
+                <td colspan = 3 >This event has been canceled. Contact CELTS if you have any questions.</td>
```
AndersonStettner: We should also show this message, or a similar one to admins. Also, we should say "Contact {{programManager}}..." if the program has a manager. If there is not program manager then it is fine to keep it as CELTS. We should also turn the program managers name into a mailTo link.

AndersonStettner (changes requested):
If there is an event in the future that has been canceled we should make sure the program cards on the landing page do not say there is an upcoming event. 

Also, all the query formatting you changed should be reverted. If there is anything after the select (join, where order_by, etc...) they should be in line with the select.

app/templates/admin/createEvent.html
```diff
@@ -252,13 +252,40 @@
...
+<!-- Modal for Cancel Event -->
+<div class="modal fade" id="cancelWarning" tabindex="-1" aria-labelledby="cancelModel" aria-hidden="true">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title" id="cancelModel">Cancel Event</h5>
+        <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+      </div>
+      <div class="modal-body">
+        <p> Are you sure you want to cancel this event? <br> By proceeding with the cancellation, you understand that this event will no longer take place on the proposed date, but the event information will be kept for potential future use or reference.</p>
```
AndersonStettner: I think I would prefer something like "A canceled event will no longer take place during its scheduled time but the information will be kept for future reference."

Also, the first sentence and second sentence should be in two different `<p>` tags. This change should also be made for the delete events message.

AndersonStettner (changes requested):

app/controllers/admin/routes.py
```diff
@@ -245,9 +245,15 @@ def eventDisplay(eventId):
...
-                eventData["nextRecurringEvent"] = eventSeriesList[eventIndex + 1]
-
+            # Find the next uncanceled event
+            nextUncanceledEvent = None
+            for nextEvent in eventSeriesList[eventIndex + 1:]:
+                if not nextEvent.isCanceled:  
+                    nextUncanceledEvent = nextEvent
+                    break
+            if nextUncanceledEvent:
+                eventData["nextRecurringEvent"] = nextUncanceledEvent
```
AndersonStettner, line 255: The above logic should be a function in a logic folder.

AndersonStettner (changes requested):
You have not accounted for the "Upcoming Events" accordion on the user profile. You need to make sure everywhere we do queries for events we are making sure to include canceled event checks if necessary.

app/templates/eventNav.html
```diff
@@ -19,8 +19,12 @@
             {{ eventheader(page_title, eventData, 'large', isNewEvent) }}
         {% endif %}
         {% if eventPast %}
-        <div class="alert-danger mb-2 pb-2 pt-2" role="alert">This event is in the past.</div>
+            <div class="alert-danger mb-2 pb-2 pt-2" role="alert">This event is in the past.</div>
         {% endif %}
+        {% if eventData.isCanceled %}
+            <div class="alert-warning mb-2 pb-2 pt-2" role="alert">This event has been canceled.</div>
+        {% endif %}
```
AndersonStettner: We should think about changing the display if an event was canceled and it is in the past. Currently there are two messages shown, we probably want to make it only one.

AndersonStettner (changes requested):

app/templates/events/event_list.html
```diff
@@ -81,9 +81,9 @@ <h1 class="text-center">Events List for {{selectedTerm.description}}</h1>
     <tbody>
       {% for event in events %}
           {% if event.isPast %}
-            <tr class="table-light showlist" style="display: none;">
+            <tr height = '50' class="table-light showlist" style="display: none;">
```
zachneill, line 84: you could use the BS5 class "d-none" instead of that style tag who even taught you to code

tests/code/test_events.py
```diff
@@ -389,6 +389,36 @@ def test_saveEventToDb_update():
...
+        
+        testingEvent = Event.get(Event.name == "Testing canceled event")
+        eventId = testingEvent.id
+        
+
+        with app.app_context():
+            g.current_user = User.get_by_id("ramsayb2")
+            cancelEvent(eventId)
+            event = Event.get(Event.id == eventId)
+        assert event.isCanceled == True
```
zachneill: i'm too lazy to clone this repo but i'm sure you can just do `assert event.isCanceled 
`since it returns a boolean, it will evaluate correctly anyways

tests/code/test_event_list.py
```diff
@@ -172,7 +172,7 @@ def test_training_events(training_events):
                                            isStudent = True,
                                            isFaculty = False,
                                            isStaff = False,
-                                           isCeltsAdmin = False,
+                                           isCeltsAdmin = False,
```
zachneill, line 175: ?????

zachneill (changes requested):
btw brian and scott, moodle 4.0 rolls out soon with a few changes. 

also i found some issues with this pr.

app/templates/admin/createEvent.html
stevensonmichel: Fixed

app/templates/events/event_list.html
stevensonmichel: thank you for your help, Brian and Scott but unfortunately your coding experience has been trumped by the almighty and all-powerful god that is Zach Neil.

tests/code/test_event_list.py
stevensonmichel: ??????????????????

tests/code/test_events.py
stevensonmichel: Good catch!!!

app/controllers/admin/routes.py
BrianRamsay: You should just use the same code as before but add something to the where clause to exclude canceled events

app/controllers/admin/routes.py
```diff
@@ -262,6 +268,17 @@ def eventDisplay(eventId):
                                 image = image,
                                 pageViewsCount= pageViewsCount)
 
+
+@admin_bp.route('/event/<eventId>/cancel', methods=['POST'])
+def cancelRoute(eventId):
```
BrianRamsay, line 269: You need a permissions check here, or anyone can cancel an event by hitting this url with a POST request.

app/templates/admin/createEvent.html
```diff
@@ -252,13 +252,41 @@
...
+      </div>
+      <div class="modal-footer justify-content-between">
+          <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Close</button>
+          <form action="/event/{{eventData.id}}/cancel" method='post'>
+              <button type="submit" class="btn btn-warning">Cancel Event</button>
+          </form>
+      </div>
+      <div>
+        <input type="hidden" value="{{eventData.id}}" name="cancel_id">
+      </div>
```
BrianRamsay: Doesn't look like you use this hidden input anymore

app/templates/events/event_list.html
BrianRamsay: Silly @zachneill . `d-none` is just the style tag for people who want don't want to feel dirty. also in this case they need the style because jquery's toggle() manipulates the style.

app/templates/events/event_list.html
```diff
@@ -96,12 +96,16 @@ <h1 class="text-center">Events List for {{selectedTerm.description}}</h1>
...
                 <td><a href="/event/{{event.id}}/edit" class="{{link_class}}">{{event.name}}</a></td>
             {% else %}
-              {% set link_class = 'link-secondary' if event.isPast else 'link-primary' %}
-              <td><a href="/event/{{event.id}}/view" class="{{link_class}}">{{event.name}}</a></td>
+              <td><a href="/event/{{event.id}}/view" class="link-primary">{{event.name}}</a></td>
             {% endif %}
-            <td nowrap>{{event.startDate.strftime('%m/%d/%Y')}}</td>
-            <td nowrap>{{event.timeStart.strftime('%-I:%M %p')}} - {{event.timeEnd.strftime('%-I:%M %p')}}</td>
-            <td class="text-dark">{{event.location}}</td>
+            {% set ProgramManager = 'programmanager' if isProgramManager  else 'CELTS' %}
```
BrianRamsay: This is putting the text "programmanager" as the contact instead of the actual contact info. Also, variables shouldn't start with a capital letter.

tests/code/test_events.py
```diff
@@ -389,6 +389,36 @@ def test_saveEventToDb_update():
...
+                                      isCanceled = False)
+        
+        testingEvent = Event.get(Event.name == "Testing canceled event")
+        eventId = testingEvent.id
+        
+
+        with app.app_context():
+            g.current_user = User.get_by_id("ramsayb2")
+            cancelEvent(eventId)
+            event = Event.get(Event.id == eventId)
```
BrianRamsay: Let's pull this line out of the block. The result is the same, but really it's connected with the assert, not with the app_context

BrianRamsay (changes requested):

app/templates/events/event_list.html
stevensonmichel: Fixed

app/templates/eventNav.html
stevensonmichel: Fixed

app/templates/events/event_list.html
stevensonmichel: Fixed

tests/code/test_events.py
stevensonmichel: Fixed

app/templates/admin/createEvent.html
stevensonmichel: Fixed

app/templates/admin/createEvent.html
```diff
@@ -290,7 +319,7 @@ <h5 class="modal-title" id="warningModel">Delete Event</h5>
        </div>
         {% else %}
         <div class="modal-body">
-          <p>This action is irreversible.</p>
+          <p>This action is irreversible. <br> By deleting this event, all information regarding this event will be removed from the website. </p>
```
AndersonStettner: This should be two separate p tags

app/controllers/admin/routes.py
stevensonmichel: Fixed

Mercy-Eze: I restored the previous code. We cannot exclude canceled events by using the where clause in the query because then canceled events would not be passed at all and that therefore forfeits the purpose of ability to view information about canceled event

AndersonStettner (changes requested):
We need to create an admin log when an event is canceled.

app/templates/events/event_list.html
```diff
@@ -96,12 +96,20 @@ <h1 class="text-center">Events List for {{selectedTerm.description}}</h1>
...
-            <td nowrap>{{event.timeStart.strftime('%-I:%M %p')}} - {{event.timeEnd.strftime('%-I:%M %p')}}</td>
-            <td class="text-dark">{{event.location}}</td>
+              {% if event.isCanceled %}
+                {% if managersProgramDict[event.program] and managersProgramDict[event.program]["managers"] | length > 2 %} 
+                  <td colspan = 3 >This event has been canceled. Contact <a href="mailto:{{event.contactEmail}}" class="text-dark">{{managersProgramDict[event.program]["managers"]}}</a> if you have any questions.</td>
+                {% else %}
+                  <td colspan = 3 >This event has been canceled. Contact <a href="mailto:celts@berea.edu;" class="text-dark">CELTS</a> if you have any questions.</td>
+                {% endif %}
+              {% else %}
+              {% set id = event.id %}
```
AndersonStettner: Why are you doing this set? You never use it and it would be better to just use event.id if you needed to use it since it would be more readable.

AndersonStettner (changes requested):

app/templates/eventNav.html
```diff
@@ -18,9 +18,17 @@
...
-        <div class="alert-danger mb-2 pb-2 pt-2" role="alert">This event is in the past.</div>
+        {% set alertClass = "alert-danger" if eventPast else "alert-warning"%}
+        {% set display = "" if eventPast or eventData.isCanceled else "d-none"%}
+        <div class = "{{alertClass}} {{display}} mb-2 pb-2 pt-2" role="alert">
+        {% if eventPast and eventData.isCanceled %}
+            This is a past event and it was canceled
+        {% elif eventPast %}
+            This is a past event
+        {% elif eventData.isCanceled %}
+            This event has been canceled
```
AndersonStettner: Add some punctuation to the messages.

app/logic/events.py
```diff
@@ -21,6 +22,21 @@
...
+    event = Event.get_or_none(Event.id == eventId)
+    if event: 
+        event.isCanceled = True
+        event.save()
+
+    program = event.program
+    if program:
+        createAdminLog(f"Canceled <a href= \"{url_for('admin.eventDisplay', eventId = event)}\" > \"{event.name}\" </a> for {program.programName}, which had a start date of {datetime.datetime.strftime(event.startDate, '%m/%d/%Y')}.")
+    else:
+        createAdminLog(f"Canceled a non-program event, <a href= \"{url_for('admin.eventDisplay', eventId = event)}\" > \"{event.name}\"  </a>, which had a start date of {datetime.datetime.strftime(event.startDate, '%m/%d/%Y')}.")
```
AndersonStettner: Don't put the event name in quotations.

AndersonStettner (changes requested):
When canceled events show up in `Upcoming Events` on the `User Profile` we need to indicate that they are canceled in the table.

AndersonStettner (dismissed):

app/controllers/admin/routes.py
BrianRamsay: This is the WHERE that I used to avoid that error: `.where((Event.isCanceled == False) | (Event.id == event.id))`

BrianRamsay (dismissed):

PR #1015, opened 2023-07-27
996 refactor dietary restriction

### LINKED ISSUE #996: Refactor `Dietary Restrictions` tab
The tab should be turned into a general `Volunteer Information` tab. 

The new tab should show volunteer/RSVP information based on selected information. The user should be able to choose what group they want information about (RSVP, RSVP + waitlist, etc...) and the displayed data (dietary restrictions, phone number, emergency contact, etc...). 

There also needs to be the option to print/save  the page.

azabeli:
Issue: The client requested that the previous dietary restrictions page be made into a page where they can print information about their volunteers such as dietary restrictions emergency contact info, etc.

Fixes: We implemented a refactor on the page which included two available view modes, a card view, and a table view. We added checkboxes so users could choose which information they view from the volunteers. We also added a print button that will automatically print the table or card view based on which view the user is currently on.  Additionally, implemented Data Table for displaying the data in a table view.

Test: Add Event participants, and rsvps, rsvps should include both people on the waitlist and those not on the waitlist. There should be some members with dietary restrictions. Ensure proper rendering of data, and that checkboxes properly toggle the respective data to be shown or hidden.

Resolves #996


app/controllers/admin/volunteers.py
```diff
@@ -97,34 +96,37 @@ def trackVolunteersPage(eventID):
...
-
-    #get unique list of users for each category waitlist/notwaitlist
-    volunteerUser = list(set([obj.user for obj in participantsAndRsvp if not obj.rsvpWaitlist and obj.user.dietRestriction]))
-    waitlistUser = list(set([obj.user for obj in participantsAndRsvp if obj.rsvpWaitlist and obj.user.dietRestriction]))
-
+    eventParticipantUsers = [obj.user for obj in eventParticipantData]
+    eventNonAttendedData = [obj.user for obj in eventRsvpData if obj.user not in eventParticipantUsers]
+    
+    #get unique list of users for each category waitlist/notwaitlist,rsvped/attended
+    volunteerUser = list(set([obj.user for obj in eventParticipantData if not obj.rsvpWaitlist]))
```
AndersonStettner: You have essentially replaced this with `attendedUser`. You also never use this anywhere else so if it is not needed remove it.

AndersonStettner (changes requested):

app/templates/events/volunteerInformation.html
```diff
@@ -0,0 +1,161 @@
...
+        <label for="rsvpSelect">RSVP Information</label><br>
+        {% endif %}
+        {% if waitlistUser %}
+        <input class="displayCheckbox noprint" type="checkbox" name="selected_items" id="waitlistSelect" checked>
+        <label for="waitlistSelect">Waitlist</label><br>
+        {% endif %}
+        {% if attendedUser %}
+        <input class="displayCheckbox noprint" type="checkbox" name="selected_items" id="attendedSelect" checked>
+        <label for="attendedSelect">Attended</label><br>
+        {% endif %}
```
AndersonStettner: It is a nitpick but why not just do elif down instead of everything having individual if statements.

app/templates/events/volunteerInformation.html
```diff
@@ -0,0 +1,161 @@
...
+          <h4 class="nameSelect" nowrap><b>{{ participant.firstName }} {{ participant.lastName }}</b></h4>
+          <ul>
+            <div class="phoneSelect" nowrap>{{ participant.phoneNumber }}</div>
+            <div class="emailSelect" nowrap>{{ participant.email }}</div>
+            <div class="dietRestrictionSelect">Dietary Restrictions: {{ participant.dietRestriction }}</div>
+          </ul>
+        </div>
+      {% endif %}
+    {% endfor %}
+    </div>
```
AndersonStettner: Same as previous comment, why not just elif?

app/templates/events/volunteerInformation.html
```diff
@@ -0,0 +1,161 @@
...
+            <td class="border p-1 px-3 nameSelect" nowrap><b>{{ participant.firstName }} {{ participant.lastName }}</b></td>
+            <td class="border p-1 px-3 phoneSelect" nowrap>{{ participant.phoneNumber }}</td>
+            <td class="border p-1 px-3 emailSelect" nowrap>{{ participant.email }}</td>
+            <td class="border p-1 px-3 dietRestrictionSelect">{{ participant.dietRestriction }}</td>
+          </tr>
+          {% endif %}
+        </tbody>
+      {% endfor %}
+      </table>
+
```
AndersonStettner: Fix indentation please.

AndersonStettner (changes requested):

app/controllers/admin/volunteers.py
andersoncedu: You are correct, we've removed volunteerUser.

app/templates/events/volunteerInformation.html
andersoncedu: The elif won't work with our current implementation in JavaScript as if there was a person who was both an RSVP and an attended user they would only show up for one of those categories.

app/templates/events/volunteerInformation.html
andersoncedu: Answered above.

app/templates/events/volunteerInformation.html
andersoncedu: We have fixed the indentation errors.

app/controllers/admin/volunteers.py
```diff
@@ -97,34 +96,37 @@ def trackVolunteersPage(eventID):
...
+    eventRsvpData = list(EventRsvp.select().where(EventRsvp.event==event))
     eventParticipantData = list(EventParticipant.select().where(EventParticipant.event==event))
     participantsAndRsvp = (eventParticipantData + eventRsvpData)
-
-    #get unique list of users for each category waitlist/notwaitlist
-    volunteerUser = list(set([obj.user for obj in participantsAndRsvp if not obj.rsvpWaitlist and obj.user.dietRestriction]))
-    waitlistUser = list(set([obj.user for obj in participantsAndRsvp if obj.rsvpWaitlist and obj.user.dietRestriction]))
-
+    print('/'*100)
+    print(participantsAndRsvp)
```
AndersonStettner: remove debug statements.

AndersonStettner (changes requested):
Rename `Track Volunteers` to `Manage Volunteers` and `Volunteer Information` to `Volunteer Details`.

app/controllers/admin/volunteers.py
andersoncedu: I have removed the debug output.

AndersonStettner (changes requested):
The table in `Table View` should be a striped table. Also, While in `Card View` if the user does not have a phone number we should "No number" and in `Table View`, leave the column blank if it is "None" instead of putting "None" in.

AndersonStettner (changes requested):
Add a header to the top of the page when you are printing that has the name and start date of the event and possibly the start time as well.

AndersonStettner (changes requested):
If you select or unselect options from the "Included Information:" section while in `Table View`  and then go to a different page of the table, the new page of the table does not have the correct information displayed.

app/controllers/admin/volunteers.py
```diff
@@ -57,8 +57,8 @@ def trackVolunteersPage(eventID):
     if not (g.current_user.isCeltsAdmin or (g.current_user.isCeltsStudentStaff and isProgramManager)):
         abort(403)
 
-    eventParticipantData = list(EventParticipant.select().where(EventParticipant.event==event))
-    eventRsvpData = list(EventRsvp.select().where(EventRsvp.event==event).order_by(EventRsvp.rsvpTime))
+    eventParticipantData = list(EventParticipant.select(EventParticipant, User).where(EventParticipant.event==event).join(User))
```
AndersonStettner: joins should always be before the where

AndersonStettner (changes requested):

azabeli:
This issue#996 is almost done, the only thing left is to make sure the barcode on the Cardview is correct and can be scanned. When we checked using a scanner, none of them could be scanned with the phone. Probably, the barcode was too tinny and that could be the reason, but it can be more than that.

PR #1022, opened 2023-08-02
992 emergency contact information

### LINKED ISSUE #992: Add Emergency Contact and Insurance Information to User Profile
Using the document below, create a new page linked off of the User Profile to a new emergency contact and insurance information form. All of the input fields on this new form should be saved automatically (like phone number on User Profile) and we should create a printout of the data so that it looks as close to the original form as possible. 
[request.pdf](https://github.com/BCStudentSoftwareDevTeam/celts/files/12028236/request.pdf)

hoerstl:
Issue #992 

Issue - This was an enhancement to create a new method to add an emergency contact as well as a user's insurance information to their profiles. This information should then be used to create a printout of the data so that it looks just like the form here: [https://github.com/BCStudentSoftwareDevTeam/celts/files/12028236/request.pdf](https://github.com/BCStudentSoftwareDevTeam/celts/files/12028236/request.pdf)

Solution - We created two data tables to store this information. The user profile now has a select on the top right which allows navigation to the three new pages we've created. One for Emergency Contact, another for Insurance Information, and finally another to print the form. 

To Test - Go to any user profile as the user who owns it. Now, open the 'Choose an Action' select and choose the options. Submit data to the emergency contact and insurance information pages. Now, got to 'Print Travel Form' and ensure that the form prompts you to print correctly with the information you've just filled in. Now go to another user's profile as someone else (e.x. neillz checking out ramsayb2's profile). Go to their edit contact info page and ensure that it's view only.


app/controllers/main/routes.py
```diff
@@ -156,10 +160,103 @@ def viewUsersProfile(username):
...
+    This loads the Emergency Contact Page
+    """
+    if not (g.current_user.username == username or g.current_user.isCeltsAdmin):
+        abort(403)
+
+
+    if request.method == 'GET':
+
+        # Beans: Q, Should student staff be able to see the emergency contact and insurance info of everyone?
+        readOnly = False if g.current_user.username == username else True
```
zachneill: u can just do `readOnly = g.current_user.username != username`

it evaluates the way you want it to. the only problem is readability which tbh is a you problem

app/controllers/main/routes.py
```diff
@@ -156,10 +160,103 @@ def viewUsersProfile(username):
...
+        
+
+        contactInfo = EmergencyContact.get_or_none(EmergencyContact.user_id == username)
+        return render_template ("/main/emergencyContactInfo.html",
+                                username=username,
+                                contactInfo=contactInfo,
+                                readOnly = readOnly
```
zachneill: the other parameters don't have   s p a c e s    b e t w e e n    t h e     e q u a l s     s i g n

app/controllers/main/routes.py
```diff
@@ -156,10 +160,103 @@ def viewUsersProfile(username):
...
+def insuranceInfo(username):
+    """
+    This loads the Insurance Information Page
+    """
+    if not (g.current_user.username == username or g.current_user.isCeltsAdmin):
+            abort(403)
+
+    if request.method == 'GET':
+        # Beans: Q, Should student staff be able to see the emergency contact and insurance info of everyone?
+        readOnly = False if g.current_user.username == username else True
```
zachneill: see previous comment

app/models/emergencyContact.py
```diff
@@ -0,0 +1,13 @@
+from app.models import *
```
zachneill, line 1: do you need this import?

app/static/css/travelForm.css
```diff
@@ -0,0 +1,38 @@
...
+
+
+.line{
+    width: 100%;
+    font-family: "Lucida Console", monospace;
+    padding-left: 10px;
+    padding-right: 10px;
+    border-bottom: 1px solid #000;
+    padding-bottom: -5px;
+    margin-bottom: 0px;
```
zachneill: if it's a 0 then you don't need to specify px, em, pt, etc...

app/static/css/travelForm.css
```diff
@@ -0,0 +1,38 @@
+
+html * {
+    font-size: 13px;
```
zachneill, line 3: idk if your chad supervisors care but avoid using px for font-size. it poses a problem for accessibility. when a user changes their browser default font size, rem and em values change accordingly but px values stay the same

here's a reference. https://joshcollinsworth.com/blog/never-use-px-for-font-size

app/static/js/base.js
```diff
@@ -73,6 +73,19 @@ function processPhoneSetup (editButtonId, phoneInputId, username, action) {
   }
 }
 
+function printDocument(path){
+  // KNOWN ISSUE: Firefox and Chrome load pages differently, due to this we had to add a timeout that as a hack workaround 
+  // but it still has some issues. 
+  var printProposal = window.open(path);
```
zachneill: use let or const instead of var in this function

app/templates/main/emergencyContactInfo.html
```diff
@@ -0,0 +1,85 @@
...
+    <script type="text/javascript" src="{{url_for('static', filename='js/emergencyContactInfo.js') }}?u={{lastStaticUpdate}}"></script>
+{% endblock %}
+
+{% block app_content %}
+
+<div class="text-center">
+    <h1> Emergency Contact </h1> <br>
+</div>
+
+<div id='username' hidden value='{{contactInfo.user}}'></div>
```
zachneill: if you're using this hidden tag the way i think you are, change this from a div to an input tag. if nothing else, it's better practice

app/templates/main/emergencyContactInfo.html
```diff
@@ -0,0 +1,85 @@
...
+        <div class='form-group'>
+            <label class='form-label' for='address'><strong> Home Address </strong></label>
+            <input {{'disabled' if readOnly else ''}} class='form-control' type='text' id='address' value="{{contactInfo.homeAddress if contactInfo else ''}}" name='homeAddress'/>
+        </div>
+
+    </div>
+
+    <br>
+
+    <div class='form-group' style='text-align:right'>
```
zachneill: bootstrap class="text-end"

app/templates/main/insuranceInfo.html
```diff
@@ -0,0 +1,91 @@
...
+
+
+
+{% block app_content %}
+
+<div class="text-center">
+    <h1> Insurance Information </h1> <br>
+</div>
+
+<div id='username' hidden value='{{userInsuranceInfo.user}}'></div>
```
zachneill: prob change to hidden input tag

app/templates/main/travelForm.html
```diff
@@ -0,0 +1,163 @@
...
+    <a>Insurance Company: <u>&nbsp;{{userData.insuranceCompany}}{{('&nbsp;' * (66 - (userData.insuranceCompany | length))) | safe}}</u></a><br><br>
+    <a>Policy Number: <u>&nbsp;{{userData.policyNumber}}{{('&nbsp;' * (66 - (userData.policyNumber | length))) | safe}}</u></a><br><br>
+    <a>Group Number (if applicable): <u>&nbsp;{{userData.groupNumber}}{{('&nbsp;' * (66 - (userData.groupNumber | length))) | safe}}</u></a><br><br>
+    <br>
+    <a>
+        Please list any Health Issues that the Berea College CELTS program staff need to know (use additional sheets, if needed
+        and update as appropriate):
+    </a>
+    <br>
+    <!-- {% set lineCharLength = 87 %}
```
zachneill: take this out if ur not using it. totally ruined the moment

zachneill (changes requested):
i get emails every time you git push plus another one from your github code coverage bot immediately afterwards. so ur next commits better be avengers-level threat good

app/models/emergencyContact.py
hoerstl: yeah, we need baseModel from the app models so we can inherit from it.

AndersonStettner (changes requested):
We should improve our email validation, the following screenshot was accepted as a valid email address:
image

app/static/css/travelForm.css
AndersonStettner: In this case it should be fine since this is a css file to print out or download a pdf of a travel form and the only time they will see the page is when they are printing or downloading the form.

app/templates/main/emergencyContactInfo.html
```diff
@@ -0,0 +1,85 @@
...
+
+    </div>
+
+    <br>
+
+    <div class='form-group text-end'>
+        {% if readOnly %}
+            <a href='/profile/{{username}}' class='btn btn-primary'>Exit</a>           
+        {% else %}
+            <a id='exit' href='/profile/{{username}}' class='btn btn-secondary'>Cancel</a>
```
AndersonStettner: Cancel should be left aligned.

app/templates/main/insuranceInfo.html
```diff
@@ -0,0 +1,91 @@
...
+
+    </div>
+
+    <br>
+
+    <div class='form-group text-end'>
+        {% if readOnly %}
+            <a href='/profile/{{username}}' class='btn btn-primary'>Exit</a>
+        {% else %}
+            <a id='exit' href='/profile/{{username}}' class='btn btn-secondary'>Cancel</a>
```
AndersonStettner: Cancel should be left aligned.

AndersonStettner (changes requested):
We should be more specific with the ids of the emergency contact info and insurance info. 

We might want to add an 'inputmask()` to the phone numbers on the emergency contact info page and. 

Also, did yall look into changing the logo that is used on the travel form print? If you decide to stick with the current logo that is fine but we should get a better screen grab so that we are not cutting off part of the "s" in CELTS.

PR #1039, opened 2023-08-15
Attendance Limit for RSVP

### LINKED ISSUE #1017: Attendance Limit for RSVP is allowed to be negative
On the view/edit event page when creating or editing an event, the attendance limit is allowed to be negative. The value should probably be locked to 1 at the lowest as there should never be 0 or negative for the attendance limit when rsvp is required.

stevensonmichel:
While creating a new event, the user is able to check the RSVP toggle and add an attendance limit for that event. This pull request addresses the instance where the user chooses a negative number for attendance limit.

Changes:
-add "min" attribute to the code in createEvent.html file to set a limit for attendance limit (which is zero in that case).
-Resolves issue #1017

### DIFF
--- a/app/templates/admin/createEvent.html
+++ b/app/templates/admin/createEvent.html
@@ -222,7 +222,7 @@
                     {% set hide = "" if eventData.isRsvpRequired else "display: none" %}
                     <div class="form-group" id="limitGroup" style= "{{hide}}">
                       <label class="form-label mt-2" for='limitRsvp'><strong>Attendance Limit</strong></label>
-                      <input class="form-control mb-2" type="number" id="limitRsvp" name="rsvpLimit" value="{{eventData.rsvpLimit}}" placeholder="No Limit"></input>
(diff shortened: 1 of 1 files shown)


app/templates/admin/createEvent.html
```diff
@@ -222,7 +222,7 @@
                     {% set hide = "" if eventData.isRsvpRequired else "display: none" %}
                     <div class="form-group" id="limitGroup" style= "{{hide}}">
                       <label class="form-label mt-2" for='limitRsvp'><strong>Attendance Limit</strong></label>
-                      <input class="form-control mb-2" type="number" id="limitRsvp" name="rsvpLimit" value="{{eventData.rsvpLimit}}" placeholder="No Limit"></input>
+                      <input class="form-control mb-2" type="number" id="limitRsvp" name="rsvpLimit" value="{{eventData.rsvpLimit}}" min="0" placeholder="No Limit"></input>
```
AndersonStettner: Lets also remove the ability to put in a number in scientific notation (1.02e20).

AndersonStettner (changes requested):

AndersonStettner (approved):

PR #1041, opened 2023-08-16
Add upcoming event count to student led accordion.

### LINKED ISSUE #1040: Indicate there are no upcoming events on the outside of a programs accordion.
Currently, if there are no future events for a program but there are past events we have an accordion that a user has to open up to see there is nothing to see. We should add an indicator on the outside of the accordion, to the right of the Program Name, letting users know there are no upcoming events. Possibly just "0 upcoming events." or "Currently, there are no more events planned."

AndersonStettner:
Added a counter for the upcoming events to the accordion header and removed the extra `Other CELTS` option from the create event page. 

Changed `Dietary Restrictions` text box so that it has a placeholder and if there are no dietary restrictions entered we no longer just say "None".

To Test:
Run the test suite and verify that all the tests pass. 
Run the application and create multiple events in a program, make sure some are in the past and some are canceled, verify the count that is given in the accordion header is accurate to upcoming events that are not canceled.


AndersonStettner:
[Indicate there are no upcoming events on the outside of a programs accordion. #1040](https://trello.com/c/V3PFkdWq/682-indicate-there-are-no-upcoming-events-on-the-outside-of-a-programs-accordion-1040)

BrianRamsay (approved):

PR #1042, opened 2023-08-17
Creation of log for dietary restrictions

### LINKED ISSUE #1030: Create a log for Dietary Restrictions.
We should log every time a dietary restriction is updated. We need to know who updated it, when it was updated, and what was updated.

stevensonmichel:
Prior to that PR, dietary restrictions updates were not recorded on Admin Logs. 
This PR display dietary restrictions in Admin Log, while also handle the case where the user deletes his dietary restrictions.

-Added conditional statements inside getDietInfo() function in routes.py file and create admin log if necessary.
-Resolves issue #1030


AndersonStettner:
[Create a log for Dietary Restrictions. #1030](https://trello.com/c/2iD4ioF0/675-create-a-log-for-dietary-restrictions-1030)

AndersonStettner (approved):

Issue #1020, opened 2023-08-01
Change the flash message when adding multiple volunteers to an event. 


AndersonStettner:
Currently, if you add multiple volunteers to an event at the same time through the `Add Volunteers` button, every volunteer that is added gets their own flash message and it covers the whole page.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Change the flash message when adding multiple volunteers to an event. #1020](https://trello.com/c/b5KkgAYT/668-change-the-flash-message-when-adding-multiple-volunteers-to-an-event-1020)

PR #1052, opened 2023-08-23
Change of flash message for multiple added volunteers

stevensonmichel:
If multiple volunteers were added for an event, there will be multiple flash messages whose numbers are tantamount to that of the volunteers. 
This pull request ensures that there is a single flash message, regardless of the number of volunteers who were added. Hence, the user will be able to see all the names of the volunteers in a single flash message that will displayed at the top of the screen.

- Changes volunteer.py file by using three lists that keep track of volunteers and triggering three conditional statements for this PR objective.
- Resolves issue #1020

### DIFF
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -142,16 +142,33 @@ def addVolunteer(eventId):
     usernameList = request.form.getlist("volunteer[]")
 
     successfullyAddedVolunteer = False
+    alreadyAddedList = []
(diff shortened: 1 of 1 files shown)


AndersonStettner:
[Change the flash message when adding multiple volunteers to an event. #1020](https://trello.com/c/b5KkgAYT/668-change-the-flash-message-when-adding-multiple-volunteers-to-an-event-1020)

AndersonStettner (approved):

PR #1058, opened 2023-08-29
Added Emergency contact info to Volunteer Details

### LINKED ISSUE #994: Link emergency contact to event volunteer information
When student staff look at the "Volunteer Information" we should show volunteers emergency contact info. The emergency contact info should be on the volunteer info printout at as well.

andersoncedu:
Issue: Request to add emergency contact information to volunteer details page.

Fixes: Added important emergency contact details to User model as property of the user. Emergency contact details can be added to volunteer details page from already imported users.

Tests: Add emergency contact details to users by logging in as the user. Add users to an event as volunteers check volunteer details to ensure that emergency contact name and number appear as they're supposed to on the volunteer details page.

Resolves: #994


andersoncedu:
[Link emergency contact to event volunteer information #994](https://trello.com/c/z7tKcmLl/659-link-emergency-contact-to-event-volunteer-information-994)

app/models/user.py
```diff
@@ -40,6 +41,18 @@ def isBonnerScholar(self):
     @property
     def fullName(self):
         return f"{self.firstName} {self.lastName}"
+    
+    @property
+    def getEmergencyContact(self):
```
AndersonStettner: Since this is a property the function should not be named `get...`, it should be descriptive as to what it is. Like above, its not `getFullName` but `fullName`.

app/models/user.py
```diff
@@ -40,6 +41,18 @@ def isBonnerScholar(self):
...
+
+        from app.models.emergencyContact import EmergencyContact # Must defer import until now to avoid circular reference
+        useremergencyContact = EmergencyContact.get_or_none(user=self.username)
+        if useremergencyContact:
+            emergencyName = useremergencyContact.name or None
+            emergencyNumber = useremergencyContact.cellPhone or EmergencyContact.homePhone or EmergencyContact.workPhone or None
+        else:
+            emergencyName, emergencyNumber = None, None
+        return emergencyName, emergencyNumber
```
AndersonStettner: Cache the values so we don't run the queries twice in one request.

app/models/user.py
```diff
@@ -40,6 +41,18 @@ def isBonnerScholar(self):
     @property
     def fullName(self):
         return f"{self.firstName} {self.lastName}"
+    
+    @property
+    def getEmergencyContact(self):
+
+        from app.models.emergencyContact import EmergencyContact # Must defer import until now to avoid circular reference
+        useremergencyContact = EmergencyContact.get_or_none(user=self.username)
```
AndersonStettner: To be consistent with everywhere else, change `self.username` to `self`.

AndersonStettner (changes requested):

app/models/user.py
solijonovam: we've removed property altogether

app/templates/events/volunteerDetails.html
```diff
@@ -19,23 +19,29 @@
...
-      <td class="phoneSelect" nowrap>{{ participant.phoneNumber if participant.phoneNumber else "" }}</td>
-      <td class="emailSelect" nowrap>{{ participant.email }}</td>
-      <td class="dietRestrictionSelect">{{ participant.dietRestriction if participant.dietRestriction else ''}}</td>
+    <tr class="{{volunteerType}}Select volunteerInfoEntries" data-user="{{participant.user.username}}">
+      <td class="nameSelect" nowrap><b>{{ participant.user.firstName }} {{ participant.user.lastName }}</b></td>
+      <td class="phoneSelect" nowrap>{{ participant.user.phoneNumber if participant.user.phoneNumber else "" }}</td>
+      <td class="emailSelect" nowrap>{{ participant.user.email }}</td>
+      <td class="dietRestrictionSelect">{{ participant.user.dietRestriction if participant.user.dietRestriction else ''}}</td>
+      <td class="emergencyContactSelect">{{participant.emergencycontact.name if participant.emergencycontact}} 
+        {{participant.emergencycontact.cellPhone or participant.emergencycontact.homePhone or participant.emergencycontact.workPhone if participant.emergencycontact}} </td>
```
stevensonmichel, line 44: You might need to consider handling the case in which there is no emergency contact.

andersoncedu: The case for no emergency contact is handled by leaving the table cell blank in the table view, or as "None specified." in the card view.

app/models/user.py
andersoncedu: This is no longer relevant as the queries were redesigned to be done as join so that additional queries are no longer needed.

app/controllers/admin/volunteers.py
```diff
@@ -111,16 +112,16 @@ def volunteerDetailsPage(eventID):
     if not (g.current_user.isCeltsAdmin or (g.current_user.isCeltsStudentStaff and g.current_user.isProgramManagerForEvent(event))):
         abort(403)
 
-    eventRsvpData = list(EventRsvp.select().where(EventRsvp.event==event))
-    eventParticipantData = list(EventParticipant.select().where(EventParticipant.event==event))
+    eventRsvpData = list(EventRsvp.select(EmergencyContact, EventRsvp).where(EventRsvp.event==event).join(EmergencyContact, JOIN.LEFT_OUTER, on=(EmergencyContact.user==EventRsvp.user)))
+    eventParticipantData = list(EventParticipant.select(EmergencyContact, EventParticipant).where(EventParticipant.event==event).join(EmergencyContact, JOIN.LEFT_OUTER, on=(EmergencyContact.user==EventParticipant.user)))
```
AndersonStettner: Your `.join()` should come before your `.where()`

AndersonStettner (changes requested):

app/controllers/admin/volunteers.py
andersoncedu: Changed the order of the query methods.

Karina-Agliullova:
There is an issue with entering an email address. I tried to enter random letters without providing @ and whatever goes after like @gmail.com. It asked me to enter a valid email address and did not let me save it, which is good. However, right after that message, I try to enter a valid email address with @gmail.com or @berea.edu, and it doesn't save the real email address either. It keeps telling me that the email address is not valid even when I enter my actual email address.
Screenshot (1470)

azabeli:
Overall this looks good, but when the emergency contact name is long, in a card view, the phone number gets wrapped from its middle. I would suggest fixing the wrapping when the names get longer than usual.

andersoncedu:
> There is an issue with entering an email address. I tried to enter random letters without providing @ and whatever goes after like @gmail.com. It asked me to enter a valid email address and did not let me save it, which is good. However, right after that message, I try to enter a valid email address with @gmail.com or @berea.edu, and it doesn't save the real email address either. It keeps telling me that the email address is not valid even when I enter my actual email address. Screenshot (1470)

Made into it's own issue

AndersonStettner (dismissed):

app/controllers/admin/volunteers.py
```diff
@@ -111,16 +112,16 @@ def volunteerDetailsPage(eventID):
...
+    eventParticipantUsers = [obj for obj in eventParticipantData]
+    eventNonAttendedData = [obj for obj in eventRsvpData if obj not in eventParticipantUsers]
     
     #get unique list of users for each category waitlist/notwaitlist,rsvped/attended
-    waitlistUser = list(set([obj.user for obj in participantsAndRsvp if obj.rsvpWaitlist]))
-    rsvpUser = list(set([obj.user for obj in eventRsvpData if not obj.rsvpWaitlist ]))
-    attendedUser= list(set([obj.user for obj in eventParticipantData if obj.user not in eventNonAttendedData]))
+    waitlistUser = list(set([obj for obj in participantsAndRsvp if obj.rsvpWaitlist]))
+    rsvpUser = list(set([obj for obj in eventRsvpData if not obj.rsvpWaitlist ]))
+    attendedUser= list(set([obj for obj in eventParticipantData if obj not in eventNonAttendedData]))
```
BrianRamsay: We're mixing `EventRsvp` and `EventParticipant` now instead of using all User objects. That's fine if everything still works, but this variable arrangement doesn't completely make sense anymore if they aren't all lists of users. For example, line 118 is an unnecessary list comprehension, because the two lists are the same now. Also, for line 124, `attendedUser` will always be the same as `eventParticipantData`, because there will never be `EventParticipant` objects in `eventNonAttendedData`.

BrianRamsay (changes requested):

app/controllers/admin/volunteers.py
andersoncedu: Removed the weird unused list comprehensions.

AndersonStettner (approved):

PR #1061, opened 2023-08-31
Upload Course Participants Refactor

BrianRamsay:
Some refactoring of the UI and code for the uploading of course participant.

To test the "Upload Course Participants" button from the Manage Service Learning admin page, use the files in `tests/files/` for examples of valid and invalid XLS documents.


app/controllers/admin/routes.py
```diff
@@ -28,7 +30,10 @@
 from app.logic.fileHandler import FileHandler
 from app.logic.bonner import getBonnerCohorts, makeBonnerXls, rsvpForBonnerCohort
 from app.controllers.admin import admin_bp
-from app.logic.serviceLearningCoursesData import parseUploadedFile, courseParticipantPreviewSessionCleaner
+from app.logic.manageSLFaculty import getCourseDict
+from app.logic.courseManagement import unapprovedCourses, approvedCourses
+from app.logic.utils import getRedirectTarget, setRedirectTarget
```
AndersonStettner: This is the second import from utils, these imports should probably be moved to the same line.

app/controllers/admin/routes.py
```diff
@@ -362,14 +367,49 @@ def addCourseFile():
...
+    if not g.current_user.isCeltsAdmin:
+        abort(403) 
+
+    if request.method =='POST' and "submitParticipant" in request.form:
+        saveCourseParticipantsToDatabase(session.pop('cpPreview', {}))
+        flash('Courses and participants saved successfully!', 'success')
+        return redirect(url_for('admin.manageServiceLearningCourses'))
+
+  
+    courseDict = getCourseDict()
```
AndersonStettner: I think we could rename the `getCourseDict` to be more descriptive/accurate to what the function is doing.

AndersonStettner (changes requested):

app/controllers/admin/routes.py
BrianRamsay: Yes, this was a terribly named function. It's not even close to what it is actually doing!

PR #1063, opened 2023-09-01
Student admin

### LINKED ISSUE #984: Add a new `Student Admin` role
We need a new `Student Admin` role that will have more permissions than normal student staff but less than admin. They should have the same privileges as program managers for all programs aside from Bonner. 

These people should not be the manager displayed as the "Contact" on the program cards on the landing page.

qasema:
Did: fixes #984 and added a new role called student Admin. Student admin should have the same privileges as program managers for all programs aside from Bonner. 

Test: 
- Switch to branch studentAdmin
- Reset the data to ensure the role in the database (test data).
- In the bottom left corner switch the current user to Student Admin: bledsoef
- Go through events, create, delete, and edit events. Make sure the user has access and can edit for events associated with all programs except for Bonner 
- Switch to user to Admin and go to settings and then user management. Try to add and remove a new student Admin to make sure it works as expected.
-   review code


app/controllers/admin/routes.py
```diff
@@ -40,13 +40,12 @@ def switchUser():
...
     print(f"Switching user from {g.current_user} to",request.form['newuser'])
     session['current_user'] = model_to_dict(User.get_by_id(request.form['newuser']))
-
     return redirect(request.referrer)
 
 
 @admin_bp.route('/eventTemplates')
 def templateSelect():
-    if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff:
+    if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff or g.current_user.isCeltsStudentAdmin :
```
BrianRamsay: This is what `isAdmin` is for

app/controllers/admin/routes.py
```diff
@@ -59,7 +58,7 @@ def templateSelect():
 
 @admin_bp.route('/eventTemplates/<templateid>/<programid>/create', methods=['GET','POST'])
 def createEvent(templateid, programid):
-    if not (g.current_user.isAdmin or g.current_user.isProgramManagerFor(programid)):
+    if not (g.current_user.isAdmin or g.current_user.isCeltsStudentAdmin or g.current_user.isProgramManagerFor(programid)):
```
BrianRamsay: student admins already count as program managers

app/controllers/admin/routes.py
```diff
@@ -144,7 +143,7 @@ def rsvpLogDisplay(eventId):
     eventData = model_to_dict(event, recurse=False)
     eventData['program'] = event.program
     isProgramManager = g.current_user.isProgramManagerFor(eventData['program'])
-    if g.current_user.isCeltsAdmin or (g.current_user.isCeltsStudentStaff and isProgramManager):
+    if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentAdmin or (g.current_user.isCeltsStudentStaff and isProgramManager):
```
BrianRamsay: This doesn't look like the right place for this. they aren't equivalent to admins as far as we know.

app/controllers/admin/routes.py
```diff
@@ -169,7 +168,7 @@ def eventDisplay(eventId):
         print(f"Unknown event: {eventId}")
         abort(404)
 
-    notPermitted = not (g.current_user.isCeltsAdmin or g.current_user.isProgramManagerForEvent(event))
+    notPermitted = not (g.current_user.isCeltsAdmin or g.current_user.isProgramManagerForEvent(event) or g.current_user.isCeltsStudentAdmin)
```
BrianRamsay: unnecessary, because they should count as program managers

app/controllers/admin/routes.py
```diff
@@ -268,7 +267,7 @@ def eventDisplay(eventId):
 
 @admin_bp.route('/event/<eventId>/cancel', methods=['POST'])
 def cancelRoute(eventId):
-    if g.current_user.isAdmin:
+    if g.current_user.isAdmin or g.current_user.isCeltsStudentAdmin:
```
BrianRamsay: you're already using `isAdmin`

BrianRamsay (changes requested):
Re-evaluate all of the boolean expressions you have added this to. So far all of them are wrong.

gahimbaref:
The Student Admin doesn't have access to the edit page

image

gahimbaref:
When an admin creates a bonner event, a student admin has access to it when you switch roles
image

app/templates/main/userProfile.html
```diff
@@ -492,6 +492,7 @@ <h5 class="modal-title">Add Notes</h5>
                   {% if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff %}
                     <option value='2'>Admin/Student Staff</option>
                   {% endif %}
+                  <!-- Ala, we mightt need to add something here -->
```
AndersonStettner: Ala, do we know if we need something here yet?

AndersonStettner (changes requested):
Need to create tests for new logic.

app/templates/main/userProfile.html
ojmakinde: thanks for pointing that out!

AndersonStettner (changes requested):
Student Admin Program table on user profile is broken: 
image

ojmakinde:
> Student Admin Program table on user profile is broken: image

Solved. Thanks.

app/controllers/admin/userManagement.py
```diff
@@ -40,12 +40,24 @@ def manageUsers():
...
+                flash(user.firstName + " " + user.lastName + " has been added as a CELTS Student Admin", 'success')
     elif method == "removeCeltsAdmin":
         removeCeltsAdmin(user)
         flash(user.firstName + " " + user.lastName + " is no longer a CELTS Admin ", 'success')
     elif method == "removeCeltsStudentStaff":
         removeCeltsStudentStaff(user)
         flash(user.firstName + " " + user.lastName + " is no longer a CELTS Student Staff", 'success')
+    elif method == "removeCeltsStudentAdmin":
+        removeCeltsStudentAdmin(user)
+        flash(user.firstName + " " + user.lastName + " is no longer a CELTS Student Admin", 'success')
```
AndersonStettner: All of these flash messages should end in a period. Also, it might be good to change all of the flashes into f strings and use the `fullName` user property.

app/logic/userManagement.py
```diff
@@ -20,19 +20,29 @@ def addCeltsStudentStaff(user):
...
     user = User.get_by_id(user)
     user.isCeltsStudentStaff = False
     user.save()
     createAdminLog(f'Removed {user.firstName} {user.lastName} from a CELTS student staff member.')
+    
+def removeCeltsStudentAdmin(user):
+    user = User.get_by_id(user)
+    user.isCeltsStudentAdmin = False
+    user.save()
+    createAdminLog(f'Removed {user.firstName} {user.lastName} from a CELTS student Admin member.')
```
AndersonStettner: I think it would be good to use the `fullName` property for all of these as well.

app/templates/admin/userManagement.html
```diff
@@ -17,64 +17,86 @@
...
-    </h3>
-    {% set show = "show" if not visibleAccordion or visibleAccordion == "user" %}
-    <div id="collapseOne" class="accordion-collapse collapse {{show}}" aria-labelledby="headingOne" data-bs-parent="#adminManagement">
-      <div class="accordion-body">
-        <div class="container-fluid col-10">
-          <div class="row d-flex justify-content-center mb-3">
-            <div class="col-md-6 mb-3">
-              <div>
-                {{createInputsButtons("searchCeltsAdminInput", "Add Celts Admin")}}<br>
+  {% if g.current_user.isCeltsAdmin and not g.current_user.isCeltsStudentAdmin %}
```
AndersonStettner: Why are you excluding `isCeltsStudentAdmin`? It should be fine to just have the if `isCeltsAdmin` check right?

AndersonStettner (changes requested):
Why are you specifically excluding `isCeltsStudentAdmin` all over the place in checks?

app/templates/admin/userManagement.html
ojmakinde: We excluded the student admins here, as we felt they shouldn't possess the ability to add/remove admins or other student admins. Do you want us to provide that access?

BrianRamsay: I think that's an ok assumption, that the student admins shouldn't be able to mess with that.

app/controllers/admin/routes.py
```diff
@@ -148,7 +147,7 @@ def rsvpLogDisplay(eventId):
     eventData = model_to_dict(event, recurse=False)
     eventData['program'] = event.program
     isProgramManager = g.current_user.isProgramManagerFor(eventData['program'])
-    if g.current_user.isCeltsAdmin or (g.current_user.isCeltsStudentStaff and isProgramManager):
+    if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentAdmin or (g.current_user.isCeltsStudentStaff and isProgramManager):
```
BrianRamsay: You don't need the `isCeltsStudentStaff` check here, because all we care about is whether they are a program manager (and that will likely mean student staff is redundant)

app/controllers/admin/volunteers.py
```diff
@@ -55,7 +55,7 @@ def manageVolunteersPage(eventID):
 
     isProgramManager = g.current_user.isProgramManagerForEvent(event)
     bannedUsers = [row.user for row in getBannedUsers(event.program)]
-    if not (g.current_user.isCeltsAdmin or (g.current_user.isCeltsStudentStaff and isProgramManager)):
+    if not (g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentAdmin or (g.current_user.isCeltsStudentStaff and isProgramManager)):
```
BrianRamsay: Same as above. the `isCeltsStudentStaff` check is redundant with `isProgramManager`

app/controllers/admin/volunteers.py
```diff
@@ -109,7 +109,7 @@ def volunteerDetailsPage(eventID):
         print(f"No event found for {eventID}", e)
         abort(404)
 
-    if not (g.current_user.isCeltsAdmin or (g.current_user.isCeltsStudentStaff and g.current_user.isProgramManagerForEvent(event))):
+    if not (g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentAdmin or (g.current_user.isCeltsStudentStaff and g.current_user.isProgramManagerForEvent(event))):
```
BrianRamsay: And again here

app/templates/eventNav.html
```diff
@@ -33,7 +33,7 @@
     {% endblock %}
 
     {% block navbar %}
-        {% if g.current_user.isCeltsAdmin or (g.current_user.isCeltsStudentStaff and g.current_user.isProgramManagerFor(eventData['program'])) %}
+        {% if g.current_user.isCeltsAdmin or (g.current_user.isCeltsStudentAdmin and eventData["program"].id != 5) or (g.current_user.isCeltsStudentStaff and g.current_user.isProgramManagerFor(eventData['program'])) %}
```
BrianRamsay: we have a flag you can use here instead of id.

app/templates/sidebar.html
```diff
@@ -38,14 +38,14 @@
         </a>
       </li>
       <li>
-        {% if not g.current_user.isCeltsStudentStaff %}
+        {% if not g.current_user.isCeltsStudentStaff and not g.current_user.isCeltsStudentAdmin %}
```
BrianRamsay: This is equivalent to `if g.current_user.isCeltsAdmin`

app/templates/sidebar.html
```diff
@@ -56,18 +56,23 @@
         </nav>
         <div class="{{ 'collapse-show' if request.path in ['/adminLogs','/manageServiceLearning','/admin','/courseManagement','/manageBonner'] else 'collapse'}}" id="navbarToggleExternalContent">
           <div class="bg-dark ps-3">
-            <a href="/adminLogs" class="nav-link text-white {{ 'active' if request.path =='/adminLogs' }}">
-              Activity Logs
-            </a>
+            {% if g.current_user.isCeltsAdmin and not g.current_user.isCeltsStudentAdmin %}
```
BrianRamsay: those two things are never the same. So the first expression is the only one that matters.

app/templates/sidebar.html
```diff
@@ -56,18 +56,23 @@
...
             <a href="/admin" class="nav-link text-white {{'active' if request.path =='/admin'}}" {{"aria-current='page'" if request.path =='/admin'}}>
               Settings
             </a>
-            <a href="/manageServiceLearning" class="nav-link text-white {{ 'active' if request.path =='/manageServiceLearning' }}">
-              Course Management
-            </a>
-            <a href="/manageBonner" class="nav-link text-white {{ 'active' if request.path =='/manageBonner' }}">
-              Bonner Management
-            </a>
+            {% if g.current_user.isCeltsAdmin and not g.current_user.isCeltsStudentAdmin %}
```
BrianRamsay: Same as above. The first bit is the only one that matters for this expression.

BrianRamsay (changes requested):

stevensonmichel:
Please, resolve this merge conflict.

qasema:
> Please, resolve this merge conflict.

resolved @stevensonmichel

Karina-Agliullova:
Finn is set as a student admin by default, even though in the database, he exists as faculty. Faculty cannot be student admin. It should not be possible for faculty to be student admin. This should be fixed and the actual student from the database should be set as a student admin by default, or if you are going to use Finn, make him a student and not faculty so that he can be set as a student admin.

azabeli:
Other than what Karina said, everything looks and works as described in the PR comment!

app/models/user.py
```diff
@@ -28,7 +29,7 @@ def __init__(self,*args, **kwargs):
         
     @property
     def isAdmin(self):
-        return (self.isCeltsAdmin or self.isCeltsStudentStaff)
+        return (self.isCeltsAdmin or self.isCeltsStudentAdmin or self.isCeltsStudentStaff)
```
BrianRamsay, line 32: this is just `self.isAdmin`

app/templates/admin/userManagement.html
BrianRamsay: Everyone with `isCeltsAdmin` true will have `isCeltsStudentAdmin` false, right? So Anderson was right after all, that the second part is redundant.

BrianRamsay (changes requested):

gahimbaref:
image
double-check if search loads up students properly when adding a new student admin, it doesn't always load student names

app/logic/userManagement.py
```diff
@@ -59,14 +68,16 @@ def getAllowedPrograms(currentUser):
...
+        return Program.select().where(Program.id !=5).order_by(Program.programName)
     else:
         return Program.select().join(ProgramManager).where(ProgramManager.user==currentUser).order_by(Program.programName)
 
 
 
 def getAllowedTemplates(currentUser):
     """Returns a list of all visible templates depending on who the current user is. If they are not an admin it should always be none."""
-    if currentUser.isCeltsAdmin:
+    if currentUser.isCeltsAdmin or currentUser.isCeltsStudentAdmin:
```
bledsoef, line 95: Should this allow them to see Bonner events?

tests/code/test_userManagement.py
```diff
@@ -78,6 +78,26 @@ def test_modifyCeltsStudentStaff():
         transaction.rollback()
 
 
+@pytest.mark.integration
+def test_modifyCeltsStudentAdmin():
```
bledsoef, line 223: Need to change this to test what happens when you try it on a non-student

bledsoef:
Placing on hold because we need to reevaluate what permissions student admins should be able to have.

bledsoef:
#1181 creating spreadsheet here

bledsoef:
Spreadsheet has been created, still waiting to hear back from @BrianRamsay on whether not the permissions are correct so we can finalize this PR.

PR #1064, opened 2023-09-01
View past event toggle

azabeli:
Task: For the past events the view past event toggle needs to be disabled and checked, for both present and future terms, the toggle needs to be unchecked and enabled.

Fixes: We checked if selectedTerm is future or current and if it is then we enabled the toggle but unchecked; if selectedTerm is past then we disabled the toggle but checked, meaning that the past events should be displayed.

Test: To test this, you select the past term in the Events Lists then the toggle should be checked and disabled. If you select the current term or future then the toggle should be enabled but unchecked.

Resolved Issue# 1047


AndersonStettner (changes requested):
This branch has been made off of `email-ashley-slc-submitted`, you should make a new, fresh, branch off of development and cherry pick they changes relevant to your issue onto it.

Issue #1047, opened 2023-08-22
Fix `View Past Events` toggle


AndersonStettner:
Currently, we have a check at the bottom of `event_list.html` to which sets the past events toggle, we should change the check to compare against the new `termOrder` instead of `g.current_term`.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix View Past Events toggle #1047](https://trello.com/c/rCkgCuWw/688-fix-view-past-events-toggle-1047)

Mercy-Eze:
Issue resolved on #1069

PR #1069, opened 2023-09-06
Fixed the toggle issue for viewing past events

Karina-Agliullova:
Task: For the past events the view past event toggle needs to be disabled and checked, for both present and future terms, the toggle needs to be unchecked and enabled.

Fixes: We checked if selectedTerm is future or current and if it is then we enabled the toggle but unchecked; if selectedTerm is past then we disabled the toggle but checked, meaning that the past events should be displayed.

Test: To test this, you select the past term in the Events Lists then the toggle should be checked and disabled. If you select the current term or future then the toggle should be enabled but unchecked.

Resolved Issue #1047


doucoureb:
pr-review
Is also the button supposed to be checked and disabled even when there are no past events but it is a past term?

andersoncedu:
> pr-review Is also the button supposed to be checked and disabled even when there are no past events but it is a past term?


Has this been addressed?

Karina-Agliullova:
> pr-review Is also the button supposed to be checked and disabled even when there are no past events but it is a past term?

- I checked it again. It works correctly because Summer 2021 is set to the current term by default, which is why it shows the behavior of the current term. Fall 2021 is the future term because it goes after Summer 2021, so the behavior would be the same as the current term. So the toggle should be enabled but unchecked. For any other term before Summer 2021, the behavior would be for past terms, which is checked and disabled toggle.

azabeli:
> > pr-review Is also the button supposed to be checked and disabled even when there are no past events but it is a past term?
> 
> Has this been addressed?

Yes, you can refer to Karina's reply to the comment.

AndersonStettner (approved):

Issue #1019, opened 2023-08-01
Make Remove Button functional on New Proposal For Service-Learning Course Page


azabeli:
On the New Proposal For Service-Learning Course Page when a course instructor is added to a new course Proposal, they can't be removed because the "Remove" button isn't functional. This needs to be changed and make the remove button functional.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Make Remove Button functional on New Proposal For Service-Learning Course Page #1019](https://trello.com/c/UBI3Y7aS/670-make-remove-button-functional-on-new-proposal-for-service-learning-course-page-1019)

PR #1072, opened 2023-09-11
Fixed the remove button

Karina-Agliullova:
Issue: When creating a new course proposal, you need to select course instructors. Once you select a course instructor or instructors, their name appears and the button "Remove" on the right of their name appears. The button doesn't do anything.

Fix: Fixed the functionality of the button. Now it removes the instructor that was added. 

Test: In the Course Proposal, create a new proposal, and on the second page of the process of creating the proposal, add any instructor(s). Once the instructor's name and button "Remove" appear, click the button and see if the instructor will be removed from the list.

Issue #1019


app/static/js/slcNewProposal.js
```diff
@@ -121,7 +121,7 @@ $(document).ready(function(e) {
   // -----------------------------------------
       $("#instructorTable").on("click", "#remove", function() {
         let closestRow =  $(this).closest("tr")
-        $("#instructorTableNames input[value="+closestRow.data('username')+"]").remove()
+        let username = closestRow.data('username');
```
ojmakinde: The variable "username" is not used after its instantiation

Karina-Agliullova: Fixed it!

AndersonStettner (changes requested):
This works to fix the case mentioned in the issue but it breaks the remove button for courses that have already been created.

AndersonStettner (changes requested):
There is some id mismatch happening somewhere. If you have multiple instructors and you delete one, there are cases where the one you tried to delete is still in the instructor list and one you were not trying to delete is deleted. 

EX: 
- Course has just Scott as instructor.
- Add Brian and Finn as instructors. 
- Delete the instructor you added last.
- Save and Exit
- The instructor you tried to delete will still be an instructor with Scott. 

Also, if you are editing a proposal and you are changing the instructor, if you delete the instructor and add a new one and then also delete the newly added instructor you will be allowed to save the proposal without an instructor because the second instructor you deleted was not actually deleted it seems.

AndersonStettner (dismissed):
I can still remove an instructor after adding them and save the form with no instructor but the instructor that I removed is saved.

stevensonmichel:
I created a new Course Proposal, add Scott Heggen as instructor, then added Brian Ramsey. After, I remove Scott Heggen as instructor and added him back again. It appears that after submitting the proposal, Scott's name appears twice both in the Proposal Page and in the courseinstructor table in the database. It might be wise to look at how the remove process is executed.
Screenshot 2023-11-15 115457

doucoureb:
Capture3
Is it necessary for the list to add the name of the instructor and the creator twice if that's the same person? 
Also, I removed Scott from the list and he is still showing.

app/static/js/instructorTable.js
```diff
@@ -0,0 +1,66 @@
...
+$(document).ready(function() {
+    $("#instructorTable").on("click", ".removeButton", function() {
+        let closestRow =  $(this).closest("tr");
+        let username = closestRow.data('username');
+
+        $("#instructorTableNames input[value='" + username + "']").remove();
+        closestRow.remove();
+    });
+
+
```
azabeli: remove the unnecessary(empty) lines

app/static/js/slcManagement.js
```diff
@@ -140,3 +140,54 @@ function unapproveProposal(el){
...
+  // Add each instructor to the table
+  instructors.forEach(function(instructor) {
+    var newRow = createInstructorRow(instructor);
+    $("#instructorTableBody").append(newRow);
+  });
+}
+
+function createInstructorRow(instructor) {
+  // Create a new row element based on the instructor data
+  var row = `<tr data-username="${instructor.username}">
```
azabeli: I think the variable name 'row' should be more specific to what the row contains. It looks to be containing instructors' information, so maybe something that specifies that.

app/static/js/slcNewProposal.js
```diff
@@ -118,25 +118,39 @@ $(document).ready(function(e) {
...
+
+    $("#courseInstructor").on('input', function() {
+        searchUser("courseInstructor", createNewRow, true, null, "instructor");
+    });
+
+    // for each row in instructorTable that has an instructor, pass that instructors phone data to setupPhoneNumber
+    $('#instructorTable tr').each(function(){
+      var username = getRowUsername(this)
+      var edit = "#editButton-" + username
+      var input = "#inputPhoneNumber-" + username
```
azabeli, line 140: Make a more specific variable name; I think 'input' is not  a so specific variable. Same thing for 'edit' variable, both needs to be more specific to what they are inputting or editing.

azabeli (dismissed):

azabeli:
image


I created the new course proposal and added, Bryan then removed him, and then added Scott and Finn respectively. After saving, Bryan's name still appeared on the course proposal table as shown. see the attached.

bledsoef (approved):

bledsoef:
[Make Remove Button functional on New Proposal For Service-Learning Course Page #1019](https://trello.com/c/UBI3Y7aS/670-make-remove-button-functional-on-new-proposal-for-service-learning-course-page-1019)

Issue #1016, opened 2023-08-01
Fix the search button on `Student Search`


azabeli:
The magnifying glass of the Search button doesn't work(it is not clickable) and nothing pops up until the user hits enter. It should be clickable and pop relevant information to what the user is searching for, without necessarily waiting for the user to hit enter.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Refactor the code for Student Search Page #1016](https://trello.com/c/bAVjVtIW/672-refactor-the-code-for-student-search-page-1016)

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix the search button on Student Search #1016](https://trello.com/c/bAVjVtIW/672-fix-the-search-button-on-student-search-1016)

PR #1074, opened 2023-09-13
fixed the search button functionality

azabeli:
Issue: The search Icon(magnifying glass) on the student search Page was not functional(not clickable). It only worked after clicking the enter key on the keyboard. 

Task: We had to make the icon functional and work without exclusively relying on the enter key on the keyboard.  

solution: We solved this by creating an id in HTML for the search icon which we used in a javascript file to make its functionality work. "

Test: You can test this by entering a  full name in the search bar and instead of hitting enter click on the search Icon on your right. It should flash a message if the name is not in the database or show student information if the name is in the database.   

Solve issue #1016

### DIFF
--- a/app/static/js/searchStudent.js
+++ b/app/static/js/searchStudent.js
@@ -6,3 +6,9 @@ function callback(selected) {
 $("#searchStudentsInput").on("input", function() {
   searchUser("searchStudentsInput", callback);
 });
+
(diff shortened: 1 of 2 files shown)


qasema:
it looks good

AndersonStettner (approved):

Issue #1033, opened 2023-08-10
Update event cover image


AndersonStettner:
We need to force all displayed images on events to be square and we should decide which file should be displayed based on a set name and filetype. e.g., only event-cover.jpg or event-cover.png

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Modify the dimensions of an image that is displayed on an event and force it to always be square. #1033](https://trello.com/c/6qtDiVZo/678-modify-the-dimensions-of-an-image-that-is-displayed-on-an-event-and-force-it-to-always-be-square-1033)

PR #1082, opened 2023-09-25
Event images

qasema:
issue: we are trying to allow the admin to decide what image of the event attachments to be a cover for the event. (fixes issue #1033 )

Did: we add checkboxes next to the image so that the user can only select one and that image will be displayed as a cover for the event. other attachments such as pdf should be disabled when it comes to selection 

Test: 
- switch to branch "event_images"
- reset the database and run test cases to make sure they all run 
- create an event and add attachments to it (add images) then go to edit event and select one of them to display. View the event and see if the cover is updated 
- review the code


qasema:
do not review yet We need to add test cases

andersoncedu:
image
Consider changing these to radio buttons, checkboxes usually imply that you can make more than one choice.

app/static/js/displayFilesMacro.js
```diff
@@ -18,4 +18,38 @@ $(document).ready(function(){
...
+    $(document).ready(function() {
+        $('.attachmentCheck').change(function() {
+            // Uncheck all checkboxes
+            $('.attachmentCheck').prop('checked', false);
+            // Check the selected checkbox
+            $(this).prop('checked', true);
+    
+            var attachmentId = $(this).data('id');
+            var isChecked = $(this).is(':checked');
+            console.log(isChecked);
```
andersoncedu: Is this still necessary?

app/logic/fileHandler.py
```diff
@@ -79,6 +79,7 @@ def saveFiles(self, saveOriginalFile=None):
     def retrievePath(self,files):
         pathDict={}
         for file in files:
+            
             pathDict[file.fileName] = ((self.path+"/"+ file.fileName)[3:], file)
```
andersoncedu: Please space evenly before and after the additions in your string concatenation.

app/templates/displayFilesMacro.html
```diff
@@ -1,21 +1,36 @@
 {% macro displayFiles(filepaths, titleName, deleteLink, databaseId)  %}
-<label class="mb-2"><strong>{{titleName}}</strong></label>
 <table>
+    <tr>
+        <td class="py-2" ><strong>Display</strong></td>
+        <td class="py-2 px-3"><strong>{{ titleName }}</strong></td>
```
andersoncedu, line 5: This might be nitpicking, but add the same amount of space after quotation marks '"' and before the greater than symbol '>'.

app/templates/displayFilesMacro.html
```diff
@@ -1,21 +1,36 @@
...
-            {% set filename = key[:8]+ "..." +key[-10:] if key| length > 25 else key  %}
-                <a class="mr-5 fileName" data-filename='{{key}}' href="{{value[0]}}" target="_blank">{{filename}}</a>
+        <tr id="attachment_{{ value[1] }}" class="attachmentRow_{{ key }}" data-id="{{ key }}">
+            <td class="py-2 px-3">
+                {% if value[-1].isDisplayed %}
+                    <input type="checkbox" class="attachmentCheck" data-id="{{ value[1] }}" checked>
+                
+                {% else %}
+                {% if not (value[0].endswith('.png') or value[0].endswith('.jpg') or value[0].endswith('.svg') or value[0].endswith('.jpeg') or value[0].endswith('.gif')) %}
+                        <input type="checkbox" class="attachmentCheck" data-id="{{ value[1] }}" disabled >
```
andersoncedu: Instead of adding a disabled checkbox, maybe a "-" or just omitting the checkbox altogether would be a better option.

qasema:
> image Consider changing these to radio buttons, checkboxes usually imply that you can make more than one choice.

That is correct, but I discussed this with Scott and Anderson. Both suggested checkboxes!

app/templates/displayFilesMacro.html
qasema: I think we should keep them for consistency. what do you think?

AndersonStettner:
> > image Consider changing these to radio buttons, checkboxes usually imply that you can make more than one choice.
> 
> That is correct, but I discussed this with Scott and Anderson. Both suggested checkboxes!

I don't remember saying you had to do checkboxes, it was just an option that was brought up. I do think radio could work and look better.

app/templates/displayFilesMacro.html
AndersonStettner: I think it makes sense to not have a selector or, as Chris suggests, something like a "-", for options that are not able to be selected. 

If they do end up looking different then we should put all similar attachments together. This means all images should be together and all non images should be together in the list, we should should not have a list that goes back and forth between the two.

app/static/js/displayFilesMacro.js
```diff
@@ -18,4 +18,37 @@ $(document).ready(function(){
...
+
+            $.ajax({
+                url: '/displayEventFile',
+                method: 'POST',
+                data: {
+                    id: attachmentId,
+                    checked: isChecked
+                },
+                success: function(response) {
+                    msgFlash("Attachment is updated successfully", "success")
```
AndersonStettner: Since we are doing the flash in javascript we have the option to use toasts. I think they might work better in this case, what do you think?

app/logic/fileHandler.py
```diff
@@ -94,3 +94,9 @@ def deleteFile(self, fileId):
         if not AttachmentUpload.select().where(AttachmentUpload.fileName == file.fileName).exists():
             path = os.path.join(self.path, file.fileName)
             os.remove(path)
+    
+    def displayFile(self, fileId):
+        file = AttachmentUpload.get_by_id(fileId)
+        AttachmentUpload.update(isDisplayed=True).where(AttachmentUpload.id == fileId).execute()
+        AttachmentUpload.update(isDisplayed=False).where(AttachmentUpload.event == file.event, AttachmentUpload.id != fileId).execute()
+        return 0
```
AndersonStettner: Everywhere else we don't really need to return anything we use empty string `""`. To be consistent it would be good to change the `0` to `""`

app/logic/fileHandler.py
```diff
@@ -94,3 +94,9 @@ def deleteFile(self, fileId):
         if not AttachmentUpload.select().where(AttachmentUpload.fileName == file.fileName).exists():
             path = os.path.join(self.path, file.fileName)
             os.remove(path)
+    
+    def displayFile(self, fileId):
+        file = AttachmentUpload.get_by_id(fileId)
+        AttachmentUpload.update(isDisplayed=True).where(AttachmentUpload.id == fileId).execute()
+        AttachmentUpload.update(isDisplayed=False).where(AttachmentUpload.event == file.event, AttachmentUpload.id != fileId).execute()
```
AndersonStettner: While it might not matter much in an application like ours, this is not very good practice. There is no reason to update every file in an event when, at most, we will  be unselecting 1 and selecting 1. So currently, instead of modifying the 2 files that are being changed, we are modifying all of them.

app/logic/fileHandler.py
```diff
@@ -94,3 +94,9 @@ def deleteFile(self, fileId):
         if not AttachmentUpload.select().where(AttachmentUpload.fileName == file.fileName).exists():
             path = os.path.join(self.path, file.fileName)
             os.remove(path)
+    
+    def displayFile(self, fileId):
```
AndersonStettner: This is a rather small nitpick but it might be good to modify the function name. 

Sure, we call this function when selecting a file to be displayed but, the function also removes a file from display in the case where one is already being displayed and we change to a different file.

tests/code/test_fileHandler.py
```diff
@@ -150,5 +150,30 @@ def test_deleteFile():
...
+
+        eventfile = FileHandler(eventId=1)
+
+        # display image1 as an event cover 
+        eventfile.displayFile(image1)
+
+        assert AttachmentUpload.get_by_id(image1).isDisplayed ==True
+        assert AttachmentUpload.get_by_id(image2).isDisplayed ==False
+
+        # display image1 as an event cover
```
AndersonStettner: If you are going to have comments, please make sure they are accurate.

AndersonStettner (changes requested):

AndersonStettner (approved):

PR #1087, opened 2023-10-12
Countdown on event view

### LINKED ISSUE #998: Show relative dates for events
We should add a relative date to the Event View. Something like: 
"Event is today at __" 
"Event is in _ days" 
"Event is this Saturday"

doucoureb:
Issue: We are trying to add a relative date to the Event View, just like a countdown before the event starts. (fixes issue #998)

Did: We calculated and displayed a countdown to the specified event, considering both the date and time of the event. Depending on the remaining time, it shows different messages indicating how soon the event is happening.

Test:

- Switch to branch "countdownOnEventView"
- reset the database to test
- Select a term on the drop-down of the event list (I usually just pick fall 2021) or you can just create an event
-  View the event and check if the countdown is accurate 
- Review the code


ojmakinde:
For events that start in more than 24 hours but less than 48, there's an off-by-one error. An event created for the 25th says "tomorrow" on the 23rd, because its start time is less than 48 hours away. I think using the (in x hours and y minutes) format for anything less than 24 hours is great, but anything in more than 24 hours should be at least 2 days. Additionally, Brian also suggested using a datetime library to avoid dealing with the complications associated with date and time.

qasema (changes requested):
This could be a nitpick, we should display "hours" only if it is greater than 0 

Screenshot 2023-10-26 at 3 26 08 PM

qasema (changes requested):
create some test cases for the function eventCountDown

qasema (dismissed):
Looks good

AndersonStettner (dismissed):
"Stop trying to do this yourself, this is a solved problem, use a library" - Brian

Karina-Agliullova:
The test case for test_events.py fails when running a test suite

qasema (changes requested):
when setting a date for the event more than 30 days away, the countdown starts counting from zero. it does not go beyond 30 days

qasema (dismissed):
This might be a small change. when events start time passes, It starts counting minutes in negative values to indicate time passed since the start

andersoncedu:
You have test cases failing.

app/logic/events.py
```diff
@@ -501,3 +502,77 @@ def getEventRsvpCount(eventId):
...
+    # Hours until
+    if timeUntilEvent.hours:
+        if timeUntilEvent.minutes:
+            return f"{hourString} and {minuteString}"
+        return f"{hourString}"
+    # Minutes until
+    elif timeUntilEvent.minutes:
+        return f"{minuteString}"
+    # Seconds until
+    return "<1 minute"
```
Karina-Agliullova, line 581: The case with <1 minute is not working. I was trying to set an event that is happening in less than a minute and I only got "(1 minute)" that was displayed.

hoerstl: Thanks! Good catch!

app/logic/events.py
```diff
@@ -2,6 +2,7 @@
 from peewee import DoesNotExist, fn, JOIN
 from dateutil import parser
 from datetime import timedelta, date
+from dateutil.relativedelta import relativedelta
 import datetime
```
bledsoef: I know this isn't all you guys but since you use this it would potentially be good if removed this import and added `datetime` to the `from datetime` import statement. That way we could make some of the `datetime.datetime` calls in your function and other places be cleaner and we could remove this import.

app/logic/events.py
```diff
@@ -507,6 +508,79 @@ def getEventRsvpCount(eventId):
     """
     return len(EventRsvp.select().where(EventRsvp.event_id == eventId))
 
+def getCountdownToEvent(event, *, currentDatetime=None):
+    """
+    Given an event, this function returns a string that conveys the amount of time left
+    until the start of the event.
+
+    Note about dates:
+    The way dates are communicated is unintuitive. There are two major rules that govern how we discuss dates.
```
bledsoef: I think we could remove this line from the docstring. I think it can be a little confusing what you are talking about when you say "the way dates are communicated is unintuitive", plus the docstring flows just fine without it.

app/logic/events.py
```diff
@@ -507,6 +508,79 @@ def getEventRsvpCount(eventId):
...
+    - If an event happens tomorrow but less than 24 hours away from us we still say that it happens 
+    tomorrow with no mention of the hour. 
+    - If an event happens tomorrow but more than 24 hours away from us, we'll count the number of days 
+    and hours in actual time.
+
+    E.x. if the current time of day is greater than the event start's time of day, we give a number of days
```
bledsoef: Since I have already left other comments I will add this nitpicky thing, I believe it is E.g. not E.x. if you are meaning to abbreviate "for example".

bledsoef (dismissed):
I think everything looks good and functions well. I added a few changes that I would make but I am approving this since it would be fine regardless. Feel free to make the changes or merge it yourself if you disagree.

app/logic/events.py
hoerstl: I saw this and thought about it. Thank you for bringing it up so I can fix it!

app/logic/events.py
hoerstl: Fixed!

app/logic/events.py
hoerstl: Finn, I see that leaving comments that aren't absolutely 100% correct isn't in your skillset. Good catch

hoerstl:
[Countdown on Event View](https://trello.com/c/CruRR9HM/753-countdown-on-event-view)

gahimbaref (dismissed):
looks good.

Karina-Agliullova (dismissed):
Good job!

AndersonStettner (dismissed):
Please resolve merge conflicts.

AndersonStettner (approved):
Please resolve merge conflicts.

PR #1089, opened 2023-10-12
FLASK_ENV Changes.

### LINKED ISSUE #1079: Find FLASK_ENV alternative
In the current version of Flask, FLASK_ENV has been deprecated and, in the newer versions, it has been completely removed. To keep our Flask up to date we need to remove FLASK_ENV and figure out a new way to do the environment switching that we currently do with FLASK_ENV.

ojmakinde:
Issue: As FLASK_ENV has been deprecated in later versions, we were tasked with getting rid of the environment variable, and replacing its functionality with up-to-date methods in Flask.

Solution: Using FLASK_DEBUG, an environment variable which controls the debugger and reloader, we replaced FLASK_ENV. As our production configuration does not require a source_setup, this approach solves the problem directly. Additionally, we added the APP_ENV environment variable to load the development configuration file.

To test: Enter a shell session, using source setup.sh, and use the flask run command. You will no longer see the message about FLASK_ENV's deprecation. Additionally, as configuration files were changed, existing test scripts were modified to account for the new changes.

Fixes #1079.


app/__init__.py
```diff
@@ -8,9 +8,11 @@
 
 # Initialize our application
 app = Flask(__name__, template_folder="templates")
+# debug = os.getenv('FLASK_DEBUG', False)
+env = os.environ.get('APP_ENV', 'development')
```
BrianRamsay: This needs to default to production, or else it won't work in production where we don't use the environment variables.

app/__init__.py
```diff
@@ -8,9 +8,11 @@
...
 
-# Set the correct configuration according to the environment
-load_config_files(app, get_env())
+# app.config['APP_ENV'] = env
```
BrianRamsay: Remove commented code

BrianRamsay (changes requested):

app/__init__.py
ojmakinde: Great catch. Thanks.

AndersonStettner (dismissed):

app/__init__.py
```diff
@@ -1,20 +1,25 @@
...
+load_config_files(app, env)
 
 # set the secret key after configuration is set up
 app.secret_key = app.config['secret_key']
 
+# Overwrite the get_env() method to return our environment, stored in APP_ENV
+def new_env():
+    return os.environ.get('APP_ENV', 'production')
+helpers.get_env = new_env
+
```
BrianRamsay: Do this earlier and then you don't need to change line 13 at all.

BrianRamsay (changes requested):

solijonovam (changes requested):
getting a name error on get_env()

qasema:
> error

fixed

andersoncedu:
When I try to access the landing page in your branch:
image

andersoncedu:
There are unaddressed comments from Brian.

andersoncedu:
It also seems like FLASK_ENV still exists in the repo, I'm not sure if this is intended.
image

PR #1091, opened 2023-10-13
Renew/Reschedule Canceled Events Issue 986

### LINKED ISSUE #986: Add a Renew/Reschedule button to Canceled Events
Add Renew or Reschedule button to Canceled events to start a new event creation with the same data as the previous and with a link between the events. Extend the previous volunteer functionality to get RSVPs from the canceled event into the track volunteer tab of the new one.

andersoncedu:
Issue: 
There was a request to have a button to be able to renew/reschedule a canceled event.

Fix:
Added modal which allows supervisors to reschedule previously canceled events. We take data from the page of the canceled event and send that data to a route which uses the current event's information to copy the event contents over.

Test:
1. Reset database
2. Create a new event from the create event option in the sidebar
3. Add RSVP and waitlister students from the track volunteers page
4. cancel the event from the event view page
5. Try different dates to see if the new event reflects changes and ensure rsvps were properly copied to new event

Fixes #986


qasema (changes requested):
I think you need to validate the start and end times of the event. It should not allow the start time to be after the end time

qasema (dismissed):
when renewing an event I get this error!
Screenshot 2023-10-26 at 7 44 18 PM

 could you double-check that?

app/controllers/admin/routes.py
```diff
@@ -157,6 +159,76 @@ def rsvpLogDisplay(eventId):
...
         abort(403)
 
+@admin_bp.route('/event/<eventId>/renew', methods=['POST'])
+def renewEvent(eventId):
+    try: 
+        formData=request.form
+        try:
+            assert formData['timeStart'] < formData['timeEnd']
+        except AssertionError:
+            flash("End time must be after start time", 'warning')
```
gahimbaref, line 164: When the start time is after the event end time or end time before the start time, it might be better to make the error message `danger` (I believe) instead of `warning` to make it more clear right away that the event was not saved since the modal gets closed

app/controllers/admin/routes.py
```diff
@@ -157,6 +159,76 @@ def rsvpLogDisplay(eventId):
...
+                assert formData['dateStart'] < formData['dateEnd']
+        except AssertionError:
+            flash("End date must be after start date", 'warning')
+            return redirect(url_for('admin.eventDisplay', eventId = eventId))
+
+
+        priorEvent = Event.get_by_id(eventId)
+        rsvpInfo = list(EventRsvp.select().where(EventRsvp.event == priorEvent).execute())
+        participantInfo = list(EventParticipant.select().where(EventParticipant.event == priorEvent).execute())
+        newEvent = Event(
```
AndersonStettner: We already have a function that creates new events and what you have above is basically a repeat. Is there a reason you are doing the new event creation this way instead of modifying the existing function?

AndersonStettner (changes requested):

app/controllers/admin/routes.py
andersoncedu: The reason this was made as a warning alert instead of a danger alert is that on the create event page, when the start time is after the end time, they also use a warning alert. We just copied the format for the modal.

stevensonmichel:
From my consultation with our supervisors, the Renew Button should be functional for only Canceled Events. 
However, it can be triggered for Past Events. One idea would be to disable that button is the event is not a Canceled Event.

app/controllers/admin/routes.py
andersoncedu: Changed route to use the existing function.

ojmakinde:
It appears the current logic does not handle recurring events. I created, canceled and renewed multiple recurring events, and I kept getting the error below.
Screenshot 2023-11-27 163019

gahimbaref (dismissed):
image

A student shouldn't be able to rsvp for an event when it has been cancelled.

gahimbaref:
This might be fine but an admin can still add volunteers on a cancelled event.

andersoncedu:
> image
> 
> A student shouldn't be able to rsvp for an event when it has been cancelled.

This comment has been addressed.

app/logic/events.py
```diff
@@ -350,24 +353,27 @@ def validateNewEventData(data):
...
     if data['timeEnd'] <= data['timeStart']:
         return (False, "Event end time must be after start time.")
-
+    
     # Validation if we are inserting a new event
     if 'id' not in data:
 
-        event = (Event.select()
+        sameEventList = list((Event.select()
                       .where((Event.name == data['name']) &
```
AndersonStettner: Please fix formatting of query.

app/logic/events.py
```diff
@@ -350,24 +353,27 @@ def validateNewEventData(data):
...
                              (Event.location == data['location']) &
                              (Event.startDate == data['startDate']) &
-                             (Event.timeStart == data['timeStart'])))
+                             (Event.timeStart == data['timeStart'])).execute()))
+        
+        for item in sameEventList:
+            if item.isCanceled or item.recurringId:
+                sameEventList.remove(item)
```
AndersonStettner: It is really nitpicky but I think it  would read a lot better to use `event` instead of `item` for the `for` loop variable.

app/logic/events.py
```diff
@@ -501,3 +507,18 @@ def getEventRsvpCount(eventId):
...
+        newRsvp = EventRsvp(
+                user = student.user,
+                event = newEvent,
+                rsvpWaitlist = student.rsvpWaitlist
+            )
+        newRsvp.save()
+        rsvpCopies=len(rsvpInfo)
+
+    if rsvpCopies:
+        createRsvpLog(newEvent, f"Copied {rsvpCopies} Rsvp's from {priorEvent.name} to {newEvent.name}")
```
AndersonStettner: The name of the renewed event should link to the new event like how we link to newly created/canceled events. 

image

AndersonStettner (changes requested):
Some of the indentation and spacing is a little wonky in the html, please clean it up a bit.

app/controllers/admin/routes.py
```diff
@@ -157,6 +159,62 @@ def rsvpLogDisplay(eventId):
...
+        try:
+            if formData.get('dateEnd'):
+                assert formData['dateStart'] < formData['dateEnd']
+        except AssertionError:
+            flash("End date must be after start date", 'warning')
+            return redirect(url_for('admin.eventDisplay', eventId = eventId))
+
+
+        priorEvent = Event.get_by_id(eventId)
+        newEvent, message = attemptSaveEvent({
```
AndersonStettner: Get your dictionary from the prior event and update with your new form data. That new dict should be passed to `attemptSaveEvent` after its ID is removed.

AndersonStettner (changes requested):

app/controllers/admin/routes.py
andersoncedu: I have resolved this.

app/logic/events.py
andersoncedu: I have resolved this.

app/logic/events.py
andersoncedu: Done.

app/logic/events.py
andersoncedu: Done.

AndersonStettner (dismissed):
I am getting the following error again. This occurred while trying to renew a normal canceled event for the next day. 

image

bledsoef:
> I am getting the following error again. This occurred while trying to renew a normal canceled event for the next day.
> 
> image

We also got the same error for any kind of renewal.

azabeli:
image

seems the error is still there!

andersoncedu:
The error should be resolved.

gahimbaref:
image


You can only renew a canceled event once. If you cancel it the second time and try to renew it without changing any details, you get this error message.

gahimbaref:
image

We can only see half of the calendar pop up.

gahimbaref:
image

It creates separate entries for the same renewed event, is this the expected behavior?

app/controllers/admin/routes.py
```diff
@@ -157,6 +159,54 @@ def rsvpLogDisplay(eventId):
...
+            assert formData['timeStart'] < formData['timeEnd']
+        except AssertionError:
+            flash("End time must be after start time", 'warning')
+            return redirect(url_for('admin.eventDisplay', eventId = eventId))
+        
+        try:
+            if formData.get('dateEnd'):
+                assert formData['dateStart'] < formData['dateEnd']
+        except AssertionError:
+            flash("End date must be after start date", 'warning')
```
gahimbaref, line 171: Since the renew event modal does not have an end date, there is no instance where the end date would be before the start date. So this try/except might not be necessary.

app/controllers/admin/routes.py
```diff
@@ -157,6 +159,54 @@ def rsvpLogDisplay(eventId):
...
+            return redirect(url_for('admin.eventDisplay', eventId = eventId))
+
+        copyRsvp(priorEvent['id'], newEvent[0].id)
+        createAdminLog(f"Renewed {priorEvent['name']} as <a href='event/{newEvent[0].id}/view'>{newEvent[0].name}</a>.")
+        flash("Event successfully renewed.", "success")
+        return redirect(url_for('admin.eventDisplay', eventId = newEvent[0].id))
+
+
+    except Exception as e:
+        print("Error while trying to renew event:", e)
```
gahimbaref, line 200: remove print statement

andersoncedu: To my understanding, this is the proper way to do this and the same way it is done in other places to show exceptions in flask.

app/controllers/admin/routes.py
andersoncedu: I would like to keep this in here because if the start date ever ends up after the end date for whatever reason, this would catch that error.

andersoncedu:
> image
> 
> You can only renew a canceled event once. If you cancel it the second time and try to renew it without changing any details, you get this error message.

This has been resolved.

andersoncedu:
> image
> 
> We can only see half of the calendar pop up.

I couldn't replicate this.

bledsoef (changes requested):
Screenshot_20240212_112343

When trying to renew an event with rsvp'd volunteers and waitlisted volunteers, we get two different errors. The first is that "An error occured, contact system support." The other is "this event already exists" but it leaves you on the edit event screen of the cancelled event.

app/controllers/admin/routes.py
```diff
@@ -157,6 +159,54 @@ def rsvpLogDisplay(eventId):
...
+        flash("There was an error with your selection. Please try again or contact Systems Support.", 'danger')
```
bledsoef: I feel this flash statement could be a bit more descriptive. Does not really say what the error was, what does it mean by "selection"? Just a little confusing.

app/logic/events.py
```diff
@@ -501,3 +506,18 @@ def getEventRsvpCount(eventId):
         Returns the number of RSVP'd participants for a given eventId.
     """
     return len(EventRsvp.select().where(EventRsvp.event_id == eventId))
+
+def copyRsvp(priorEvent, newEvent):
```
bledsoef: Please leave comments on this code. Confused :'(

azabeli:
image


I find it confusing to have both "Save" and "Renew Event" buttons when editing a canceled event. As in which button to click on since they both are saving changes. Also, Why do we need to save changes for canceled events?  In my opinion, I would think having just the Renew Event button should be good for canceled events instead of confusing the user.

bledsoef (changes requested):
Will look back over code more in depth once the errors are resolved, overall looks alright, will say it would be great if some more comments and docstrings could be added in. Some of the copying stuff is not the most intuitive. Good work though!

app/controllers/admin/routes.py
andersoncedu: Change the test to read; there was an error renewing the event.

app/logic/events.py
andersoncedu: I left a docstring.

app/controllers/admin/routes.py
```diff
@@ -157,6 +159,54 @@ def rsvpLogDisplay(eventId):
...
+                    'endDate': f'{formData["endDate"][-4:]}-{formData["endDate"][0:-5]}',
+                    'isRecurring': bool(priorEvent['recurringId'])
+                    })
+        newEvent, message = attemptSaveEvent(newEventDict, renewedEvent = True)
+        if message:
+            flash(message, "danger")
+            return redirect(url_for('admin.eventDisplay', eventId = eventId))
+
+        copyRsvpToNewEvent(priorEvent, newEvent[0])
+        createAdminLog(f"Renewed {priorEvent['name']} as <a href='event/{newEvent[0].id}/view'>{newEvent[0].name}</a>.")
```
bledsoef, line 194: Should we link to both the priorEvent and the newEvent or is just the newEvent necessary? Feels like it couldn't hurt to link to both.

app/logic/events.py
```diff
@@ -350,24 +352,28 @@ def validateNewEventData(data):
...
-                      .where((Event.name == data['name']) &
-                             (Event.location == data['location']) &
-                             (Event.startDate == data['startDate']) &
-                             (Event.timeStart == data['timeStart'])))
+        sameEventList = list((Event.select().where((Event.name == data['name']) &
+                                                   (Event.location == data['location']) &
+                                                   (Event.startDate == data['startDate']) &
+                                                   (Event.timeStart == data['timeStart'])).execute()))
+        
+        copyList = sameEventList.copy()
```
bledsoef: I think this variable could be named to be more descriptive.

app/logic/events.py
```diff
@@ -350,24 +352,28 @@ def validateNewEventData(data):
...
             Term.get_by_id(data['term'])
         except DoesNotExist as e:
             return (False, f"Not a valid term: {data['term']}")
-
-        if event.exists():
+        if sameEventList:
             return (False, "This event already exists")
-
+        
     data['valid'] = True
```
bledsoef, line 377: Does this line have any effect?

app/templates/admin/createEvent.html
```diff
@@ -333,4 +352,37 @@ <h5 class="modal-title" id="warningModel">Delete Event</h5>
...
+  <div class="modal fade" id="renewWarning" tabindex="-1" aria-labelledby="renewModel" aria-hidden="true">
+    <div class="modal-dialog">
+      <div class="modal-content">
+        <div class="modal-header">
+          <h5 class="modal-title" id="renewModel">Renew Event</h5>
+          <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+        </div>
+        <div class="modal-body">
+          <p> Are you sure you want to renew this event?</p>
+          <p> A renewed event will use the current events RSVP, and location information to create a new event.</p>
```
bledsoef: This is a little grammatically incorrect. It should be "A renewed event will use the current event's RSVP and location information to create a new event."

app/templates/admin/createEvent.html
```diff
@@ -42,6 +42,82 @@
...
+              aria-describedby="timeIcon1" placeholder="Pick a start time"  
+              id="startTime-{{pageLocation}}" data-page-location="{{pageLocation}}" required />
+            <span class="input-group-text timeIcons" id="timeIcon1-{{pageLocation}}" hidden><i class="bi bi-clock"></i></span>
+          </div>
+        </div>
+        <div class='form-group col-md-6'>
+          <label class="form-label me-2" for="endTime-{{pageLocation}}"><strong>End Time</strong></label>
+          <div class="input-group">
+            {% if eventData.timeEnd %}
+              {% set end_time = eventData.timeEnd %}
```
bledsoef: Other variables in this file are camel case.

app/templates/admin/createEvent.html
```diff
@@ -42,6 +42,82 @@
...
+    <!-- Start and End Time -->
+    <div class="mb-4">
+      <div class="row col-md-12" >
+        <div class="form-group col-md-6">
+          <label class="form-label me-2" for="startTime-{{pageLocation}}"><strong>Start Time</strong></label>
+          <div class="input-group">
+            {% if eventData.timeStart %}
+              {% set start_time = eventData.timeStart %}
+            {% else %}
+              {% set start_time = "12:00" %}
```
bledsoef: Other variables in this file are camel case. Lets keep them consistent.

tests/code/test_events.py
```diff
@@ -896,3 +897,48 @@ def test_getEventRsvpCount():
             assert rsvpd_user_count == (index + 1)
 
         transaction.rollback()
+
+
+@pytest.mark.integration
+def test_copyRsvpToNewEvent():
+    with mainDB.atomic() as transaction:
+        with app.app_context():
+            g.current_user = "heggens"
```
bledsoef, line 906: Why do we set g.current_user. Doesn't seem to be used anywhere else?

bledsoef (changes requested):
All pretty small things, I think otherwise it looks and works really well!

app/controllers/admin/routes.py
andersoncedu: This was actually a decision that was made pretty early on in the design process. I think I spoke with either Anderson or Brian about it; I can't remember who for sure.

app/logic/events.py
andersoncedu: From my understanding, the valid field is added so that when we run the saveEventToDb function, we verify that it's vent through the validation function because we check if this field is there.

app/logic/events.py
andersoncedu: Changed variable name to sameEventListCopy.

app/templates/admin/createEvent.html
andersoncedu: Fixed

app/templates/admin/createEvent.html
andersoncedu: Fixed.

app/templates/admin/createEvent.html
andersoncedu: Fixed

tests/code/test_events.py
andersoncedu: It's used by the createRsvpLog function called in copyRsvpToNewEvent.

app/logic/events.py
```diff
@@ -501,3 +506,20 @@ def getEventRsvpCount(eventId):
...
+    for student in rsvpInfo:
+        newRsvp = EventRsvp(
+                user = student.user,
+                event = newEvent,
+                rsvpWaitlist = student.rsvpWaitlist
+            )
+        newRsvp.save()
+    numRsvps = len(rsvpInfo)
+    if numRsvps:
+        createRsvpLog(newEvent, f"Copied {numRsvps} Rsvp's from {priorEvent['name']} to {newEvent.name}")
```
bledsoef: I'm sorry Chris, but "Rsvp's" should be "Rsvps".

app/static/js/createEvents.js
```diff
@@ -8,14 +9,15 @@ function updateDate(obj) {
   var newMonth = dateToChange.getMonth();
   var newYear = dateToChange.getFullYear();
   var newDay = dateToChange.getDate();
-  if(obj.id == "startDatePicker") {
-    $("#endDatePicker").datepicker({minDate: new Date(  newYear, newMonth, newDay)});
-    $("#endDatePicker").datepicker( "option", "minDate", new Date(  newYear, newMonth, newDay));
+  if(obj.className.includes("startDatePicker")) {
+    
+    $("#endDatePicker-"+$(obj).data("page-location")).datepicker({minDate: new Date(  newYear, newMonth, newDay)});
+    $("#endDatePicker-"+$(obj).data("page-location")).datepicker( "option", "minDate", new Date(  newYear, newMonth, newDay));
```
bledsoef: Formatting on this line is a little wacky.

app/static/js/createEvents.js
```diff
@@ -8,14 +9,15 @@ function updateDate(obj) {
...
   }
 
-  if (obj.id == "endDatePicker") {
-    $("#startDatePicker").datepicker({maxDate: new Date(  newYear, newMonth, newDay)});
-    $("#startDatePicker").datepicker("option", "maxDate", new Date(  newYear, newMonth, newDay));
+  if (obj.className.includes("endDatePicker")) {
+    $("#startDatePicker-"+$(obj).data("page-location")).datepicker({maxDate: new Date(  newYear, newMonth, newDay)});
```
bledsoef: Same on these lines with the two spaces after Date(

app/controllers/admin/routes.py
```diff
@@ -153,6 +156,54 @@ def rsvpLogDisplay(eventId):
...
+                    'startDate': f'{formData["startDate"][-4:]}-{formData["startDate"][0:-5]}',
+                    'endDate': f'{formData["endDate"][-4:]}-{formData["endDate"][0:-5]}',
+                    'isRecurring': bool(priorEvent['recurringId'])
+                    })
+        newEvent, message = attemptSaveEvent(newEventDict, renewedEvent = True)
+        if message:
+            flash(message, "danger")
+            return redirect(url_for('admin.eventDisplay', eventId = eventId))
+
+        copyRsvpToNewEvent(priorEvent, newEvent[0])
```
bledsoef, line 193: You use this function but it is not imported and causes an error.

bledsoef (changes requested):
Code looks good but it breaks because of a missing import.

app/controllers/admin/routes.py
andersoncedu: Yes I messed up when resolving a merge conflict.

app/static/js/createEvents.js
andersoncedu: Fixed

app/static/js/createEvents.js
andersoncedu: Fixed

app/logic/events.py
andersoncedu: Fixed

bledsoef (approved):
Everything seems to work and look good, couldn't break it :disappointed: I would just talk to Brian about whether a new issue is necessary to renew past events. Nice work!

PR #1096, opened 2023-10-24
Minor page 989

bledsoef:
# Changes
- Created a new html page at `/profile/neillz/cceMinor`
- Created new logic and controller functions in the respective `minor.py` and `routes.py` files.
- Added the functionality to view engagement history for courses and programs.

# Testing
- `database/reset_database.sh test`
- `tests/run_tests.sh`
- Visit the user profile page for any user
- Click on actions in the top right, and select "Manage CCE Minor"
- Test the express interest button, verify that selecting it updates your status to interest expressed.
- Click on the accordion and verify you see a selection of entries, for instance Summer 2021 should show Databases
- Click on an entry in the accordion and for courses, verify you see the course information, and for programs, see the volunteer's participation history.


qasema:
I think the page design could be a bit better. I would say try to group related elements together.

qasema (changes requested):
both accordions are initially open but the content is not displayed. I would say have the first accordion open or have them both closed

bledsoef:
> both accordions are initially open but the content is not displayed. I would say have the first accordion open or have them both closed

Fixed.

bledsoef:
> I think the page design could be a bit better. I would say try to group related elements together.

Gave the page a little bit of a redesign

qasema (changes requested):
Screenshot 2023-11-30 at 7 26 48 PM

Checkbox alignment is off. Try changing the page size to test it

app/controllers/minor/routes.py
```diff
@@ -1,13 +1,22 @@
-from flask import Flask, g
-
+from flask import Flask, g, render_template, request
+from app.logic.minor import updateMinorInterest, getProgramEngagementHistory, getCourseInformation
+from app.models.user import User
+from app.models.course import Course
```
qasema: Is Course going to be used later?

app/controllers/minor/routes.py
```diff
@@ -1,13 +1,22 @@
-from flask import Flask, g
-
+from flask import Flask, g, render_template, request
+from app.logic.minor import updateMinorInterest, getProgramEngagementHistory, getCourseInformation
+from app.models.user import User
+from app.models.course import Course
 from app.controllers.minor import minor_bp
+from app.logic.minor import getCommunityEngagementByTerm
+from playhouse.shortcuts import model_to_dict
```
qasema: same with model_to_dict

app/controllers/minor/routes.py
```diff
@@ -16,6 +25,18 @@ def identifyCommunityEngagement(username):
...
+def getEngagementInformation(username, type, id, term):
+    """
+        For a particular engagement activity (program or course), get the participation history or course information respectively.
+    """
+    if type == "program":
+        information = getProgramEngagementHistory(id, username, term)
+    else:
+        information = getCourseInformation(id)
+
+    return information
```
qasema, line 37: I think there could be a case where the student does not have engagement. In that case, we should return something indicating that. But you can handle that in the front end by checking if information is empty

qasema (changes requested):
We should also add an extra layer of validation for the current user in the controllers. Now everyone can view anyone else's minor page

app/controllers/minor/routes.py
bledsoef: Brian had let me know that I just shouldn't display anything if they have nothing

qasema:
looks good to me

qasema (dismissed):

app/controllers/minor/routes.py
```diff
@@ -1,13 +1,21 @@
...
 from app.controllers.minor import minor_bp
+from app.logic.minor import getCommunityEngagementByTerm
 
 @minor_bp.route('/profile/<username>/cceMinor', methods=['GET'])
 def viewCceMinor(username):
     """
         Load minor management page with community engagements and summer experience
     """
-    pass
+    if not (g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff):
```
BrianRamsay: use isAdmin if you want both cases (we'll figure out the exact permissions needed later)

app/controllers/minor/routes.py
```diff
@@ -30,7 +50,7 @@ def removeCommunityEngagement(username):
     """
     pass
 
-@minor_bp.route('/cceMinor/<username>/requestOtherCommunityEngagement', methods=['GET,POST'])
+@minor_bp. route('/cceMinor/<username>/requestOtherCommunityEngagement', methods=['GET,POST'])
```
BrianRamsay: no space

app/logic/minor.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+
+    # retrieve the course and the course instructors
+    course = model_to_dict(Course.get_by_id(id))
+
+    courseInstructors = (CourseInstructor.select(CourseInstructor.user)
+                         .join(Course)
+                         .where(Course.id == id))
+    
+    courseInformation = {"instructors": [(instructor.user.firstName + " " + instructor.user.lastName) for instructor in courseInstructors], "course": course}
```
BrianRamsay, line 39: This will be doing a separate query for each instructor. Join User in the query.

app/logic/minor.py
```diff
@@ -0,0 +1,90 @@
...
+    # initialize default dict to store term descriptions as keys mapping to each
+    # engagement's respective type, name, id, and term.
+    terms = defaultdict(list)
+    for course in courses:
+        terms[(course.term.description, course.term.id)].append({"name":course.courseName, "id":course.id, "type":"course", "term":course.term})
+
+    events = (Event.select(Event)
+                       .join(EventParticipant, on=(Event.id == EventParticipant.event))
+                       .where(EventParticipant.user == username)
+                       .group_by(Event.program, Event.term))
```
BrianRamsay: Anderson doesn't like the query indentation. Line up your periods.

app/logic/minor.py
```diff
@@ -0,0 +1,90 @@
...
+def getProgramEngagementHistory(program_id, username, term_id):
+    """
+        Given a program_id, username, and term_id, return an object containing all events in the provided program 
+        and in the given term along with the program name.
+    """
+
+    # execute a query that will retrieve all events in which the user has participated
+    # that fall under the provided term and programs.
+    eventsInProgramAndTerm = (Event.select(Event.id, Event.name)
+                               .join(Program, JOIN.LEFT_OUTER).switch()
```
BrianRamsay: left join not needed, as the where clause prevents the null case anyway

app/templates/main/userProfile.html
```diff
@@ -42,6 +42,9 @@ <h1>{{volunteer.firstName}} {{volunteer.lastName}}</h1>
             <option value="Insurance Information"> {{'Edit' if volunteer == g.current_user else 'View'}} Insurance Information </option>
             <option value="Print Travel Form"> Print Travel Form </option>
         {% endif %}
+        {% if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff %}
```
BrianRamsay: is `isAdmin` to cover both cases

BrianRamsay (changes requested):
click for more information should open in a new window

community engagement ui feedback
- highlight and click on the whole row, not just links
- links out to the event page should open in a new window (`target="_blank"`)
- include Course abbreviation with course name
- Remove "Information" from course info header
- Program event list should include hours for each event, and the total for the program

gahimbaref:
image

clicking on the checkboxes doesn't seem to be doing anything unless you click the activity name, is there a reason for the checkboxes?

bledsoef:
> image
> 
> 
> 
> clicking on the checkboxes doesn't seem to be doing anything unless you click the activity name, is there a reason for the checkboxes? 

That functionality is part of a different issue

gahimbaref:
Centering the title "Zach Neill's CCE Minor" might look a bit better, it would also be consistent with the other pages

bledsoef:
> Centering the title "Zach Neill's CCE Minor" might look a bit better, it would also be consistent with the other pages

I was basing this off of the user profile which has its title aligned to the left.

BrianRamsay (approved):

PR #1101, opened 2023-10-27
Remove unnecessary queries issue1100

### LINKED ISSUE #1100: Reduce amount of unnecessary queries that are run when loading the landing page.
Currently, we run ~120 queries when loading the landing page if you have reset you database `from-backup`, this is way to many and should be, at most, about 15 to 20.

andersoncedu:
Issue: 
There was an issue where when accessing the landing page with production data, 120+ queries were being made.

Fix: 
To fix this issue, added two joins and a switch to queries to prevent them from needing to query when loading additional user and program data for the program managers.
Test:
Reset the database with backup data "from-backup" and load into the landing page. Check the flask terminal window to make sure queries are a reasonable size (<6). 

fix #1100


app/logic/landingPage.py
```diff
@@ -2,15 +2,18 @@
...
 from app.models.program import Program
+from app.models.user import User
 
 
 def getManagerProgramDict(user):
     
-    managerRows = ProgramManager.select()
+    managerRows = (ProgramManager.select(ProgramManager, User, Program)
+                                 .join(User)
+                                 .switch(ProgramManager).join(Program))
```
AndersonStettner: the last `.join()` should be in line with everything else

AndersonStettner (changes requested):

app/controllers/main/routes.py
```diff
@@ -43,12 +43,26 @@ def redirectToLogout():
...
+    # Optimize the query to fetch programs with non-canceled, non-past events in the current term
+    programsWithEventsQuery = (Program
+        .select()
+        .join(Event)
+        .where(
+            (Event.term == g.current_term) &
+            (Event.isCanceled == False) &
+            (Event.isPast == False)
+        )
+        .distinct()  # Ensure only unique programs are included
```
AndersonStettner: Please put the query in the same format as all other queries in the application.

AndersonStettner (changes requested):
Nice job, this is a lot better. We still have 3 identical queries (4, 5, and 6) so please check and see if we can reduce those as well.

image

app/controllers/main/routes.py
andersoncedu: Done.

app/controllers/main/routes.py
```diff
@@ -43,12 +43,19 @@ def redirectToLogout():
...
 
-    return render_template("/main/landingPage.html", managerProgramDict = managerProgramDict,
-                                                     term = g.current_term,
-                                                     programsWithEventsList = programsWithEventsList)
+    # Optimize the query to fetch programs with non-canceled, non-past events in the current term
+    programsWithEventsList = list(Program.select(Program, Event)
+                                         .join(Event)
+                                         .where((Event.term == g.current_term) and (Event.isCanceled == False) and (Event.isPast == False))
+                                         .distinct()
+                                         .execute())  # Ensure only unique programs are included
```
AndersonStettner, line 52: Neither of the comments on this are really needed.

AndersonStettner (changes requested):

app/controllers/main/routes.py
BrianRamsay: I agree, especially since the second one is on the wrong line, but in the interest of keeping PRs moving I am going to go ahead and merge it

BrianRamsay (approved):

PR #1102, opened 2023-11-17
884 generalize spreadsheet

### LINKED ISSUE #884: Generalize Spreadsheet queries and write tests for them
There is a logic file, `spreadsheet.py`, that queries the database for data that CELTS has asked for. The three main tasks for this issue are to generalize the functions, write tests for the new logic file, and create UI to download the data within the application. 

### Task one:
Currently, a few of the queries are too specific, the terms for getting retention rates are the main ones I can think of. Using the retention rate as an example, we want to allow the user to select which terms they would like to be used to calculate the retention rate instead of hardcoding them in the logic like they are now. 

### Task two: 
Write new tests for the new `spreadsheet.py` logic.

andersoncedu:
Issue: Testing functions from `spreadsheet.py`

Fix: created a new file in test folder called `test_spreadsheet.py`, and tested almost every single function in `spreadsheet.py` 

Also created an initial UI using a sidebar button, "Create Spreadsheet", and route for that button, however, this was stated to be outside the scope of our issue so it was not completed, only moved to a new issue. 

Test:
1. `source setup.sh`
2. `git checkout 884-Generalize-Spreadsheet`
3. `./database/reset_database.sh test`
4. `tests/monitor.sh`
all the tests should be passing. 

fix #884


app/logic/spreadsheet.py
```diff
@@ -143,9 +145,9 @@ def termParticipation(termDescription):
 
     return programParticipationDict
 
-def removeNullParticipants(bla):
-    # loop through the list and remove all entries that do not have a participant
-    return list(filter(lambda participant: bool(participant), bla))
+def removeNullParticipants(participant_list):
+    # loop through the dictionary and remove all entries that do not have a participant
+    return list(filter(lambda participant: bool(participant), participant_list))
```
AndersonStettner: If it is a dictionary then it should be called `participant_dict` not `participant_list`. If it is not a dictionary then the comment should be changed back.

app/logic/spreadsheet.py
```diff
@@ -96,7 +96,9 @@ def volunteerMajorAndClass(column, reorderClassLevel=False):
 
 def repeatVolunteersPerProgram():
     # Get people who participated in events more than once (individual program)
-    repeatPerProgramQuery = (EventParticipant.select(fn.CONCAT(User.firstName, " ", User.lastName),Program.programName.alias("programName"),fn.Count(EventParticipant.event_id).alias('event_count'))
+    repeatPerProgramQuery = (EventParticipant.select(fn.CONCAT(User.firstName, " ", User.lastName),
+                                              Program.programName.alias("programName"),
+                                              fn.Count(EventParticipant.event_id).alias('event_count'))
```
AndersonStettner: `Program` and the `fn.Count` are a part of the select so they should be tabbed over to be in line with the `fn.CONCAT`

AndersonStettner (changes requested):
Brian and I have some problems with the tests, talk to us when you come in so we can go over everything.

app/logic/spreadsheet.py
andersoncedu: Made requested change.

app/logic/spreadsheet.py
andersoncedu: Made requested change.

app/controllers/admin/stats.py
```diff
@@ -0,0 +1,30 @@
...
+                           academicYears = academicYears)
+
+@admin_bp.route("/createSpreadsheet/", methods=["POST"])
+def createSpreadsheet():
+    if not g.current_user.isCeltsAdmin:
+        abort(403)
+
+    formData = request.form
+    print(formData.get("academicYear"))
+    # create_spreadsheet(formData.get("academicYear"), formData.get("academicTerm"))
```
AndersonStettner: Remove commented-out code.

app/static/js/spreadsheetMaker.js
```diff
@@ -0,0 +1,15 @@
+$(document).ready(function(){
+	$("#createSpreadsheet").on("click", function(){
+
+		$.ajax({
+			method: "POST",
+			url: "/createSpreadsheet/",
+			data: {"academicYear": $("#academicYear").val(),
+				   "academicTerm": $("#academicTerm").val()},
+			success: function(response) {
+				console.log(data)
```
AndersonStettner: I am guessing this is a debug statement, if so it should be removed. Also, please fix the spacing for `data`

AndersonStettner (dismissed):

app/controllers/admin/stats.py
andersoncedu: Done

app/static/js/spreadsheetMaker.js
andersoncedu: Done

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,154 @@
...
+@pytest.mark.integration
+def test_termParticipation():
+    assert termParticipation('Fall 2020')=={'Adopt-a-Grandparent': ['khatts'], 'CELTS-Sponsored Event': [None]}
+
+@pytest.mark.integration
+def test_getRetentionRate():
+    assert getRetentionRate("2020-2021") == [('Adopt-a-Grandparent', '0.0%'), ('CELTS-Sponsored Event', '0.0%')]
+
+@pytest.mark.integration
+def test_repeatVolunteers():
```
bledsoef, line 97: Test does not pass for me but it does for Azabe.

/home/fbledsoe/SSDT/celts/tests/code/test_spreadsheet.py:44: AssertionError: assert [('Sreynit Kh...ch Neill', 2)] == [('Sreynit Kh...r Parton', 2)]

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,154 @@
...
+    with mainDB.atomic() as transaction:
+        assert list(onlyCompletedAllVolunteer("2020-2021").execute()) == []
+        EventParticipant.create(user = 'partont',  
+                                event = 14,
+                                hoursEarned = 1)
+        assert list(onlyCompletedAllVolunteer("2021-2022").execute()) == []
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_volunteerProgramHours():
```
bledsoef, line 282: This test also doesn't pass for me. It passes for Azabe though. I have checked out and pulled and reset database a million times and have rerun tests a bunch too. Don't know why.

/home/fbledsoe/SSDT/celts/tests/code/test_spreadsheet.py:103: AssertionError: assert [('Adopt-a-Gr...artont', 5.0)] == [('Hunger Ini...artont', 5.0)]

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,154 @@
...
+
+from app.logic.spreadsheet import *
+
+
+@pytest.mark.integration
+def test_create_spreadsheet():
+    create_spreadsheet("2020-2021")
+
+@pytest.mark.integration
+def test_calculateRetentionRate():
```
bledsoef, line 16: I think this test could be better. You only have one test case that tests to see if two programs have a retention rate of 0.0. I feel like additional assertions where the retention rate is not zero would make this test more robust and more helpful.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,154 @@
...
+    fall_Dict=({'Adopt-a-Grandparent': ['khatts'], 'CELTS-Sponsored Event': [None]})
+    spring_dict=({'Hunger Initiatives': ['neillz', 'khatts', 'ayisie', 'neillz', 'partont']})
+    assert calculateRetentionRate(fall_Dict,spring_dict) == {'Adopt-a-Grandparent': 0.0, 'CELTS-Sponsored Event': 0.0}
+	
+@pytest.mark.integration
+def test_removeNullParticipants():
```
bledsoef, line 30: This test could be better. This does not really test what the function does since no null participants are ever included in the test. I would add some more cases that better reflect what is in the function.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,154 @@
...
+	# Test case 1: Dictionary with null participants
+    input_dict_1 = ['khatts']
+    assert removeNullParticipants(input_dict_1) == ['khatts']
+
+@pytest.mark.integration
+def test_termParticipation():
```
bledsoef, line 38: I think this test could be a bit better. I think maybe creating some new data and ensuring that the term participants are properly updating and are being caught by the function.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,154 @@
...
+    assert termParticipation('Fall 2020')=={'Adopt-a-Grandparent': ['khatts'], 'CELTS-Sponsored Event': [None]}
+
+@pytest.mark.integration
+def test_getRetentionRate():
```
bledsoef, line 61: Same comment as on calculateRetentionRate.

tests/code/test_spreadsheet.py
bledsoef: Additionally, I would recommend testing to make sure that someone who has only participated once is not caught. For instance create a new EventParticipant entry for someone who has never participated in an event and make sure they are not present in the list.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,154 @@
...
+
+        assert list(repeatVolunteers().execute()) == [('Sreynit Khatt', 4), ('Zach Neill', 2)]
+        EventParticipant.create(user = 'partont',  
+                                event = 2,
+                                hoursEarned = 1)
+        assert list(repeatVolunteers().execute()) == [('Sreynit Khatt', 4), ('Zach Neill', 2), ('Tyler Parton', 2)]
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_repeatVolunteersPerProgram():
```
bledsoef, line 143: Same comment on this as repeatVolunteers.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,154 @@
...
+        EventParticipant.create(user = 'partont',  
+                                event = 2,
+                                hoursEarned = 1)
+        assert list(repeatVolunteersPerProgram().execute()) == [('Zach Neill', 'Hunger Initiatives', 2), ('Tyler Parton', 'Hunger Initiatives', 2),
+                                                                ('Sreynit Khatt', 'Adopt-a-Grandparent', 3)]
+
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_volunteerMajorAndClass():
```
bledsoef, line 185: This is good but is not quite testing for all cases. We never get to test volunteerMajorAndClass when User.classLevel is passed or when the reorderClassLevel flag is set to true.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,154 @@
...
+    with mainDB.atomic() as transaction:
+        assert list(volunteerHoursByProgram().execute()) == [('Adopt-a-Grandparent', 9.0), ('Berea Buddies', 0.0), ('Hunger Initiatives', 11.0)]
+        EventParticipant.create(user = 'partont',  
+                                event = 2,
+                                hoursEarned = 1)
+        assert list(volunteerHoursByProgram().execute()) == [('Adopt-a-Grandparent', 9.0), ('Berea Buddies', 0.0), ('Hunger Initiatives', 12.0)]
+        transaction.rollback()
+        
+@pytest.mark.integration
+def test_onlyCompletedAllVolunteer():
```
bledsoef, line 250: The function actually does work correctly. The issue is that you are testing with Tyler, who has participated in another event. If you create an entirely fresh user, it will work, since the function is testing that the volunteer has only completed the All Volunteer Training and nothing else.

bledsoef (changes requested):
In general, I would just reevaluate the test cases in general to make sure you are testing all possible scenarios.

tests/code/test_spreadsheet.py
andersoncedu: I can't recreate the test failing.

andersoncedu: The test you've recommended is already being tested for, partont has only participated in one event so he is not counted in the first assertion, but when he's added for another event, he shows up in the second assertion.

tests/code/test_spreadsheet.py
andersoncedu: I can't recreate this.

tests/code/test_spreadsheet.py
andersoncedu: This is more robust now.

tests/code/test_spreadsheet.py
andersoncedu: We addressed this!

tests/code/test_spreadsheet.py
andersoncedu: This has been updated to be more robust.

tests/code/test_spreadsheet.py
andersoncedu: I think, same as repeatVolunteers, this tests cases where a person isn't a repeated volunteer and then is.

tests/code/test_spreadsheet.py
andersoncedu: I've updated the test to be more robust and test different amounts of retention.

tests/code/test_spreadsheet.py
andersoncedu: Added new user and added them to all volunteer training to test a user who's only been in one all-volunteer training, and that's it.

tests/code/test_spreadsheet.py
andersoncedu: I implemented your suggested changes.

app/logic/spreadsheet.py
```diff
@@ -143,9 +145,9 @@ def termParticipation(termDescription):
 
     return programParticipationDict
 
-def removeNullParticipants(bla):
+def removeNullParticipants(participant_list):
```
gahimbaref: use camelCase to stay consistent

app/logic/spreadsheet.py
```diff
@@ -158,8 +160,8 @@ def calculateRetentionRate(fall_dict, spring_dict):
             retention_rate = len(fall_participants & spring_participants) / len(fall_participants)
         except ZeroDivisionError:
             pass
-
         retention_dict[program] = retention_rate
+
   
     return retention_dict
```
Karina-Agliullova, line 168: there is a huge piece of code with two functions, it doesn't look clean. Do you want to either use it or get rid of it?

gahimbaref:
Perhaps consider testing invalid inputs as well, seems like the tests only account for valid inputs, this might make the test suite more robust.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,196 @@
...
+                                event = 2,
+                                hoursEarned = 1)
+        assert list(volunteerHoursByProgram().execute()) == [('Adopt-a-Grandparent', 9.0), ('Berea Buddies', 0.0), ('Hunger Initiatives', 12.0)]
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_onlyCompletedAllVolunteer():
+    #seems like onlyCompletedAllVolunteer isn't working properly
+    with mainDB.atomic() as transaction:
+        assert list(onlyCompletedAllVolunteer("2020-2021").execute()) == []
```
gahimbaref: Did you mean to fix it eventually? (comment says it doesn't work properly). Also, this function in particular could use testing invalid inputs.

Karina-Agliullova:
Why did you create a sidebar button, "Create Spreadsheet"? It says it was out of the scope of this issue, so why do we need it? Also, it is not working. I tested it and it doesn't generate any spreadsheet for any academic year but instead throws a TypeError message when I inspect the page

tests/code/test_spreadsheet.py
andersoncedu: This function has been changed to work as intended, the comment has been removed. There are negative tests in this function as there are people from 2020-2021 who have not completed all volunteer events and are not showing up as intended.

app/logic/spreadsheet.py
andersoncedu: I don't understand this comment.

andersoncedu: If you're talking about the commented-out code, Anderson had instructed me to leave them in there.

app/logic/spreadsheet.py
andersoncedu: I refactored the entire spreadsheets.py to be more consistent, since on the page, both camelCase and underscores were used.

app/logic/spreadsheet.py
hoerstl: After verbal confirmation with Karina, she was referring to the commented code

app/templates/sidebar.html
```diff
@@ -68,6 +68,9 @@
             <a href="/manageBonner" class="nav-link text-white {{ 'active' if request.path =='/manageBonner' }}">
               Bonner Management
             </a>
+            <a href="/spreadsheetMaker" class="nav-link text-white d-none {{ 'active' if request.path =='/spreadsheetMaker' }}">
```
bledsoef: I feel like this shouldn't be in here if it is hidden. Can we move this code out into its own PR?

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,194 @@
...
+        EventParticipant.create(user = 'ayisie',
+                                event = 14,
+                                hoursEarned = 1)
+        assert termParticipation('Fall 2020')=={'Adopt-a-Grandparent': ['khatts', 'partont'], 'CELTS-Sponsored Event': ['ayisie']}
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_getRetentionRate():
+    with mainDB.atomic() as transaction:
+        assert getRetentionRate("2020-2021") == [('Adopt-a-Grandparent', '0.0%'), ('CELTS-Sponsored Event', '0.0%')]
```
bledsoef: It looks like this working okay, but do you think we could make it so that we are manually adding someone to a fall term, and then manually adding that same person to a program event that next term. Not necessarily mandatory but I feel it may be best to do this so it is more clear what the test is doing and so we are not relying just on test data.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,194 @@
...
+                                event = testEvent,
+                                hoursEarned = 1)
+        assert getRetentionRate("2020-2021") == [('Adopt-a-Grandparent', '100.0%'), ('CELTS-Sponsored Event', '0.0%')]
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_repeatVolunteers():
+    with mainDB.atomic() as transaction:
+
+        assert list(repeatVolunteers().execute()) == [('Sreynit Khatt', 4), ('Zach Neill', 2)]
```
bledsoef: I think like we may have talked about, and like my previous comment which is not necessarily mandatory, but it makes it more clear what the test is doing if we add someone who has not participated in any events to a single event, and verify they do not show up in our first assertion. Then we can add them to another event and verify they do in fact show up in our second. Again, just a little hesitant on relying on test data.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,194 @@
...
+        EventParticipant.create(user = 'partont',
+                                event = 2,
+                                hoursEarned = 1)
+        assert list(repeatVolunteers().execute()) == [('Sreynit Khatt', 4), ('Zach Neill', 2), ('Tyler Parton', 2)]
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_repeatVolunteersPerProgram():
+    with mainDB.atomic() as transaction:
```
bledsoef, line 145: Reiterating last comment.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,194 @@
...
+                                hoursEarned = 1)
+        assert list(repeatVolunteersPerProgram().execute()) == [('Zach Neill', 'Hunger Initiatives', 2), ('Tyler Parton', 'Hunger Initiatives', 2),
+                                                                ('Sreynit Khatt', 'Adopt-a-Grandparent', 3)]
+
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_volunteerMajorAndClass():
+    with mainDB.atomic() as transaction:
+        assert list(volunteerMajorAndClass(User.major).execute()) == [('Biology', 1), ('Chemistry', 1), ('Computer Science', 2), ('Psychology', 1)]
```
bledsoef: Can we run each of the assertions before and after the User is created to make sure each one works as expected?

tests/code/test_spreadsheet.py
bledsoef: An alternative could be to add a comment stating that Tyler Parton belongs to an event already. I would recommend just either make these more explicit by defining things that don't exist, or add comments to make sure people know what we are doing, or even both.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,194 @@
...
+                                event = 2,
+                                hoursEarned = 1)
+        assert list(volunteerHoursByProgram().execute()) == [('Adopt-a-Grandparent', 9.0), ('Berea Buddies', 0.0), ('Hunger Initiatives', 12.0)]
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_onlyCompletedAllVolunteer():
+    with mainDB.atomic() as transaction:
+        assert list(onlyCompletedAllVolunteer("2020-2021").execute()) == []
+        EventParticipant.create(user = 'partont',
```
bledsoef: Like I have mentioned, can we inform the viewer that Tyler has already participated in an event, or use someone else where we define that they have participated in another event before the assertion?

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,194 @@
...
+    with mainDB.atomic() as transaction:
+        assert list(totalVolunteerHours().execute()) == [(20.0,)]
+        EventParticipant.create(user = 'qasema',
+                                event = 2,
+                                hoursEarned = 1)
+        assert list(totalVolunteerHours().execute()) == [(21.0,)]
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_getVolunteerProgramEventByTerm():
```
bledsoef, line 316: Maybe test adding someone to an event here to make sure this is working properly.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,194 @@
...
+                                                                                  ('Zach Neill', 'neillz', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'), 
+                                                                                  ('Zach Neill', 'neillz', 'Hunger Initiatives', 'Hunger Hurts'),
+                                                                                  ('Tyler Parton', 'partont', 'Hunger Initiatives', 'Hunger Hurts')])
+
+    assert list(getVolunteerProgramEventByTerm(Term.get_by_id(4)).execute()) == [('Alex Bryant', 'bryanta', 'Berea Buddies', 'Tutoring'),
+                                                                                 ('Sreynit Khatt', 'khatts', 'Adopt-a-Grandparent', 'Adoption 101'),
+                                                                                 ('Sreynit Khatt', 'khatts', 'Adopt-a-Grandparent', 'Meet & Greet with Grandparent')]
+
+@pytest.mark.integration
+def test_getUniqueVolunteers():
```
bledsoef, line 370: Same comment as last.

bledsoef (changes requested):
Tests mostly look good, just a couple changes that could be made. More comments or docstrings would be great as well.

app/templates/sidebar.html
andersoncedu: This has been broken into a new branch

tests/code/test_spreadsheet.py
andersoncedu: This has been addressed

app/logic/spreadsheet.py
```diff
@@ -115,15 +117,15 @@ def repeatVolunteers():
...
-    springParticipationDict = termParticipation("Spring 2023")  
+    (fall, spring) = academicYear.split("-")
+    fallParticipationDict = termParticipation(f"Fall {fall}")
+    springParticipationDict = termParticipation(f"Spring {spring}")  
 
     # calculate the retention rate using the defined function
-    retention_rate_dict = calculateRetentionRate(fallParticipationDict, springParticipationDict)
-    for program, retention_rate in retention_rate_dict.items():
+    retentionRateDict = calculateRetentionRate(fallParticipationDict, springParticipationDict)
+    for program, retention_rate in retentionRateDict.items():
```
bledsoef: Snake case -> camel case

app/logic/spreadsheet.py
```diff
@@ -115,15 +117,15 @@ def repeatVolunteers():
...
          retentionDict.append((program, str(round(retention_rate * 100, 2)) + "%"))
 
     return  retentionDict
```
bledsoef: Double spaces

app/logic/spreadsheet.py
```diff
@@ -115,15 +117,15 @@ def repeatVolunteers():
     
     return repeatAllProgramQuery.tuples()
 
-def getRetentionRate():
+def getRetentionRate(academicYear):
     retentionDict = []
-
-    fallParticipationDict = termParticipation("Fall 2022")
-    springParticipationDict = termParticipation("Spring 2023")  
+    (fall, spring) = academicYear.split("-")
```
bledsoef: Do you need parenthesis here?

app/logic/spreadsheet.py
```diff
@@ -137,31 +139,31 @@ def termParticipation(termDescription):
...
-        programParticipationDict[prog_name].append(participant)
+        programParticipationDict[progName].append(participant)
 
     return programParticipationDict
 
-def removeNullParticipants(bla):
+def removeNullParticipants(participantList):
     # loop through the list and remove all entries that do not have a participant
-    return list(filter(lambda participant: bool(participant), bla))
+    return list(filter(lambda participant: bool(participant), participantList))
```
bledsoef: Is the cast to bool necessary here?

app/logic/spreadsheet.py
```diff
@@ -137,31 +139,31 @@ def termParticipation(termDescription):
...
 # function to calculate the retention rate for each program
-def calculateRetentionRate(fall_dict, spring_dict):
-    retention_dict = {}
-    for program in fall_dict.keys():
-        fall_participants = set(removeNullParticipants(fall_dict[program]))
-        spring_participants = set(removeNullParticipants(spring_dict.get(program, [])))
-        retention_rate = 0.0
+def calculateRetentionRate(fallDict, springDict):
+    retentionDict = {}
+    for program in fallDict.keys():
```
bledsoef: Maybe use .items() here so we don't have to say fallDict[program] in the line below. But that is not mandatory.

app/logic/spreadsheet.py
```diff
@@ -263,7 +265,7 @@ def makeDataXls(getData, columnTitles, sheetName, workbook):
         setColumnWidth = max(len(str(x)) for x in columnData)
         worksheet.set_column(column, column, setColumnWidth + 3)
 
-def create_spreadsheet(): 
+def create_spreadsheet(academicYear):
```
bledsoef: Change function name to be camel case.

app/logic/spreadsheet.py
```diff
@@ -279,18 +281,18 @@ def create_spreadsheet():
     onlyCompletedAllVolunteerColumns = ["Username","Full Name "]
```
bledsoef: I know this isn't you, but adding a space between elements and removing the extra space at the end would be nice.

app/logic/spreadsheet.py
```diff
@@ -279,18 +281,18 @@ def create_spreadsheet():
...
-    makeDataXls(getUniqueVolunteers(), uniqueVolunteersColumns, "Unique Volunteers", workbook)
+    makeDataXls(getRetentionRate(academicYear), volunteerProgramRetentionRateAcrossTermColumns, "Retention Rate By Semester", workbook)
+    makeDataXls(getUniqueVolunteers(academicYear), uniqueVolunteersColumns, "Unique Volunteers", workbook)
     makeDataXls(totalVolunteerHours(), totalVolunteerHoursColumns, "Total Hours", workbook)
     makeDataXls(volunteerProgramHours(), volunteerProgramHoursColumns, "Volunteer Hours By Program", workbook)
-    makeDataXls(onlyCompletedAllVolunteer(), onlyCompletedAllVolunteerColumns , "Only All Volunteer Training", workbook)
-    makeDataXls(getVolunteerProgramEventByTerm(Term.get_by_id(8)), volunteerProgramEventByTerm, "Fall 2022", workbook)
+    makeDataXls(onlyCompletedAllVolunteer(academicYear), onlyCompletedAllVolunteerColumns , "Only All Volunteer Training", workbook)
+    makeDataXls(getVolunteerProgramEventByTerm(Term.get_or_none(f"Fall {academicYear.split('-')[0]}")), volunteerProgramEventByTerm, f"Fall {academicYear.split('-')[0]}", workbook)
```
bledsoef: Does this get_or_none ever not return none?

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,356 @@
...
+from app.models import mainDB
+from app.models.user import User
+from app.models.term import Term
+from app.models.eventParticipant import EventParticipant
+
+from app.logic.spreadsheet import *
+
+
+@pytest.mark.integration
+def test_create_spreadsheet():
```
bledsoef: Make this match the "test_functionName" format like other tests.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,356 @@
...
+    create_spreadsheet("2020-2021")
+
+@pytest.mark.integration
+def test_calculateRetentionRate():
+    fall_dict=({'Adopt-a-Grandparent': ['khatts'], 'CELTS-Sponsored Event': [None]})
```
bledsoef: Make these variable names camel case and add spaces around the = sign.

bledsoef: In this entire function.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,356 @@
...
+    fall_dict=({'Hunger Initiatives': ['neillz', 'khatts', 'ayisie', 'partont']})
+    assert calculateRetentionRate(fall_dict,spring_dict) == {'Hunger Initiatives': 1.0}
+
+    fall_dict=({'Hunger Initiatives': ['neillz', 'khatts', 'ayisie', 'partont']})
+    spring_dict=({'Hunger Initiatives': ['neillz', 'khatts', 'ayisie']})
+    assert calculateRetentionRate(fall_dict,spring_dict) == {'Hunger Initiatives': 0.75}
+
+@pytest.mark.integration
+def test_removeNullParticipants():
+    test_input_list = ['khatts']
```
bledsoef: Snake case -> camel case

bledsoef: In the whole function.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,356 @@
...
+
+@pytest.mark.integration
+def test_termParticipation():
+    with mainDB.atomic() as transaction:    
+        assert termParticipation('Fall 2020')=={'Adopt-a-Grandparent': ['khatts'], 'CELTS-Sponsored Event': [None]}
+
+        EventParticipant.create(user = 'partont',
+                                event = 10,
+                                hoursEarned = 1)
+        assert termParticipation('Fall 2020')=={'Adopt-a-Grandparent': ['khatts', 'partont'], 'CELTS-Sponsored Event': [None]}
```
bledsoef: Spaces around == in whole function.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,356 @@
...
+                    classLevel = 'Sophomore')
+        testEvent = Event.create(name="All Volunteer Training",
+                                 term=1,
+                                 program=9,
+                                 isTraining=1,
+                                 isAllVolunteerTraining=1)
+        EventParticipant.create(user = 'solijonovam', #Not participated in event
+                                event = testEvent, #Added to all volunteer training event
+                                hoursEarned = 1)
+        assert list(onlyCompletedAllVolunteer("2020-2021").execute()) == [('solijonovam', 'Madinabonu Solijonova')]
```
bledsoef: Maybe add her to another event and make sure she disappears.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,356 @@
...
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_volunteerProgramHours():
+    with mainDB.atomic() as transaction:
+        assert list(volunteerProgramHours().execute()) == ([('Hunger Initiatives', 'neillz', 4.0),
```
bledsoef: Why can someone have 0 hours and someone have None hours?

bledsoef (changes requested):
Screenshot_20240318_125859

It all looks pretty good besides some variable naming and spacing stuff. You are getting some failures in tests on my end. The arrays that are asserting as not equal contain the same information but in different orders. I would just sort them.

app/logic/spreadsheet.py
andersoncedu: No, I removed it.

app/logic/spreadsheet.py
andersoncedu: This has been changed to use the Term.description.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+                    classLevel = 'Sophomore')
+        testProgram = Program.create(programName = "Test Program",
+                                     programDescription = "A good program")
+        testEvent = Event.create(name="Test Event",
+                                 term=1, #Fall 2020
+                                 program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent,
+                                hoursEarned=1)
+        assert getRetentionRate("2020-2021") == [('Adopt-a-Grandparent', '0.0%'),
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+                                                 ('CELTS-Sponsored Event', '0.0%'), 
+                                                 ('Test Program', '0.0%')]
+        
+        testEvent2 = Event.create(name="Test Event2",
+                                  term=2, #Fall 2021
+                                  program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent2,
+                                hoursEarned=1)
+        assert getRetentionRate("2020-2021") == [('Adopt-a-Grandparent', '0.0%'),
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+                    classLevel = 'Sophomore')
+        testProgram = Program.create(programName = "Test Program",
+                                     programDescription = "A good program")
+        testEvent = Event.create(name="Test Event",
+                                 term=1, #Fall 2020
+                                 program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent,
+                                hoursEarned=1)
+        assert list(repeatVolunteers().execute()) == [('Sreynit Khatt', 4), ('Zach Neill', 2)]
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+        testEvent2 = Event.create(name="Test Event2",
+                                 term=2, #Fall 2021
+                                 program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent2,
+                                hoursEarned=1)
+        assert list(repeatVolunteers().execute()) == [('Sreynit Khatt', 4), ('Zach Neill', 2), ('Madinabonu Solijonova', 2)]
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+                    classLevel = 'Sophomore')
+        testProgram = Program.create(programName = "Test Program",
+                                     programDescription = "A good program")
+        testEvent = Event.create(name="Test Event",
+                                 term=1, #Fall 2020
+                                 program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent,
+                                hoursEarned=1)
+        assert list(repeatVolunteersPerProgram().execute()) == [('Zach Neill', 'Hunger Initiatives', 2),
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+                                hoursEarned=1)
+        assert list(repeatVolunteersPerProgram().execute()) == [('Zach Neill', 'Hunger Initiatives', 2), 
+                                                                ('Sreynit Khatt', 'Adopt-a-Grandparent', 3)]
+        testEvent2 = Event.create(name="Test Event2",
+                                 term=2, #Fall 2021
+                                 program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent2,
+                                hoursEarned=1)
+        assert list(repeatVolunteersPerProgram().execute()) == [('Zach Neill', 'Hunger Initiatives', 2),
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+        assert list(repeatVolunteersPerProgram().execute()) == [('Zach Neill', 'Hunger Initiatives', 2), 
+                                                                ('Sreynit Khatt', 'Adopt-a-Grandparent', 3),
+                                                                ('Madinabonu Solijonova', 'Test Program', 2)]
+
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_volunteerMajorAndClass():
+    with mainDB.atomic() as transaction:
+        assert list(volunteerMajorAndClass(User.major).execute()) == [('Biology', 1), ('Chemistry', 1), ('Computer Science', 2), ('Psychology', 1)]
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+        assert list(volunteerMajorAndClass(User.classLevel).execute()) == [('Junior', 1), ('Senior', 3), ('Sophomore', 1)]
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+        assert list(volunteerMajorAndClass(User.classLevel, True).execute()) == [('Sophomore', 1), ('Junior', 1), ('Senior', 3)]
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+                    phoneNumber = '732-384-3469',
+                    firstName = 'Madinabonu',
+                    lastName  = 'Solijonova',
+                    isStudent = True,
+                    major = 'Agriculture',
+                    classLevel = 'Sophomore')
+        EventParticipant.create(user = 'solijonovam',
+                                event = 2,
+                                hoursEarned = 2)
+        assert list(volunteerMajorAndClass(User.major).execute()) == [('Agriculture', 1), ('Biology', 1), ('Chemistry', 1), ('Computer Science', 2), ('Psychology', 1)]
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+        assert list(volunteerMajorAndClass(User.classLevel).execute()) == [('Junior', 1), ('Senior', 3), ('Sophomore', 2)]
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+        assert list(volunteerMajorAndClass(User.classLevel, True).execute()) == [('Sophomore', 2), ('Junior', 1), ('Senior', 3)]
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+                                hoursEarned=3
+                                )
+        testEvent3 = Event.create(name="Test Event",
+                                  term=4,
+                                  program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent3,
+                                hoursEarned=5
+                                )
+        assert list(volunteerHoursByProgram().execute()) == [('Adopt-a-Grandparent', 9.0), ('Berea Buddies', 0.0), ('Hunger Initiatives', 11.0), ('Test Program', 10.0)]
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+        EventParticipant.create(user = 'solijonovam', #Not participated in event
+                                event = testEvent2, #Added to all volunteer training event
+                                hoursEarned = 1)
+        assert list(onlyCompletedAllVolunteer("2020-2021").execute()) == []
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_volunteerProgramHours():
+    with mainDB.atomic() as transaction:
+        assert list(volunteerProgramHours().execute()) == ([('Hunger Initiatives', 'neillz', 4.0),
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+        assert list(volunteerProgramHours().execute()) == ([('Hunger Initiatives', 'neillz', 4.0),  
+                                                            ('Hunger Initiatives', 'khatts', 2.0),
+                                                            ('Berea Buddies', 'bryanta', 0.0),
+                                                            ('Adopt-a-Grandparent', 'khatts', 9.0),
+                                                            ('Hunger Initiatives', 'ayisie', None),
+                                                            ('Hunger Initiatives', 'partont', 5.0)])
+        EventParticipant.create(user = 'qasema',
+                                event = 2,
+                                hoursEarned = 1)
+        assert list(volunteerProgramHours().execute()) == ([('Hunger Initiatives', 'neillz', 4.0),
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+                                event = 2,
+                                hoursEarned = 1)
+        assert list(totalVolunteerHours().execute()) == [(21.0,)]
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_getVolunteerProgramEventByTerm():
+    with mainDB.atomic() as transaction:
+        assert list(getVolunteerProgramEventByTerm(Term.get_by_id(3)).execute()) == ([])
+        assert list(getVolunteerProgramEventByTerm(Term.get_by_id(2)).execute()) == ([('Ebenezer Ayisi', 'ayisie', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'),
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+                                                                                      ('Sreynit Khatt', 'khatts', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'),
+                                                                                      ('Zach Neill', 'neillz', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'), 
+                                                                                      ('Zach Neill', 'neillz', 'Hunger Initiatives', 'Hunger Hurts'),
+                                                                                      ('Tyler Parton', 'partont', 'Hunger Initiatives', 'Hunger Hurts')])
+
+        assert list(getVolunteerProgramEventByTerm(Term.get_by_id(4)).execute()) == [('Alex Bryant', 'bryanta', 'Berea Buddies', 'Tutoring'),
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+                                  term=2,
+                                  program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent2)
+        testEvent3 = Event.create(name="Test Event",
+                                  term=4,
+                                  program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent3)
+        assert list(getVolunteerProgramEventByTerm(Term.get_by_id(3)).execute()) == [('Madinabonu Solijonova', 'solijonovam', 'Test Program', 'Test Event')]
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+        assert list(getVolunteerProgramEventByTerm(Term.get_by_id(2)).execute()) == [('Ebenezer Ayisi', 'ayisie', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'),
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+        assert list(getVolunteerProgramEventByTerm(Term.get_by_id(2)).execute()) == [('Ebenezer Ayisi', 'ayisie', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'), 
+                                                                                     ('Sreynit Khatt', 'khatts', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'), 
+                                                                                     ('Zach Neill', 'neillz', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'), 
+                                                                                     ('Zach Neill', 'neillz', 'Hunger Initiatives', 'Hunger Hurts'), 
+                                                                                     ('Tyler Parton', 'partont', 'Hunger Initiatives', 'Hunger Hurts'), 
+                                                                                     ('Madinabonu Solijonova', 'solijonovam', 'Test Program', 'Test Event')]
+
+        assert list(getVolunteerProgramEventByTerm(Term.get_by_id(4)).execute()) == [('Alex Bryant', 'bryanta', 'Berea Buddies', 'Tutoring'),
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+        assert list(getVolunteerProgramEventByTerm(Term.get_by_id(4)).execute()) == [('Alex Bryant', 'bryanta', 'Berea Buddies', 'Tutoring'), 
+                                                                                     ('Sreynit Khatt', 'khatts', 'Adopt-a-Grandparent', 'Adoption 101'), 
+                                                                                     ('Sreynit Khatt', 'khatts', 'Adopt-a-Grandparent', 'Meet & Greet with Grandparent'), 
+                                                                                     ('Madinabonu Solijonova', 'solijonovam', 'Test Program', 'Test Event')]
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_getUniqueVolunteers():
+    with mainDB.atomic() as transaction:
+        assert list(getUniqueVolunteers("2021-2022").execute()) == ([('bryanta', 'Alex Bryant', 'B00708826'),
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+                                                                    ('khatts', 'Sreynit Khatt', 'B00759107')])
+    
+        assert list(getUniqueVolunteers("2020-2021").execute()) == ([('ayisie', 'Ebenezer Ayisi', 'B00739736'),
```
bledsoef: These are at risk for failing due to comparing two lists.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,366 @@
...
+                    major = 'Agriculture',
+                    classLevel = 'Sophomore')
+        testEvent = Event.create(name="Test Event",
+                                 term=1,
+                                 program=1)
+        EventParticipant.create(user = 'solijonovam',
+                                event = testEvent,
+                                hoursEarned = 1)
+        
+        assert list(getUniqueVolunteers("2020-2021").execute()) == [('ayisie', 'Ebenezer Ayisi', 'B00739736'),
```
bledsoef: These are at risk for failing due to comparing two lists.

bledsoef (changes requested):
Better safe than sorry, try to make sure any time we compare two lists they are both sorted so we never get weird errors.

tests/code/test_spreadsheet.py
andersoncedu: These are sorted a specific way with the query

Karina-Agliullova:
Consider this comment https://github.com/BCStudentSoftwareDevTeam/celts/pull/1102#issuecomment-1922160078

app/logic/spreadsheet.py
```diff
@@ -137,31 +139,31 @@ def termParticipation(termDescription):
```
hoerstl, line 140: I know this wasn't you guys but since a left_outer join here would include all events regardless of whether or not they have event participants, this is returning non-null events with the null participants attached to them. I think if we remove this `JOIN.LEFT_OUTER` condition then it will default to an inner join and resolve us getting null participants.

If this change is made we should consider whether the function removeNullParticipants is necessary and probably remove it from everywhere its used.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+def test_removeNullParticipants():
+    testInputList = ['khatts']
+    assert removeNullParticipants(testInputList) == ['khatts']
+    testInputList = ['khatts', '', 'ayisie']
+    assert removeNullParticipants(testInputList) == ['khatts', 'ayisie']
+
+@pytest.mark.integration
+def test_termParticipation():
+    with mainDB.atomic() as transaction:    
+        assert termParticipation('Fall 2020') == {'Adopt-a-Grandparent': ['khatts'], 'CELTS-Sponsored Event': [None]}
```
hoerstl, line 41: Once you change the left_outer join to an inner join you'll need to recalculate this dictionary for testing. It shouldn't have a `'CELTS-Sponsored Event'` key at all after that.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+
+        EventParticipant.create(user = 'partont',
+                                event = 10,
+                                hoursEarned = 1)
+        termParticipationResult = termParticipation('Fall 2020')
+        for program in termParticipationResult:
+            termParticipationResult[program].sort()
```
hoerstl: maybe:
`for participantList in termParticipationResult.values():`
`    participantList.sort()`
Would be cleaner

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+        assert termParticipationResult == {'Adopt-a-Grandparent': ['khatts', 'partont'], 'CELTS-Sponsored Event': [None]}
+
+        EventParticipant.create(user = 'ayisie',
+                                event = 14,
+                                hoursEarned = 1)
+        termParticipationResult = termParticipation('Fall 2020')
+        for program in termParticipationResult:
+            termParticipationResult[program].sort()
```
hoerstl: maybe:
`for participantList in termParticipationResult.values():`
`    participantList.sort()`
Would be cleaner

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                                 program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent,
+                                hoursEarned=1)
+        assert sorted(getRetentionRate("2020-2021")) == [('Adopt-a-Grandparent', '0.0%'), 
+                                                         ('CELTS-Sponsored Event', '0.0%'), 
+                                                         ('Test Program', '0.0%')]
+        
+        testEvent2 = Event.create(name="Test Event2",
+                                  term=2, #Fall 2021
```
hoerstl: This comment is inaccurate and doesn't make any sense if it were correct. It should read #Spring 2021
It would also be better to rename this event something like `spring2021event`

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                    email = 'solijonovam@berea.edu',
+                    phoneNumber = '732-384-3469',
+                    firstName = 'Madinabonu',
+                    lastName  = 'Solijonova',
+                    isStudent = True,
+                    major = 'Agriculture',
+                    classLevel = 'Sophomore')
+        testProgram = Program.create(programName = "Test Program",
+                                     programDescription = "A good program")
+        testEvent = Event.create(name="Test Event",
```
hoerstl: I think it would be clearer to name this event to something like `fall2020event` so it's clear that you're putting someone in from the fall and showing the difference in retention when they stay in the spring

app/logic/spreadsheet.py
```diff
@@ -115,53 +117,53 @@ def repeatVolunteers():
     
     return repeatAllProgramQuery.tuples()
 
-def getRetentionRate():
+def getRetentionRate(academicYear):
     retentionDict = []
```
hoerstl: This is atrocious and it hurts my soul to see this.  (Please change this variable's name to imply a list)

hoerstl: Could you also write a docstring for this function that explains what this function returns. I'm usually not the first to say that they need it but the return type of this function is kind of unintuitive.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                                 term=1, #Fall 2020
+                                 program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent,
+                                hoursEarned=1)
+        assert sorted(list(repeatVolunteers().execute())) == [('Sreynit Khatt', 4), ('Zach Neill', 2)]
```
hoerstl: the `.execute()` here is unnecessary if we're going to cast it to a list anyway.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+        testEvent2 = Event.create(name="Test Event2",
+                                 term=2, #Fall 2021
```
hoerstl: Same notes as above about renaming the variable. Also, this comment is incorrect. In the testing data, term 2 is Spring 2021.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                                 program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent2,
+                                hoursEarned=1)
+        assert sorted(list(repeatVolunteers().execute())) == [('Madinabonu Solijonova', 2), ('Sreynit Khatt', 4), ('Zach Neill', 2)]
+        transaction.rollback()
```
hoerstl, line 140: Can you add in some more testing here that ensures that you can get someone else as a repeat volunteer if they attend two events from different programs in the same fall term. I don't think you need to add people to both a fall and spring event at all to get results from this function so it's making me confused about the goal of this function the way its written now.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+        testEvent = Event.create(name="Test Event",
+                                 term=1, #Fall 2020
+                                 program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent,
+                                hoursEarned=1)
+        assert sorted(list(repeatVolunteersPerProgram().execute())) == [('Sreynit Khatt', 'Adopt-a-Grandparent', 3),
+                                                                        ('Zach Neill', 'Hunger Initiatives', 2)]
+        testEvent2 = Event.create(name="Test Event2",
+                                 term=2, #Fall 2021
```
hoerstl: Incorrect comment again

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                    classLevel = 'Sophomore')
+        testProgram = Program.create(programName = "Test Program",
+                                     programDescription = "A good program")
+        testEvent = Event.create(name="Test Event",
+                                 term=1, #Fall 2020
+                                 program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent,
+                                hoursEarned=1)
+        assert sorted(list(repeatVolunteersPerProgram().execute())) == [('Sreynit Khatt', 'Adopt-a-Grandparent', 3),
```
hoerstl: We don't need the `.execute()` here if we're casting to a list

tests/code/test_spreadsheet.py
hoerstl: You can also just add Madina to an event in the same term

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                                                                        ('Zach Neill', 'Hunger Initiatives', 2)]
+        testEvent2 = Event.create(name="Test Event2",
+                                 term=2, #Fall 2021
+                                 program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent2,
+                                hoursEarned=1)
```
hoerstl, line 172: I would like Madina first added to an event outside of the testing program to show that this function behaves differently than simple repeat volunteers. It should, in fact, only add those who have volunteered more than once in a specific program. We need more tests that show that only those who volunteer more than once in the same program are selected by this function.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+        assert sorted(list(repeatVolunteersPerProgram().execute())) == [('Madinabonu Solijonova', 'Test Program', 2),
+                                                                        ('Sreynit Khatt', 'Adopt-a-Grandparent', 3),
+                                                                        ('Zach Neill', 'Hunger Initiatives', 2)]
+
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_volunteerMajorAndClass():
+    with mainDB.atomic() as transaction:
+        assert list(volunteerMajorAndClass(User.major).execute()) == [('Biology', 1), ('Chemistry', 1), ('Computer Science', 2), ('Psychology', 1)]
```
hoerstl: We don't need a `.execute()` if we're casting to a list.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                    lastName  = 'Solijonova',
+                    isStudent = True,
+                    major = 'Agriculture',
+                    classLevel = 'Sophomore')
+        EventParticipant.create(user = 'solijonovam',
+                                event = 2,
+                                hoursEarned = 2)
+        assert list(volunteerMajorAndClass(User.major).execute()) == [('Agriculture', 1), ('Biology', 1), ('Chemistry', 1), ('Computer Science', 2), ('Psychology', 1)]
+        assert list(volunteerMajorAndClass(User.classLevel).execute()) == [('Junior', 1), ('Senior', 3), ('Sophomore', 2)]
+        assert list(volunteerMajorAndClass(User.classLevel, True).execute()) == [('Sophomore', 2), ('Junior', 1), ('Senior', 3)]
```
hoerstl: We don't need `.execute()`s on any of these.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                                hoursEarned=3
+                                )
+        testEvent3 = Event.create(name="Test Event",
+                                  term=4,
+                                  program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent3,
+                                hoursEarned=5
+                                )
+        assert list(volunteerHoursByProgram().execute()) == [('Test Program', 10.0)]
```
hoerstl: I suppose this isn't a problem but because of python's typing, 10.0 == 10 so you really don't need to specify it as a float.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+        User.create(username = 'solijonovam',
+                    bnumber = 'B00769465',
+                    email = 'solijonovam@berea.edu',
+                    phoneNumber = '732-384-3469',
+                    firstName = 'Madinabonu',
+                    lastName  = 'Solijonova',
+                    isStudent = True,
+                    major = 'Agriculture',
+                    classLevel = 'Sophomore')
+        testEvent = Event.create(name="All Volunteer Training",
```
hoerstl: You could rename this variable to make it more clear that it's an all volunteer training. Something like `allVolunteerEvent`.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                                 isTraining=1,
+                                 isAllVolunteerTraining=1)
+        EventParticipant.create(user = 'solijonovam', #Not participated in event
+                                event = testEvent, #Added to all volunteer training event
+                                hoursEarned = 1)
+        assert list(onlyCompletedAllVolunteer("2020-2021").execute()) == [('solijonovam', 'Madinabonu Solijonova')]
+        testEvent2 = Event.create(name="Test Event",
+                                  program=1,
+                                  term=1)
+        EventParticipant.create(user = 'solijonovam', #Not participated in event
```
hoerstl: Change this comment to say something like 
`#Only participated in a single all volunteer event`
since at this point they have participated in an event and you're about to put them in another

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                                event = testEvent2, #Added to all volunteer training event
```
hoerstl: Remove this comment since `testEvent2` is not an all volunteer training event

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+
+@pytest.mark.integration
+def test_volunteerProgramHours():
+    with mainDB.atomic() as transaction:
+        EventParticipant.delete().execute()
+        assert sorted(list(volunteerProgramHours().execute())) == ([])
+        EventParticipant.create(user = 'qasema',
+                                event = 2,
+                                hoursEarned = 1)
+        assert sorted(list(volunteerProgramHours().execute())) == ([('Hunger Initiatives', 'qasema', 1.0)])
```
hoerstl: no need for the `.execute()`

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                                event = testEvent2, #Added to all volunteer training event
+                                hoursEarned = 1)
+        assert list(onlyCompletedAllVolunteer("2020-2021").execute()) == []
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_volunteerProgramHours():
+    with mainDB.atomic() as transaction:
+        EventParticipant.delete().execute()
+        assert sorted(list(volunteerProgramHours().execute())) == ([])
```
hoerstl: no need for the `.execute()` on line 254. you do need it on 253

tests/code/test_spreadsheet.py
hoerstl: Also, remove the parentheses surrounding `([('Hunger Initiatives', 'qasema', 1.0)])`. It doesn't do anything yet just makes me a bit confused about the data type.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+    with mainDB.atomic() as transaction:
+        EventParticipant.delete().execute()
+        # Empty the EventParticipant table and check to make sure hours displayed as 'None'
+        assert list(totalVolunteerHours().execute()) == [(None,)]
+        # Adding 1 volunteer hour
+        EventParticipant.create(user = 'qasema',
+                                event = 2,
+                                hoursEarned = 1)
+        # Checking that the total volunteer hours has increased by 1
+        assert list(totalVolunteerHours().execute()) == [(1.0,)]
```
hoerstl: no need for the `.execute()`

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                                hoursEarned = 1)
+        assert sorted(list(volunteerProgramHours().execute())) == ([('Hunger Initiatives', 'qasema', 1.0)])
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_totalVolunteerHours():
+    with mainDB.atomic() as transaction:
+        EventParticipant.delete().execute()
+        # Empty the EventParticipant table and check to make sure hours displayed as 'None'
+        assert list(totalVolunteerHours().execute()) == [(None,)]
```
hoerstl: no need for the `.execute()`

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+        # Adding 1 volunteer hour
+        EventParticipant.create(user = 'qasema',
+                                event = 2,
+                                hoursEarned = 1)
+        # Checking that the total volunteer hours has increased by 1
+        assert list(totalVolunteerHours().execute()) == [(1.0,)]
+        transaction.rollback()
```
hoerstl, line 313: I'd like to see a little more on this test where someone else is added to a different event and it's shown that their hours are also added to the total volunteer hours.

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+
+@pytest.mark.integration
+def test_getVolunteerProgramEventByTerm():
+    with mainDB.atomic() as transaction:
+        assert list(getVolunteerProgramEventByTerm(Term.get_by_id(3)).execute()) == ([])
+        assert sorted(list(getVolunteerProgramEventByTerm(Term.get_by_id(2)).execute())) == ([('Ebenezer Ayisi', 'ayisie', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'),
```
hoerstl: no need for the `.execute()`

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                                                                                             ('Sreynit Khatt', 'khatts', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'),
+                                                                                             ('Tyler Parton', 'partont', 'Hunger Initiatives', 'Hunger Hurts'),
+                                                                                             ('Zach Neill', 'neillz', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'), 
+                                                                                             ('Zach Neill', 'neillz', 'Hunger Initiatives', 'Hunger Hurts')])
+
+        assert sorted(list(getVolunteerProgramEventByTerm(Term.get_by_id(4)).execute())) == [('Alex Bryant', 'bryanta', 'Berea Buddies', 'Tutoring'),
```
hoerstl: no need for the `.execute()`

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                                  term=2,
+                                  program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent2)
+        testEvent3 = Event.create(name="Test Event",
+                                  term=4,
+                                  program=testProgram)
+        EventParticipant.create(user='solijonovam',
+                                event=testEvent3)
+        assert list(getVolunteerProgramEventByTerm(Term.get_by_id(3)).execute()) == [('Madinabonu Solijonova', 'solijonovam', 'Test Program', 'Test Event')]
```
hoerstl: no need for the `.execute()`

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+        assert sorted(list(getVolunteerProgramEventByTerm(Term.get_by_id(2)).execute())) == [('Ebenezer Ayisi', 'ayisie', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'),
```
hoerstl: no need for the `.execute()`

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+        assert sorted(list(getVolunteerProgramEventByTerm(Term.get_by_id(2)).execute())) == [('Ebenezer Ayisi', 'ayisie', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'), 
+                                                                                             ('Madinabonu Solijonova', 'solijonovam', 'Test Program', 'Test Event'),
+                                                                                             ('Sreynit Khatt', 'khatts', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'),
+                                                                                             ('Tyler Parton', 'partont', 'Hunger Initiatives', 'Hunger Hurts'), 
+                                                                                             ('Zach Neill', 'neillz', 'Hunger Initiatives', 'Empty Bowls Spring Event 1'), 
+                                                                                             ('Zach Neill', 'neillz', 'Hunger Initiatives', 'Hunger Hurts')]
+
+        assert sorted(list(getVolunteerProgramEventByTerm(Term.get_by_id(4)).execute())) == [('Alex Bryant', 'bryanta', 'Berea Buddies', 'Tutoring'),
```
hoerstl: no need for the `.execute()`

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                                                                                             ('Sreynit Khatt', 'khatts', 'Adopt-a-Grandparent', 'Meet & Greet with Grandparent')]
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_getUniqueVolunteers():
+    with mainDB.atomic() as transaction:
+        assert sorted(list(getUniqueVolunteers("2021-2022").execute())) == ([('bryanta', 'Alex Bryant', 'B00708826'),
+                                                                             ('khatts', 'Sreynit Khatt', 'B00759107')])
+    
+        assert sorted(list(getUniqueVolunteers("2020-2021").execute())) == ([('ayisie', 'Ebenezer Ayisi', 'B00739736'),
```
hoerstl: no need for the `.execute()`

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+        assert sorted(list(getVolunteerProgramEventByTerm(Term.get_by_id(4)).execute())) == [('Alex Bryant', 'bryanta', 'Berea Buddies', 'Tutoring'), 
+                                                                                             ('Madinabonu Solijonova', 'solijonovam', 'Test Program', 'Test Event'),
+                                                                                             ('Sreynit Khatt', 'khatts', 'Adopt-a-Grandparent', 'Adoption 101'), 
+                                                                                             ('Sreynit Khatt', 'khatts', 'Adopt-a-Grandparent', 'Meet & Greet with Grandparent')]
+        transaction.rollback()
+
+@pytest.mark.integration
+def test_getUniqueVolunteers():
+    with mainDB.atomic() as transaction:
+        assert sorted(list(getUniqueVolunteers("2021-2022").execute())) == ([('bryanta', 'Alex Bryant', 'B00708826'),
```
hoerstl: no need for the `.execute()`

tests/code/test_spreadsheet.py
```diff
@@ -0,0 +1,360 @@
...
+                    major = 'Agriculture',
+                    classLevel = 'Sophomore')
+        testEvent = Event.create(name="Test Event",
+                                 term=1,
+                                 program=1)
+        EventParticipant.create(user = 'solijonovam',
+                                event = testEvent,
+                                hoursEarned = 1)
+        
+        assert sorted(list(getUniqueVolunteers("2020-2021").execute())) == [('ayisie', 'Ebenezer Ayisi', 'B00739736'),
```
hoerstl: no need for the `.execute()`

hoerstl (dismissed):
Mostly good! Just a bunch of unnecessary `.executes()` and some incorrect/misleading comments but good job. I would like to see some more testing on the test_totalVolunteerHours() but otherwise great work!

app/logic/spreadsheet.py
andersoncedu: I think that tracking events that don't have participants is important as well though, if this were implemented they would be excluded.

tests/code/test_spreadsheet.py
andersoncedu: Again, I think the left outer join is important.

azabeli (commented):
It looks good to me, I would suggest adding some doc string for some of the function definitions.

tests/code/test_spreadsheet.py
bledsoef: I agree with the first part of what Lawrence is saying which doesn't look like is implemented. Would be great if we could a test verifying it works across programs.

bledsoef (dismissed):
I looked through everything and it looks good. If you make this one change I will go ahead and approve it. Nice work!

AndersonStettner (approved):
LGTM! - Finn

PR #1107, opened 2023-12-22
1053 Improve logging for import_users.py

### LINKED ISSUE #1053: Write `import_user.py` output to a file.
When `import_user.py` is run, we save the result to the file `cron.log`. However, the problem is that the output we are providing is very poor. We need to modify the output that is being written to the log to be more descriptive and actually say what the script is doing. That way, if it breaks, or malfunctions we can know where.

If you are curious how the output is being written to a new file, reference line 35 in `crontab`.

daizenj:
Summary of changes:
- fix #1053 
- Updated `import_user.py` script to use the `logging` module instead of print statements.
- When the script is run, log output is directed to both the terminal and a file named `import_users_log.txt`.

Test:
- Switch to the `production` branch.
- Run the `import_user.py` script.
- Verify that an `import_users_log.txt` file is created.
- Check that the content of the `import_users_log.txt` file matches the output from running the `import_user.py` script.

Please note that this change hasn't been tested on prod as it clearly states:
"DO NOT RUN THIS SCRIPT ON PRODUCTION UNLESS YOU REALLY REALLY KNOW WHAT YOU ARE DOING"
PS: I do not know what I am doing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ app/static/files/
 local-override.yml
 geckodriver.log
 .coverage
+import_users_log.txt
+
--- a/app/scripts/import_users.py
+++ b/app/scripts/import_users.py
@@ -1,3 +1,5 @@
+import logging
+import sys
 import pyodbc
 from ldap3 import Server, Connection, ALL
 import peewee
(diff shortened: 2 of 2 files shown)


bledsoef:
@DaizeNjounkeng Unfortunately, there seemed to be a miscommunication on our end and the issue description was incorrect. So while your functionality does just what the issue asked, the issue description was just corrected so this functionality is no longer necessary. Feel free to reference the new issue and make modifications, or close the PR and a member of the SSDT will pick it up. Sorry about that!

bledsoef:
No progress has been made and a new direction is needed, closing this PR.

PR #1109, opened 2023-12-27
Fix lingering email validity message

### LINKED ISSUE #1078: Issue editing emergency contact email address
There is an issue with entering an email address. I tried to enter random letters without providing @ and whatever goes after like @gmail.com. It asked me to enter a valid email address and did not let me save it, which is good. However, right after that message, I try to enter a valid email address with @gmail.com or @berea.edu, and it doesn't save the real email address either. It keeps telling me that the email address is not valid even when I enter my actual email address.
Screenshot (1470)

_Originally posted by @Karina-Agliullova in https://github.com/BCStudentSoftwareDevTeam/celts/issues/1058#issuecomment-1708815902_

daizenj:
### Issue ###
The error message `Please enter a valid email address` lingered even after putting a valid email address.
image

### Fix ###
fix #1078 
- Refactored code that was causing this lingering error message: The `formIsValid` function should not have been setting the custom validity for the `emergencyContactEmail` field within its scope, as the custom validity for this field is already handled separately in the `$('#emergencyContactEmail').on('input', ...)` block.
- Added regex to validate email address.

### Test ###

- Login as a student. 
- Go to My Profile
- From the `Choose an Action drop-down menu`, select `Edit Emergency Contact`
- Fill in valid values for every other field except the email address. 
- Fill in an incorrect email address an ensure that you are prompted to **"Please enter a valid email address"**
- Try saving and ensure that this error pesists. 
- Now fix that email address field with a typical valid filed e.g ...@gmail.com or ...@berea.edu
- Try saving and ensure that saving is successful.

### DIFF
--- a/app/static/js/emergencyContactInfo.js
+++ b/app/static/js/emergencyContactInfo.js
@@ -1,12 +1,23 @@
 $(document).ready(function(){
-
-    $('input.phone-input').inputmask('(999)-999-9999')
-
+    
     $('#emergencyContactEmail').on('input', function(){
-        if($(this).val() == '') {
-            this.setCustomValidity('')
+        let emailAddress = $(this).val();
+        if (emailAddress === '') {
+            this.setCustomValidity('');
+        } else if (!isValidEmailAddress(emailAddress)) {
+            this.setCustomValidity('Please enter a valid email address.');
+            this.reportValidity();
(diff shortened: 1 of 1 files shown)


bledsoef (approved):
Looks good to me, great work!

Issue #1059, opened 2023-08-30
Create course participants UI 


AndersonStettner:

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Create course participants UI #1059](https://trello.com/c/ujuECWK6/696-create-course-participants-ui-1059)

BrianRamsay:
We have added the functionality to submit course participants and import courses for the registrar but there is no way to access that information outside of the database.

We need to:
- See the imported courses and renew or edit (likely with the same kind of view we use for in-progress and approved courses)
- See the course participants attached to a course
- Set the hours that a course gives the participants
- Make sure that everything shows up in the Service Transcript nicely

hoerstl:
We also need to add the ability to alter any information that is necessary for renewing a course as usual:
![image](https://github.com/BCStudentSoftwareDevTeam/celts/assets/89316750/7e592f5f-34d6-4afe-b1db-2311a7cec37e)
![image](https://github.com/BCStudentSoftwareDevTeam/celts/assets/89316750/f24ad34d-0c59-4a49-8af6-1174f1763c85)

bledsoef:
Closed in PR #1111

PR #1111, opened 2024-01-04
Creation of UI for Imported Courses

stevensonmichel:
**Description**
We have added the functionality to submit course participants and import courses for the registrar but there is no way to access that information outside of the database.

**Issue**
Once a course is imported using a CSV type file from the registrar, there are minimal information about that course, as the file is expected to have only the Course Abbreviation, the Term, and the Students taking that course. In that case, a more thorough access such as editing, viewing and printing some information such as Course Name, Course Credit, and Course Instructors is needed. 

**Solution**
Fixes Issue #1059 

- For that purpose, a new tab called Imported Courses was created in the Course Management Page. That tab holds all the imported courses from the registrar, while the Native Courses tab displays the Approved and Unapproved Courses
- Then, the imported courses have some of the native courses functionalities such as View, Print, Edit. 
View and Print leads to the normal Course Proposal page and Printing Page, while triggering Edit opens a modal in which the user can edit the Name, the Credit, as well as the Instructors for that course. The Course Abbreviation field is disabled for unicity  purposes. 
- If a user chooses to edit a course that already possesses preexisting information, those information should be viewable in the modal, and changes will be made according to the user will. 
- Once information are recorded, data is sent to the backend, and that specific course data is updated in the Database. 

**Files Changed or Added**
- `courseManagement.py`
- `routes.py in admin`
- `slcManagement.js`
- `manageServiceLearningCourses.html`
- `test_course_management.py`
- `InstructorTable.js` ((file added)

**Tests**
- Initial setup of environment, followed by activation of the virtual environment using `source setup.sh` command
- Run `database/reset_database.sh` test to reset the test data
- Run flask run to start the Celts app
- On the navigation bar, click on Admin, then Course Management
- Click on Import Courses Participants, upload the `parseUpload_ValidTest.xlsx` file from `celts/tests/files`
- Click on Preview and then Save
- After, navigate in the tab Imported Courses. If a term has imported courses, those will be shown in a similar layout as Approved and Unapproved Courses. 
- Perform multiple actions, such as View, Print, and Edit
- For Edit, check on the Database (MySQL Workbench) to see if the course was updated with the correct data. 

**Remarks**
- Removing, adding, or submitting the same instructors seem to be a challenge, even in other areas such as Adding a New Course. This solution used some of the preexisting structures of manipulating instructors for Approved/Unapproved courses.
- The field instructor is allowed to be empty because that information is usually omitted when uploading the csv file.


azabeli:
Looks good to me, however, in the documentation I would suggest adding the fact that when testing, we need to be in a specific term(semester) the courses imported are in. Otherwise, the course won't be seen. 

Regarding the remove button on the course proposal, it only works for existing proposals but for the new ones it does not. 
Screenshot (168)

app/logic/courseManagement.py
```diff
@@ -84,6 +136,10 @@ def updateCourse(courseData, attachments=None):
             instructorList = []
             if 'instructor[]' in courseData:
                 instructorList = courseData.getlist('instructor[]')
+                print("--------------------------------")
+                print(instructorList)
+                print(courseData)
+                print("--------------------------------")
```
BrianRamsay: DEBUG MESSAGES!!

azabeli:
Screenshot (176)
The name of Hours Earned on Completion should be changed to "Service Hours" and the input box should not allow the letters input but float or integers!

Fixed

app/templates/admin/manageServiceLearningFaculty.html
```diff
@@ -111,11 +146,139 @@ <h5 class="ms-3">There are no approved courses for {{ term.description }}.</h5>
...
+          <td>{{course.courseName or (course.courseAbbreviation + " (name unspecified)")}}</td>
+          <td>{{course.instructors}}</td>
+          <td>
+            <a class="status-key text-dark" id="statusKey" href="javascript:void(0);" data-toggle="popover"
+              title="Status Key" data-content="{{course.status.status}}">{{course.status.status}}</a>
+          </td>
+          <td>
+            <select class="form-select" id="{{course.id}}" onchange='changeAction(this)'>
+              <option value="---" disabled selected>Select Action</option>
+              <option value="Alter">Edit</option>
```
bledsoef: So I know that there is already an Edit value somewhere else so this value can't be Edit as well, but I think having the actions Edit and Alter do different things but be synonyms may be a little confusing. Is there a better name we could give this that is more descriptive?

app/templates/admin/manageServiceLearningFaculty.html
```diff
@@ -111,11 +146,139 @@ <h5 class="ms-3">There are no approved courses for {{ term.description }}.</h5>
...
+                  <tr>
+                    <td class="fw-bold">Course Name</td>
+                    <td><input class="form-control" id="courseName" name="courseName"></td>
+                  </tr>
+                  <tr>
+                    <td class="fw-bold"> Course Abbreviation</td>
+                    <td><input class="input form-control" id="courseAbbreviation" disabled></input></td>
+                  </tr>
+                  <tr>
+                    <td class="fw-bold">Hours Earned on Completion</td>
```
bledsoef: Directing you to what Azabe was talking about in her comment. This should be changed to something like Service Hours and shouldn't be attached to Credits. Those things are independent of each other. May need to talk to Brian more to figure out exactly how the functionality should work.

app/logic/courseManagement.py
```diff
@@ -84,6 +136,10 @@ def updateCourse(courseData, attachments=None):
             instructorList = []
             if 'instructor[]' in courseData:
                 instructorList = courseData.getlist('instructor[]')
+                print("--------------------------------")
```
bledsoef: Remove debug statements.

app/static/js/instructorTable.js
```diff
@@ -0,0 +1,79 @@
...
+    // Create new table row and update necessary attributes
+    let lastRow = tableBody.find("tr:last");
+    let newRow = lastRow.clone();
+  
+    let instructorName = newRow.find("td:eq(0) p")
+    instructorName.text(instructor);
+  
+    let phoneInput = newRow.find("td:eq(0) input")
+    phoneInput.val(phone);
+    console.log("Phone input is", phoneInput.val(phone))
```
bledsoef: Remove debug statement.

app/static/js/slcManagement.js
```diff
@@ -140,3 +144,102 @@ function unapproveProposal(el){
...
+              </tr>`;
+  return row;
+}
+
+
+function getCourseInstructors() { // this function gets usernames out of the table rows from editButton class and transform the object into an array
+  var instructorUsernames = $("#instructorTableBody tr").map(function() {
+      return $(this).find('.editButton').data('username');
+  }).get();
+  console.log("Mesdames et Messieurs", instructorUsernames)
```
bledsoef: Remove debug statement.

app/controllers/admin/routes.py
```diff
@@ -407,6 +412,44 @@ def removeFromSession():
 
     return ""
 
+
+@admin_bp.route('/manageServiceLearning/imported/<courseID>', methods = ['POST', 'GET'])
+def alterImportedCourse(courseID):
+    """
+    Goals:
+    This route is meant to be called via ajax from the course management page in order to fill out the scanty information
+    we possess on imported courses so that they can edit them in our modal.
```
bledsoef: Modify this comment to be more concise and talking about what the code is actually doing instead of what it should be doing.

app/controllers/admin/routes.py
```diff
@@ -407,6 +412,44 @@ def removeFromSession():
...
+    POST: on a post we should update their course entry into our course table and update all of the previously listed information
+    
+    BEANS: This is a big one, we need to abstract most of this functionality into a logic file and test it!!!
+    """
+    if request.method == 'GET':
+        try:
+            targetCourse = Course.get_by_id(courseID)
+            targetInstructors = CourseInstructor.select().where(CourseInstructor.course == targetCourse)
+            courseData = model_to_dict(targetCourse, recurse=False)
+            instructors = [model_to_dict(ci.user) for ci in targetInstructors]
```
bledsoef: Nit picky but ci is not a great variable name for this comprehension. Also maybe change the variable name instructors to better represent what these instructors actually are. courseInstructors maybe?

app/logic/courseManagement.py
```diff
@@ -45,6 +47,56 @@ def approvedCourses(termId):
...
+    """
+        This function will take in courseData for the SLC proposal page and a dictionary
+        of instructors assigned to the imported course after that one is edited 
+        and update the information in the db.
+    """
+
+    with mainDB.atomic() as transaction:
+        try:
+            course = Course.get_by_id(courseData["courseId"])
+            (Course.update(
```
bledsoef: Channeling my inner Anderson to tell you to make the indentations all on the same line for this query.

app/logic/courseManagement.py
```diff
@@ -45,6 +47,56 @@ def approvedCourses(termId):
...
+                        .join(User, JOIN.LEFT_OUTER).switch(Course)
+                        .join(CourseStatus).switch(Course)
+                        .join(Term)
+                        .where(Term.id == termId, Course.status == CourseStatus.IMPORTED)
+                        .group_by(Course, Term, CourseStatus))
+
+    return importedCourses
+
+
+def editImportedCourses(courseData, attachments=None):
```
bledsoef: Can we try and add comments to this function. Some of the code in here isn't necessarily intuitive.

bledsoef (changes requested):
Will finish my review later.

app/logic/courseManagement.py
stevensonmichel: Fixed.

app/static/js/instructorTable.js
stevensonmichel: Fixed

app/static/js/slcManagement.js
stevensonmichel: Fixed

app/logic/courseManagement.py
stevensonmichel: Fixed.
Don't let him see. It would be too much of a praise.

app/logic/courseManagement.py
stevensonmichel: Fixed

app/controllers/admin/routes.py
```diff
@@ -407,6 +412,44 @@ def removeFromSession():
...
+    we possess on imported courses so that they can edit them in our modal.
+    There are two paths, GET and POST
+    GET: On a get we should return a dictionary containing 
+    - course name, 
+    - course abbreviation, 
+    - hours earned on completion, 
+    - list of the instructors
+    POST: on a post we should update their course entry into our course table and update all of the previously listed information
+    
+    BEANS: This is a big one, we need to abstract most of this functionality into a logic file and test it!!!
```
azabeli: I do not understand how "BEANS" are relevant here.

azabeli (changes requested):

app/controllers/admin/routes.py
stevensonmichel: Fixed.

app/controllers/admin/routes.py
stevensonmichel: Fixed.

app/templates/admin/manageServiceLearningFaculty.html
stevensonmichel: Fixed

app/controllers/admin/routes.py
stevensonmichel: Fixed

app/templates/admin/manageServiceLearningFaculty.html
stevensonmichel: Fixed

app/controllers/admin/routes.py
```diff
@@ -407,6 +413,43 @@ def removeFromSession():
...
+    if request.method == 'GET':
+        try:
+            targetCourse = Course.get_by_id(courseID)
+            targetInstructors = CourseInstructor.select().where(CourseInstructor.course == targetCourse)
+            
+            courseData = model_to_dict(targetCourse, recurse=False)
+            serviceHours = list(CourseParticipant.select()
+                                             .where(CourseParticipant.course_id == targetCourse.id)
+                                             )[0].hoursEarned 
+
```
andersoncedu: Please correct the formatting.

andersoncedu:
Change modal by making title bold and add course abbreviation

andersoncedu:
Phone number is broken and route should go back to same term. Instructor/faculty should be called the same thing across modal and table.

app/static/js/manageServiceLearningFaculty.js
```diff
@@ -37,7 +39,7 @@ $(document).ready( function () {
   });
   
 
-  /******** Preview Events **************/
+  /************** Preview Events **************/
   $('#modalPreview button[data-bs-dismiss="modal"]').click(function () {
     $('#modalPreview').removeClass('show d-block');
```
andersoncedu: modalPreview is not very descriptive.

app/static/js/manageServiceLearningFaculty.js
```diff
@@ -62,7 +64,7 @@ $(document).ready( function () {
     })
   });
 
-  /******** Course Participant Stuff **************/
+  /************** Course Participant Stuff **************/
     $("#modalCourseParticipant").on("click", function () {
       $("#modalSubmit").modal("toggle");
```
andersoncedu: modalSubmit is not a good name.

app/templates/admin/manageServiceLearningFaculty.html
```diff
@@ -111,11 +146,139 @@ <h5 class="ms-3">There are no approved courses for {{ term.description }}.</h5>
...
+                          <th scope="col">Instructor</th>
+                          <th scope="col"></th>
+                        </tr>
+                      </thead>
+                      <tbody id="instructorTableBody">
+                          <tr data-username="" hidden>  <!-- This is a hidden table row. To add new rows to the instructor table, we copy this one and fill out the information -->
+                              <td>
+                                <p class="mb-0"></p>
+                                <input type="text" style="border: none" size="14" class="form-control-sm" id="" name="courseInstructorPhone" aria-label="Instructor Phone" data-value="" value="" placeholder="Phone Number" />
+                                <a class="text-decoration-none primary editButton" tabindex="0" data-username="" id=""  type="button">Edit</a>
```
andersoncedu: If you're not going to assign a value to id, value, etc., you probably don't need to include it here.

app/static/js/slcManagement.js
```diff
@@ -140,3 +144,101 @@ function unapproveProposal(el){
...
+  $('#alterModal form input[name="instructor[]"]').remove(); // Clear existing hidden instructor inputs
+  var instructordata = getCourseInstructors(); // Get current instructors from the table
+
+  // Append new hidden inputs for each instructor
+  instructordata.forEach(function(instructor) {
+      $('<input>').attr({
+          type: 'hidden',
+          name: 'instructor[]',
+          value: instructor
+      }).appendTo('#alterModal form');
```
andersoncedu: The formatting here is pretty weird.

app/controllers/admin/routes.py
stevensonmichel: Solved

app/static/js/slcManagement.js
stevensonmichel: It is to show that data are being added to a dictionary.

app/templates/admin/manageServiceLearningFaculty.html
andersoncedu: This is actually necessary for functionality, it doesn't work without it.

app/static/js/manageServiceLearningFaculty.js
stevensonmichel: Fixed

app/static/js/manageServiceLearningFaculty.js
stevensonmichel: Fixed

app/controllers/admin/routes.py
```diff
@@ -387,16 +393,32 @@ def manageServiceLearningCourses(term=None):
...
-                            courseInstructors = getInstructorCourses(),
-                            unapprovedCourses = unapprovedCourses(term),
-                            approvedCourses = approvedCourses(term),
-                            terms = selectSurroundingTerms(g.current_term),
-                            term = manageTerm,
-                            cpPreview= session.get('cpPreview',{}),
-                            cpPreviewErrors = session.get('cpErrors',[])
-                           )
+    courseID = session.get("alterCourseId")
+    print(courseID)
```
azabeli: remove print statement!

app/controllers/admin/routes.py
```diff
@@ -407,6 +429,47 @@ def removeFromSession():
...
+            
+            courseData = model_to_dict(targetCourse, recurse=False)
+            serviceHours = (list(CourseParticipant.select()
+                                                  .where(CourseParticipant.course_id == targetCourse.id))[0]
+                                                  .hoursEarned)
+            
+            courseInstructors = [model_to_dict(instructor.user) for instructor in targetInstructors]
+            courseData['instructors'] = courseInstructors
+            courseData["hoursEarned"] = serviceHours
+
```
azabeli, line 517: remove some unnecessary lines, ones that are empty!

app/static/js/slcManagement.js
```diff
@@ -140,3 +145,104 @@ function unapproveProposal(el){
...
+                <td class="align-middle">
+                  <button type="button" class="btn btn-danger removeButton">Remove</button>
+                </td>
+              </tr>`;
+  return row;
+}
+
+function emptyInstructorTable() {
+  saveRow = $("#instructorTableBody tr")[0];
+  console.log("The saved row is", saveRow)
```
azabeli: remove the print statement!

app/templates/admin/manageServiceLearningFaculty.html
```diff
@@ -77,11 +109,15 @@ <h5 class="ms-3">There are no unapproved courses for {{ term.description }}.</h5
     </table>
   {% endif %}
 
+
+
```
azabeli: remove the unnecessary lines!

azabeli (changes requested):

app/logic/courseManagement.py
```diff
@@ -45,6 +48,58 @@ def approvedCourses(termId):
 
     return approvedCourses
 
+
+def importedCourses(termId):
+    '''
+    Queries the database to get all the necessary information for 
+    imported courses.
+    '''
+    importedCourses = (Course.select(Course, Term, CourseStatus, fn.GROUP_CONCAT(" " ,User.firstName, " ", User.lastName).alias('instructors'))
```
bledsoef: Is there a better way to separate instructors than the first " " in the group_concat? It means when there is only one instructor it looks a little bit funky.

app/controllers/admin/routes.py
```diff
@@ -387,16 +393,32 @@ def manageServiceLearningCourses(term=None):
...
-                            approvedCourses = approvedCourses(term),
-                            terms = selectSurroundingTerms(g.current_term),
-                            term = manageTerm,
-                            cpPreview= session.get('cpPreview',{}),
-                            cpPreviewErrors = session.get('cpErrors',[])
-                           )
+    courseID = session.get("alterCourseId")
+    print(courseID)
+    if courseID:
+        session.pop("alterCourseId")
```
bledsoef: Here is a good place to add a comment, I personally am not familiar with session variables so explaining what you are doing here could be good.

app/logic/courseManagement.py
```diff
@@ -45,6 +48,58 @@ def approvedCourses(termId):
...
+                   .where(Course.id == course.id).execute()) # Update the data (Course Name) of the course in the database
+
+            (CourseParticipant.update(hoursEarned=courseData["hoursEarned"])
+                              .where(CourseParticipant.course_id == course.id).execute())
+            
+            instructorList = []
+            if 'instructor[]' in courseData:
+                instructorList = courseData.getlist('instructor[]') # Fetch the list of course instructors from CourseData
+            
+            if instructorList:
```
bledsoef: A little nitpick but this will never execute if the previous conditional does not run, so I think you should combine these conditionals in some way.

bledsoef (changes requested):
Looks good for the most part, just a couple small things between Azabe and I.

app/controllers/admin/routes.py
stevensonmichel: Fixed

app/controllers/admin/routes.py
stevensonmichel: Fixed

app/static/js/slcManagement.js
stevensonmichel: It is a console.log statement. Fixed

app/templates/admin/manageServiceLearningFaculty.html
stevensonmichel: Fixed

app/logic/courseManagement.py
stevensonmichel: Fixed.

app/logic/courseManagement.py
stevensonmichel: This format for concatenation is used across Approved Courses and Unapproved Courses. More information will be provided after consultation with Brian or Anderson.

app/controllers/admin/routes.py
andersoncedu: Added comment as suggested.

app/controllers/admin/routes.py
```diff
@@ -407,6 +432,45 @@ def removeFromSession():
...
+                                                  .hoursEarned)
+            
+            courseInstructors = [model_to_dict(instructor.user) for instructor in targetInstructors]
+            courseData['instructors'] = courseInstructors
+            courseData["hoursEarned"] = serviceHours
+
+            return jsonify(courseData)
+        
+        except Exception as e:
+            flash("Course not found or something else went wrong")
```
AndersonStettner: You never use `e` for anything so why not just `except Exception`?  Also, if you are just checking whether or not a course exists you should do `except DoesNotExist` to catch the specific exception you are expecting. Doing blanket exception handling without printing the exception is not great and can lead to very annoying debugging down the line.

Also, for the flash message, just say "Course not found" or "Could not find Course".

AndersonStettner (changes requested):

AndersonStettner (dismissed):

app/static/js/slcManagement.js
```diff
@@ -140,3 +145,103 @@ function unapproveProposal(el){
...
+  saveRow = $("#instructorTableBody tr")[0];
+  $("#instructorTableBody").empty().html(saveRow);
+}
+
+function getCourseInstructors() { // this function gets usernames out of the table rows from editButton class and transform the object into an array
+  var instructorUsernames = $("#instructorTableBody tr").map(function() {
+      return $(this).find('.editButton').data('username');
+  }).get();
+  return instructorUsernames;
+}
```
BrianRamsay: These functions are all duplicated with our instructor-adding code in the slc proposal, right? We need to see where we can re-use or replace existing code, rather than add new, very similar code.

BrianRamsay (changes requested):

bledsoef:
[Create course participants UI #1059](https://trello.com/c/ujuECWK6/696-create-course-participants-ui-1059)

app/static/js/slcManagement.js
stevensonmichel: Fixed

bledsoef:
Screenshot_20240322_145635
The hover on Imported status shows undefined which I believe is not the expected functionality.

app/controllers/admin/routes.py
```diff
@@ -432,16 +438,35 @@ def manageServiceLearningCourses(term=None):
...
+    # retrieve and store the courseID of the imported course from a session variable if it exists. 
+    # This allows us to export the courseID in the html and use it.
+    courseID = session.get("alterCourseId")  
+    
+    if courseID:
+        # delete courseID from the session if it was retrieved, for storage purposes.
+        session.pop("alterCourseId")  
+        return render_template('/admin/manageServiceLearningFaculty.html',
+                                courseInstructors = getInstructorCourses(),
+                                unapprovedCourses = unapprovedCourses(term),
```
bledsoef: I think in your merge conflict you may have overwritten a bug patch we put in. All calls using `term` as a parameter need to use `manageTerm` instead.

bledsoef (changes requested):
I am also having a problem where if you there is only one instructor and you attempt to delete it it will still be there.

app/controllers/admin/routes.py
bledsoef: Here and in the render_template below it

stevensonmichel: Fixed

app/static/js/slcManagement.js
```diff
@@ -140,3 +187,113 @@ function unapproveProposal(el){
...
+      }
+      
+      if (callback) callback();
+    } 
+  });
+}
+
+// Instructor manipulation functions
+
+function updateInstructorList(instructorData) { // This function fetches instructor usernames and attached the list of usernames to the form submission
```
AndersonStettner: The comment is not accurate to what the function does.

app/static/js/slcManagement.js
```diff
@@ -140,3 +187,113 @@ function unapproveProposal(el){
...
+  // Add each instructor to the table
+  instructors.forEach(function(instructor) {
+    var newRow = createInstructorRow(instructor);
+    $("#instructorTableBody").append(newRow);
+
+    $('<input>').attr({
+      type: 'hidden',
+      name: 'instructor[]',
+      value: instructor.username
+    }).appendTo('#instructorTableNames');
```
AndersonStettner: Why are you maintaining two different lists of the same information? the other list is on line 251. If you need to maintain two lists then you should have a function that does the list stuff and call it where you need it rather than having the same code in two separate spots.

app/templates/admin/manageServiceLearningFaculty.html
```diff
@@ -111,37 +145,162 @@ <h5 class="ms-3">There are no approved courses for {{ term.description }}.</h5>
...
+    {% endif %}
+    </div>
+  </div>
+  {% set downloadAndSend = "disabled" if approvedCourses | length == 0 else "" %} 
+
+    <!-- Modal for altering courses -->
+    <div class="modal fade" id="alterModal">
+      <div class="modal-dialog">
+        <div class="modal-content">
+          <h5 class="modal-header fw-bold" id="importedModalTitle"></h5>
```
AndersonStettner: Is this an `alterModal` or an `importedModal`?

AndersonStettner (changes requested):
When your trailing comments are long, put them in front of or after instead.

app/static/js/slcManagement.js
stevensonmichel: Fixed

app/templates/admin/manageServiceLearningFaculty.html
stevensonmichel: Fixed

Karina-Agliullova:
Screenshot (1760)
I have an issue with the title at the top of this page.

When I view the imported course, it takes me to tha page to create a new proposal. Technically, the data in it is correct regarding the course that is imported because it contains all the information and it is good that it is greyed out so that we can't edit it in the view mode. However, the title says "New Proposal for Service-Learning Course." The title is confusing because it is not a new course proposal that we are creating

gahimbaref:
Might not be a big deal but the user can put negative values for service hours in the edit modal.

Karina-Agliullova:
Screenshot (1762)
Since we are in the tab called "Imported Courses" and viewing the table called "Imported," is it really relevant to display status "Imported" on that table? Sounds like a lot of repetition. The name of the table "Imported" seems to already let the user know what is the status of the course, especially, since we cannot change or indicate any other status for these courses

Karina-Agliullova:
When we print the imported course, it lets me review the file that is about to get printed. The file displays the information correctly; however, it is missing the service hours. Don't they also need to be included into that file to keep track of the service hours?

app/static/js/instructorTable.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export function createNewRow(selectedInstructor) {
+  let instructor = (selectedInstructor["firstName"]+" "+selectedInstructor["lastName"]+" ("+selectedInstructor["email"]+")");
+  let username = selectedInstructor["username"];
+  let phone = selectedInstructor["phoneNumber"];
+  let tableBody = $("#instructorTable").find("tbody");
+
+  if(tableBody.prop('outerHTML').includes(instructor)){
+    msgFlash("Instructor is already added.", "danger");
```
Karina-Agliullova, line 15: For your modal, when I edit it and add instructors to the instructor's table, if I try to add an instructor that was already added to the table, it doesn't let me, which is good. However, the flash message "Instructor is already added." is displayed behind the modal, so I actually cannot see it, which is a bad user experience. I see only the corners of the flash message, and the rest is blurred because we are displayed with the modal. It would be nice to show the flash message on top of modal, so that the user can actually see it

stevensonmichel:
> Screenshot (1760) When I view the imported course, it takes me to tha page to create a new proposal. Technically, the data in it is correct regarding the course that is imported because it contains all the information and it is good that it is greyed out so that we can't edit it in the view mode. However, the title says "New Proposal for Service-Learning Course." The title is confusing because it is not a new course proposal that we are creating

The same thing can be observed for non Imported Courses, whether they are approved or not. So, a new issue will be created for that purpose.

app/static/js/instructorTable.js
stevensonmichel: Fixed

stevensonmichel:
> When we print the imported course, it lets me review the file that is about to get printed. The file displays the information correctly; however, it is missing the service hours. Don't they also need to be included into that file to keep track of the service hours?

Service Hours are not to be included in the file.

stevensonmichel:
> Screenshot (1762) Since we are in the tab called "Imported Courses" and viewing the table called "Imported," is it really relevant to display status "Imported" on that table? Sounds like a lot of repetition. The name of the table "Imported" seems to already let the user know what is the status of the course, especially, since we cannot change or indicate any other status for these courses

For consistency purpose, we keep that column marking Imported because these courses can be edited in Course Proposals. Then, their status would become In Progress.

stevensonmichel:
> Might not be a big deal but the user can put negative values for service hours in the edit modal.

That is a general issue, also observed for any other proposal. Chris has created an issue for that concern.

app/controllers/admin/routes.py
```diff
@@ -431,16 +437,35 @@ def manageServiceLearningCourses(term=None):
...
+        # delete courseID from the session if it was retrieved, for storage purposes.
+        session.pop("alterCourseId")  
+        return render_template('/admin/manageServiceLearningFaculty.html',
+                                courseInstructors = getInstructorCourses(),
+                                unapprovedCourses = unapprovedCourses(manageTerm),
+                                approvedCourses = approvedCourses(manageTerm),
+                                importedCourses = importedCourses(manageTerm),
+                                terms = selectSurroundingTerms(g.current_term),
+                                term = manageTerm,
+                                cpPreview= session.get('cpPreview',{}),
```
hoerstl: Please put a space after this keyword argument for consistency.

app/controllers/admin/routes.py
```diff
@@ -463,6 +488,42 @@ def removeFromSession():
 
     return ""
 
+@admin_bp.route('/manageServiceLearning/imported/<courseID>', methods = ['POST', 'GET'])
+@admin_bp.route('/manageServiceLearning/imported/<courseID>', methods = ['POST'])
```
hoerstl: I believe that the former line here is a superset of the latter and makes the second unnecessary.

app/controllers/admin/routes.py
```diff
@@ -463,6 +488,42 @@ def removeFromSession():
...
+    The POST request updates a specific imported course (course name, course abbreviation, 
+    hours earned on completion, list of instructors) in the database with new information 
+    coming from the imported courses modal. 
+    """
+    if request.method == 'GET':
+        try:
+            targetCourse = Course.get_by_id(courseID)
+            
+            targetInstructors = CourseInstructor.select().where(CourseInstructor.course == targetCourse)
+            serviceHours = list(CourseParticipant.select().where(CourseParticipant.course_id == targetCourse.id))[0].hoursEarned
```
hoerstl: I don't know if this would ever be the case for an imported course but I though I should note that if there exists an imported course that no one has taken before then this line will raise an error and prevent us from editing the imported course. 

I'd consider this one a pretty serious bug to fix

app/controllers/admin/routes.py
```diff
@@ -463,6 +488,42 @@ def removeFromSession():
...
+    The GET request provides preexisting information of an imported course in a modal. 
+    The POST request updates a specific imported course (course name, course abbreviation, 
+    hours earned on completion, list of instructors) in the database with new information 
+    coming from the imported courses modal. 
+    """
+    if request.method == 'GET':
+        try:
+            targetCourse = Course.get_by_id(courseID)
+            
+            targetInstructors = CourseInstructor.select().where(CourseInstructor.course == targetCourse)
```
hoerstl, line 507: Consider adding an order_by clause on this so that instructors appear in a predictable and perhaps alphabetical order by name. We don't sort this elsewhere.

app/controllers/admin/routes.py
```diff
@@ -463,6 +488,42 @@ def removeFromSession():
...
+            serviceHours = list(CourseParticipant.select().where(CourseParticipant.course_id == targetCourse.id))[0].hoursEarned
+            
+            courseData = model_to_dict(targetCourse, recurse=False)
+            courseData['instructors'] = [model_to_dict(instructor.user) for instructor in targetInstructors]
+            courseData["hoursEarned"] = serviceHours
```
hoerstl: You think you could use consistent quotation marks here?

app/logic/serviceLearningCourses.py
```diff
@@ -115,6 +115,21 @@ def approvedCourses(termId: int) -> List[Course]:
 
     return approvedCourses
 
+def importedCourses(termId: int) -> List[Course]:
```
hoerstl: Consider renaming this function `getImportedCourses` because it's a little ambiguous what this function does from its name.

app/static/js/instructorTable.js
```diff
@@ -0,0 +1,51 @@
+// Instructor manipulation functions
+// -------------------------------------
+
+export function getRowUsername(element) {
+    return $(element).closest("tr").data("username")
+}
+
+export function createNewRow(selectedInstructor) {
+  let instructor = (selectedInstructor["firstName"]+" "+selectedInstructor["lastName"]+" ("+selectedInstructor["email"]+")");
```
hoerstl: Use a formatted JS string here like so:
\`${selectedInstructor["firstName"] } ${selectedInstructor["lastName"]} (${selectedInstructor["email"]})\`

app/static/js/instructorTable.js
```diff
@@ -0,0 +1,51 @@
...
+
+  let removeButton = newRow.find("td:eq(1) button")
+  let editLink = newRow.find("td:eq(0) a")
+  editLink.attr("id", "editButton-" + username);
+
+  editLink.attr("data-username", username)
+  newRow.prop("hidden", false);
+  lastRow.after(newRow);
+
+  phoneInput.attr("data-value", phone)
```
hoerstl: This should really be a call like this to data:
`phoneInput.data('value', phone)`
This is kind of a roundabout way to change the data attribute named 'value'

app/static/js/instructorTable.js
```diff
@@ -0,0 +1,51 @@
...
+  var edit = "#editButton-" + username
+  var input = "#inputPhoneNumber-" + username
+  if (username){
+    setupPhoneNumber(edit, input)
```
hoerstl: I have some problems with the readability of lines 39, 40, and 42. Mostly, my problem is that you call them edit and input when they are not the objects themselves nor their ids. You should probably rename them to names like
`editSelector`
and
`inputSelector`

while you're at it, I'd love it if you could also rename the parameters of `setupPhoneNumber` in base.js to address the same naming concern.

app/static/js/instructorTable.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  $("#instructorTableNames").append('<input hidden name="instructor[]" value="' + username + '"/>')
```
hoerstl: Again, we can use a formatted JS string here:
``$("#instructorTableNames").append(`<input hidden name="instructor[]" value="${username}"/>`)``

app/static/js/slcManagement.js
```diff
@@ -24,9 +28,49 @@ $(document).ready(function() {
 
       } else if ($(this).attr('data-content') == "In Progress") {
         return "This proposal has not been submitted for review."
+
+      } else if ($(this).attr('data-content') == "Imported") {
+        return "This proposal has been submitted and is waiting for review."
+
       }
```
hoerstl, line 35: I suppose that this isn't bad to leave in here but ever since you removed the status column from the imported courses tab, this hover logic is no longer necessary. So we should probably remove this.

app/static/js/slcManagement.js
```diff
@@ -24,9 +28,49 @@ $(document).ready(function() {
...
+  $("#instructorTable").on("click", ".removeButton", function() {
+    $(this).closest("tr").find(".editButton").each(function() {
+      let username = $(this).data('username');
+  
+      // Remove input with matching username value
+      $("#instructorTableNames input[value='" + username + "']").remove();
+  
+      // Remove the closest tr
+      $(this).closest("tr").remove();
+  });
```
hoerstl: Please indent this one level to match its opening on line 41

app/static/js/slcManagement.js
```diff
@@ -24,9 +28,49 @@ $(document).ready(function() {
...
+  });
+
+  $("#courseInstructor").on('input', function() {
+      searchUser("courseInstructor", createNewRow, true, null, "instructor");
+      setTimeout(function() {
+        $(".ui-autocomplete").css("z-index", 9999);
+    }, 500);
```
hoerstl: Please indent this one level to match the opening indentation level on line 54

app/static/js/slcManagement.js
```diff
@@ -42,8 +86,8 @@ function updateRenewModal(courseID){
 }
 
 function changeAction(action){
-  courseID = action.id;
-  courseAction = action.value
+  let courseID = action.id;
+  let courseAction = action.value
```
hoerstl: Could we give both of these lines semicolons?

app/static/js/slcManagement.js
```diff
@@ -140,3 +187,111 @@ function unapproveProposal(el){
...
     })
 }
+
+/************** Imported Courses Modal Functions **************/
+function showAlterModalWithCourse(courseID) {
+  getImportedCourseInfo(courseID, function() {
+    $('#alterModal #alterCourseId').val(courseID);
+
+    $('#alterModal').modal('show');
+  });
```
hoerstl, line 196: Can you remove the newline on 195 between the lines of this anonymous function since its so small? It just spreads out a compact bit of logic

app/static/js/slcManagement.js
```diff
@@ -140,3 +187,111 @@ function unapproveProposal(el){
...
+    success: function(courseDict) {
+      if (Object.keys(courseDict).length !== 0){
+        // update the imported course modal
+        $('#courseName').val(courseDict['courseName']);
+        $('#courseAbbreviation').val(courseDict['courseAbbreviation']);
+
+        let modalTitle = courseDict['courseName'] ? courseDict['courseName'] : courseDict['courseAbbreviation'] 
+        
+        // Find the element with class "modal-header" and "fw-bold" using jQuery
+        $("#alterModalTitle").text("Edit " + modalTitle);
```
hoerstl: JS formatted string?
`` `Edit ${modalTitle}` ``

app/static/js/slcManagement.js
```diff
@@ -140,3 +187,111 @@ function unapproveProposal(el){
...
+    $('#alterModal #alterCourseId').val(courseID);
+
+    $('#alterModal').modal('show');
+  });
+
+  $('#saveButton').on('click', function(event) {
+    event.preventDefault();
+    var instructorData = getCourseInstructors();
+    // Assuming you need to clear specific instructor inputs for some reason before proceeding.
+    $('#alterModal form input[name="instructor[]"]').remove();
```
hoerstl: To address the comment on line 202, no you don't. Line 203 is the exact same line that's called at the very beginning of updateInstructorList on line 243 so you can remove lines 202 and 203.

app/static/js/slcManagement.js
```diff
@@ -140,3 +187,111 @@ function unapproveProposal(el){
...
+      
+      if (callback) callback();
+    } 
+  });
+}
+
+// Instructor manipulation functions
+
+function updateInstructorList(instructorData) { // This function attaches the list of usernames to the form submission
+  $('#alterModal form input[name="instructor[]"]').remove();
```
hoerstl, line 240: I like this logic but if this function is going to clear out the list before adding what you pass into it then it should really be called `setInstructorList`. We shouldn't call this function an update because an update to other kinds of structures is always additive. We should also rename `instructorData` to something like `newInstructorList` or, simpler, `instructorList`.

app/static/js/slcManagement.js
```diff
@@ -140,3 +187,111 @@ function unapproveProposal(el){
...
+function showAlterModalWithCourse(courseID) {
+  getImportedCourseInfo(courseID, function() {
+    $('#alterModal #alterCourseId').val(courseID);
+
+    $('#alterModal').modal('show');
+  });
+
+  $('#saveButton').on('click', function(event) {
+    event.preventDefault();
+    var instructorData = getCourseInstructors();
```
hoerstl, line 200: You should rename this variable `instructorList` since it's a list and not a dict/general object.

app/templates/admin/manageServiceLearningFaculty.html
```diff
@@ -111,37 +145,157 @@ <h5 class="ms-3">There are no approved courses for {{ term.description }}.</h5>
...
 
+    {########################### Imported Table #################################}
+    <div class="tab-pane fade {{'active show' if courseID else ''}}" id="importedCourseTables" role="tabpanel">
+      <div class="container">
+        <div class="row">
+          <div class="col">
+            <h2>Imported</h2>
+          </div>
+        </div>
+      </div>
```
hoerstl: See line 115 but I don't think we need all these divs surrounding the Imported h2.
(I'm pretty sure I'm the one that typed this code ^u^)

hoerstl: I believe it was there in case we wanted to add a button next to the imported header but for now I guess we can remove it.

app/templates/admin/manageServiceLearningFaculty.html
```diff
@@ -111,37 +145,157 @@ <h5 class="ms-3">There are no approved courses for {{ term.description }}.</h5>
...
+            <input type="hidden" id="alterTermId" name="termId" value="{{term}}">
+            <div class="modal-body"> 
+              <table class="table table-striped ml-4">
+                <tbody>
+                  <tr>
+                    <td class="fw-bold">Course Name</td>
+                    <td><input class="form-control" id="courseName" name="courseName"></td>
+                  </tr>
+                  <tr>
+                    <td class="fw-bold"> Course Abbreviation</td>
```
hoerstl: Can you take away the space before Course or put one after Abbreviation?

app/templates/admin/manageServiceLearningFaculty.html
```diff
@@ -111,37 +145,157 @@ <h5 class="ms-3">There are no approved courses for {{ term.description }}.</h5>
...
+                            </td>
+                            <td class="align-middle">
+                                <button id="remove" type="button" class="btn btn-danger removeButton">Remove</button>
+                            </td>
+                          </tr>
+                          {% endfor %}
+                      </tbody>
+                    </table>
+                    <!-- End of Instructor Table -->
+                  </tr>
```
hoerstl, line 223: This doesn't close on the same indentation level as it opens on line 242. Do you think you could fix that?

hoerstl (changes requested):
I give it a pass for all functional uses but there is one pretty severe bug I found that you need to address. I also threw in a bunch of formatting suggestions (some renaming of variables and parameter kinds of things).

app/controllers/admin/routes.py
stevensonmichel: Fixed

app/controllers/admin/routes.py
stevensonmichel: Fixed

app/controllers/admin/routes.py
stevensonmichel: Sorting is not needed for that purpose because instructors are listed in a chronological order.

app/controllers/admin/routes.py
stevensonmichel: Fixed

app/templates/admin/manageServiceLearningFaculty.html
stevensonmichel: Fixed

app/templates/admin/manageServiceLearningFaculty.html
stevensonmichel: Fixed

app/static/js/slcManagement.js
stevensonmichel: That does not return a list.

app/static/js/slcManagement.js
stevensonmichel: Fixed.

app/controllers/admin/routes.py
stevensonmichel: Also, we use the same layout as everywhere else where instructors are added in a table.

app/templates/admin/manageServiceLearningFaculty.html
stevensonmichel: Fixed

app/controllers/admin/routes.py
stevensonmichel: Presumably, Imported Courses are submitted along with participants. However, a try/except block has been added to verify that matter.

app/static/js/instructorTable.js
andersoncedu: I'm changing it in instructortable.js, but changing this in base.js is out of the scope of the issue and just adds another file changed which would make it more difficult to review, you could create an issue to refactor base,js

app/static/js/slcManagement.js
andersoncedu: This is still important for the course proposals view, which you can navigate to from the sidebar.

andersoncedu:
Lawrence mentioned that the new functions in serviceLearningCourses.py didn't have a test, so this will need to be added since it's a logic file.

bledsoef:
Viewing and Creating course proposals are both broken.

bledsoef:
Suggestion: Remove the imported tab and add it is a section header. 
Screenshot_20240424_111916

app/static/js/slcManagement.js
```diff
@@ -24,9 +28,49 @@ $(document).ready(function() {
 
       } else if ($(this).attr('data-content') == "In Progress") {
         return "This proposal has not been submitted for review."
+
+      } else if ($(this).attr('data-content') == "Imported") {
```
bledsoef, line 32: This may not be the best status message for imported courses.

bledsoef (changes requested):
Screenshot_20240424_112339
Imported courses have a blank name in the faculty Course Proposals page.


Fixed

app/static/js/slcManagement.js
stevensonmichel: Fixed

AndersonStettner (approved):

PR #1113, opened 2024-01-05
Add test for saveOtherEngagementRequest

daizenj:
Issue: Missing test for the function saveOtherEngagementRequest in logic/minor.py.

Fix: Added a test for saveOtherEngagementRequest. 
Also, addressed attribute typos that were causing inconsistencies in the requestOtherEngagement.html. e.g `descirption`

### DIFF
--- a/tests/code/test_minor.py
+++ b/tests/code/test_minor.py
@@ -190,15 +190,14 @@ def test_saveOtherEngagementRequest():
         }
 
         # Get the actual saved request from the database (the most recent one)
-        saved_request_id = CommunityEngagementRequest.select().order_by(CommunityEngagementRequest.id.desc()).first().id
-        saved_request = CommunityEngagementRequest.get_by_id(saved_request_id)
+        saved_request = CommunityEngagementRequest.select().order_by(CommunityEngagementRequest.id.desc()).first()
 
         # Check that the saved request matches the expected values
         for key, expected_value in expected_values.items():
             if key == "user":
-                actual_value = getattr(saved_request.user, 'username')
+                actual_value = 'ramsayb2'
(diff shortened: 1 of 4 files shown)


tests/code/test_minor.py
```diff
@@ -148,10 +149,58 @@ def test_getCommunityEngagementByTerm():
...
+            'company': testInfo['company'],
+            'weeklyHours': testInfo['hours'],
+            'weeks': testInfo['weeks'],
+            'filename': testInfo['attachment'],
+            'status': testInfo['status']
+        }
+
+        # Get the actual saved request from the database (the most recent one)
+        saved_request_id = CommunityEngagementRequest.select().order_by(CommunityEngagementRequest.id.desc()).first().id
+        saved_request = CommunityEngagementRequest.get_by_id(saved_request_id)
```
BrianRamsay: Once you use `.first()` you already have a `CommunityEngagementRequest` object, so you don't need to get the id as an intermediate step.

tests/code/test_minor.py
```diff
@@ -148,10 +149,58 @@ def test_getCommunityEngagementByTerm():
...
+
+        # Check that the saved request matches the expected values
+        for key, expected_value in expected_values.items():
+            if key == "user":
+                actual_value = getattr(saved_request.user, 'username')
```
BrianRamsay: Instead of using `getattr` here, let's just reference the username member directly. Same below with term and id. It's nice to be consistent with `getattr` on all of them, but that means we're being consistent with a more confusing way of getting the value. `getattr` is only necessary in the last case.

BrianRamsay (changes requested):

PR #1116, opened 2024-01-16
1115 event attachment filenames

### LINKED ISSUE #1115: Event Attachment Filenames
Behind the scenes, when we save a file for an event, we save it in the following format:
.../eventattachments/"{eventID}/{filename}"
where the file is actually being saved as "{eventID}/{filename}" in the event attachments directory.
This is all fine and well, solving one of our major problems for file uploads but when the uploaded files are displayed to the user, they have the event id prepended to their names.
image

Things to change:
- Remove the event id and forward slash from the beginning of all uploaded files
- Add a tooltip that allows you to hover over a shortened filename and view its full name. Currently, when we make the name shorter by adding the "..." in the middle, we lose the ability to distinguish the file exactly.

hoerstl:
Issue: Fixes #1115. When files are displayed to the user they are displayed with the event ID as part of the file's path. This ought not to be visible to users and adds too much irrelevant information to the screen. We'd like to remove this event ID and use the original filename. 
Additionally, we abbreviate the filenames of uploaded files for formatting which makes us unable to properly distinguish them later. To fix this, we needed to add a tooltip with the full filename listed.

Solution: I've removed the odd prefix from the filenames by altering the displayFiles macro. Additionally, I've added the tooltip and fixed a small bug where duplicate filenames didn't stop you from adding a file until you saved the page. Now they stop you from uploading a file and it provides you with immediate feedback.

To Test: Go to any event and upload a file. Save the event and ensure that what is displayed is the filename (possibly shortened) without the event ID. Check the tooltip and assert that it correctly displays the unshortened filename. 

Additionally, prepare a different file and upload it once before saving the event. Now that it's saved, try to upload it again and ensure that you can't and that a message is displayed communicating the file duplication as the root cause of the failure.

As there are two places where the display files macro is used, a tester should also navigate to the course proposals page and upload a file to any course proposal. I shouldn't have changed anything here so play with the name and make sure that no prefix is removed from these files.  We only add the folder to the filename behind the scenes for event attachments. If someone wants to upload a weird filename to the slc proposal page we should let them.

### DIFF
--- a/app/static/js/displayFilesMacro.js
+++ b/app/static/js/displayFilesMacro.js
@@ -1,4 +1,5 @@
 $(document).ready(function(){
+    $("a.fileName").tooltip()
     $(".removeAttachment").on("click", function(){
         let fileId=  $(this).data("id")
         let deleteLink = $(this).data("delete-url")
--- a/app/templates/admin/createEvent.html
+++ b/app/templates/admin/createEvent.html
@@ -244,7 +244,8 @@
           {% if filepaths %}
           <div class="form-group mb-5">
             {% from 'displayFilesMacro.html' import displayFiles %}
-            {{ displayFiles(filepaths, 'Event Attachments', '/deleteEventFile', '{{course.id}}') }}
+            <!-- {{eventData.id}} -->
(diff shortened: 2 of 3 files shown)


bledsoef (approved):
LGTM! Nice work!

Issue #944, opened 2023-06-16
Refactor Track Volunteers controller and template


AndersonStettner:
Refactor the controller (`/eventsList/<eventID>/track_volunteers`) to simplify the data controller gets and the template needs.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Refactor Track Volunteers controller and template #944](https://trello.com/c/z4UURNhs/637-refactor-track-volunteers-controller-and-template-944)

daizenj:
Request for more Info/Update: Couldn't locate route` /eventsList/<eventID>/track_volunteers` either this has already been refactored or it no longer exists. Please provide more information or update the issue if there have been changes.

PR #1119, opened 2024-01-24
Refactor Manage Volunteers Controller

bledsoef:
## Issue Description
Fixes issue #944 
In the issue description, we were tasked with refactoring the `/event/<eventID>/manage_volunteers` route.

## Changes
- The route was originally broken into separate `POST` and `GET` controllers, but we thought it was cleanest to combine them into one route, especially since they reuse a bit of code. 
- There were some variables that were unused either in the controller or passed into the template and never accessed, we removed those.
- Some variables were defined unnecessarily like `program = event.program` which could just be accessed as `event.program` anywhere necessary in the code.
- Adjust variable names to be more descriptive, example: `trainedParticipantsList` -> `trainedParticipantsForProgramAndTerm`
- Added a docstring and comments to the code to make it a little more readable. Also broke into sections to make it easier to navigate.
- Modified the `getEventParticipants` logic function to return a better representation for this controller since this is the only place it is used.
- Updated tests for `getEventParticipants` accordingly to make sure tests are passing.
- All throughout the `admin/routes` and `admin/volunteers` controllers and the templates for each respective route there was the usage of `eventData` which always just duplicated the `event` object's data. We removed all instances of eventData when unnecessary from each controller and template.
- In order to get `createEvent.html` to render properly, since it does not have an `event` object associated with it and it still extends `eventNav.html`, we modified the `createEvent.html` file to only extend `eventNav.html` when it is in `edit` mode and not `create` mode since we know it will have an `event` object then.

## Testing
- `git checkout development`
- `git pull origin development`
- `git checkout 944_Refactor_Track_Volunteers`
- `git pull origin 944_Refactor_Track_Volunteers`
- `source setup.sh`
- `./database/reset_database.sh test`
- `./tests/run_tests.sh` and ensure all tests pass (the last test in `test_participants.py` sometimes fails and that is unrelated to our issue).
- Navigate to `/app/controllers/admin/volunteers` and look over the `manageVolunteersPage` route to ensure that it looks good and there are no unnecessary pieces of code there.
- `flask run`
- Create an event.
- Once the event is created go to its view page, and then toggle through all pages in the event navigation bar (Manage Volunteers, RSVP Log, etc.) and verify each page loads properly and functions like it should.
- Pay special attention to the Manage Volunteers tab and make sure everything works properly.
- Replicate previous UI testing steps but with a past event instead.


app/controllers/admin/volunteers.py
```diff
@@ -1,5 +1,4 @@
 from flask import request, render_template, redirect, url_for, request, flash, abort, g, json, jsonify
```
gahimbaref: `request` is imported twice

app/controllers/admin/volunteers.py
```diff
@@ -25,81 +23,90 @@ def getVolunteers(query):
...
     return json.dumps(searchUsers(query))
 
-@admin_bp.route('/event/<eventID>/manage_volunteers', methods=['POST'])
-def updateVolunteerTable(eventID):
-    try:
-        event = Event.get_by_id(eventID)
-    except DoesNotExist as e:
-        print(f"No event found for {eventID}")
-        abort(404)
+@admin_bp.route('/event/<eventID>/manage_volunteers', methods=['GET', 'POST'])
```
gahimbaref, line 26: This might be far fetched but:

I like that the POST and GET are refactored into one controller but I think grouping related functionalities together might be more readable. For instance, placing all the POST request handling logic in a separate function and all GET logic in another and calling both functions in the controller with if statements ` if request.method == "POST":  //call POST function` and same for GET. 
 This will have both POST and GET in one controller but still keeping the logic separated.

gahimbaref (dismissed):

gahimbaref:
In the manage volunteers page, this error shows up when trying to add a new volunteer

image

app/controllers/admin/volunteers.py
bledsoef: So like instead of a bunch of function calls in the controller a single logic function for each request that just contains a lot of that stuff inside of it?

app/controllers/admin/volunteers.py
```diff
@@ -25,81 +23,71 @@ def getVolunteers(query):
...
-@admin_bp.route('/event/<eventID>/manage_volunteers', methods=['GET'])
-def manageVolunteersPage(eventID):
+    GET: retrieves all necessary participant lists and dictionaries and categorizes
+    the participants/volunteers into their respective participation statuses. Then 
+    renders the manageVolunteers.html template.
+    """
     try:
         event = Event.get_by_id(eventID)
     except DoesNotExist as e:
         print(f"No event found for {eventID}", e)
```
gahimbaref, line 41: This print statement for debugging?

bledsoef: I believe it was already in there I just moved it around. I think we do some other similar stuff other places in the code base to alert us when the parameters passed in on request fail.

app/templates/eventNav.html
```diff
@@ -1,12 +1,12 @@
...
 
-{% macro eventheader(page_title, eventData, size, isNewEvent) %}
+{% macro eventheader(page_title, event, size, isNewEvent) %}
     {% set bigh = 'h1' if size == 'large' else 'h3' %}
     {% set smallh = 'h5' if size == 'large' else 'p' %}
 
     <{{bigh}} class="mb-0" id="title">{{page_title}}</{{bigh}}>
     {% if not isNewEvent %}
-        {% set programName = "CELTS-Sponsored" if eventData["program"] == None else eventData["program"].programName %}
+        {% set programName = "CELTS-Sponsored" if event.program == None else event.program.programName %}
```
andersoncedu: This shouldn't be needed because there should never be an event without a program.

tests/code/test_volunteers.py
```diff
@@ -4,6 +4,7 @@
 from app import app
 from app.logic.volunteers import getEventLengthInHours, updateEventParticipants, addUserBackgroundCheck
 from app.models.eventParticipant import EventParticipant
+from app.models.eventRsvp import EventRsvp
```
andersoncedu: This was added but not used.

app/templates/events/manageVolunteers.html
```diff
@@ -119,7 +119,7 @@ <h2 class="accordion-header">
                             {% if participant.hoursEarned == None %}
                               value=""
                             {% else %}
-                              value="{{participant.hoursEarned}}"
+                              value={{participant.hoursEarned}}
```
andersoncedu: While it's true you might not need to wrap the value in quotes in this particular case, the value should be wrapped in quotes in HTML to match other places on the same page, such as on line 99.

andersoncedu (commented):
It's not a big deal, but when you go from any of the other pages to the edit page on the eventNav bar, it slightly shifts, this might be outside the scope of you issue, or not even worth looking at.

app/controllers/admin/volunteers.py
```diff
@@ -25,81 +23,71 @@ def getVolunteers(query):
...
+        # ------- Grab the different lists of participants -------
+        trainedParticipantsForProgramAndTerm = trainedParticipants(event.program, event.term)
 
-    allRelevantUsers = [participant.user for participant in eventVolunteerData + eventNonAttendedData + eventWaitlistData]
-    completedTrainingInfo = getParticipationStatusForTrainings(program, allRelevantUsers, event.term)
+        bannedUsersForProgram = [bannedUser.user for bannedUser in getBannedUsers(event.program)]
+  
+        eventNonAttendedData, eventWaitlistData, eventVolunteerData, eventParticipants = sortParticipantsByStatus(event)
+        
+        allRelevantUsers = [participant.user for participant in (eventParticipants + eventNonAttendedData + eventWaitlistData)]
```
stevensonmichel, line 67: Does that handle duplicated users?

bledsoef: There should never be a duplicate in any of those, or else something broke somewhere else. If necessary I could convert to a set and back, but I don't think that is necessary.

Karina-Agliullova (approved):
Good job! Everything works as expected, the code looks clean, all test cases pass, no errors detected.

gahimbaref (approved):
looks good

PR #1160, opened 2024-03-19
Clean up Test Data and refactor CELTS Labor tests to not rely on test data

AndersonStettner:
Refactored test_celtsLabor.py to not be reliant on test_data and fixed some unusual data in test_data


bledsoef (changes requested):
Don't need a lot of the verification left over from when we were using test data. Example in test_refreshCeltsLaborRecords.

BrianRamsay (approved):

PR #1169, opened 2024-04-02
1133 bonner cohort control

### LINKED ISSUE #1133: Change how Bonner Cohort Years are added to the Cohort Management section.
Currently, a new cohort year is added automatically at a certain date, we should change it to have a button to manually add more cohort years instead of having it happen automatically. 

Screenshot_20240219_140833

hoerstl:
# Add New Cohorts to Bonner

## Issue

Resolves #1133  
Currently, a new cohort year is added automatically at a certain date, we should change it to have a button to manually add more cohort years instead of having it happen automatically.

This allows administrators to add people to a cohort by hand as early as they know who should be in it.

## Solution

We added a button that adds a new cohort to the Bonner Management page.
image
This button does not permanently add a cohort but rather makes it visible for the current session so that administrators can add students to it which then makes the cohort persistent.

### Setup

From the `celts` directory
- `git checkout 1133-Bonnor-Cohort-Control` (yes, I know I spelled Bonner incorrectly in the branch title but in my defense, that's how it should be spelled)
- `database/reset_database.sh from-backup`
- `flask run`
- Navigate to Bonner Management using the side menu
image

### Testing

- Add users to a cohort of your choice and refresh the page, ensuring that they persist and no functionality is lost.
- Add a few new cohorts and add students to some of them ensuring that the cohorts are displayed on page load when they have non-zero student counts.
- Remove everyone from the most recently added cohort (furthest in the future) and the page will reload automatically. That cohort should disappear. Ensure that all empty cohorts after the current year are not displayed.
- Ensure that this system feels natural to use and isn't jarring when cohorts are removed on page reload if they're empty.


hoerstl:
[Change how Bonner Cohort Years are added to the Cohort Management section. #1133](https://trello.com/c/KqAKqJhW/717-change-how-bonner-cohort-years-are-added-to-the-cohort-management-section-1133)

Karina-Agliullova:
Screenshot (1763)
Before requesting a review for your PR, make sure that the page you were working on or any other related one doesn't break. I couldn't load a Bonner management page, as shown in the image above. Also, 4 checks fail out of 5. One test case fails

app/logic/bonner.py
```diff
@@ -61,7 +61,10 @@ def getBonnerCohorts(limit=None, currentYear=date.today().year):
     defaultStart = currentYear-4
     firstYear = years[0].year if len(years) and years[0].year < defaultStart else defaultStart
 
-    cohorts = { year: [] for year in range(firstYear, currentYear+1) }
+    defaultEnd = currentYear
```
bledsoef: I like that this variable name is descriptive but I think it may be a little unnecessary. Couldn't we just use currentYear wherever we need to use defaultEnd?

app/templates/admin/bonnerManagement.html
```diff
@@ -15,7 +15,7 @@
 {% block app_content %}
 
     {# Figure out if we are showing a particular tab inside of the accordion #}
-    {% set subTab =  visibleAccordion[7:] if visibleAccordion and "cohort-" in visibleAccordion else "newest" %}
+    {% set subTab =  visibleAccordion[7:] if visibleAccordion and "cohort-" in visibleAccordion and visibleAccordion[7:]|int <= cohorts.keys()|sort|last else "newest" %}
```
bledsoef: Got a bit of a weird space after the = here.

app/static/js/bonnerManagement.js
```diff
@@ -69,7 +72,30 @@ function addRequirement() {
     addRequirementsRowHandlers()
     newRow.find("input").focus()
 }
-
+function addCohort(){
```
bledsoef, line 80: Some additional comments in this function could be great

app/templates/admin/bonnerManagement.html
```diff
@@ -32,12 +32,13 @@ <h3 class="accordion-header" id="headingOne">
             <!---------------- Cohort management ---------------->
             <div class="d-flex justify-content-center">
                 <div class="nav flex-column nav-pills me-3" id="v-pills-tab" role="tablist" aria-orientation="vertical">
-                {% set subTab = cohorts.keys()|last if subTab == "newest" else subTab %}
-                {% for year in cohorts.keys()|reverse %}
-                    {% set show = "active" if subTab|int == year else "" %}
-                    {% set aria = "true" if subTab|int == year else "false" %}
-                    <button class="nav-link {{show}}" id="v-pills-{{year}}-tab" data-bs-toggle="pill" data-bs-target="#v-pills-{{year}}" type="button" role="tab" aria-controls="v-pills-{{year}}" aria-selected="{{aria}}">{{year}} - {{year + 1}}</button>
-                {% endfor %}
+                    <button class="btn-success py-2 mb-2 rounded boarder-0" id="addCohort">Add New Cohort</button>
```
bledsoef: I think we should probably use the default button styling for this button. The different styling sticks out a little bit here.

app/static/js/bonnerManagement.js
```diff
@@ -69,7 +72,30 @@ function addRequirement() {
...
+            <div class="input-group mb-3">
+                <input type="search" id="search-${newCohortYear}" name="search-${newCohortYear}" class="form-control" data-year="${newCohortYear}" placeholder="Add Student" autocomplete="off" style="width:50%" />
+                <span class="input-group-text me-1"><span class="bi bi-search"></span></span>
+            </div>
+            <table class="w-100 table table-striped">
+                <tr><td>No students added.</td></tr>
+            </table>
+        </div>
+    </div>`)
+    $(`#search-${newCohortYear}`).on("input", function(){
```
bledsoef: I could definitely be wrong but I believe this code might be unnecessary. I think this is duplicating functionality already being done on line 22.

bledsoef (changes requested):
Works great! Had a couple issues with syntax, commenting, and redundant code but otherwise is great.

app/static/js/bonnerManagement.js
```diff
@@ -69,7 +72,30 @@ function addRequirement() {
...
     newRow.find("input").focus()
 }
-
+function addCohort(){
+    let years = $('#v-pills-tab .nav-link').map((i, element) => {return Number($(element).data('year'))}).get();
+    let newCohortYear = Math.max(...years) + 1;
+    // Remove class 'active' from the currently active tab
+    $('#v-pills-tab .active').removeClass('active');
+    $('#v-pills-tabContent .tab-pane').removeClass('show active')
+    $('#v-pills-tab #addCohort').after(`<button class="nav-link active" id="v-pills-${newCohortYear}-tab" data-bs-toggle="pill" data-bs-target="#v-pills-${newCohortYear}" type="button" role="tab" data-year="${newCohortYear}" aria-controls="v-pills-${newCohortYear}" aria-selected="{{aria}}">${newCohortYear} - ${newCohortYear + 1}</button>`)
```
azabeli, line 89: Finding it hard to understand what's going on here, I think adding some comments might be more helpful.

app/logic/bonner.py
hoerstl: This is a good suggestion, I refactored the way we were calculating these in their entirety to remove both default variables since we were essentially performing a conditional `min()` and `max()` operation on these.

app/templates/admin/bonnerManagement.html
hoerstl: good catch!

app/static/js/bonnerManagement.js
hoerstl: Comments added!`

app/static/js/bonnerManagement.js
hoerstl: Good eyes on that, we did copy that logic from line 22 but we still need it. Since line 22 runs on page load and we're adding in new elements after page load, we need to rerun the logic that gives that input its searching capabilities. That said, we are copying code which is bad practice! I've abstracted this logic into a function called `addSearchCapabilities(inputElement)` which should solve the readability issues!

app/static/js/bonnerManagement.js
hoerstl: Done!

app/templates/admin/bonnerManagement.html
hoerstl: Thanks for teaching me how to use bootstrap like a pro

app/logic/bonner.py
```diff
@@ -56,18 +56,23 @@ def getBonnerCohorts(limit=None, currentYear=date.today().year):
         Return a dictionary with years as keys and a list of bonner users as values. Returns empty lists for
         intermediate years, or the last 5 years if there are no older records.
     """
-    years = list(BonnerCohort.select(BonnerCohort, User).join(User).order_by(BonnerCohort.year).execute())
+    bonnorCohorts = list(BonnerCohort.select(BonnerCohort, User).join(User).order_by(BonnerCohort.year).execute())
 
-    defaultStart = currentYear-4
-    firstYear = years[0].year if len(years) and years[0].year < defaultStart else defaultStart
+    firstYear = currentYear-4
+    if bonnorCohorts:
```
bledsoef: The refactor looks good, but I think you could combine these two `if bonnorCohorts:` conditionals into one.

app/static/js/bonnerManagement.js
```diff
@@ -14,14 +14,21 @@ function cohortRequest(year, method, username){
   })
 }
 
+function addSearchCapabilities(inputElement){
+    inputElement = $(inputElement);
```
bledsoef: Nitpicky I apologize but I think this line could be removed and we just call .on() on $(inputElement).

app/static/js/bonnerManagement.js
```diff
@@ -14,14 +14,21 @@ function cohortRequest(year, method, username){
...
+
+
 /*** Run After Page Load *************************************/
 $(document).ready(function(){
-    $("input[type=search]").on("input", function(){
-        let year = $(this).data('year')
-        searchUser(this.id, s => cohortRequest(year, "add", s.username), false, null, "student")
-    });
+    $("#addCohort").on('click', addCohort);
+    $("input[type=search]").each((i, e) => addSearchCapabilities(e));
```
bledsoef: It would be great if we get more descriptive variable names here and across the board for each's and maps's. e, s, and element can often be a little confusing.

app/static/js/bonnerManagement.js
```diff
@@ -14,14 +14,21 @@ function cohortRequest(year, method, username){
...
     $(".removeBonner").on("click",function(){
-        let year = $(this).data('year')
+        let year = $(this).data('year');
```
bledsoef, line 30: I like that this function now completely has semicolons for line endings but it seems like it is a little inconsistent in other locations where you have written JS code. I would recommend trying to either keep it consistent or just not worry too much about it in areas where you may not have made a change.

bledsoef (dismissed):
Good work, just a couple more minor things!

app/static/js/bonnerManagement.js
hoerstl: Gotcha

app/static/js/bonnerManagement.js
hoerstl: I like that change

azabeli (approved):

bledsoef (approved):
LGTM

Issue #1159, opened 2024-03-15
Mark trainings as completed when there are none


bledsoef:
On the Manage Volunteers page, if there are no required training for the event then the icon is still an “x” but it should be a check since they have technically completed all trainings. Same on the User Profile where we can see all of the programs and whether they have completed the trainings. We don't want it to seem like they haven't completed some required trainings when there aren't any.

![Screenshot_20240315_161826](https://github.com/BCStudentSoftwareDevTeam/celts/assets/89262886/0a2a7599-f6da-4811-baee-f3124f58b0b4)

bledsoef:
![](https://github.trello.services/images/mini-trello-icon.png) [Mark trainings as completed when there are none](https://trello.com/c/JphALRtR/748-mark-trainings-as-completed-when-there-are-none)

PR #1172, opened 2024-04-04
Events that have no requirements now display a check instead of an X

bledsoef:
## Issue
Resolves issue #1159 
Previously if there were no program trainings for an event or a program it would display an X, which didn't make much sense.

## Changes
- Check in `userProfile.html` and `manageVolunteers.html` if there are no program trainings. If there aren't we display a check.

## To Test
- checkout 'development'
- delete the All Volunteer Training in Fall 2021.
- Go to Manage Volunteers and User Profile and make sure that for programs with no trainings an x is displayed instead of a check (doesn't make sense).
- checkout `1159_program_trainings_bug`
- those Xs should now be checks


bledsoef:
[Mark trainings as completed when there are none](https://trello.com/c/JphALRtR/748-mark-trainings-as-completed-when-there-are-none)

Karina-Agliullova (approved):
Good work!

gahimbaref (approved):

PR #1175, opened 2024-04-09
Fix data table responsiveness on the minor admin page 

### LINKED ISSUE #1148: Minor Management page tables need to be more reactive
When adjusting the size of the Minor Management page, the tables aren't responsive so it can look pretty poor.

## To Replicate
Inspect the page and increase and decrease the size of the page. This should cause the tables to conflict with the sidebar as seen below. 

Screenshot_20240313_122119

AndersonStettner:


AndersonStettner:
[Minor Management page tables need to be more reactive](https://trello.com/c/2BcWog9J/738-minor-management-page-tables-need-to-be-more-reactive)

BrianRamsay (approved):

Issue #658, opened 2022-08-04
Selectively remove service transcript entry for banned programs


AndersonStettner:
Add a checkbox to the ban modal dialog to choose whether to completely remove the program from the service transcript (are they in Good Standing or not)

Make the checkbox actually affect the service transcript

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Selectively remove service transcript entry for banned programs #658](https://trello.com/c/1y7PJAEM/457-selectively-remove-service-transcript-entry-for-banned-programs-658)

BrianRamsay:
Completed

PR #1176, opened 2024-04-09
Selectively remove service transcript entry for banned programs

gahimbaref:
Issue: #658 
Add a checkbox to the ban modal dialog to choose whether to completely remove the program from the service transcript or keep it in the service transcript (in case if the checkbox is not clicked). The checkbox on the modal should affect the service transcript directly and remove or keep the programs, whose modals were checked/unchecked.

Did:

- Moved the checkbox outside the modal on its own column "On Trancript"
- Added a new boolean column `removeFromTranscript` in programBan table 
- Added a new route to update programBan table when checkbox state changes
- The fixes are reflected on the database and the service transcript page
- Added a message next to the checkbox letting users know if they "added" or "removed" the program from transcript whenever they check/uncheck it
- Modified the getTotalHours function in transcript.py to calculate hours correctly for banned/unbanned programs
- Made the unbanned programs show on the transcript by default (checked and greyed out)

Test:

- checkout the branch `serviceTEntry658`
- reset the database and run flask
- Go to student search and search for a student, like "khatts"
- In the Choose Action dropdown, select view service transcript, note which programs appear on the transcript
- Go back to the student profile, and under "Programs" in accordion, click on the pencil edit button that will open a modal Ban/Unban
- Ban a student from another program like "Hunger Initiatives" (keep in mind that the program will show up in the service transcript only if the student earned hours for that program)
- Now that the program is banned, you can now modify the checkmark. Try to uncheck the On Transcript box and see if the message "removed" appears
- Go back to view the service transcript, and check if that program appears (it should not appear). Also make sure that the Total Event Hours is correct
- Try to check the checkbox and see if the messaged "added" appears and if the program appears again on the service transcript. Also check the hours again to make sure it is correct
- Review Code in these files: userProfile.js, userProfile.html, main/routes.py, transcript.py


hoerstl:
[Selectively remove service transcript entry for banned programs #658](https://trello.com/c/1y7PJAEM/457-selectively-remove-service-transcript-entry-for-banned-programs-658)

app/controllers/main/routes.py
```diff
@@ -449,6 +449,48 @@ def serviceTranscript(username):
                             startDate = startDate,
                             userData = user)
 
+@main_bp.route('/profile/<username>/removeFromTranscript/<program_id>', methods=['GET'])
+def checkRemoved(username, program_id):
```
hoerstl: The name check, to me, sounds like we're going to be doing some logic behind the scenes and then throwing and catching some exceptions if it doesn't pan out. `isRemovedFromTranscript` would be a better name I think since it implies that we'd be returning some boolean information.

app/controllers/main/routes.py
```diff
@@ -449,6 +449,48 @@ def serviceTranscript(username):
...
+    # Check if the user is banned from the program
+    user = User.get_or_none(User.username == username)
+    if user is None:
+        abort(404)
+
+    try:
+        program_ban = ProgramBan.get((ProgramBan.program == program_id) & (ProgramBan.user == user))
+        # If the user is banned, check if it's marked for removal from transcript
+        if program_ban.removeFromTranscript:
```
hoerstl: A variable name using underscores here is a bit jarring considering the other variables' names in the file.

app/controllers/main/routes.py
```diff
@@ -449,6 +449,48 @@ def serviceTranscript(username):
...
+    data = request.json
+
+    # Retrieve removeFromTranscript value from the request data
+    removeFromTranscript = data.get('removeFromTranscript')
+
+    # Update the ProgramBan object matching the program_id and username
+    try:
+        program_ban = ProgramBan.get((ProgramBan.program == program_id) & (ProgramBan.user == user))
+        program_ban.removeFromTranscript = removeFromTranscript
+        program_ban.save()
```
hoerstl: To reiterate, the underscores on `program_ban` should probably be changed.

app/logic/transcript.py
```diff
@@ -1,77 +1,88 @@
...
+                      .where(EventParticipant.user == username)
+                      .group_by(Event.program, Event.term)
+                      .order_by(Event.term)
+                      .having(fn.SUM(EventParticipant.hoursEarned > 0)))
+
+    # Fetch all ProgramBan objects for the user
+    BannedProgramsForParticipant = ProgramBan.select().where(ProgramBan.user == username)
+
+    # Create a set of program IDs to remove from transcript
+    programsToremoveFromTranscript = {bannedProgram.program_id for bannedProgram in BannedProgramsForParticipant if bannedProgram.removeFromTranscript}
```
hoerstl: Forgot to capitalize Remove in the variable name:
`programsToRemoveFromTranscript`

app/logic/transcript.py
```diff
@@ -1,77 +1,88 @@
...
+
+    EventData = (Event.select(Event, fn.SUM(EventParticipant.hoursEarned).alias("hoursEarned"))
+                      .join(EventParticipant)
+                      .where(EventParticipant.user == username)
+                      .group_by(Event.program, Event.term)
+                      .order_by(Event.term)
+                      .having(fn.SUM(EventParticipant.hoursEarned > 0)))
+
+    # Fetch all ProgramBan objects for the user
+    BannedProgramsForParticipant = ProgramBan.select().where(ProgramBan.user == username)
```
hoerstl: I would use camel case here and lower the case of the 'b' in 'Banned'.
`bannedProgramsForParticipant`

app/logic/transcript.py
```diff
@@ -1,77 +1,88 @@
...
+    programsToremoveFromTranscript = {bannedProgram.program_id for bannedProgram in BannedProgramsForParticipant if bannedProgram.removeFromTranscript}
+    transcriptData = {}
+
+    # Iterate through EventData and populate transcriptData
+    for event in EventData:
+        if event.program.id not in programsToremoveFromTranscript:  # Check if program is not in programs to be removed from transcript
+            if event.program in transcriptData:
+                transcriptData[event.program].append([event.term.description, event.hoursEarned])
+            else:
+                transcriptData[event.program] = [[event.term.description, event.hoursEarned]]
```
hoerstl: This behavior here is what a `defaultdict` is perfect for. If you do this at the start of the file:
`from collections import defaultdict`
Then on line 32 you do this:
`transcriptData = defaultDict(list)`
Then every time you try to access a key that doesn't exist in the dictionary (like when line 39's else clause runs) it will create a new empty list and assign it as a value to that key. Now you can replace lines 37 through 40 with just line 38:
`transcriptData[event.program].append([event.term.description, event.hoursEarned])`

app/logic/transcript.py
```diff
@@ -1,77 +1,88 @@
...
+
+    return transcriptData
```
hoerstl: If you go with a default dict approach, it's best to cast it back to a dict so others aren't working with a data type they don't expect.
`return dict(transcriptData)`

app/logic/transcript.py
```diff
@@ -1,77 +1,88 @@
...
+    """
+
+    slCourses = (Course.select(Course, fn.SUM(CourseParticipant.hoursEarned).alias("hoursEarned"))
+                       .join(CourseParticipant, on=(Course.id == CourseParticipant.course))
+                       .where(CourseParticipant.user == username)
+                       .group_by(Course.courseName, Course.term))
+
+    return slCourses
+
+def getTotalHours(username):
```
hoerstl: We talked about this in person but I noticed that when programs are removed from the transcript, their hours are still displayed in the sum total of hours at the bottom of the page. The culprit is this function. You just need to filter out the programs that are set to be removed from the transcript.

app/static/js/userProfile.js
```diff
@@ -1,323 +1,367 @@
...
+    }).attr('readonly','readonly');
+  });
+
+    /*
+     * Ban Functionality
+     */
+
+
+  var programID;
+  $(".ban").click(function() {
```
hoerstl: I'm pretty sure this wasn't you all but I think you should be the changemakers to pave our bright future and change the name of this class from `ban` to `banEdit` or something to that effect. The class really doesn't seem like it should be applied to the edit button. See line 205 in `userProfile.html` for the code that sets this class. I think its only referenced here from what I could tell so it should be a quick change.

app/static/js/userProfile.js
```diff
@@ -1,323 +1,367 @@
...
+      type: "GET",
+      success: function(response) {
+          // Check if the program is marked for removal from transcript
+          if (response.removedFromTranscript) {
+              // Handle the case where program is marked for removal
+              $('.removeFromTranscriptCheckbox').prop('checked', true);
+          } else {
+              // Handle the case where program is not marked for removal
+              $('.removeFromTranscriptCheckbox').prop('checked', false);
+          }
```
hoerstl: I think you could simplify this if-else block. Also, you should probably be using the jquery selector for the id of the object instead of the class.
Lines 108 - 114 could be:
`$('#removeFromTranscriptCheckbox').prop('checked', response.removedFromTranscript)`
Great comments on this by the way :)

app/static/js/userProfile.js
```diff
@@ -1,323 +1,367 @@
...
+      error: function(error, status) {
+          console.log(error, status);
+      }
+    });
+  
+    var banButton = $("#banButton")
+    var banEndDateDiv = $("#banEndDate") // Div containing the datepicker in the ban modal
+    var banEndDatepicker = $("#banEndDatepicker") // Datepicker in the ban modal
+    var banNoteDiv = $("#banNoteDiv") // Div containing the note displaying why the user was banned previously
+                                     //Should only diplay when the modal is going to unban a user
```
hoerstl: Since the comment is on this line, you should move the line of code that hides the banNoteDiv up to line 125. It's currently on line 138.

app/static/js/userProfile.js
```diff
@@ -1,323 +1,367 @@
...
+    var banNote = $("#banNote")
+
+    banButton.text($(this).val() + " Volunteer");
+    programID = $(this).data("programid"); // Assign value to programID variable
+    banButton.data("programID", $(this).data("programid"))
```
hoerstl: `programID` is initialized on 129 and not used anywhere in this function though could be used on line 130 in place of `$(this).data("programid")`. Personally, I like having it extracted into a variable like this. You should consider keeping it and use it on line 130.

app/static/js/userProfile.js
```diff
@@ -1,323 +1,367 @@
...
+    banButton.data("username", $("#notifyInput").data("username"))
```
hoerstl: While this object does have the username of the user, it feels weird that you're using something which isn't really meant to be global to the page and that's not totally related to the ban model you're using it in. The whole thing is a refactor. It's odd we don't have a hidden input on the user's profile page with their username in it but we have 10 data-usernames. You could use the edit button which currently has the class '.ban' which you'll hopefully be changing soon since that's a little more related and still has a data-username attribute already attached to it.

app/static/js/userProfile.js
```diff
@@ -1,323 +1,367 @@
...
+    banButton.data("banOrUnban", $(this).val());
+    banEndDateDiv.show();
+    banEndDatepicker.val("")
+    $(".modal-title").text($(this).val() + " Volunteer");
```
hoerstl: This selector is a problem since there's more than just one modal on the page. You're also changing the titles of all other modals. See the Add Notes Modal under the Bonner Program:
image

app/static/js/userProfile.js
```diff
@@ -1,323 +1,367 @@
...
+    $("#modalProgramName").text("Program: " + $(this).data("name "));
+    $("#banModal").modal("toggle");
+    banNoteDiv.hide();
+    $("#removeFromTranscriptDiv").hide();
+    $("#banNoteTxtArea").val("");
+    $("#banButton").prop("disabled", true);
+    if( $(this).val()=="Unban"){
```
hoerstl: Some odd spacing here around the if statement

app/static/js/userProfile.js
```diff
@@ -1,323 +1,367 @@
...
+      }
+    });
+  
+    var banButton = $("#banButton")
+    var banEndDateDiv = $("#banEndDate") // Div containing the datepicker in the ban modal
+    var banEndDatepicker = $("#banEndDatepicker") // Datepicker in the ban modal
+    var banNoteDiv = $("#banNoteDiv") // Div containing the note displaying why the user was banned previously
+                                     //Should only diplay when the modal is going to unban a user
+    var banNote = $("#banNote")
+
```
hoerstl: You use `$(this).val()` quite a bit in this function. I think it holds the value of either 'Ban' or 'Unban'. You should consider making a variable for it up here so the code below is clearer.

hoerstl (dismissed):
Pretty good! I found a couple of bugs but they don't seem too difficult to iron out. Once you fix the non-displayed program hours ending up in the transcript total and that one `modal-title` selector changing all modal titles on the user's profile this looks good!

app/static/js/userProfile.js
Karina-Agliullova: Fixed

app/static/js/userProfile.js
Karina-Agliullova: Fixed

app/static/js/userProfile.js
Karina-Agliullova: fixed

app/logic/transcript.py
gahimbaref: fixed

app/logic/transcript.py
gahimbaref: fixed

app/logic/transcript.py
gahimbaref: fixed

bledsoef:
So I found an issue of when you ban a user, remove that program from their transcript, and then unban the user, it is still removed from the transcript.

ojmakinde (dismissed):
The PR functionality and code look good. However, I have a few suggestions:
- It would be good to add some confirmation after (un)removing a program from a student's transcript. The only way I could tell changes were being made was looking at the POST requests on my flask session. A flash or basic confirmation would improve feedback.
- Using a radio instead of a checkbox would also look better.
- Adding the "Remove from Transcript" to the main page instead of the Edit Modal would also make the functionality better. Greying out the function until the event is banned would be a viable way to implement its current functionality.

bledsoef (dismissed):
Generally LGTM. There are lots of lines changed that did not need to be changed and there are also a couple variable nitpicks that it is taking every ounce of willpower to not request changes on but it looks like it works perfectly so I will abstain. Great work!

app/logic/transcript.py
```diff
@@ -47,8 +56,14 @@ def getTotalHours(username):
     """
     Get the toal hours from events and courses combined.
     """
+    transcriptsRemoved = [program.program_id for program in ProgramBan.select().where((ProgramBan.user == username) and (ProgramBan.unbanNote.is_null(True)) and (ProgramBan.removeFromTranscript == 1))]
```
BrianRamsay: Let's pull the query out into a variable so that the query can be formatted more nicely and our list comprehension is a little more readable.

app/templates/main/userProfile.html
```diff
@@ -193,6 +193,8 @@ <h3 class="accordion-header" id="headingThree">
                 <th scope="col">Training</th>
                 {% if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff%}
                   <th scope="col">Eligibility</th>
+                  <th scope="col">On Transcript</th>
+                  
                 {% endif %}
                 <th scope="col"></th>
```
BrianRamsay: There's an extra `<th>` now

BrianRamsay (changes requested):
This looks good! Just a couple of minor changes in addition to the code comment. 

- The edit button is a little too low, so it's not in line with the other elements in the row.
- 'Removed' shows up on a separate line for me. Let's replace 'Added!' and 'Removed!' with 'Saved' or 'Saving...', I think. It will be the same for both cases, then.

BrianRamsay (approved):

PR #1185, opened 2024-06-10
ali works now

ali-ramazani:


BrianRamsay (changes requested):
You have changes in files that are outside of the contributors list.

ali-ramazani:
Added our names to developer page

BrianRamsay (approved):
works for me

PR #1191, opened 2024-06-11
We added our names to the list of contributors in default.yml file.

seedyjahateh:
We added our names to the list of contributors in default.yml file.


Karina-Agliullova (approved):
Good work

PR #1198, opened 2024-06-11
changed 'faculty' to 'instructor' in the main page proposal

vungc:
We went into the slcManagement.html file and changed <th> html tag from "Faculty" to "instructor" in line 23 of the html.


vungc:
We changed "Instructor" to "Instructor(s)" since some courses have more than one instructors and also changed "Courses" to "Course."

PR #1202, opened 2024-06-12
Aa issue1

AravDe:
We changed the peewee query as it was returning false always. We used datetime.date.today( ) to check if the event was past due.  We double checked on the app itself and now the no upcoming events button updates on the presence of an event to view upcoming events. The redirection is also good.


app/controllers/main/routes.py
```diff
@@ -47,18 +47,22 @@ def redirectToLogout():
...
     # Optimize the query to fetch programs with non-canceled, non-past events in the current term
+  
     programsWithEventsList = list(Program.select(Program, Event)
                                          .join(Event)
-                                         .where((Event.term == g.current_term) and (Event.isCanceled == False) and (Event.isPast == False))
+                                         .where((Event.term == g.current_term) & (Event.isCanceled == False))
                                          .distinct()
                                          .execute())  # Ensure only unique programs are included
+    
+    futureEvents = [p for p in programsWithEventsList if not p.event.isPast] #List Comprehension: filtered upcoming events from programWithEventsList
```
BrianRamsay: The comment should add clarity for future programmers about the intent of the developer or a step of an algorithm. Something like `# Limit returned list to events in the future` is better because it summarizes what the list comprehension is doing.

I also (usually) prefer comments on their own line before the code they are clarifying, so they can be read in-line with the code.

app/controllers/main/routes.py
```diff
@@ -139,6 +143,7 @@ def viewUsersProfile(username):
         backgroundTypes = list(BackgroundCheckType.select())
 
         eligibilityTable = []
+        current_time = datetime.datetime.now()
```
BrianRamsay: Are you using this variable?

BrianRamsay (changes requested):
You have some failing tests from the term change.

BrianRamsay (approved):

PR #1204, opened 2024-06-12
Made buttons to be not congested and for them to be responsive to smaller screens

vungc:
We changed the slcNewProposal.css to made the buttons change size on smaller screens and changed slcNewProposal.html to make sure the buttons stay in a line and not wrap.


BrianRamsay (approved):
nice work!

PR #1205, opened 2024-06-12
Create Event delete button correction

JohnCox2211:
The delete button is no longer of the create event page. Added if conditions to only make events that have already been made have the delete button (on the edit page)


BrianRamsay (approved):
Nice work!

PR #1209, opened 2024-06-12
Uncheck event file

Ndiayem14:
We added a function to our script to allow users to check and uncheck selected file attachments. This task proved more challenging than expected and took us five hours to complete. During this time, we extensively researched various platforms such as W3Schools, Stack Overflow, YouTube tutorials, jQuery documentation, Codehim. and ChatGPT as well for debugging.
Initially, we encountered issues that caused the script to malfunction. After commenting on sections of the HTML and testing each part of the script, we identified the mistakes. We meticulously reviewed the code line by line, testing the functionality after each change.
Through persistent debugging and multiple attempts, we finally resolved the issue. The key was ensuring that the script correctly handled the toggling of the checkboxes, allowing users to check and uncheck the file attachments as needed.


README.md
```diff
@@ -15,7 +15,7 @@
  * Mac OS or Linux
  * mysql ```root``` user is accessible by a non-root OS user, with password ```root``` (in order to run ```reset_database.sh```)
  * default python is Python 3
-
+<!-- Ndiayem -->
```
BrianRamsay, line 18: We don't want this in the README

app/templates/displayFilesMacro.html
```diff
@@ -37,5 +37,67 @@
             </td>
         </tr>
     {% endfor %}
+</table> -->
```
BrianRamsay, line 40: Don't leave things commented out. If the code is not needed it can be removed.

app/templates/displayFilesMacro.html
```diff
@@ -37,5 +37,67 @@
...
+    <tr>
+        <td class="py-2"><strong>Display</strong></td>
+        <td class="py-2 px-3"><strong>{{ titleName }}</strong></td>
+    </tr>
+    {% for key, value in filePaths.items() %}
+        <tr id="attachment_{{ value[1] }}" class="attachmentRow_{{ key }}" data-id="{{ key }}">
+            <td class="py-2 px-3">
+                <div class="input-group">
+                    <div class="input-group-prepend">
+                        <div class="input-group-text">
```
BrianRamsay, line 51: I don't think we need all of these input-group classes

app/templates/displayFilesMacro.html
```diff
@@ -37,5 +37,67 @@
...
+        });
+    });
+ 
+    function uncheckAllExcept(current) {
+        document.querySelectorAll('.attachmentCheck').forEach(checkbox => {
+            if (checkbox !== current) {
+                checkbox.checked = false;
+            }
+        });
+    }
```
BrianRamsay, line 100: This function could be simpler if it just unchecked all of them. You are setting the one you want to be checked right after this anyway.

app/templates/displayFilesMacro.html
```diff
@@ -37,5 +37,67 @@
...
 </table>
+
+<script>
+    document.querySelectorAll('.attachmentCheck').forEach(item => {
+        item.addEventListener('click', event => {
+            if (item.checked) {
+                console.log("checked");
+                item.checked = false; // This toggles the checkbox off if it was already checked
+
+                console.log(item.checked);
```
BrianRamsay, line 86: We don't need debug output anymore.

app/templates/displayFilesMacro.html
```diff
@@ -37,5 +37,67 @@
...
+                    type="button"
+                    class="removeAttachment btn btn-danger btn-sm">
+                    <i class="bi bi-trash h5 align-middle"></i>
+                </button>
+            </td>
+        </tr>
+    {% endfor %}
 </table>
+
+<script>
```
BrianRamsay, line 79: Javascript should be in the included javascript for the page, rather than in the template file

BrianRamsay (changes requested):
Good work getting the checkbox selection to work!

Removing the cover image doesn't work, however. If you uncheck the checkbox, the cover image stays the same.

README.md
Ndiayem14: Resolved

app/templates/displayFilesMacro.html
Ndiayem14: Resolved

app/templates/displayFilesMacro.html
Ndiayem14: Resolved

app/templates/displayFilesMacro.html
Ndiayem14: Resolved

app/templates/displayFilesMacro.html
Ndiayem14: Resolved

app/templates/displayFilesMacro.html
Ndiayem14: Resolved

README.md
BrianRamsay: It is still here.

app/static/js/displayFilesMacro.js
```diff
@@ -21,10 +21,18 @@ $(document).ready(function(){
         });
 
     $('.attachmentCheck').change(function() {
+        // Store the current checkbox state
+        var isChecked = $(this).is(':checked');
```
BrianRamsay, line 25: It's a good idea to keep variable creation as close as possible to where it is used. In this case, that means right before the `if`

app/templates/displayFilesMacro.html
BrianRamsay: commented code still here

BrianRamsay (changes requested):
It seems like most of the comments I made are unresolved. Are you missing some commits?

BrianRamsay:
Please don't leave old PRs open

PR #1211, opened 2024-06-13
fixed date past rendering issue

### LINKED ISSUE #1166: The warning message for past event shows when the event is currently happening
The warning message that we have on the View Event Page says "This is a past event." This message is applicable for the events that are in the past. However, it can also display this message on the events that are happening right now, which is a problem. We should fix that by removing that warning message from showing up for the events that are happening right now. "This is a past event" warning should only show up after the end time of the event has passed.
image

ali-ramazani:
There was an issue in the isPast property which was responsible for displaying a box with the text "Event already passed". We figured out that the property was returning True or False only based on current time comparison with start time of the event. To fix this, we compared current time with end date of the event. 

Resolves issue #1166


WackyWeaver:
For the property eventPast, it was defined by being greater than the start date and time. Logically, the event has only already pasted when the current time is greater than the end date and time.

ali-ramazani:
We noticed a build error after changing the logic of the isPast Event property. To fix the issue, we added a new property with the new logic.

app/models/event.py
```diff
@@ -35,7 +35,11 @@ def noProgram(self):
 
     @property
     def isPast(self):
-        return datetime.now() >= datetime.combine(self.startDate, self.timeStart)
+        return datetime.now() >= datetime.combine(self.startDate, self.timeStart)  
+
+    @property
+    def isEventPast(self):
+        return datetime.now() >= datetime.combine(self.endDate, self.timeEnd)
```
BrianRamsay, line 42: We definitely don't want two different properties that seem like they do the same thing. You either need to rename them to make the distinction clear or replace the existing one with the new logic (if we are needing that everywhere).

BrianRamsay (changes requested):
It is hard to review this as I have no idea what issue you are addressing. There's no issue linked to the PR, and I can't find a Trello card in PRs Ready for Review for this one either.

app/models/event.py
WackyWeaver: I renamed them to make a much clearer distinction. Also, we needed to change some of the logic. From our understanding, much of the logic could stay the same

tests/code/test_events.py
```diff
@@ -34,7 +34,8 @@
 @pytest.mark.integration
 def test_event_model():
     event = Event.get_by_id(11)
-    assert event.isPast
+    assert event.isPastStart
+    assert event.isPastEnd
```
BrianRamsay: We should beef up this test a little bit. Test for the negative cases (event in the future), and where an event has started and not finished yet.

BrianRamsay (changes requested):

ali-ramazani:
There were 4 variables with very identical names, which made understanding the code harder. We refactored the whole codebase and only used 2 variables to make the event time more intuitive. Specifically, we are using isPastStart and isPastEnd to refer to the start and end of an event in the past.

BrianRamsay (approved):

PR #1212, opened 2024-06-13
Added full name to the emergency contact updates in admin logs | Kafui Gle and Eun Sung

Kafui123:
1. Fixed the activity logs notification to display the full name instead of the username.
2. Changed the code of the line that displayed "username's emergency information"  to "username's insurance information" after updating the Insurance Information.
3. Changed a line code that would allow the user to update their Insurance Information without making any changes. It was another error or issue found. If the user tried to update their Insurance Information without making any changes, it would give an Error Page, so the line of code we changed helped to fix this error.


Karina-Agliullova:
Whenever I try to edit the emergency information and press continue to go to the next page OR try to save and exit, it gives me this error
Screenshot (2)

Issue #1164, opened 2024-03-26
Instructors are labeled as Faculty for the main course proposals page.


andersoncedu:
The wording for the table should say "Instructors" instead of "Faculty", or something to that effect since staff who aren't faculty are allowed to be added to the SLC proposals.

andersoncedu:
![](https://github.trello.services/images/mini-trello-icon.png) [Instructors are labeled as Faculty for the main course proposals page](https://trello.com/c/dxOMFhy2/750-instructors-are-labeled-as-faculty-for-the-main-course-proposals-page)

bledsoef:
PR was merged.

PR #1214, opened 2024-06-14
Changed faculty and course labels and made the course name to also show the course number 

### LINKED ISSUE #1206: Add Course Abbreviations for courses on SLC Proposals
- Course should be structured (ex: CSC 330 - Databases)
Inside logic:
- add abbreviations in the courseDict (from the courses data)
- change the else statement for "name" so that course abbreviation does not repeat
- Find unit test and modify it

MImran2002:
The changes we made is that we created a file in models that would produce an attribute for course name and number. We also created a test for the function and tested it when both course name and number exist, when none exit and when either one of them exist. We also went to html page for slcmanagement.html and added some jinja codes. We also have to add import statements across the pages.

Fixes #1206 and #1164


app/logic/serviceLearningCourses.py
```diff
@@ -43,7 +44,7 @@ def getSLProposalInfoForUser(user: User) -> Dict[int, Dict[str, Any]]:
 
         courseDict[course.id] = {"id":course.id,
                                  "creator":f"{course.createdBy.firstName} {course.createdBy.lastName}",
-                                 "name":course.courseName if course.courseName else course.courseAbbreviation,
+                                 "coursename": nameNumCombo(course.courseName, course.courseAbbreviation),
```
BrianRamsay: Let's leave `name` and add `abbr` as keys, in addition to a `displayname` key. That way the template has more info to use.

app/models/courseNameAndNumber.py
```diff
@@ -0,0 +1,17 @@
+from app.models import * 
+
+def nameNumCombo(name, abbreviation):
```
BrianRamsay: This function is not a model. It should either go in a `logic/` file, or in the `Course` model

tests/code/test_serviceLearningCourses.py
```diff
@@ -39,6 +42,20 @@ def test_getServiceLearningCoursesData():
     assert 'Spring 2021' in courseDict[4]['term'].description
     assert "Scott Heggen"  == courseDict[4]['creator']
```
BrianRamsay, line 45: The new key is never checked in any of these tests.

BrianRamsay (changes requested):

vungc:
Modifications:

1. Added “name” and “abbr” back into the dictionary in logic/serviceLearningCourses.py and changed the combination of name and abbreviation as “displayName”

2. “namenumcombo” function changed to “displayName” and moved to logic

3. changed slcMangement.html to use displayName from the courseDict

4. added tests to test the displayName from the logic files displayName and serviceLearningCourses

app/logic/displayName.py
```diff
@@ -0,0 +1,17 @@
+from app.models import * 
+
+def displayName(name, abbreviation):
```
BrianRamsay: This function needs a better name. The name should give you enough information that when you encounter it in code you have an understanding of what it does and how to use it without needing to look up the documentation.

Should be an action name, and should mention course.

app/logic/displayName.py
```diff
@@ -0,0 +1,17 @@
+from app.models import *
```
BrianRamsay: This function doesn't need it's own file. It is course related so should probably go with other course logic.

BrianRamsay (changes requested):

BrianRamsay:
@MImran2002 , is this an old PR? There are unresolved comments, still.

app/logic/serviceLearningCourses.py
MImran2002: we have added both name, abbr and displayname as keys so that it can be used anywhere

tests/code/test_serviceLearningCourses.py
```diff
@@ -38,6 +43,21 @@ def test_getServiceLearningCoursesData():
...
+
+@pytest.mark.integration
+def test_createCourseDisplayName():
+    '''tests for the successful implementation of combining course name and number to proper format'''
+    
+    assert 'Databases' == createCourseDisplayName("Databases", '')
+    assert 'Databases' == createCourseDisplayName("Databases", "")
+    assert 'FRN 103 - Frenchy Help' == createCourseDisplayName("Frenchy Help", 'FRN 103')
+    assert 'FRN 103' == createCourseDisplayName("", 'FRN 103')
+    assert '' == createCourseDisplayName ("", '')
```
WackyWeaver, line 58: one extra whitespace on 58

app/logic/displayName.py
MImran2002: We have named it displaycourse and createcoursedisplayname as a function name

app/logic/displayName.py
MImran2002: We have added the function under servicelearningcourse.py so it's not in its own file

tests/code/test_serviceLearningCourses.py
MImran2002: Thanks for the review will look into and remove the whitespace

WackyWeaver:
I would approve this. I ran the unit tests and checked for where there was change. It was looking great as far as I could see 👍

BrianRamsay (dismissed):
BAM

BrianRamsay (approved):

PR #1215, opened 2024-06-14
Event logic comments

JohnCox2211:
Updated many docstrings and comments in events.py. Added docstrings and many comments to improve code readability.

### DIFF
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -413,27 +413,26 @@ def getUpcomingEventsForUser(user, asOf=datetime.now(), program=None):
 
 def getParticipatedEventsForUser(user):
     """
-        Get all the events a user has participated in.
         :param user: a username or User object
-        :param asOf: The date to use when determining future and past events.
-                      Used in testing, defaults to the current timestamp.
+
+        Get all the events a user has participated in.
+        
         :return: A list of Event objects
     """
-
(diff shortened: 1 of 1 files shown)


app/logic/events.py
```diff
@@ -173,18 +229,34 @@ def saveEventToDb(newEventData, renewedEvent = False):
...
+        :param term: Object that gives a range of time (fall 2024, spring 2024..)
+        as an integer ID, 1 being the oldest term in DB.
+
+        :returns programs: dictionary containing all student led Events for given term
+        and which programs they belong to.
+
+    """
+    #lists program identifying numbers with their corresponding events saved in the DB
+    #sorted by date and time
+    #student led programs included Berea Buddies(program 2) and adopt a grandparent (program 3)
```
WackyWeaver, line 228: Could be a block comment if there need to be 3 lines

WackyWeaver:
It looks pretty good, but there are few changes I would make if I was you. Most of the comments do not have a space between the hashtag and the first word. Also, some of the method comments are filled with redundant information.

bledsoef:
Going to close this. Generally looks good but I think a lot of the comments may actually be overkill. Will keep this linked in the issue as it is a great reference but will likely not get merged.

PR #1216, opened 2024-06-14
deleted the mediaquery from slcNewProposal.css and added bootstrap breakpoints for slcnewProposal.html

MImran2002:
we deleted the media query in css and remove width = "100 px" in the html SLCProposal.html file because of the content style mix up. The we added the breakpoints and managed the different column breakpoint and different flexbox breakpoint.


BrianRamsay (approved):

BrianRamsay (approved):
nice work!

PR #1218, opened 2024-06-17
Graduation flag

### LINKED ISSUE #1141: Add a graduated flag and display it on the CCE Minor Progress table
We want to be able to filter and track students who did the CCE minor and have graduated. We should add a graduated flag in the database and display this next to their name with a graduation cap or something. This is just an idea though, experiment with different ways to display the flag.

Here is what it might look like (replace the square with something like a graduation cap or anything that will show that they graduated):

Screenshot_20240313_110346

We should add a filter as well that can allow them to view only those who have graduated or those who haven't.

ali-ramazani:
We added a column to the User module for the hasGraduated and then we tested it using the test database. Graduation cap has been added to the minor management page, and there's a icon next to names of those have graduated. 

resolves #1141


MImran2002:
The code works! We manually added the true boolean in  database and it shows.
image

BrianRamsay (dismissed):

Karina-Agliullova:
There is a problem with a class level and hasGraduated functionality. So, if I change test data and make one of the students a graduated student, it shows a graduation flag, which is what we want. However, the class level is not being checked, so even if we flag a junior student as a graduated student, it would still show us a graduation flag, even though on the profile, we can see, that the person is still a junior student. 
Screenshot (1)
Screenshot (2)

bledsoef:
[Add a graduated flag and display it on the CCE Minor Progress table](https://trello.com/c/PYlLDfvm/731-add-a-graduated-flag-and-display-it-on-the-cce-minor-progress-table)

bledsoef:
@BrianRamsay @Karina-Agliullova Going to approve and merge this and put in a new issue to allow admin to manually add students as graduated.

bledsoef (approved):
Looks good and works, will need additional functionality to be useful but good work.

PR #1220, opened 2024-06-17
Renew btn issue

ali-ramazani:
The renew button had an issue and the renew function wasn't catching the value sent from the UI. The problem was the ajax call wasn't being made and the function wasn't returning anything. We fixed it by adding   $('#renewBtn').on('click', renew); to the .ready() function at the top to capture ajax call.


BrianRamsay (changes requested):
There are a bunch of unrelated changes that we need to remove (or make a new branch with just the relevant changes).

ali-ramazani:
This issue and the solution has been moved to issue number 1257 (https://github.com/BCStudentSoftwareDevTeam/celts/pull/1258)

PR #1226, opened 2024-06-18
Check uncheck display final branch

Ndiayem14:
We fixed the problem where the 'unchecked' event file attachment is still displayed. We also enabled users to add more image types for event file attachments.


BrianRamsay:
Is this old? I thought we'd already merged this into development. Please close PRs that are not needed

Issue #1170, opened 2024-04-03
Negative numbers can be added to slc as number of service hours.


andersoncedu:
The SLC module will need to be changed to disallow negative numbers of service hours.

andersoncedu:
![](https://github.trello.services/images/mini-trello-icon.png) [Negative numbers can be added to SLC as a number of service hours.](https://trello.com/c/akjLrRPk/757-negative-numbers-can-be-added-to-slc-as-a-number-of-service-hours)

JohnCox2211:
Created fix for issue https://github.com/BCStudentSoftwareDevTeam/celts/pull/1227

bledsoef:
Closed in #1227

PR #1227, opened 2024-06-18
Negative service hours on SLC proposals

JohnCox2211:
Added onkeyup attribute to the credit entry text box that checks for invalid key entries with an if statement. Negative value entries are now automatically changed to 0 upon entry.

Screenshot (45)


JohnCox2211:
https://github.com/BCStudentSoftwareDevTeam/celts/issues/1170 - issue in question

BrianRamsay (approved):

Issue #1228, opened 2024-06-18
Indicate if an event is currently in progress


BrianRamsay:
Since we have both `isPastStart` and `isPastEnd`, we can show more detail than if an event is in the past. We should also add a banner on all the event views that indicates if an event is in progress.

BrianRamsay:
![](https://github.trello.services/images/mini-trello-icon.png) [Indicate if an event is currently in progress](https://trello.com/c/49ucZV8o/767-indicate-if-an-event-is-currently-in-progress)

bledsoef:
Closed by PR #1231

PR #1231, opened 2024-06-20
Indicate Event in progress 

AravDe:
The event edit page did not show if the event was currently taking place, we added this feature. We also added more details for when the event has already passed.

Issue: #1228


BrianRamsay (changes requested):
We don't need to do our own check of the time. Event.isPastStart and Event.isPastEnd already exist.

app/controllers/admin/routes.py
```diff
@@ -211,10 +211,13 @@ def renewEvent(eventId):
 @admin_bp.route('/event/<eventId>/edit', methods=['GET','POST'])
 def eventDisplay(eventId):
     pageViewsCount = EventView.select().where(EventView.event == eventId).count()
+    current_time = datetime.now().time()
+    current_date = datetime.now().date()
```
BrianRamsay: Are they needed? If so, let's use our `g.currentTime` variable. If not, GONE!!!

app/controllers/admin/routes.py
```diff
@@ -211,10 +211,13 @@ def renewEvent(eventId):
...
     if request.method == 'GET' and request.path == f'/event/{eventId}/view':
         viewer = g.current_user
         event = Event.get_by_id(eventId)
         addEventView(viewer,event) 
+        eventCountdown = getCountdownToEvent(event)
```
BrianRamsay: Seems like this isn't needed, as it is set down below in the 'view' case.

app/controllers/admin/routes.py
```diff
@@ -280,6 +283,8 @@ def eventDisplay(eventId):
     # Event Edit
     if 'edit' in rule.rule:
         return render_template("admin/createEvent.html",
+                                current_time = current_time,
+                                current_date = current_date,
```
BrianRamsay: Don't need to pass this anymore, since we have our global

app/templates/eventView.html
```diff
@@ -3,6 +3,7 @@
 {% set isPastEnd = event.isPastEnd %} 
 {% set tabName = 'view' %}
 {% extends "eventNav.html"%}
+{% set eventPast = event.isPastStart if not isNewEvent else False %}
```
BrianRamsay, line 7: Rather than add `eventPast` here, when it is used once later, let's just make that conditional use `event.isPastStart`

app/__init__.py
```diff
@@ -84,6 +84,11 @@ def load_currentTerm():
         session['current_term'] = model_to_dict(term)
         g.current_term = term
 
+import datetime
+@app.before_request
+def load_currentDateTime():
+    g.currentTime = datetime.datetime.now()
```
BrianRamsay: Let's name this variable so that it isn't ambiguous what the data type is. `g.currentDateTime`

BrianRamsay (changes requested):

app/templates/eventNav.html
```diff
@@ -18,13 +18,20 @@
...
+        {% set formatted_date = event.endDate.strftime("%B %d %Y") if currentYear > event_year else event.endDate.strftime("%B %d") %}
+
         <div class = "{{alertClass}} {{display}} mb-2 pb-2 pt-2" role="alert">
-        {% if isPastEnd and event.isCanceled %}
+        {% if eventPast and event.isCanceled %}
             This is a past event and it was canceled.
-        {% elif isPastEnd %}
-            This is a past event.
+        {% elif isInProg %}
+            This event is happening right now
```
BrianRamsay: Let's change this to say, "This event is in-progress! It started at X."

app/templates/eventNav.html
```diff
@@ -18,13 +18,20 @@
...
         <div class = "{{alertClass}} {{display}} mb-2 pb-2 pt-2" role="alert">
-        {% if isPastEnd and event.isCanceled %}
+        {% if eventPast and event.isCanceled %}
             This is a past event and it was canceled.
-        {% elif isPastEnd %}
-            This is a past event.
+        {% elif isInProg %}
+            This event is happening right now 
+        {% elif eventPast %}
+            This event ended on {{ formatted_date }} at {{ event.timeEnd.strftime("%I:%M %p") }}
```
BrianRamsay: End sentences with a period, please. Or at least match what the other messages had, since we don't always obey grammar rules on websites.

app/templates/eventView.html
```diff
@@ -66,7 +67,7 @@ <h5 id="location"style="margin: 0;">Location:</h5>
                     <br>
                     <div>
                         {% if eventData.isRsvpRequired %}
-                            {% if not eventPast %}
+                            {% if eventPast != event.isPastEnd %}
```
BrianRamsay: Seems like the current logic only shows the RSVP stuff if the event is in progress (when isPastStart isn't equal to isPastEnd)

app/templates/eventView.html
BrianRamsay: This is a good case for a comment explaining why we need to set eventPast here.

app/templates/eventView.html
BrianRamsay: `if not event.isPastStart`

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #1232, opened 2024-06-21
Ask for additional information for Community-Engaged Summer Experience | Eun Sung & Kafui

### LINKED ISSUE #1152: Ask for additional information for Community-Engaged Summer Experience
Right now we are asking for very little information for the creation of a Community-Engaged Summer Experience. 
Screenshot_20240313_124937

We should update the desired information to reflect what is in the file below:
[CCEminorForm1online.docx](https://github.com/BCStudentSoftwareDevTeam/celts/files/14591009/CCEminorForm1online.docx)

UPDATE:

We have decided it would be best to turn the profile page into three different tabs instead of the one page it currently is. The first tab would be for Sustained Community Engagements, the second would be for a summer experience, and the third would be for requesting other engagement.

We will also want to still show on the first page the status of their summer experience.

eunsungw:
1. Created three CCE Minor Profile Page tabs: Sustained Community Engagements, Summer Experience, and Requesting Other Engagement. We utilized the code of line from event_list.html. We set the Sustained Community Engagements tab as a default.
2. Added a download link for the Proposal Form of Community-Engaged Summer Experience at the Summer Experience Tab

Resolves #1152


BrianRamsay (changes requested):
The tabs look good!

The ticket is asking to create a form like the Request Other Engagement form, but for Summer Experience. Both forms should be in their tabs, rather than accessible through another button.

Issue #1171, opened 2024-04-03
"New Proposal For Service-Learning Course" is shown for every slc course.


andersoncedu:
This should probably be changed to reflect the status of the course, i.e. imported, approved.
![image](https://github.com/BCStudentSoftwareDevTeam/celts/assets/97543993/946ff129-578e-4218-a988-9d0828b27607)

andersoncedu:
![](https://github.trello.services/images/mini-trello-icon.png) ["New Proposal For Service-Learning Course" is shown for every slc course](https://trello.com/c/HnfSP8xU/758-new-proposal-for-service-learning-course-is-shown-for-every-slc-course)

seedyjahateh:
@Ndiayem14 and I are working on this issue.

bledsoef:
Resolved in PR #1233

PR #1233, opened 2024-06-21
Slc status final branch

seedyjahateh:
We added the status of slc's when on view. The status changes dynamically based on the status of the proposal.

Fixing issue #1171.


Ndiayem14:
We added the status of slc's when on view. The status changes dynamically based on the status of the proposal.

app/controllers/serviceLearning/routes.py
```diff
@@ -54,6 +54,7 @@ def slcEditProposal(courseID):
     if g.current_user.isCeltsAdmin or g.current_user in courseInstructors or isCourseCreator:
         course = Course.get_by_id(courseID)
         courseStatus = CourseStatus.get_by_id(course.status)
+        print("Type of courseStatus:", type(courseStatus))
```
BrianRamsay: Don't leave debug output in the final code

app/models/course.py
```diff
@@ -3,6 +3,7 @@
 from app.models.courseStatus import CourseStatus
 from app.models.note import Note
 from app.models.user import User
+from peewee import CharField, FloatField, BooleanField, ForeignKeyField, TextField
```
BrianRamsay: These imports should not be necessary. These symbols are already imported in app.models

app/templates/serviceLearning/slcProposal.html
```diff
@@ -198,3 +205,7 @@ <h2 class="text-center pt-4 pb-2"> New Proposal For Service-Learning Course</h2>
   </div>
   <input value={{course.id}} name="courseID" hidden/>
 </div>
+
+<script src="https://code.jquery.com/jquery-3.3.1.slim.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.14.7/umd/popper.min.js"></script>
+<script src="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/js/bootstrap.min.js"></script>
```
BrianRamsay: You definitely didn't need these just to show the status. Where did they come from?

app/templates/serviceLearning/slcProposal.html
```diff
@@ -1,4 +1,11 @@
-<h2 class="text-center pt-4 pb-2"> New Proposal For Service-Learning Course</h2>
+ <h2 class="text-center pt-4 pb-2">Proposal For Service-Learning Course</h2>
+
+      <h3 class="text-center pb-4">Proposal Status:
+        <span style="color: rgb(12, 8, 245);">
+                {{ courseStatus.status }}
+            </span>
```
BrianRamsay: We don't need a color unless it is going to change for different statuses

app/templates/serviceLearning/slcProposal.html
```diff
@@ -1,4 +1,11 @@
-<h2 class="text-center pt-4 pb-2"> New Proposal For Service-Learning Course</h2>
+ <h2 class="text-center pt-4 pb-2">Proposal For Service-Learning Course</h2>
+
+      <h3 class="text-center pb-4">Proposal Status:
```
BrianRamsay: "Proposal Status" is redundant. "Status" should be fine.

app/models/courseStatus.py
```diff
@@ -1,4 +1,7 @@
-from app.models import*
+# from app.models import*
+from app.models import baseModel
+from peewee import CharField
```
BrianRamsay: We can leave the previous import statement as is.

BrianRamsay (changes requested):

app/controllers/serviceLearning/routes.py
seedyjahateh: Removed

app/models/course.py
seedyjahateh: Removed.

app/templates/serviceLearning/slcProposal.html
seedyjahateh: Removed.

app/templates/serviceLearning/slcProposal.html
seedyjahateh: Changed.

app/models/courseStatus.py
seedyjahateh: Now using 'from app.models import*' only.

app/templates/serviceLearning/slcProposal.html
seedyjahateh: removed

BrianRamsay (approved):

PR #1237, opened 2024-06-24
fixed datepicker issue

ali-ramazani:
The date picker wasn't working on Firefox browser, we fixed the issue by changing the input field type from 'text' to 'date'. Also, the JavaScript file wasn't correctly linked to the HTML file. Fixing that made the behavior of the date picker similar across both browsers.

### DIFF
--- a/app/logic/emailHandler.py
+++ b/app/logic/emailHandler.py
@@ -281,3 +281,6 @@ class EmailHandler:
                                      recipient_name="{recipient_name}",
                                      relative_time=event.relativeTime)
         return new_body
+    
+
+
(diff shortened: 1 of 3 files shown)


Ndiayem14:
Firefox did not allow users to pick a date, but we fixed it by changing the input type from "text" to "date."

BrianRamsay (changes requested):
Has all of the logging changes. We need a PR without those, that just has the change from text -> date

ali-ramazani:
We moved this issue to a new branch 'date-picker-browser-issue'

PR #1240, opened 2024-06-24
Enhanced 'import_users.py' script using logging module

ali-ramazani:
Fixes issue #1053 

We enhanced the `import_users.py` script by integrating Python's logging module to provide detailed and descriptive log messages. These improvements will make the `cron.log` file significantly more informative, aiding in monitoring the script's execution and diagnosing any issues that arise from crontab scheduled tasks.


Ndiayem14:
We have upgraded the `import_users.py` script by incorporating Python's logging module to generate comprehensive and detailed log messages. This enhancement will make the `cron.log` file much more insightful, facilitating better monitoring of the script's performance and easier troubleshooting of any issues that emerge from the crontab-scheduled tasks.

WackyWeaver:
This PR seemed to be pretty straightforward. I saw obvious problems with the logging, and I ran unit testing for this. This was not client side, so I saw no need to try to break it on that end. I would approve this PR.

bledsoef:
This fixes issue #1053

vungc:
Can you resolve the merge conflicts that exist? @ali-ramazani

vungc (dismissed):
Need to resolve merge conflicts.

hoerstl:
I was running this in a devcontainer and I couldn't check this script with tracy but I looked over it and everything looks to be good. Let's give this one final run before we merge it.

WackyWeaver (commented):
I approve it, but I would also want someone to try running it with TRACY locally.

BrianRamsay (commented):
Based on https://docs.python.org/3/library/logging.html#levels, the log levels should be better defined. Log level DEBUG should show all of the output, INFO should give a summary of what the script did (count of users updated/inserted, etc), and WARNING should be when something unexpected happens, or we can't match a user.

We should be able to control the log levels in an argument to the script.

The output should go to stdout, not to a file. Let crontab control where the output goes.

app/scripts/import_users.py
```diff
@@ -78,81 +125,94 @@ def getMssqlCursor():
...
 
 def addToDb(userList):
+    usersAdded = 0
+    usersUpdated = 0
     for user in userList:
         try:
             User.insert(user).execute()
-
+            logger.debug(f"Inserted user {user['bnumber']}")
+            userAdded += 1
```
BrianRamsay: You're using `userAdded` in the loops, but the variable is initialized as `usersAdded`

BrianRamsay (dismissed):
It doesn't seem like there's an intermediate level of detail, but maybe that's because the first updates don't work right now. For ldap at least I either get only errors, or all of the info output.

app/scripts/import_users.py
stevensonmichel: Fixed

.devcontainer/Dockerfile
```diff
@@ -8,7 +8,11 @@ RUN date
 RUN pip install --upgrade pip
 
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
-     && apt-get -y install --no-install-recommends default-mysql-client ack bash-completion unixodbc unixodbc-dev
+     && apt-get -y install --no-install-recommends default-mysql-client ack bash-completion unixodbc unixodbc-dev \
+     build-essential libssl-dev libffi-dev freetds-dev freetds-bin unixodbc-dev tdsodbc
+
+RUN sudo bash -c "echo -e '[FreeTDS]\nDescription=FreeTDS Driver\nDriver=/usr/lib/x86_64-linux-gnu/odbc/libtdsodbc.so\
+\nSetup=/usr/lib/x86_64-linux-gnu/odbc/libtdsS.so' >> /etc/odbcinst.ini"
```
hoerstl, line 15: I couldn't find a way to break this into two lines while keeping the indentation style so we're just going to have to live with it.

hoerstl (dismissed):
I've fully tested the changes I made to the dockerfile and the logging code works perfectly.

BrianRamsay (approved):
made some changes. let's goooo

PR #1245, opened 2024-06-26
fixed action issue in course review

ali-ramazani:
## Issue
- [ ] The reviewProposal.html could not access the variables and functions in the JS file because the JS was a module
- [ ] Functions and variables were locally scoped to the module
- [ ] The JavaScript file included a few lines of code that manually set some functions to the window object to make them globally accessible

## Changes
- [ ] I defined an allExports object to localize all the functions for easier addition/removal
- [ ] A for loop is used to attach all functions on the window object


Ndiayem14:
We fixed the issue in the course review proposal, which needed to be fixed on approved/unapproved.

BrianRamsay (changes requested):
We need to remove the extra `routes.py`

ali-ramazani:
We moved this issue to a new pull request

PR #1246, opened 2024-06-26
Removed rows from canceled course proposals

### LINKED ISSUE #1210: mySQL creating empty rows in Course Proposals
When you click "Create Course Proposal" and then click "Cancel", a new line is still added in mySQL with no data.

Screenshot 2024-06-12 160333

image

Empty rows in mySQL:
image

WackyWeaver:
We added an ajax request onto the cancelButton event handler that routes to a new route 'canceledProposal'. If the proposal has no course name and abbreviation, then the course row is deleted in the database as well as the dependent row from the course questions table. 

Resolves #1210

### DIFF
--- a/app/controllers/serviceLearning/routes.py
+++ b/app/controllers/serviceLearning/routes.py
@@ -93,7 +93,7 @@ def slcCancelProposal():
     courseID = request.form.get('courseID')
     course = Course.get_by_id(courseID)
     if not course.courseName and not course.courseAbbreviation:
-        CourseQuestion.delete().where(CourseQuestion.course_id == courseID).execute()
+        CourseQuestion.delete().where(CourseQuestion.course == course).execute()
         course.delete_instance()
     return "Proposal Canceled"
           
--- a/app/controllers/serviceLearning/routes.py
+++ b/app/controllers/serviceLearning/routes.py
@@ -17,7 +17,6 @@ from app.logic.serviceLearningCourses import getSLProposalInfoForUser, withdrawP
 from app.logic.downloadFile import *
 from app.logic.utils import getRedirectTarget, setRedirectTarget
 from app.controllers.serviceLearning import serviceLearning_bp
-global courseIDGLOBAL
 
 @serviceLearning_bp.route('/serviceLearning/courseManagement', methods = ['GET'])
 @serviceLearning_bp.route('/serviceLearning/courseManagement/<username>', methods = ['GET'])
@@ -94,7 +93,7 @@ def slcCancelProposal():
     courseID = request.form.get('courseID')
     course = Course.get_by_id(courseID)
     if not course.courseName and not course.courseAbbreviation:
-        CourseQuestion.delete().where(CourseQuestion.course_id == courseID).execute()
+        CourseQuestion.delete().where(CourseQuestion.course == course)
         course.delete_instance()
     return "Proposal Canceled"
           
--- a/app/controllers/serviceLearning/routes.py
+++ b/app/controllers/serviceLearning/routes.py
@@ -13,11 +13,11 @@ from app.models.courseQuestion import CourseQuestion
 from app.models.attachmentUpload import AttachmentUpload
 from app.logic.utils import selectSurroundingTerms, getFilesFromRequest
 from app.logic.fileHandler import FileHandler
-from app.logic.serviceLearningCourses import getSLProposalInfoForUser, withdrawProposal, renewProposal, updateCourse, createCourse, approvedCourses
+from app.logic.serviceLearningCourses import getSLProposalInfoForUser, withdrawProposal, renewProposal, updateCourse, createCourse, approvedCourses, deleteCourseObject
 from app.logic.downloadFile import *
 from app.logic.utils import getRedirectTarget, setRedirectTarget
 from app.controllers.serviceLearning import serviceLearning_bp
-
+global courseIDGLOBAL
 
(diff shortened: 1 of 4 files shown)


app/controllers/serviceLearning/routes.py
```diff
@@ -13,11 +13,11 @@
 from app.models.attachmentUpload import AttachmentUpload
 from app.logic.utils import selectSurroundingTerms, getFilesFromRequest
 from app.logic.fileHandler import FileHandler
-from app.logic.serviceLearningCourses import getSLProposalInfoForUser, withdrawProposal, renewProposal, updateCourse, createCourse, approvedCourses
+from app.logic.serviceLearningCourses import getSLProposalInfoForUser, withdrawProposal, renewProposal, updateCourse, createCourse, approvedCourses, deleteCourseObject
 from app.logic.downloadFile import *
 from app.logic.utils import getRedirectTarget, setRedirectTarget
 from app.controllers.serviceLearning import serviceLearning_bp
-
+global courseIDGLOBAL
```
BrianRamsay: We aren't using this global var anymore

app/controllers/serviceLearning/routes.py
```diff
@@ -87,9 +87,17 @@ def slcEditProposal(courseID):
...
     course = createCourse(g.current_user)
-
     return redirect(url_for('serviceLearning.slcEditProposal', courseID = course.id))
 
+@serviceLearning_bp.route('/serviceLearning/canceledProposal', methods=['POST'])
+def slcCancelProposal():
+    courseID = request.form.get('courseID')
+    course = Course.get_by_id(courseID)
+    if not course.courseName and not course.courseAbbreviation:
+        CourseQuestion.delete().where(CourseQuestion.course_id == courseID).execute()
```
BrianRamsay: Once we have gotten a Course object, we should use it everywhere, rather than reverting back to a value passed in to the function (`courseID`). And with peewee, you don't need to compare id to id. It's smart enough to compare the objects themselves. So, `CourseQuestion.course == course` would be the preferred conditional here.

app/controllers/serviceLearning/routes.py
```diff
@@ -87,9 +87,17 @@ def slcEditProposal(courseID):
...
+        course.delete_instance()
```
BrianRamsay, line 98: This is redundant if you are using delete and execute above. Since you already have a course, you could just use `delete_instance` without the query above.

app/static/js/slcNewProposal.js
```diff
@@ -78,6 +78,18 @@ $(document).ready(function(e) {
...
   $("#cancelButton").on("click", function() {
+    $.ajax({
+      url: '/serviceLearning/canceledProposal',
+      method: 'POST',
+      data: {courseID : document.getElementById('courseID').value},
+      success: function(response) {
+          msgToast("Proposal Canceled", "Successfully canceled the proposal edit/creation.")
+      },
+      error: function(error) {
+          msgFlash('Error: ', error)
```
BrianRamsay: Is there a reason we are using flash on error and toast on success? we probably want to be consistent.

app/templates/serviceLearning/slcProposal.html
```diff
@@ -198,5 +198,5 @@ <h2 class="text-center pt-4 pb-2"> New Proposal For Service-Learning Course</h2>
       </div>
     </div>
   </div>
-  <input value={{course.id}} name="courseID" hidden/>
+  <input value={{course.id}} id="courseID" name="courseID" hidden/>
```
BrianRamsay, line 206: We don't use this id.

BrianRamsay (changes requested):

app/controllers/serviceLearning/routes.py
WackyWeaver: The reason for the query is because CourseQuestion is dependent with the Course row of the same ID. We can definitely delete_instance for CourseQuestion though.

app/static/js/slcNewProposal.js
WackyWeaver: I will can change both to msgFlash, but is there a type that you would prefer us to be consistent with?

app/static/js/slcNewProposal.js
```diff
@@ -78,7 +78,15 @@ $(document).ready(function(e) {
   });
 
   $("#cancelButton").on("click", function() {
-      window.location.replace($(this).val());
+    var cancelButtonContext = this
```
BrianRamsay: let's just call this `cancelButton`. You could even convert it to jquery here, `$(this)`

BrianRamsay (changes requested):

app/static/js/slcNewProposal.js
WackyWeaver: changed to cancelButton and moved the jquery to $(this)

app/controllers/serviceLearning/routes.py
WackyWeaver: Changed it to CourseQuestion.course == course

app/logic/serviceLearningCourses.py
```diff
@@ -189,7 +189,12 @@ def withdrawProposal(courseID) -> None:
     except DoesNotExist:
         print(f"File, {AttachmentUpload.fileName}, does not exist.")
 
-    # delete course object
+    # deletes course
```
AravDe, line 208: This comment is not needed, the line right after describes what is going on pretty well in my opinion

BrianRamsay (approved):

PR #1249, opened 2024-06-27
Redesign the display for uploaded files in events and courses | Eun Sung & Mahak

### LINKED ISSUE #1138: Redesign the display for uploaded files in events and courses.
The file upload looks great, but the display for when files are uploaded and you have the option to display them could definitely look better. Here are examples:

Screenshot_20240312_144557

Screenshot_20240312_144725

- I think we shouldn't allow them to be able to select anything for course attachments, since when we do it has a popup which says "Event cover successfully updated.
- PDFs can't be selected, so the radio button doesn't exist, maybe make it disabled because it doesn't look great when its a dash next to other radio buttons.
- We should probably change the wording on "Display", since that isn't super descriptive.
- I think it just needs a little bit of a redesign.

eunsungw:
1. Redesigned the display for the files in the View & Edit Event tabs by adding an Icon for the specific files the user uploaded.
2. Successfully rebased the changes we've made into the development to work on additional features.
3. Got rid of the Check Box for all the files other than image files in the Edit Event tab. 
4. Added padding to the margin so the Icon and file are not too close to each other for better display and visibility to the user.

Resolves #1138

### DIFF
--- a/app/static/css/base.css
+++ b/app/static/css/base.css
@@ -103,7 +103,3 @@ select.empty {
 .rsvp-btn{
     padding-top: 4em;
 }
-
-.icon {
-  margin-right: 100px; /* Adjust the value as needed */
-}
--- a/app/static/css/base.css
+++ b/app/static/css/base.css
@@ -103,3 +103,7 @@ select.empty {
 .rsvp-btn{
     padding-top: 4em;
 }
+
+.icon {
+  margin-right: 100px; /* Adjust the value as needed */
+}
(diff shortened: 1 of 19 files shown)


app/static/css/base.css
```diff
@@ -102,4 +102,8 @@ select.empty {
 
 .rsvp-btn{
     padding-top: 4em;
+}
+
+.icon {
+  margin-right: 100px; /* Adjust the value as needed */
```
BrianRamsay: We don't use this class anywhere.

BrianRamsay (changes requested):
almost!

app/static/css/base.css
eunsungw: Done!

BrianRamsay (approved):

PR #1250, opened 2024-06-27
Toggle fullscreen

MImran2002:
what has changed- code added include openfullscreen(), closefullscreen(), setfocus(), key F11 function and deleted hide function, and toggle function and remove a toggle button while added dynamically set button.


BrianRamsay (changes requested):
Back button is weirdly tall.

ESC doesn't reset the button back to Open

Issue #1180, opened 2024-04-12
Importing a course that exists will match it with the first course that it finds that matches


bledsoef:
When you go to Import Course Participants on Course Management, courses that exist in the spreadsheet you upload try to match to existing ones so there are not duplicate courses. However, when we import a course and that course exists in multiple terms, it will always match to the first one. This is not always the correct course. 

We need to figure out what the best matching technique should be if the course already exists. I think we should rework this function so that it prioritizes the current term over previous terms.

bledsoef:
![](https://github.trello.services/images/mini-trello-icon.png) [Importing a course that exists will match it with the first course that it finds that matches](https://trello.com/c/X5ubBjNL/761-importing-a-course-that-exists-will-match-it-with-the-first-course-that-it-finds-that-matches)

bledsoef:
Closed in PR #1252

PR #1252, opened 2024-06-28
Import courses

AravDe:
We changed the functionality of importing courses. When a course is imported, it checks if it already exists in the courses table. If it does exist then it copies the all information other that status(originally it was just directly set to 4, so we did not change it) and term (because term might be different). It does not check if the course exists in the same term anymore. On discussion with Karina, she pointed out that we should not completely rule out on the basis on term as there might be multiple sections of a class with the same course name and term. Maybe we can add another enhancement where it checks for the section but I don't think we are able to import sections through the spreadsheet. Have not tried to import it yet. 

Issue : #1180

***New Updates***
On discussion with Brian, we decided that we are going to ignore the case of sections for now and not add any courses that have the same course abbreviation or term. We checked for three conditions. If the Course Abbreviation and Term is the same, we add the course participants. If the Course Abbreviation is the same and the term is different there are two sub-conditions. If the term is older than that of the most recent Course Abbreviation match, a new course is added without copying any information. If the term is newer than that of the most recent Course Abbreviation match, a new course is added but with information copied from the match. If none of the Course Abbreviation or term match, we enter a new course into the table.

### DIFF
--- a/app/logic/serviceLearningCourses.py
+++ b/app/logic/serviceLearningCourses.py
@@ -73,10 +73,11 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
             termCheck = Term.get(Term.id == termObj)
 
             checkTermOrder = termCheck.termOrder
+            #dosnt work
+            CurrentTermOrder = Term.termOrder
 
-            print("$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$")
 
-            print(checkTermOrder)
+            
 
             check_for_Existing_Courses = list((Course.select()
                                         .join(Term, on =(Course.term == Term.id))
@@ -84,8 +85,10 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
                                         .order_by(Course.term.desc())
                                         .limit(1)))
 
-          
-            if not check_for_Existing_Courses:
+   
+            if not check_for_Existing_Courses :
+
+                
 
                 courseObj: Course = Course.create(courseName = "",
                                 sectionDesignation = "",
@@ -96,10 +99,15 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
                                 createdBy = g.current_user,
                                 serviceLearningDesignatedSections = "",
                                 previouslyApprovedDescription = "" )
-
-            elif Term.termOrder == checkTermOrder :
-
-                print("###################################################")
+                
+            #peewee dosnt throw an error, always returns true
+            elif CurrentTermOrder == checkTermOrder:
+                print('################################################')
+                print(CurrentTermOrder)
+                print(checkTermOrder)
+                courseObj : Course = Course.get( courseAbbreviation = course,
+                                                 term = termObj)    
+            
 
             else:
 
--- a/app/logic/serviceLearningCourses.py
+++ b/app/logic/serviceLearningCourses.py
@@ -63,39 +63,32 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
             data = Course.select().where(Course.courseAbbreviation == course).order_by(Course.term.desc()).limit(1)
             get_existing_info = list(data.dicts())
 
-
             if not get_existing_info:
-<<<<<<< HEAD
                 print("########################################################")
-                courseObj: Course = Course.get_or_create(
-=======
-                print("############################################################################################")
-                courseObj: Course = Course.create(
->>>>>>> 40fcba960e38fa07d49f286d05fd9f740784423d
-                    defaults = {"CourseName" : "",
-                                "sectionDesignation" : "",
-                                "courseCredit" : "1",
-                                "term" : termObj,
-                                "status" : 4,
-                                "createdBy" : g.current_user,
-                                "serviceLearningDesignatedSections" : "",
-                                "previouslyApprovedDescription" : "" })[0]
+                print(course)
+                courseObj: Course = Course.create(courseName = "",
+                                sectionDesignation = "",
+                                courseAbbreviation = course,
+                                courseCredit = "1",
+                                term = termObj,
+                                status = 4,
+                                createdBy = g.current_user,
+                                serviceLearningDesignatedSections = "",
+                                previouslyApprovedDescription = "" )
             
             else :
                 existing_info_dict = get_existing_info[0]
-                print("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
-                print(existing_info_dict['courseName'])
-                courseObj: Course = Course.get_or_create(
-                    #  courseAbbreviation = course,
-                    #  term = termObj, 
-                    defaults = {"CourseName" : existing_info_dict['courseName'],
-                                "sectionDesignation" : "",
-                                "courseCredit" : "1",
-                                "term" : termObj,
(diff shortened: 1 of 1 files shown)


WackyWeaver:
Reviewing this PR with Seedy

seedyjahateh:
Made code review and used unintended inputs yet it still works just fine. I recommend commenting/documenting code though.

app/logic/serviceLearningCourses.py
```diff
@@ -77,17 +77,47 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
...
-                             "term" : termObj,
-                             "status" : 4,
-                             "createdBy" : g.current_user,
-                             "serviceLearningDesignatedSections" : "",
-                             "previouslyApprovedDescription" : "" })[0]
+            termCheck = Term.get(Term.id == termObj)
+
+            checkTermOrder = termCheck.termOrder
+
+            check_for_Existing_Courses = list((Course.select()
```
BrianRamsay: This variable doesn't fit our coding style (or Python's)

app/logic/serviceLearningCourses.py
```diff
@@ -77,17 +77,47 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
...
-                             "sectionDesignation" : "",
-                             "courseCredit" : "1",
-                             "term" : termObj,
-                             "status" : 4,
-                             "createdBy" : g.current_user,
-                             "serviceLearningDesignatedSections" : "",
-                             "previouslyApprovedDescription" : "" })[0]
+            termCheck = Term.get(Term.id == termObj)
+
+            checkTermOrder = termCheck.termOrder
```
BrianRamsay: `termObj` is already the term object, and we know it exists, we can just use this. And we should name it something else, like `courseTerm`

app/logic/serviceLearningCourses.py
BrianRamsay: Also, variables names should be about the data they hold, rather than the action it took to get the data. Noun rather than verb

app/logic/serviceLearningCourses.py
```diff
@@ -77,17 +77,47 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
...
+                                courseAbbreviation = course,
+                                courseCredit = "1",
+                                term = termObj,
+                                status = 4,
+                                createdBy = g.current_user,
+                                serviceLearningDesignatedSections = "",
+                                previouslyApprovedDescription = "")
+        
+            else:
+                previous_data = check_for_Existing_Courses[0]
```
BrianRamsay: Variable needs to be mixed case, and maybe more precise, `previousCourse` or `previousMatchedCourse`

app/logic/serviceLearningCourses.py
```diff
@@ -77,17 +77,47 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
...
+
+                currentTerm = Term.get(Term.id == previous_data.term)
+                
+                
+                if (currentTerm.termOrder) == checkTermOrder:
```
BrianRamsay: we can just compare term objects, rather than comparing order.

app/logic/serviceLearningCourses.py
```diff
@@ -77,17 +77,47 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
...
+                    courseObj : Course = Course.get( courseAbbreviation = course, term = termObj)
```
BrianRamsay: in this context, `previous_data` is going to be the same as `courseObj`. We've already selected it from the database, and this is the case where the course is in the term that we want.

app/logic/serviceLearningCourses.py
```diff
@@ -77,17 +77,47 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
...
+                
+                
+                if (currentTerm.termOrder) == checkTermOrder:
+                    courseObj : Course = Course.get( courseAbbreviation = course, term = termObj) 
+               
+                else:
+                    courseObj: Course = Course.create(courseName = previous_data.courseName,
+                                    courseAbbreviation = previous_data.courseAbbreviation,
+                                    sectionDesignation = previous_data.sectionDesignation,
+                                    courseCredit = 1,
```
BrianRamsay: Let's re-use course credit from our matched course

app/logic/serviceLearningCourses.py
```diff
@@ -77,17 +77,47 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
...
+                                    term = termObj,
+                                    status = 4,
```
BrianRamsay: Use the class constant for the IMPORTED status

BrianRamsay (changes requested):
Let's copy the course proposal questions and instructors to the new course as well, like we do for renewing a course

app/logic/serviceLearningCourses.py
dixita9: We renamed it to "existingCourses"

app/logic/serviceLearningCourses.py
dixita9: We changed "termCheck" to "currentTermObj" and "termObj" to "courseTerm"

app/logic/serviceLearningCourses.py
```diff
@@ -77,28 +77,74 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
...
+                                term = courseTerm,
+                                status = CourseStatus.IN_PROGRESS,
+                                createdBy = g.current_user,
+                                serviceLearningDesignatedSections = "",
+                                previouslyApprovedDescription = "")
+                CourseQuestion.create(course = courseObj.id,
+                                questionContent= " ",
+                                questionNumber=" ")
+                CourseInstructor.create(course = courseObj.id,
+                                    user = 'ramsayb2')
```
AravDe: User attribute of the courseInstructor table cannot be left blank as it is a foreign key field. It has to beone user who is already in the user table. I have put in Brian's username for now, but I feel there should be a better alternative. I can't think of any at the moment without changing the attribute type.

app/logic/serviceLearningCourses.py
```diff
@@ -77,28 +77,74 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
...
-                             "courseCredit" : "1",
-                             "term" : termObj,
-                             "status" : 4,
-                             "createdBy" : g.current_user,
-                             "serviceLearningDesignatedSections" : "",
-                             "previouslyApprovedDescription" : "" })[0]
-
-            for userDict in courseInfo['students']:
-                if userDict['errorMsg']:
+            currentTermObj = Term.get(Term.id == courseTerm)
```
BrianRamsay: We can just use `courseTerm`, as it's the same Term Model object

app/logic/serviceLearningCourses.py
BrianRamsay: We can leave these off. No records means there is nothing entered.

app/logic/serviceLearningCourses.py
```diff
@@ -77,28 +77,74 @@ def saveCourseParticipantsToDatabase(cpPreview: Dict[str, Dict[str, Dict[str, Li
...
+                                previouslyApprovedDescription = "")
+                CourseQuestion.create(course = courseObj.id,
+                                questionContent= " ",
+                                questionNumber=" ")
+                CourseInstructor.create(course = courseObj.id,
+                                    user = 'ramsayb2')
+            else:
+                previousMatchedCourse = existingCourses[0]
+
+                currentTerm = Term.get(Term.id == previousMatchedCourse.term)
```
BrianRamsay: `currentTerm` isn't a great name, because it's not the current term in the application, nor is it necessarily the same as the `currentTermObj` term

BrianRamsay (changes requested):
so close!

BrianRamsay (approved):

PR #1254, opened 2024-07-01
Date picker browser issue #1237

ali-ramazani:
The date picker wasn't working on Firefox browser, we fixed the issue by changing the input field type from 'text' to 'date'. Also, the JavaScript file wasn't correctly linked to the HTML file. Fixing that made the behavior of the date picker similar across both browsers.

# Link to Issue


MImran2002:
Ali had decided to create a new branch and work on the issue again but I am commenting to keep a record so that he can look into the bug when he works on the  same issue with a new branch.
image
bug 1: two date picker in mozilla
bug 2 the recurring toggle button also should add end date in the page but it doesn't. (worried if this branch get push will affect the development with the recurring event.

ali-ramazani:
The issue is moved to a new branch [Link](https://github.com/BCStudentSoftwareDevTeam/celts/pull/1260)

Issue #1257, opened 2024-07-02
Renew button issue in course proposal 


ali-ramazani:
The renew button on the course proposal page doesn't work. When it's clicked on, nothing happens. A modal window should pop up and allow user to change information about the course proposal.

![image](https://github.com/BCStudentSoftwareDevTeam/celts/assets/112561860/d9f7fda3-9a94-4ae8-9219-c35ecb6ff39b)

PR #1258, opened 2024-07-02
fixed renew button issue in course proposal page

ali-ramazani:
# Issue Number: 1257 | Issue Link: [1257](https://github.com/BCStudentSoftwareDevTeam/celts/issues/1257)
The renew button had an issue and the renew function wasn't catching the value sent from the UI. The problem was the ajax call wasn't being made and the function wasn't returning anything. We fixed it by adding $('#renewBtn').on('click', renew); to the .ready() function at the top to capture ajax call.


MImran2002:
Ali, the View side is perfect so does the JavaScript file just need some clean up I feel like comments on line 45 and 48 can be deleted as they are repeating what the code below does when the codes after the comment are easy to understand.

bledsoef:
Already fixed this and the additional renew onclick you added is unnecessary, thank you though!

Issue #1259, opened 2024-07-03
Date picker browser issue 


ali-ramazani:
There's an issue with different browsers displaying date in create event page. The date picker might either only work on Chrome or Firefox.

bledsoef:
Closed in PR #1260

PR #1260, opened 2024-07-03
fixed date picker issue in create event

ali-ramazani:
# Issue Number: #1259 | Issue Link: [1259](https://github.com/BCStudentSoftwareDevTeam/celts/issues/1259)

The issue with the date picker not displaying correctly in Firefox was due to incorrect date format handling. The fix ensures the date format is consistently set to 'yy-mm-dd' for compatibility across all browsers. Additionally, the updateDate function is now properly setting the min and max dates.

### DIFF
--- a/app/static/js/createEvents.js
+++ b/app/static/js/createEvents.js
@@ -1,9 +1,8 @@
 import searchUser from './searchUser.js';
 
 // updates max and min dates of the datepickers as the other datepicker changes
-// updates max and min dates of the datepickers as the other datepicker changes
 function updateDate(obj) {
-  var selectedDate = $(obj).val(); // No need for / for Firefox compatibility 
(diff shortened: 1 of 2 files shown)


MImran2002:
The page works on other browsers but not on Mozilla. It does show date picker, but the calendar button doesn't work. However, it could be an alternative by detecting Mozilla for the JavaScript side and if it's true you can hide the button.

BrianRamsay (changes requested):
Let's change the date format displayed to mm-dd-yy, (experiment a bit with slashes, but ok if it is too much work).

In Firefox, the date icon should bring up the calendar.

Create a new github and trello issue for 24 hour time showing in firefox.

BrianRamsay (approved):

PR #1266, opened 2024-07-10
TimeStr variable undefined on 'Create Events' Page

thealphagurlux:
When loading into the Create Event page, looking at the console you could see a undefined error for 'timeStr' variable. This is because the function ,format24to12HourTime, using this variable would run, but its parameter would be empty causing a error to be thrown. The issue was solved by adding a check to see if the parameter is empty, and if so don't run the rest of the function.


BrianRamsay (changes requested):
There are a lot of other changes in this PR beyond the fix for the undefined variable

Issue #1265, opened 2024-07-10
Add asterisks for the required input fields in the Create Events page 


vungc:
On the the create event page, event name, locaiton, date, and description are required and the user cannot move on until they are filled out. We need to add red asterisks and "Required" next to the name of those fields so that the user know they are required.

Here are the areas that need required labels:
![image](https://github.com/BCStudentSoftwareDevTeam/celts/assets/112138502/edcae1a1-0ff7-40e8-bc7a-288bc62c4777)
 This is what we are thinking to implement:
![image](https://github.com/BCStudentSoftwareDevTeam/celts/assets/112138502/65a574ba-87aa-4ebc-8e1d-64b8faa7e902)

Mbeweg:
![](https://github.trello.services/images/mini-trello-icon.png) [Add labels](https://trello.com/c/HRMKF92L/778-add-labels)

vungc:
We asked the room and they all agreed that just the red asterissk (pictured below) is enough to show that it is required. We decided we don't need to waste extra time trying to achieve the example picture if it is not necessary.

![image](https://github.com/BCStudentSoftwareDevTeam/celts/assets/112138502/bf6ff25d-99bf-4f10-a3f7-496812d2abec)

bledsoef:
Closed in PR #1269

PR #1269, opened 2024-07-10
Added Asterisks on required fields

vungc:
Created a class in base.css and applied the class to events name start date, location, and description on the create event page.


vungc:
Resolves issue #1265

BrianRamsay (changes requested):
End date is also required if it is shown, right? Like for recurring events.

BrianRamsay (approved):

Issue #1270, opened 2024-07-11
View Past Event Toggle


vungc:
“View past event” toggle in the events list page untoggles when the term is changed. The toggle needs to be updated to stay toggled even when the term is changed. It also cannot be toggled at all for Fall 2020 and Spring 2021 (pictured below)

![image](https://github.com/BCStudentSoftwareDevTeam/celts/assets/112138502/cf5aa25f-4414-4ca9-ba52-381f746f5089)
![image](https://github.com/BCStudentSoftwareDevTeam/celts/assets/112138502/d8ed7cbf-9acb-4f63-98cd-f55219dfec61)

Mbeweg:
![](https://github.trello.services/images/mini-trello-icon.png) [The “View past event” toggle](https://trello.com/c/VuoJ90qx/780-the-view-past-event-toggle)

vungc:
So we saw this code on the event_list.html **var g_isPastTerm ={{"false" if selectedTerm.isFutureTerm or selectedTerm.isCurrentTerm else "true" }};** and figured out that the disabling of the toggle for the terms Fall 2020 and Spring 2021 were because they are past terms. We did not realize that the current term in the system was Summer 2021. We are leaving that alone since it is not an issue. We will now be focusing on making the toggle consistent throughout the current and future terms.

bledsoef:
Closed in PR #1290

PR #1271, opened 2024-07-12
View Past Events toggle on event list page

vungc:
We modified it to where if the toggle is checked on, then it stays on and if it is checked off, then it stays off even if terms are being changed. We also made sure that the toggle of past terms remain disabled on "on" since those events are past events.

### DIFF
--- a/app/static/js/event_list.js
+++ b/app/static/js/event_list.js
@@ -33,6 +33,8 @@ $(document).ready(function(){
     
     // Disable toggle if we are in a past term; stuck on "on"
     if (g_isPastTerm) {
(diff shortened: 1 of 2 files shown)


vungc:
Resolves issue #1270

WackyWeaver:
I don't know whether or not it is a bug or a feature. Since the view past event toggle is in local storage, even if you close the Flask app and load the page, the open stays the same. It just caught me off guard when the toggle moved true when I first ran Flask

vungc:
Closing this because it is no longer relevant. All the changes from the other branch saveCookieForToggle resolves the initial issue in 1270 and also fixes the toggle moving after page load.

PR #1275, opened 2024-07-12
Created two tables for programs and templates.

Mbeweg:
The Create Events list page needed to be split into two tables, the Programs and the templates. We have managed to make the two tables. 
Screenshot 2024-07-12 163156


BrianRamsay (changes requested):
Change `col-md-5` to `col-lg-5` on the divs, and remove the parentheses to make them always plural

BrianRamsay (approved):

Issue #797, opened 2022-11-11
Refactor Program Management section of Admin Settings


AndersonStettner:
Make a modal for program management that lets the user change the title, description, program partner, cover photo, contact email, contact name, and location of the program.

AndersonStettner:
![](https://github.trello.services/images/mini-trello-icon.png) [Refactor Program Management section of Admin Settings #797](https://trello.com/c/md8HfkpA/575-refactor-program-management-section-of-admin-settings-797)

PR #1281, opened 2024-07-16
Refactor Program Management section of Admin Settings

Kafui123:
## Issue
Issue: #797

## Changes
- Created a modal to update the information displayed for the programs on the homepage.
- Fields include program name, description, partner, contact email, contact name, and location.
- Implemented a add a new file feature, to change the image of the programs on the homepage.
- Saved changes button works as expected, and there are two ways to close the modal, cancel or cross button.
- Partner, contact email, contact name, location all stored in database.
- Program name, description, images and links  can be directly seen in the home screen.
- Added editable link for Instagram, Facebook and Berea URL.
- Refactored testing function

## Testing
- Checkout `refactorProgramMgmt`
- `./database/reset_database.sh test`
- `flask run`
- Go to Admin -> Settings
- Open up the Program Settings accordion
- Click on 3-4 programs and upload a png to 2 and a jpg to 2.
- Go to the landing page and make sure the landing page images for each program reflects what you uploaded.
- Go to `app/static/files/programattachments` and make sure that there is only one file for each program (1.jpg, 2.png, etc.)
- Once your testing has concluded delete all files that were just added and are not source controlled (pngs or any other different file types)
- Run `git restore .`
- If it does not properly add back all files that were initially there you may have to manually add them from GitHub or an earlier commit.


app/logic/fileHandler.py
```diff
@@ -2,18 +2,23 @@
...
+from app.models.program import Program
+import glob
 
 class FileHandler:
-    def __init__(self, files=None, courseId=None, eventId=None):
-        self.files = files
+    def __init__(self, files=None, courseId=None, eventId=None, programId=None):
+        self.files = files 
         self.path = app.config['files']['base_path']
+        self.path1 = app.config['files']['image_path']
```
AravDe: Better Name

app/config/default.yml
```diff
@@ -25,9 +25,12 @@ ldap:
 
 files:
   base_path: 'app/static/files'
+  image_path: 'app/static/images'
+  landing_page_path: 'landingPage'
   email_attachment_path: 'emailattachments'
   event_attachment_path: 'eventattachments'
   course_attachment_path: 'courseattachments'
+  program_image_path: 'programimages'
```
BrianRamsay: Looks like we don't use this one

app/controllers/admin/userManagement.py
```diff
@@ -62,29 +65,37 @@ def addProgramManagers():
 def removeProgramManagers():
     eventData = request.form
     try:
-        return removeProgramManager(eventData['username'],int(eventData['programID']))
+        return removeProgramManagers(eventData['username'],int(eventData['programID']))
     except Exception as e:
         print(e)
         flash('Error while removing a manager.','warning')
         abort(500,"Error while trying to remove a manager.")
```
BrianRamsay: These functions aren't used anymore. Let's get rid of them

app/controllers/admin/userManagement.py
```diff
@@ -62,29 +65,37 @@ def addProgramManagers():
...
 
+@admin_bp.route('/deleteProgramFile', methods=['POST'])
+def deleteProgramFile():
+    fileData= request.form
+    programFile=FileHandler(programId=fileData["databaseId"])
```
BrianRamsay: `databaseId` should be `programId`, because it is the id of the program.

app/controllers/admin/userManagement.py
```diff
@@ -62,29 +65,37 @@ def addProgramManagers():
...
-        return removeProgramManager(eventData['username'],int(eventData['programID']))
+        return removeProgramManagers(eventData['username'],int(eventData['programID']))
     except Exception as e:
         print(e)
         flash('Error while removing a manager.','warning')
         abort(500,"Error while trying to remove a manager.")
 
+@admin_bp.route('/deleteProgramFile', methods=['POST'])
+def deleteProgramFile():
+    fileData= request.form
```
BrianRamsay: I wouldn't use a variable here. The other two lines are getting data out of the form, so it's fine to use `request.form`

app/logic/fileHandler.py
```diff
@@ -2,18 +2,23 @@
...
         self.courseId = courseId
         self.eventId = eventId
+        self.programId = programId
         if courseId:
             self.path = os.path.join(self.path, app.config['files']['course_attachment_path'], str(courseId))
         elif eventId:
             self.path = os.path.join(self.path, app.config['files']['event_attachment_path'])
-    
+        elif programId:
+            self.image_path = os.path.join(self.image_path, app.config['files']['landing_page_path'])
```
BrianRamsay: Let's use `self.path` like the other cases

app/logic/fileHandler.py
```diff
@@ -25,17 +30,25 @@ def makeDirectory(self):
     
     def getFileFullPath(self, newfilename=''):
         try:
-            filePath = (os.path.join(self.path, newfilename))
+            if self.eventId or self.courseId:
+                filePath = (os.path.join(self.path, newfilename))
+            elif self.programId:
+                filePath = (os.path.join(self.image_path, newfilename))
```
BrianRamsay: This can go back to the old code, once you are using `self.path`

app/logic/fileHandler.py
```diff
@@ -25,17 +30,25 @@ def makeDirectory(self):
...
         except FileExistsError:
             pass
         return filePath
 
     def saveFiles(self, saveOriginalFile=None):
-        try:
+      
+        try: 
+            if not isinstance(self.files, list):
+                self.files = [self.files]
```
BrianRamsay: Handling the input to the class and translating that to member variables is what the `__init__` method is for. Do this there.

app/logic/fileHandler.py
```diff
@@ -49,11 +62,28 @@ def saveFiles(self, saveOriginalFile=None):
                     if not isFileInCourse:
                         AttachmentUpload.create(course=self.courseId, fileName=file.filename)
                         saveFileToFilesystem = file.filename
+                elif self.programId:
+                    isFileInProgram = AttachmentUpload.select().where(AttachmentUpload.program == self.programId, AttachmentUpload.fileName == file.filename).exists()
+                    if not  isFileInProgram:
+                        AttachmentUpload.create(program=self.programId, fileName=file.filename)                      
+                        name = Program.get(Program.id == self.programId)
+                        file_type = file.filename.split('.')[1]
```
BrianRamsay: We want to get the LAST bit after a `.`, not the second piece, in case they name something like "habitat.humanity.jpg"

app/logic/fileHandler.py
```diff
@@ -49,11 +62,28 @@ def saveFiles(self, saveOriginalFile=None):
...
                         AttachmentUpload.create(course=self.courseId, fileName=file.filename)
                         saveFileToFilesystem = file.filename
+                elif self.programId:
+                    isFileInProgram = AttachmentUpload.select().where(AttachmentUpload.program == self.programId, AttachmentUpload.fileName == file.filename).exists()
+                    if not  isFileInProgram:
+                        AttachmentUpload.create(program=self.programId, fileName=file.filename)                      
+                        name = Program.get(Program.id == self.programId)
+                        file_type = file.filename.split('.')[1] 
+                        current_programName = f"{str(name.programName)}.{file_type}"
+                        pattern = '*' + str(name.programName) + '*'
```
BrianRamsay: `name.programName` is always a string (database field is a varchar), so the cast isn't necessary.

app/logic/fileHandler.py
```diff
@@ -49,11 +62,28 @@ def saveFiles(self, saveOriginalFile=None):
...
+                        
+                        full_pattern = os.path.join(self.image_path, pattern)
+                        files_to_delete = glob.glob(full_pattern)
+                       
+
```
BrianRamsay, line 62: Remove extra newline

app/logic/fileHandler.py
```diff
@@ -49,11 +62,28 @@ def saveFiles(self, saveOriginalFile=None):
                     if not isFileInCourse:
                         AttachmentUpload.create(course=self.courseId, fileName=file.filename)
                         saveFileToFilesystem = file.filename
+                elif self.programId:
+                    isFileInProgram = AttachmentUpload.select().where(AttachmentUpload.program == self.programId, AttachmentUpload.fileName == file.filename).exists()
```
BrianRamsay: This check means you won't save different files with the same filename. Managing the database should mirror managing the filesystem, where you remove everything for the program and then save the new one.

app/logic/userManagement.py
```diff
@@ -38,20 +39,45 @@ def removeCeltsStudentStaff(user):
     createActivityLog(f'Removed {user.firstName} {user.lastName} from a CELTS student staff member'+ 
                    (f', and as a manager of {programManagerRoles}.' if programManagerRoles else "."))
 
-
-def changeProgramInfo(newProgramName, newContactEmail, newContactName, newLocation, programId):
-    """Updates the program info with a new sender and email."""
+def changeProgramInfo(programId, attachment, programName= None, programDescription = None, partner = None, contactEmail=None, contactName= None, location = None,instagramUrl = None, facebookUrl = None, bereaUrl = None):
```
BrianRamsay: Let's format this so that the arguments are aligned

app/models/program.py
```diff
@@ -1,21 +1,21 @@
...
     programDescription = TextField()
-    partner = ForeignKeyField(Partner, null=True)
+    partner = CharField(null=True)
     isStudentLed = BooleanField(default=False)
     isBonnerScholars = BooleanField(default=False)
     isOtherCeltsSponsored = BooleanField(default=False)
     contactName = CharField(null=True,default='')
     contactEmail = CharField(null=True,default='')
     defaultLocation = CharField(null=True,default='')
+    coverImage = BlobField(null=True)
```
BrianRamsay: If we don't use the field we should get rid of it.

app/static/css/programManagement.css
```diff
@@ -0,0 +1,57 @@
...
+
+@media (max-width: 1100px) {
+  .btn {
+    height: 80px;
+  }
+  .step {
+    width: 10px;
+    height: 10px;
+  }
+}
```
BrianRamsay: If this is necessary we should be trying to use bootstrap's grid system, otherwise let's get rid of it.

app/static/js/base.js
```diff
@@ -191,6 +191,7 @@ function handleFileSelection(fileInputId){
         $(fullTrashId).data("file", file);
         $(fullTrashId).data("file-container-id", attachedObjectContainerId);
         $(fullTrashId).on("click", function() {
+          console.log("hehe")
```
BrianRamsay: no debug output

app/static/js/displayFilesMacro.js
```diff
@@ -1,27 +1,28 @@
...
+    //         data: fileData,
+    //         success: function(){
+    //             msgFlash("Attachment removed successfully", "success");
+    //             $("#attachment_" + fileId).remove();
+    //         },
+    //         error: function(error){
+    //             msgFlash(error);
+    //         }
+    //     });
+    // });
```
BrianRamsay: We still need this code in other places.

app/static/js/userManagement.js
```diff
@@ -63,6 +63,20 @@ function submitRequest(method, username){
...
+  document.getElementById('programName').value = programName;
+  document.getElementById('programDescription').value = programDescription;
+  handleFileSelection('modalProgramImage')
+  document.getElementById('partner').value = partner;
+  document.getElementById('contactEmail').value = contactEmail;
+  document.getElementById('contactName').value = contactName;
+  document.getElementById('location').value = location;
+  document.getElementById('instagramUrl').value = instagramUrl;
+  document.getElementById('facebookUrl').value = facebookUrl;
+  document.getElementById('bereaUrl').value = bereaUrl;
```
BrianRamsay: We should use jquery like we do everywhere else

BrianRamsay: `$("#programName").val(programName)`

app/templates/admin/userManagement.html
```diff
@@ -114,36 +121,124 @@ <h3 class="accordion-header" id="headingTwo">
...
+                    <button class="btn btn-success test" 
+                    role="button" id="programManagementButton{{ loop.index }}" 
+                    data-bs-toggle="modal" 
+                    data-bs-target="#adminProgramManagement"
+                    onclick="programModal('{{ program.programName }}', '{{ program.programDescription }}', '{{ program.partner }}', '{{ program.contactEmail }}', '{{ program.contactName }}', '{{ program.defaultLocation }}', '{{ program.id }}', '{{ program.instagramUrl }}','{{ program.facebookUrl }}', '{{ program.bereaUrl }}')">Update Program Information</button>
+                    <script 
+                      type="module"> import { populateModal } from "/static/js/userManagement.js";
+                      window.programModal = function(programName, programDescription, partner, contactEmail, contactName, location, programId, instagramUrl, facebookUrl, bereaUrl){
+                      populateModal(programName, programDescription, partner, contactEmail, contactName, location, programId, instagramUrl, facebookUrl, bereaUrl)
+                      }
```
BrianRamsay: If we store the program info as a data element `data-programInfo="{{ programData }}"`, then we can add an onclick to all the buttons from javascript that grabs this data from whichever button is the `event.target`.  The snag is getting a JSON representation of the dictionary for the Program model object. I'm not sure how to do that. `model_to_dict` or some other method.

app/templates/admin/userManagement.html
```diff
@@ -114,36 +121,124 @@ <h3 class="accordion-header" id="headingTwo">
...
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title">Update Program Information</h5>
+        <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
       </div>
+      <form id="updateProgramForm" class="info" action="" method="POST" enctype="multipart/form-data">
+        <div class="modal-body">
+            <tr>
+              <!-- Left modal column -->
+              <td class="cellSpacing">
```
BrianRamsay: `<td>` is something that tables use. You want divs (probably), with bootstrap rows and cols

app/templates/admin/userManagement.html
```diff
@@ -114,36 +121,124 @@ <h3 class="accordion-header" id="headingTwo">
...
+                  <label for="contactEmail" class="form-label">Contact Email</label>
+                  <input type="email" class="form-control" id="contactEmail" name="contactEmail">
+                </div>
+                <div class="mb-3">
+                  <label for="contactName" class="form-label">Contact Name</label>
+                  <input type="text" class="form-control" id="contactName" name="contactName">
+                </div>
+                <div class="mb-3">
+                  <label for="location" class="form-label">Location</label>
+                  <input type="text" class="form-control" id="location" name="location">
```
BrianRamsay: Let's rename to Default Event Location

BrianRamsay (changes requested):
The UI needs a little work. Buttons may be better as "Edit Details", Program names shouldn't be `<input>` tags, which should fix the table column width. Buttons should be normal sized.

esw0624:
#797

app/controllers/admin/userManagement.py
eunsungw: Removed the function.

app/config/default.yml
eunsungw: We removed the code.

app/controllers/admin/userManagement.py
eunsungw: Changed the ID.

app/controllers/admin/userManagement.py
eunsungw: We got rid of the variable and used request.form

app/static/js/displayFilesMacro.js
eunsungw: Uncommented the code.

app/static/js/userManagement.js
Kafui123: We used jQuery.

app/templates/admin/userManagement.html
Kafui123: Changed it to divs

app/templates/admin/userManagement.html
Kafui123: Changed to Default Event Location

app/templates/admin/userManagement.html
eunsungw: We utilized JSON representation from the JS to directly populate the modal fields.

app/static/js/base.js
Kafui123: Got rid of the debug output.

app/static/css/programManagement.css
Kafui123: we got rid of it because it was not necessary

app/logic/fileHandler.py
Kafui123: We got rid of the extra new line

app/logic/userManagement.py
Kafui123: We have aligned the arguments.

app/models/program.py
Kafui123: CoverImage is being used as it is a foreign key of program in the database. Therefore, coverImage is used to store the attachment that contains the 'modalProgramImage'.

app/logic/fileHandler.py
Kafui123: We made the changes.

app/logic/fileHandler.py
Kafui123: Done.

app/logic/fileHandler.py
Kafui123: Resolved.

app/logic/fileHandler.py
Kafui123: Fixed.

app/logic/fileHandler.py
Kafui123: Fixed.

app/logic/fileHandler.py
Kafui123: Fixed.

app/logic/fileHandler.py
Kafui123: Fixed

app/controllers/admin/userManagement.py
```diff
@@ -52,39 +54,38 @@ def manageUsers():
 def addProgramManagers():
     eventData = request.form
     try:
-        return addProgramManager(eventData['username'],int(eventData['programID']))
+        return addProgramManagers(eventData['username'],int(eventData['programID']))
     except Exception as e:
         print(e)
         flash('Error while trying to add a manager.','warning')
         abort(500,"'Error while trying to add a manager.'")
```
BrianRamsay: This one also is not used (I'm assuming, because it calls itself, which would break).

app/logic/fileHandler.py
```diff
@@ -25,17 +32,21 @@ def makeDirectory(self):
     
     def getFileFullPath(self, newfilename=''):
         try:
-            filePath = (os.path.join(self.path, newfilename))
+            if self.eventId or self.courseId:
+                filePath = (os.path.join(self.path, newfilename))
+            elif self.programId:
+                filePath = (os.path.join(self.path, newfilename))
```
BrianRamsay, line 36: We don't need any conditional, because the code is the same in both cases.

app/logic/userManagement.py
```diff
@@ -38,20 +39,53 @@ def removeCeltsStudentStaff(user):
...
+        {Program.programName:programName,
+        Program.programDescription: programDescription, 
+        Program.partner: partner, 
+        Program.contactEmail: contactEmail, 
+        Program.contactName: contactName,
+    Program.defaultLocation: location,
+        Program.coverImage: attachment,
+        Program.instagramUrl:instagramUrl,
+        Program.facebookUrl: facebookUrl,
+    Program.bereaUrl: bereaUrl
```
BrianRamsay: These indents should all be at the same level

BrianRamsay (dismissed):
We forgot about the case when a program name changes. The current logic means it can no longer find the uploaded image.

Let's store the file path in coverImage (like we do in AttachmentUpload). This way, when the name changes we don't have to do anything. This will mean changing how the image is referenced on the home page, as well as how we delete images when we upload a new one.

The table for the programs is a little narrow. 350px minimum width would fit better. Also, the header Program should be centered.

app/controllers/admin/userManagement.py
Kafui123: Resolved.

app/logic/userManagement.py
Kafui123: Fixed.

app/logic/fileHandler.py
Kafui123: Fixed.

WackyWeaver:
For the issue with the cover images of programs, it seems like uploaded photos are not stored, but their information is saved to the file under the program's name on the landing page. The field "coverImage" under Program still stores the original file's name. 

When the landing page loads, it makes the assumption that each program has the image equal to it's program. I need to keep the photo somewhere different, or I need to correct the assumption that landingPage.py is making.

WackyWeaver:
Since I am not the owner of this PR, I could not directly address the comment on the images from Brian. I found a solution to it! By using the program ID instead of the program name to rename the landing page images, renaming the programs no longer causes issue. The program cover images need to renamed to their respective IDs to display. 

Also, the UI for the program management still needs to be improved.

vungc (dismissed):
When uploading one of the photos, it does not actually update on the home page but it still give the success message (below is the example image that did not upload)
image
Also, I am able to select multiple images to upload, but only the first one will be the one updated on the homepage. It might be a good idea to disable the upload button or something like that so users are not confused.
image

thealphagurlux:
When trying to upload a picture for a event, if you exit out of the modal and open it the file name will still be there and the same file will still be uploaded if saved.

I suggest clearing the picture and file name when the user exits the modal.

image

thealphagurlux:
For the following programs, when trying to change the event picture in settings the picture doesn't change:

-Berea Teen Mentoring
-Hispanic Outreach Program
-People Who Care

bledsoef:
> For the following programs, when trying to change the event picture in settings the picture doesn't change:
> 
> -Berea Teen Mentoring -Hispanic Outreach Program -People Who Care

@thealphagurlux I double checked these on my machine and Anna's and we could not replicate the issue you found.

thealphagurlux:
> > For the following programs, when trying to change the event picture in settings the picture doesn't change:
> > -Berea Teen Mentoring -Hispanic Outreach Program -People Who Care
> 
> @thealphagurlux I double checked these on my machine and Anna's and we could not replicate the issue you found.

thealphagurlux:
> > > For the following programs, when trying to change the event picture in settings the picture doesn't change:
> > > -Berea Teen Mentoring -Hispanic Outreach Program -People Who Care
> > 
> > 
> > @thealphagurlux I double checked these on my machine and Anna's and we could not replicate the issue you found.

I had not reset my database or restored the git branch while looking for bugs. This is what was causing the issue.

app/controllers/admin/userManagement.py
```diff
@@ -48,43 +50,32 @@ def manageUsers():
...
+            programInfo = request.form # grabs user inputs
+            uploadedFile = request.files.get('modalProgramImage')
+            changeProgramInfo(programID, uploadedFile, **programInfo) 
 
+            associatedAttachments = list(AttachmentUpload.select().where(AttachmentUpload.program == programID).execute()) 
+           
+            filePaths = FileHandler(programId=programID).retrievePath(associatedAttachments) 
+           
+          
+            file_paths = {filename: path_info[0] for filename, path_info in filePaths.items()}
```
Mbeweg: The `"filePaths" `variable and the `"file_paths" `variable should be both camelcase. The also should have different variable names.

app/logic/fileHandler.py
```diff
@@ -49,11 +58,27 @@ def saveFiles(self, saveOriginalFile=None):
...
+                    # remove the existing file
+                    deleteFileObject = AttachmentUpload.get_or_none(program=self.programId)
+                    if deleteFileObject:
+                        self.deleteFile(deleteFileObject.id)
+
+                    # add the new file
+                    fileType = file.filename.split('.')[-1]
+                    fileName = f"{self.programId}.{fileType}"
+                    AttachmentUpload.create(program=self.programId, fileName=fileName)
+                    current_programID = fileName
```
Karina-Agliullova: should be camelCase

app/controllers/admin/userManagement.py
```diff
@@ -48,43 +50,32 @@ def manageUsers():
...
+            filePaths = FileHandler(programId=programID).retrievePath(associatedAttachments) 
+           
+          
+            file_paths = {filename: path_info[0] for filename, path_info in filePaths.items()} 
             flash("Program updated", "success")
             return redirect(url_for("admin.userManagement", accordion="program"))
         except Exception as e:
-            print(e)
-            flash('Error while updating program info.','warning')
+            print("error: ", e)
```
Mbeweg: Get rid of the ``"print("error: ", e)" `we don't need it.

app/static/js/base.js
```diff
@@ -212,7 +218,12 @@ function handleFileSelection(fileInputId){
         $(fileBoxId).data("file-num", $(fileBoxId).data("file-num") + 1)
       }
       else{
+        if (single){
+          $(objectContainerId).html(fileHTML)
+        }
+        else{
         msgToast("File with filename '" + file.name + "' has already been added to this event")
```
Karina-Agliullova: indentation

app/logic/fileHandler.py
```diff
@@ -49,11 +58,27 @@ def saveFiles(self, saveOriginalFile=None):
...
+                    saveFileToFilesystem = current_programID
+
```
Mbeweg, line 74: I think using snake_case for variable naming is better. 
`deleteFileObject `to `delete_file_object.`
`fileType` to `file_type.`
`fileName` to `file_name.`

WackyWeaver: I remember we discussed this in-person last week. The code base often uses camelCase instead of snake case.

app/static/js/base.js
WackyWeaver: Added!

app/controllers/admin/userManagement.py
WackyWeaver: Removed

app/logic/fileHandler.py
WackyWeaver: Done!

app/controllers/admin/userManagement.py
WackyWeaver: Honestly both lines seem to be assigning variables to objects or dictionaries that are not used in the rest of the function. So I decided to scrap them.

Karina-Agliullova (approved):
Looking good!

PR #1282, opened 2024-07-17
Multiple offering events

### LINKED ISSUE #1156: Multiple event times of a training display as separate events
For events like All Volunteer Training, they will be offered multiple times, but if you attend only one it will say there are multiple programs you have not completed. We need to find a way to still have the ability for multiple offerings of the same program but make it so that multiple instances of the same training only appear as one for training eligibility. Here is an example of the problem.

programissue

dixita9:
The multiple offering event feature is mostly done. 

Fixes: #1156


app/controllers/admin/routes.py
```diff
@@ -98,13 +104,29 @@ def createEvent(templateid, programid):
     # Try to save the form
     if request.method == "POST":
         eventData.update(request.form.copy())
-        try:
-            savedEvents, validationErrorMessage = attemptSaveEvent(eventData, getFilesFromRequest(request))
+        if eventData.get('isMultipleOffering'):
+            multipleOfferingId = calculateNewMultipleOfferingId()
+            for event in ast.literal_eval(eventData.get('multipleOfferingData')):
```
BrianRamsay: We should use json encoding and decoding vs python evals

app/controllers/admin/routes.py
```diff
@@ -5,6 +5,10 @@
 import json
 from datetime import datetime
 import os
+import ast
+from dateutil.parser import parse
+import ast
+from dateutil.parser import parse
```
BrianRamsay: Duplicate imports.

BrianRamsay: Also it doesn't look like we use `parse`

app/controllers/admin/routes.py
```diff
@@ -113,14 +135,33 @@ def createEvent(templateid, programid):
...
+           
             if program:
-                if len(savedEvents) > 1:
+                if len(savedEvents) > 1 and eventData.get('isRecurring'):
                     createActivityLog(f"Created a recurring event, <a href=\"{url_for('admin.eventDisplay', eventId = savedEvents[0].id)}\">{savedEvents[0].name}</a>, for {program.programName}, with a start date of {datetime.strftime(eventData['startDate'], '%m/%d/%Y')}. The last event in the series will be on {datetime.strftime(savedEvents[-1].startDate, '%m/%d/%Y')}.")
+                
+                elif len(savedEventsList) >= 1 and eventData.get('isMultipleOffering'):
+                    modifiedSavedEvents = [item for sublist in savedEventsList for item in sublist]
+                    
+                    event_dates = [datetime.strptime(event_data['eventDate'], '%Y-%m-%d').date().strftime('%m/%d/%Y') for event_data in ast.literal_eval(eventData.get('multipleOfferingData'))]
```
BrianRamsay: Don't use user input this far down in the function. You've already gotten the data structure once.

Also, in this case, we should use the dates from the saved events, not from the event data that was passed in.

app/controllers/admin/routes.py
```diff
@@ -113,14 +135,33 @@ def createEvent(templateid, programid):
...
+
+                    event_list = ', '.join(f"<a href=\"{url_for('admin.eventDisplay', eventId=event.id)}\">{event.name}</a>" for event in modifiedSavedEvents)
+
+                    if len(modifiedSavedEvents) > 1:
+                        #get last event name and stick at the end before 'and' so that it reads like a sentence in admin log
+                        last_event = f"<a href=\"{url_for('admin.eventDisplay', eventId=modifiedSavedEvents[-1].id)}\">{modifiedSavedEvents[-1].name}</a>"
```
BrianRamsay: We should rework this so we don't have to duplicate the link html

app/logic/events.py
```diff
@@ -67,35 +67,39 @@ def deleteEvent(eventId):
...
+        Modified to also apply for events with multiple offerings
         """
         event = Event.get_or_none(Event.id == eventId)
         if event:
-            if event.recurringId:
-                recurringId = event.recurringId
-                allRecurringEvents = list(Event.select().where(Event.recurringId == recurringId))
+            if event.recurringId or event.multipleOfferingId:
+                recurringOrMultipleOfferingId = event.recurringId or event.multipleOfferingId
+                allRecurringEvents = list(Event.select().where(Event.recurringId == recurringOrMultipleOfferingId or Event.multipleOfferingId == recurringOrMultipleOfferingId))
```
BrianRamsay: This will be a problem if those ids overlap, because you would select unrelated events.

app/logic/events.py
```diff
@@ -136,17 +141,26 @@ def saveEventToDb(newEventData, renewedEvent = False):
...
+                                'date':newEventData['startDate'],
+                                "week":1})
+        multipleSeriesId = newEventData['multipleOfferingId']
+        
     else:
         eventsToCreate.append({'name': f"{newEventData['name']}",
                                 'date':newEventData['startDate'],
                                 "week":1})
         if renewedEvent:
             recurringSeriesId = newEventData.get('recurringId')
```
BrianRamsay, line 162: We don't handle the multiple offering case here.

BrianRamsay: Also, that `append` happens in both cases, so we should make the conditional for `isMultipleOffering` only be around the code that depends on the condition.

app/static/js/createEvents.js
```diff
@@ -49,47 +51,96 @@ function calculateRecurringEventFrequency(){
...
-  $.datepicker.setDefaults({
-    dateFormat: 'yy/mm/dd', // Ensures compatibility across browsers
-    minDate: new Date()
-  });
 
-  $(".datePicker").datepicker({
-    dateFormat: 'mm/dd/yy',
-    minDate: new Date() 
-  });
+  document.getElementById('submitParticipant').addEventListener('click', function() {
```
BrianRamsay: Since we use jquery event listener syntax elsewhere, let's do it here, too.

app/static/js/createEvents.js
```diff
@@ -49,47 +51,96 @@ function calculateRecurringEventFrequency(){
...
+});  
+    if (isEmpty){
+      $('#textNotifierPadding').addClass('pt-5');
+      $('.invalidFeedback').text("Event name or date field is empty");
+      $('.invalidFeedback').css('display', 'block');  
+      $('.invalidFeedback').on('animationend', function() {
+        $('.invalidFeedback').css('display', 'none');
+        $('#textNotifierPadding').removeClass('pt-5')
+      });
+      isEmpty = false;
```
BrianRamsay: It would be useful to highlight which input is empty. Making a thicker red border should be fine. You can save the input into isEmpty instead of true, so you have it later.

app/static/js/createEvents.js
```diff
@@ -49,47 +51,96 @@ function calculateRecurringEventFrequency(){
...
     }
-  });
 
-  // Update datepicker min and max dates on change
-  $(".startDatePicker, .endDatePicker").change(function () {
-    updateDate(this);
+    else {
+      storingMultipleOfferingEventAttributes();
```
BrianRamsay: To match our other functions, this should probably be named `storeMultipleOfferingEventAttributes`

app/static/js/createEvents.js
```diff
@@ -49,47 +51,96 @@ function calculateRecurringEventFrequency(){
...
+      $('#modalMultipleOffering').modal('hide');
+      msgFlash("Multiple time offering events saved!", "success");
+    }
   });
+//build multi-event table
+function storingMultipleOfferingEventAttributes() {
+    let entries = [];
+    $(".extraSlots").children().each(function(index, element) {
+        let rowData = $.map($(element).find("input"), (el) => $(el).val());
+        console.log("Data in row " + (index + 1) + ": " + rowData);
```
BrianRamsay: We don't need debugging output anymore

app/static/js/createEvents.js
```diff
@@ -49,47 +51,96 @@ function calculateRecurringEventFrequency(){
...
+        entries.push({
+            eventName: rowData[0],
+            eventDate: rowData[1],
+            startTime: rowData[2],
+            endTime: rowData[3]
+        });
+    });
+
+    let entriesJson = JSON.stringify(entries);
+    document.getElementById("multipleOfferingDataId").value = entriesJson
```
BrianRamsay: We use jquery everywhere else, let's use it here.

app/static/js/createEvents.js
```diff
@@ -103,20 +154,78 @@ $(document).ready(function () {
...
+  var modalOpenedByEditButton = false;
 
-  $("#checkIsRecurring").click(function () {
-    var recurringStatus = $("input[name='isRecurring']:checked").val();
+  $("#checkIsRecurring, #checkIsMultipleOffering, #edit_modal").click(function(event) { //#checkIsRecurring, #checkIsMultipleOffering are attributes for the toggle buttons on create event page
+    if(!(document.getElementById('inputEventName').value === '')){
+      document.getElementById('eventName').value = document.getElementById('inputEventName').value; //keeps main page event name for multiple event modal
+    }
+    var recurringStatus = $("input[id='checkIsRecurring']:checked").val(); //this line function is to retrive ON when its toggle for recurring event on createEvent.html line 158
+    var multipleOfferingStatus = $("input[id='checkIsMultipleOffering']:checked").val();// this line function is to retrive ON when toggle for multiple offering event button createEvent.html line 160
```
BrianRamsay: Let's put all of these end-of-line comments before the lines instead of at the end. It's easier to follow.

BrianRamsay: Also, these selectors are complicated. we don't need to use an attribute selector if we know an id, we can just use the id. That is, instead of `$("input[id='checkIsRecurring']:checked")`, use `$("#checkIsRecurring:checked")`.

Or maybe use booleans everywhere instead of 'on' or '', with `$("#checkIsRecurring").is(":checked")`.

app/static/js/createEvents.js
```diff
@@ -103,20 +154,78 @@ $(document).ready(function () {
...
+    modalOpenedByEditButton = ($(this).attr('id') === 'edit_modal');
```
BrianRamsay, line 249: `this.id` should work here.

app/static/js/createEvents.js
```diff
@@ -103,20 +154,78 @@ $(document).ready(function () {
...
+  
+  $("#cancelModalPreview, #multipleOfferingXbutton").click(function(){ //this function is to untoggle the button when the modal has cancel or close button being clicked
+    if (modalOpenedByEditButton == false) {
+    $("#checkIsMultipleOffering").prop('checked', false);
+    $('#nonMultipleOfferingTime, #nonMultipleOfferingDate').removeClass('d-none');
+    $("#multipleOfferingTableDiv").addClass('d-none');
+    $('#modalMultipleOffering').modal('hide');
+    $('.extraSlots').children().not(':first').remove();
+  }
+  });
```
BrianRamsay, line 298: Same deal with the end-of-line comments in the above chunk of code.

app/static/js/createEvents.js
```diff
@@ -103,20 +154,78 @@ $(document).ready(function () {
...
     }
+    if (multipleOfferingStatus == 'on') {
+      $(".startDatePicker").prop('required', false);
+      $("#multipleOfferingTableDiv").removeClass('d-none');
+      $("#checkIsRecurring").prop('checked', false);
+      $(".endDateStyle, #recurringTableDiv").addClass('d-none');
+      $('#modalMultipleOffering').modal('show');
+      $('#nonMultipleOfferingTime, #nonMultipleOfferingDate').addClass('d-none'); // this line hides the non multiple offering time and dates and replace them with recurring table div for multiple offering events to show
+    }
+    else{
```
BrianRamsay: I'd add a comment here that clarifies that this is the recurring case

app/static/js/createEvents.js
```diff
@@ -103,20 +154,78 @@ $(document).ready(function () {
...
+  });
+  
+  $("#cancelModalPreview, #multipleOfferingXbutton").click(function(){ //this function is to untoggle the button when the modal has cancel or close button being clicked
+    if (modalOpenedByEditButton == false) {
+    $("#checkIsMultipleOffering").prop('checked', false);
+    $('#nonMultipleOfferingTime, #nonMultipleOfferingDate').removeClass('d-none');
+    $("#multipleOfferingTableDiv").addClass('d-none');
+    $('#modalMultipleOffering').modal('hide');
+    $('.extraSlots').children().not(':first').remove();
+  }
```
BrianRamsay: Fix the indentation please.

app/static/js/createEvents.js
```diff
@@ -103,20 +154,78 @@ $(document).ready(function () {
...
+  });
+  
+  /*cloning the div with ID multipleOfferingEvent and cloning, changing the ID of each clone going up by 1. This also changes the ID of the deleteMultipleOfferingEvent so that when the trash icon is clicked, 
+  that specific row will be deleted*/
```
BrianRamsay: Let's format this comment better and line it up.

```
/* This is a multi-line comment
   that is indented and not too long */
```

app/templates/base.html
```diff
@@ -48,6 +48,7 @@
           <div class="flash_container" id="flash_container">
             {% with messages = get_flashed_messages(with_categories=true) %}
               <script> var flashMessages = {{messages | tojson}} </script>
+              {# <script> var flashMessages = {{messages | tojson}} </script> #}  {#FIXME: SHOULD BE THIS, WHY??#}
```
BrianRamsay: Not sure what this comment was doing. I'm guessing you figured this out?

BrianRamsay (changes requested):
I haven't finished, but here is some stuff to work on.

For most of the code, let's move your comments at the end of a line to the line before the code in question. It's easier to read in the flow of the code, and doesn't cause a line-wrap

app/logic/events.py
JohnCox2211: separated recurringOrMultiple valiable, ask for further elaboration

app/controllers/admin/routes.py
JohnCox2211: replaced ast evals with json in both spots and removed the ast import

app/controllers/admin/routes.py
JohnCox2211: removed these imports all together.

app/templates/base.html
JohnCox2211: figured it out, removed comment line that was unused

app/static/js/createEvents.js
JohnCox2211: reformatted

app/static/js/createEvents.js
JohnCox2211: reformatted

app/static/js/createEvents.js
JohnCox2211: corrected commenting

app/static/js/createEvents.js
JohnCox2211: replaced current way of checking if these are toggled to now be booleans as suggested.

app/static/js/createEvents.js
JohnCox2211: replaced default JS with jQuery here and in other needed locations.

app/static/js/createEvents.js
JohnCox2211: removed console log

app/static/js/createEvents.js
JohnCox2211: made the suggested change

app/static/js/createEvents.js
JohnCox2211: Changed this to jQuery listener as suggested

app/controllers/admin/routes.py
JohnCox2211: now only using one link, building the list without last_event variable and instead just parsing that and adding it to the end

app/controllers/admin/routes.py
JohnCox2211: now using saved event list to get the event date

BrianRamsay (dismissed):

BrianRamsay (changes requested):
Restore the styles from John's issue in a way that works for both features.

The modal footer has disappeared!

There's one more merge conflict

BrianRamsay (approved):

PR #1286, opened 2024-07-19
Undo button for event deletion

WackyWeaver:
Created basic undo feature in the "successfully deleted" flash message. We created a session variable that would store the most recently deleted event id or most recently deleted series of events. Then, pressing the link in the flash message html will route the user through the /event/undo route under main routes.


MImran2002:
Me and Anna are able to find different solution and implement session seperately yayyy

app/static/js/event_list.js
```diff
@@ -67,5 +67,17 @@ function removeRsvpForEvent(eventID){
...
+//     type: "POST",
+//     data: undoButton,
+//     success: function(){
+//       alert("Successfully undone!")
+//     },
+//     error: function(error, status){
+//       console.log(error, status)
+//     }
+//   })
+// }
```
BrianRamsay: If we don't need it, let's get rid of it

app/templates/events/event_list.html
```diff
@@ -84,7 +84,8 @@ <h1 class="text-center">Events List for {{selectedTerm.description}}</h1>
       </tr>
     </thead>
     <tbody>
-      {% for event in events %}
+      {% for event in events%}
+        {% if event.deletionDate == "0000-00-00 00:00:00" %}
```
BrianRamsay: We don't even need to give the deleted events to jinja in the first place.

app/logic/events.py
```diff
@@ -75,9 +76,14 @@ def deleteEventAndAllFollowing(eventId):
         if event:
             if event.recurringId:
                 recurringId = event.recurringId
-                recurringSeries = list(Event.select().where((Event.recurringId == recurringId) & (Event.startDate >= event.startDate)))
-        for seriesEvent in recurringSeries:
-            seriesEvent.delete_instance(recursive = True)
+                recurringSeries = list(Event.select(Event.id).where((Event.recurringId == recurringId) & (Event.startDate >= event.startDate)))
+                #session['lastDeletedEvent'] = []
```
BrianRamsay: don't need the commented line anymore

app/logic/events.py
```diff
@@ -75,9 +76,14 @@ def deleteEventAndAllFollowing(eventId):
...
+                deletedEventList = []                
+                for recurringEvent in recurringSeries:
+                    deletedEventList.append(recurringEvent.id)
+                Event.update({Event.deletionDate: datetime.now()}).where((Event.recurringId == recurringId) & (Event.startDate >= event.startDate)).execute()
+                Event.update({Event.deletedBy: g.current_user}).where((Event.recurringId == recurringId) & (Event.startDate >= event.startDate)).execute()
+                return deletedEventList
 
 def deleteAllRecurringEvents(eventId):
```
BrianRamsay, line 83: Let's have this function just call the previous one, with the first event in the series.

app/controllers/admin/routes.py
```diff
@@ -355,10 +355,42 @@ def cancelRoute(eventId):
...
+                    newEventNameList[-1] = f"{nameCounter}"
+                    newEventNameList = " ".join(newEventNameList)
+                    Event.update({Event.name: newEventNameList}).where(Event.id==recurringEvent.id).execute()
+                    nameCounter += 1 
+        flash("Deletion successfully undone.", "success")
+        return redirect('/eventsList/' + str(g.current_term))
+    except Exception as e:
+        print('Error while canceling event:', e)
+        return "", 500
+
```
BrianRamsay, line 379: Let's refactor this so that we process a list only, then update recurring names afterwards if it is recurring.

BrianRamsay (changes requested):

app/static/js/event_list.js
MImran2002: Anna did it.

app/static/js/base.js
```diff
@@ -1,16 +1,47 @@
...
+
+
+
 $(document).ready(function() {
     $("select[name='newuser']").on('change', function(e) {
         $(e.target).parent().submit();
     });
-
+    $('#restoredDeleted').click(function(){
+      undoGeneralEvent($('#restoredDeleted').val())
```
BrianRamsay: It doesn't look like we use this value in the function. Either we need to add an argument to the function and use it, or we can remove it here.

app/static/js/base.js
```diff
@@ -1,16 +1,47 @@
+const flashMessageResponse = function flashEventResponse(message){
+  if (message.slice(-8) == "deleted."){
+
+    return `<strong><a href="/event/undo" style="color: dark-green;" id="restoreDeleted">Undo</a></strong>`
```
BrianRamsay: We don't need this id if we're not attaching an onclick

app/static/js/base.js
```diff
@@ -1,16 +1,47 @@
...
+    type: "POST",
+    data: undoEvent,
+    success: function(){
+      alert("Successfully undone!")
+    },
+    error: function(error, status){
+      console.log(error, status)
+    }
+  })
+}
```
BrianRamsay: We've implemented this as a link now, so we don't need ajax

app/static/js/base.js
BrianRamsay: We actually don't need this click handler anymore

app/templates/serviceLearning/slcNewProposal.html
```diff
@@ -2,8 +2,6 @@
 {% extends "base.html" %}
 {% block scripts %}
   {{super()}}
-  <script type="module" src="{{url_for('static', filename ='js/slcNewProposal.js') }}"></script>
-  <script type="module" src="/static/js/displayFilesMacro.js"></script>
```
BrianRamsay: Seems like we still need these.

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #1288, opened 2024-07-22
Timepicker issue

### LINKED ISSUE #1267: Start Time and End Time time pickers are not working in Firefox
Currently, in the createEvent.html file, the user is not able to click on the time pickers to input a start time and an end time. In other words, upon clicking the icons, nothing happens. 

image

ali-ramazani:
Fixes issue #1267

We fixed the issue with the start time and end time buttons not triggering the timepicker dropdown attached to the input field. The issue was the timepicker method in the timepicker object didn't expect "show" as argument. Instead, it expected "open". Moreover, we prevent the event handler from propagating up because it would close the dropdown as in the implementation of the timepicker closes the dropdown if it's not trigged from the input field directly.


thealphagurlux:
To add on to Ali's comment, we also implemented an if else statement into the function bound to the button because the button would open the time picker scroll but clicking it again would not close the scroll box, so we used the time pickers 'closed' method in order to check if the scroll box was open or not, and close and open it appropriately.

app/static/js/createEvents.js
```diff
@@ -173,7 +162,16 @@ $(document).ready(function () {
     if ($(this).val() && $("#endDatePicker-" + $(this).data("page-location")).val()) {
       calculateRecurringEventFrequency();
     }
-  });
+  })
+  // Time picker click events
+  $(".timeIcons").on("click", function(e) {
+    e.stopPropagation()
+    console.log($(this).siblings(".timepicker").timepicker("closed") );
```
BrianRamsay: An extra console.log

app/static/js/createEvents.js
```diff
@@ -173,7 +162,16 @@ $(document).ready(function () {
...
+    if ($(this).siblings(".timepicker").timepicker("closed"))
+      $(this).siblings(".timepicker").timepicker("open"); 
+    else 
+    $(this).siblings(".timepicker").timepicker("close")
```
BrianRamsay: This should be indented. Also let's use braces around the blocks, even though they are single line

app/templates/admin/createEvent.html
```diff
@@ -26,62 +26,65 @@
...
+  <!-- Include jQuery -->
+  <script src="https://code.jquery.com/jquery-3.6.0.min.js"></script>
+  <!-- Include jQuery UI (required for datepicker) -->
+  <link rel="stylesheet" href="https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css">
+  <script src="https://code.jquery.com/ui/1.12.1/jquery-ui.js"></script>
+  <!-- Include jQuery Timepicker -->
+  <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/timepicker/1.3.5/jquery.timepicker.min.css">
+  <script src="//cdnjs.cloudflare.com/ajax/libs/timepicker/1.3.5/jquery.timepicker.min.js"></script>
+  <!-- Custom Styles -->
+  <link href="https://cdn.jsdelivr.net/gh/kartik-v/bootstrap-fileinput@5.5.0/css/fileinput.min.css" media="all" rel="stylesheet" type="text/css" />
```
BrianRamsay: This stylesheet doesn't need to be in the scripts block.

app/templates/admin/createEvent.html
```diff
@@ -26,62 +26,65 @@
...
 {% set tabName = 'edit' %}
 
-
 {% block scripts %}
-  {{super()}}
+  {{super()}} 
+  <!-- Include jQuery -->
+  <script src="https://code.jquery.com/jquery-3.6.0.min.js"></script>
+  <!-- Include jQuery UI (required for datepicker) -->
+  <link rel="stylesheet" href="https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css">
```
BrianRamsay: This shouldn't work here, but maybe the browser is letting us get away with it. It should go in the styles block if necessary.

BrianRamsay (changes requested):

app/templates/admin/createEvent.html
JohnCox2211: moved into style block

app/templates/admin/createEvent.html
JohnCox2211: moved into style block

app/static/js/createEvents.js
JohnCox2211: removed

app/static/js/createEvents.js
JohnCox2211: fixed formatting

JohnCox2211:
Corrected things mentioned in comments

app/templates/admin/createEvent.html
```diff
@@ -452,6 +451,7 @@ <h5 class="modal-title" id="renewModel">Renew Event</h5>
     </div>
   </div>
 </form>
+{% endblock %}
```
bledsoef, line 454: This endblock is messing up styling on the page.

app/static/js/createEvents.js
```diff
@@ -136,7 +134,23 @@ function storeMultipleOfferingEventAttributes() {
     let entries = [];
     $(".extraSlots").children().each(function(index, element) {
         let rowData = $.map($(element).find("input"), (el) => $(el).val());
+  // Initialize timepicker
```
bledsoef, line 137: Indentation on this timepicker block of code is weird and the .extraSlots function does not close, causing some errors on my end.

app/templates/admin/createEvent.html
bledsoef: It was initially just styling but now it is breaking completely for me

app/static/js/createEvents.js
```diff
@@ -136,7 +134,23 @@ function storeMultipleOfferingEventAttributes() {
...
+  $('.timepicker').timepicker({
+    timeFormat: 'h:mm p',
+    interval: 15,
+    minTime: '12:00am',
+    maxTime: '11:45pm',
+    defaultTime: '11:00am',
```
bledsoef, line 143: I am now hitting errors, but before I was I saw that the time for both time pickers was 11:00 am. We want to by default have the times an hour apart. For instance start time is noon and end time is 1pm.

bledsoef (changes requested):
Looks like merge conflicts might have caused some issues.

bledsoef:
There are some awful merge conflicts. Going to close this and reimplement the changes in development.

Issue #1294, opened 2024-07-23
Create Events Page: Split the toggle buttons into 2 sections, and make one of 3 required to create Event


thealphagurlux:
Spilt the toggle section under description in the create events page into 2 groups: 

Section 1: Is the event going to have food || Is RSVP required?

Section 2: make one of these three required: 
Training || Service hours || Bonner event -

The Bonner event toggle button will need to be added as well.

<img width="952" alt="image" src="https://github.com/user-attachments/assets/dc05e87b-ff16-443b-aa86-c6de8a403d5c">

bledsoef:
PR #1296 was merged

PR #1296, opened 2024-07-24
Create Events Page: Split the toggle buttons into 2 sections, and make one of 3 toggles required 

JohnCox2211:
The Bonners toggle will automatically toggle for Bonner events, but can be changed by the user if they wish and other event types can also use this toggle. The toggle state is saved between create event and edit event. The toggles are divided as requested into required and optional categories as well.

### DIFF
--- a/app/static/js/createEvents.js
+++ b/app/static/js/createEvents.js
@@ -98,40 +98,33 @@ $(document).ready(function () {
       $("#limitGroup").hide();
     }
   });
-
-  $("#checkIsTraining, #checkServiceHours, #checkBonners").on('submit', function () {
+  //Ensure only one toggle will be selected in the select one section
+  $("#checkIsTraining, #checkServiceHours, #checkBonners").on('click', function (event) {
     let trainingStatus = $("#checkIsTraining").is(":checked")
     let serviceHourStatus = $("#checkServiceHours").is(":checked")
     let bonnersStatus = $("#checkBonners").is(":checked")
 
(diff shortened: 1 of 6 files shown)


JohnCox2211:
https://github.com/BCStudentSoftwareDevTeam/celts/issues/1294 - github issue

https://trello.com/c/50TXOdHE - trello card

BrianRamsay (changes requested):
The bonner toggle only should be there for Bonner events.

Let's clean up the html classes and styles and try and simplify what controls the width of the components. Use the bootstrap grid only, if possible, for rows and columns.

app/static/js/createEvents.js
```diff
@@ -98,10 +104,25 @@ $(document).ready(function () {
       $("#limitGroup").hide();
     }
   });
+  // Determine which checkbox was clicked and its current checked status, uncheck others
+  $("#checkIsTraining, #checkServiceHours, #checkBonners").on('click', function (event) {
+    let $target = $(event.target);
+    $("#checkIsTraining, #checkServiceHours, #checkBonners").not($target).prop('checked', false);
```
BrianRamsay: we can just use `$(event.target)` in the `not()`

BrianRamsay (changes requested):
Let's clean up the css so we are relying more on bootstrap. the cutoff point should be xl, then you don't need min-width or the @media selector

BrianRamsay (approved):

PR #1297, opened 2024-07-25
Changes to the invite cohort

### LINKED ISSUE #1295: Bonner Cohort Invite
In the Bonner training event, the selected/invited cohort does not stay checked. 
image
image

Mbeweg:
### **_Issue Description_**
*******************************************
There is a bug where the Bonner cohort checkbox (`#checkBonners`) does not remain checked after creating, saving, or editing an event. While the invited Bonner scholars are successfully saved, the checkbox itself appears unchecked when revisiting the event editing page. This can cause confusion for users who may think the cohort was not properly invited.

Fixes issue #1295

### **_Changes_**
***********************************************************
- Added persistence to the Bonner cohort checkbox so that its state is saved when creating or editing an event.
- Modified the event form loading logic to correctly populate the `#checkBonners` checkbox with the previously saved state.
- Updated the event-form to retrieve the saved cohort invite status from the backend and set the checkbox accordingly.
  
### **_Testing_**
************************************************************
1. Prepare your local development environment and run the application.
   ```bash
   source setup.sh
   checkout BCI1
   database/reset_database.sh test
   flask run
   ```
2. Create a new event, select the Bonner cohort checkbox, and save the event. Verify that upon reloading the page or editing the event, the checkbox remains checked.

3. Edit an existing event without the Bonner cohort selected, invite Bonners, save, and confirm the checkbox is checked when reopening the event edit page.


bledsoef:
@Mbeweg is this in a reviewable state?

WackyWeaver:
I tried to test this a few different ways, and I followed the testing guide. The cohort checkbox did not remained checked for either a previously saved event or a new event. Also, the date picker seemed to break in the branch. I tried to pull development into the branch, but the date picker for the start date still did not seem to function properly. This was not an issue that I was having on Development.

WackyWeaver:
It looks good! I just wanted to have Finn take one last look before approval.

bledsoef (changes requested):
I think you guys did an excellent job getting it to save locally but I guess I have a few problems with the way this is currently being handled.

1. The checkbox state's are being stored in the session storage, but if I close out and reopen it, or if another user comes along and looks at it, they have no indication that the box was ever checked. I would recommend creating a new table in the database and creating a backend route to update that table.

2. I feel like once a checkbox has been clicked and saved it should be disabled and grayed out, I cannot really see a circumstance of when they would want to toggle the checkbox status since there is no action to take once it is unselected. If you do end up doing this it may be good to have an indicator (maybe a green "Invited" text next to the cohort names or something) that the cohort was invited. 

3. Whenever we save the page, even if no changes have been made to the cohort checkboxes, it will resubmit all of them meaning that the RSVP log is constantly getting updated with duplicate values.

4. Steven recommended possibly having the checkboxes and then an "invite" button that will trigger the invitations, update the database, and disable the checkboxes. That way we don't have to worry about the save button duplicating requests. We can just have the "save" form ignore the cohort stuff. This may require us not showing the "Invite cohorts" options until the edit page since we wouldn't be able to invite for an event that hasn't been created yet.

This is a lot, and more has to do with the lack of clarity in the issue than your implementation. If you need clarification feel free to find me or Brian. Nice work guys!

app/static/js/createEvents.js
```diff
@@ -484,3 +484,88 @@ $(".startDatePicker, .endDatePicker").change(function () {
...
+      cohortElement.closest('.form-check-label').addClass('text-muted');
+    }
+  });
+}
+
+function updateCohortStatus(eventId) {
+  getCohortStatus(eventId);
+}
+
+$(document).ready(function() {
```
bledsoef: there is already a document.ready defined at the start of this file. You should probably move the logic defined here into that one instead.

bledsoef: Line 264.

app/static/js/createEvents.js
```diff
@@ -484,3 +484,88 @@ $(".startDatePicker, .endDatePicker").change(function () {
...
+  const eventId = $("#eventId").val();
+  
+  if (eventId) {
+    updateCohortStatus(eventId);
+  }
+
+  $("#inviteCohortForm").on("submit", function(e) {
```
bledsoef: This handler will never run because inviteCohortForm does not exist

app/static/js/createEvents.js
```diff
@@ -484,3 +484,88 @@ $(".startDatePicker, .endDatePicker").change(function () {
...
+    e.preventDefault();
+    
+    const selectedCohorts = [];
+    $("input[name='cohorts[]']:checked").each(function() {
+      selectedCohorts.push($(this).val());
+    });
+    
+    if (selectedCohorts.length === 0) {
+      msgFlash("Please select at least one cohort to invite", "warning");
```
bledsoef: I don't have an answer to this but are we confident we always want to to invite at least one cohort?

app/static/js/createEvents.js
```diff
@@ -484,3 +484,88 @@ $(".startDatePicker, .endDatePicker").change(function () {
...
+      return;
+    }
+    
+    inviteCohorts(eventId, selectedCohorts).then(() => {
+      $("input[name='cohorts[]']:checked").each(function() {
+        $(this).prop('checked', false);
+      });
+      saveSelectedCohorts();
```
bledsoef: This function does not seem to exist anywhere in the codebase.

app/static/js/createEvents.js
```diff
@@ -484,3 +484,88 @@ $(".startDatePicker, .endDatePicker").change(function () {
...
+        </div>
+      `;
+      cohortElement.closest('.form-check').append(statusHtml);
+      cohortElement.prop('disabled', true);
+      cohortElement.closest('.form-check-label').addClass('text-muted');
+    }
+  });
+}
+
+function updateCohortStatus(eventId) {
```
bledsoef: Unless we are planning on doing some extra logic in this function we should probably just call getCohortStatus instead of defining this new function just to call getCohortStatus.

app/controllers/admin/routes.py
```diff
@@ -41,6 +42,57 @@
 from app.logic.spreadsheet import createSpreadsheet
 
 
+@admin_bp.route('/event/<eventId>/invite-cohorts', methods=['POST'])
```
bledsoef: I think we usually use camel case for route names.

app/controllers/admin/routes.py
```diff
@@ -41,6 +42,57 @@
...
+        })
+        
+    except Exception as e:
+        print(f"Error inviting cohorts: {e}")
+        return jsonify({
+            "success": False,
+            "message": "Error inviting cohorts"
+        }), 500
+
+@admin_bp.route('/event/<eventId>/cohort-status', methods=['GET'])
```
bledsoef: I think we usually use camel case for route names.

app/models/eventCohort.py
```diff
@@ -0,0 +1,15 @@
+from datetime import datetime
+from app.models import*
+from app.models.event import Event
+from app.models.bonnerCohort import BonnerCohort
+
+class EventCohort(baseModel):
+    event = ForeignKeyField(Event)
+    year = IntegerField()  
+    invited = BooleanField(default=False)
```
bledsoef: I don't mind this, but I would say we probably don't need the boolean. If they are put in this table we can infer they have been invited. Unless there is something else you were hoping to use this field for.

bledsoef (changes requested):
Looking over the code it is definitely looking better. A couple things:

The database breaks when trying to reset, this is because you are missing the following line in `migrate_db.sh`:

`pem add app.models.eventCohort.EventCohort`

I went ahead and added that line on my local machine to try and see if the rest worked and while you can successfully submit events it is not properly updating the events. This is probably due to some of the other comments I left.

Nice work guys!

app/controllers/admin/routes.py
```diff
@@ -41,6 +42,43 @@
 from app.logic.spreadsheet import createSpreadsheet
 
 
+@admin_bp.route('/event/<eventId>/inviteCohorts', methods=['POST'])
```
bledsoef: The content of this route is good but it may be better suited to go into the create/edit event routes we already have, specifically with the existing cohort logic here: https://github.com/BCStudentSoftwareDevTeam/celts/blob/af0501a75a6857ff0c85128fd15e9ca5052f977f/app/controllers/admin/routes.py#L136.

app/controllers/admin/routes.py
```diff
@@ -314,6 +352,13 @@ def eventDisplay(eventId):
     if eventData['program'] and eventData['program'].isBonnerScholars:
         requirements = getCertRequirements(Certification.BONNER)
         bonnerCohorts = getBonnerCohorts(limit=5)
+
+        invited_cohorts = list(EventCohort.select().where(
```
bledsoef: We should probably use camelcase here.

app/controllers/admin/routes.py
```diff
@@ -314,6 +352,13 @@ def eventDisplay(eventId):
...
+            EventCohort.event_id == eventId,
+        ))
+        invited_years = [inv.year for inv in invited_cohorts]
```
bledsoef: And here.

app/static/js/createEvents.js
```diff
@@ -484,3 +489,68 @@ $(".startDatePicker, .endDatePicker").change(function () {
...
+
+inviteCohorts(eventId, selectedCohorts)
+  .then(() => {
+    $("input[name='cohorts[]']:checked").prop('checked', false);
+  })
+  .catch(error => {
+    msgFlash("Error inviting cohorts: " + error.message, "danger");
+  });
+ 
+function inviteCohorts(eventId, selectedCohorts) {
```
bledsoef: If we make the necessary changes in the routes then we don't need any of the javascript logic in this function since the data is already properly being passed in to the createEvent route.

app/static/js/createEvents.js
```diff
@@ -484,3 +489,68 @@ $(".startDatePicker, .endDatePicker").change(function () {
...
+  return $.ajax({
+    type: "GET",
+    url: `/event/${eventId}/cohortStatus`,
+    success: function(response) {
+      displayCohortStatus(response.invited_cohorts);
+    }
+  });
+}
+
+function displayCohortStatus(invitedCohorts) {
```
bledsoef: This looks good.

bledsoef (changes requested):
Looking good, just have some logic that needs to be condensed but overall I think its getting pretty close.

app/templates/events/createEvent.html
```diff
@@ -237,9 +237,13 @@ <h1 id="pageTitle">{{page_title}}</h1>
           <label class="form-label fw-bold">Invite Cohort(s)</label>
           {% for year,students in bonnerCohorts.items()|reverse %}
           <div class="form-check">
-            <input class="form-check-input" type="checkbox" value="{{year}}" name="cohorts[]" id="cohort-{{year}}">
-            <label class="form-check-label text-truncate w-100" for="cohort-{{year}}">{{year}}-{{year+1}}
-              <span class="text-muted">({{students|map(attribute='fullName')|join(", ")}})</span>
+            <input class="form-check-input" type="checkbox" value="{{year}}" name="cohorts[]" id="cohort-{{year}}"
+            {{ 'checked' if invited_years and year|string in invited_years }}>
```
bledsoef: If this is checked we should probably disable it so we aren't allowing people to uncheck it since we don't have the functionality to do anything then.

app/controllers/admin/routes.py
```diff
@@ -40,6 +42,61 @@
 from app.controllers.admin import admin_bp
 from app.logic.spreadsheet import createSpreadsheet
 
+@admin_bp.route('/event/<eventId>/cohortStatus', methods=['GET'])
```
bledsoef: Could we potentially move this into the edit/create event route where it is rendering the page already? That way we don't have to do an extra request?

app/controllers/admin/routes.py
```diff
@@ -40,6 +42,61 @@
...
+            'message': 'Event not found'
+        }), 404
+    except Exception as e:
+        print(f"Error getting cohort status: {e}")
+        return jsonify({
+            'status': 'error',
+            'message': 'Internal server error'
+        }), 500
+
+def inviteCohortsToEvent(event, cohort_years):
```
bledsoef: We should move this to logic.

bledsoef (changes requested):
Generally looks good! A couple issues though:
1. We should disable the checkbox once they have checked and saved it.
2. Make sure your variables are camel case.
3. Inviting a cohort on create works well, but inviting on edit does not seem to be working. It will sometimes overwrite existing overwritten cohorts and other times not invite the ones you select at all.

Nice work!

app/controllers/admin/routes.py
stevensonmichel: Done

app/controllers/admin/routes.py
stevensonmichel: Done

app/controllers/admin/routes.py
stevensonmichel: Done

app/controllers/admin/routes.py
stevensonmichel: Done

app/static/js/createEvents.js
stevensonmichel: Done

app/controllers/admin/routes.py
stevensonmichel: Done

app/controllers/admin/routes.py
```diff
@@ -314,18 +327,34 @@ def eventDisplay(eventId):
...
         requirements = getCertRequirements(Certification.BONNER)
-        bonnerCohorts = getBonnerCohorts(limit=5)
+        rawBonnerCohorts = getBonnerCohorts(limit=5)
+        bonnerCohorts = {}
+        
+        for year, cohort in rawBonnerCohorts.items():
+            if cohort:
+                bonnerCohorts[year] = cohort
+                
+        print(1000*"*")
```
bledsoef: Debug statements >:(

app/logic/bonner.py
```diff
@@ -83,11 +84,27 @@ def rsvpForBonnerCohort(year, event):
                                       .where(BonnerCohort.year == year),
                                       [EventRsvp.user, EventRsvp.event, EventRsvp.rsvpTime]).on_conflict(action='IGNORE').execute()
     
+# def updateEventBonnerCohort(year, event):
```
bledsoef: Commented out code >:(

app/logic/events.py
```diff
@@ -706,3 +708,77 @@ def copyRsvpToNewEvent(priorEvent, newEvent):
...
+            createActivityLog(f"Added Bonner cohorts {cohortList} for newly created event {event.name}")
+
+        return True, "Cohorts successfully added to new event", invitedCohorts
+
+    except Exception as e:
+        print(f"Error inviting cohorts to new event: {e}")
+        return False, f"Error adding cohorts to new event: {e}", []
+
+
+def updateEventCohorts(event, cohortYears):
```
bledsoef, line 663: I like the functionality but we actually do not want to allow them to uninvite a cohort so a lot of this may not be necessary.

app/templates/events/createEvent.html
```diff
@@ -235,11 +235,17 @@ <h1 id="pageTitle">{{page_title}}</h1>
...
         <div class="form-group">
           <label class="form-label fw-bold">Invite Cohort(s)</label>
-          {% for year,students in bonnerCohorts.items()|reverse %}
+          {% for year, students in bonnerCohorts.items() | reverse %}
           <div class="form-check">
-            <input class="form-check-input" type="checkbox" value="{{year}}" name="cohorts[]" id="cohort-{{year}}">
-            <label class="form-check-label text-truncate w-100" for="cohort-{{year}}">{{year}}-{{year+1}}
-              <span class="text-muted">({{students|map(attribute='fullName')|join(", ")}})</span>
+            <input class="form-check-input" type="checkbox" value="{{year}}" name="cohorts[]" id="cohort-{{year}}"
+            {% if invitedYears and year|string in invitedYears %}
```
bledsoef, line 211: We need some logic to have the checkbox disabled if they have already been invited.

tests/code/test_events.py
```diff
@@ -1407,3 +1409,103 @@ def test_copyRsvpToNewEvent():
...
+                                            year = cohortYear2)
+            
+            EventCohort.create(event = event,
+                                            invited_at = datetime.now(),
+                                            year = cohortYear3)
+            
+            cohortYears = ["2020", "2022", "2023"]
+            
+            success, message, updatedCohorts = updateEventCohorts(event, cohortYears)
+            print(1000*"*")
```
bledsoef: Debug statements >:(

tests/code/test_events.py
```diff
@@ -1407,3 +1409,103 @@ def test_copyRsvpToNewEvent():
             assert len(EventRsvp.select().where(EventRsvp.event_id == newEvent)) == 2
 
             transaction.rollback()
+            
+            
+@pytest.mark.integration
+def test_inviteCohortsToEvent():
```
bledsoef, line 1311: Test looks good but we be able to verify other things? Like maybe that RSVPs were sent to all individuals in the cohort? This may require creating a new cohort...

bledsoef (changes requested):
Mostly looks good! Just some debug statements and a little bit of functionality we don't want.

app/logic/bonner.py
WackyWeaver: Fixed!

app/controllers/admin/routes.py
WackyWeaver: Resolved!

tests/code/test_events.py
WackyWeaver: Fixed!

app/logic/events.py
stevensonmichel: Fixed

app/templates/events/createEvent.html
stevensonmichel: Fixed

tests/code/test_events.py
stevensonmichel: Given that the RSVPs are sent using a function in `bonner.py`, I was expecting the tests to be in `test_bonner.py`. But it looks like the one who wrote the logic for adding RSVPs did not create tests for them.

app/templates/events/createEvent.html
stevensonmichel: Fixed

tests/code/test_events.py
stevensonmichel: I wrote the test for `addBonnerCohortToRsvp`, the one for `eventRsvp` has already been written.

bledsoef (dismissed):
LGTM!

BrianRamsay (dismissed):

PR #1305, opened 2024-07-31
Toggle correction for All Volunteer Event creation

JohnCox2211:
Not having toggles on this page does not interfere with being able to save the event you are creating (previously you would get an error even though they were not there.)

original issue: https://github.com/BCStudentSoftwareDevTeam/celts/pull/1296


bledsoef (approved):
Looks good and works fine, nice work!

PR #1332, opened 2024-09-04
Fix start date

### LINKED ISSUE #1311: "Allow Start Date to be in the Past" toggle is not working
There is a problem with the toggle that used to work. When we create an event, when we click on the toggle "Allow start date to be in the past," the feature doesn't allow the datepicker to select the past dates. The past dates are still greyed out, which is the same as if we did not select the toggle. It works for when we select the event with multiple time offerings only. Investigate features for both toggles.

vungc:
## Issue Description

Fixes #1311 
- When the allow start date to be in the past is toggled, the past dates are still disabled and greyed out.

## Changes

- modified the click function on #allowPastStart to be specific to the datepicker on the main page
image

## Testing

- run flask, create event, select from program
- select a start date, turn on the "allow start date to be in the past", and try to select a date before today


bledsoef (approved):
LGTM!

PR #1334, opened 2024-09-05
Refactor templates

### LINKED ISSUE #1131: Refactor the templates directory
Currently there is no method to the madness that is the templates directory. Why are certain templates that have "event" in their name not in any directory, much less the '"events" folder? Why are some camelcase and some snake case? Why do some have no case? So many questions, so few answers. Can we organize these a little better so we they are consistent and so we can know where to look for things?

Screenshot_20240313_135744

vungc:
## Issue Description

Fixes #1131 
- templates directory is messy and some files are in the wrong folder or inconsistent in format.

## Changes

- All events related files are now located in the events folder
- all files are formatted in camel case
- the outdated `addParticipants` html and js files have been deleted
- the outdated `outsideParticipants` model has been deleted

## Testing

- Run flask
- Run our test suite
- Check that Event List, Create Event, and other pages related to events and those under Admin still work.


bledsoef (changes requested):
I think generally, I like the refactor. However, addParticipants could likely be moved to events. Maybe also a folder for macros? That way the only things not in directories is base, sidebar, and contributors, which makes a lot of sense to me.

WackyWeaver (dismissed):
Looks good to me!

bledsoef:
@ojmakinde @vungc looks great! Have some merge conflicts and a weird settings.json file that snuck in there, but other than that just ping me and I will merge it.

bledsoef (approved):
LGTM

PR #1341, opened 2024-09-14
Datepicker issue multiple course offering

### LINKED ISSUE #1333: Datepicker Pop-Ups on Multiple Event Offerings
When trying to use the datepicker to choose a date on an event with multiple offerings, there are two issues:
1. The date picker on multiple event offerings is visually different from the one we have for creating single events.
2. After clicking the date picker on multiple event offerings, the date picker for Single Events and Multiple Events sometimes show up together, which is a strange bug.
3. Additionally, improve the date picker by making the entire body clickable. The picker only pops up if the calendar icon is clicked.

_Single Event Date Picker:_

_Multiple Event Date Picker:_

_Both showing up when clicking the date picker for multiple events:_

ali-ramazani:
## Datepicker Issue in Create Event Modal 

Fixes #1333 
- The issue was when trying to use the datepicker to choose a date on an event with multiple offerings, the date picker would show below the modal. Initially, the assumption was the datepicker in the modal was triggering the datepicker outside of the modal. After hours of investigation, I concluded that the the correct datepicker was being displayed, but there was a z-index issue with the native datepicker module. 

## Changes

- Disabled the default browser date picker. 
- Increased the z-index of the datepicker in the modal. 

## Testing

- I verified the code works for both Chrome and Firefox, two major browsers. 
- I verified the responsiveness by adjusting the screen.


ali-ramazani:
However, there's another issue that needs to be fixed in another ticket. The dynamically added events in the modal refer to the first event.

vungc:
There are merge conflicts and causing the main page to not open properly. The initial issue also does not seem to be resolved, the datepicker is not working properly for multiple offering events. For example, if I add another event and I try to change the datpicker, it changes it for the first event always. I am not sure if this is because of the merge conflicts.

vungc (changes requested):
There are merge conflicts and causing the main page to not open properly. The initial issue also does not seem to be resolved, the datepicker is not working properly for multiple offering events. For example, if I add another event and I try to change the datpicker, it changes it for the first event always. I am not sure if this is because of the merge conflicts.

WackyWeaver:
It seems like there have already been changes made to the date picker inside of multiple time offering on development since this PR was started.

Issue #1343, opened 2024-09-17
Datepicker allows past dates in multiple time offerings


WackyWeaver:
## Problem
Within the multiple time offerings modal, the time picker allows for users to create events in the past without toggling "Allow start date to be in the past"

![image](https://github.com/user-attachments/assets/88afe204-ff2a-432a-8f7f-4f25d5560f1d)

## Consider
Having the multiple time offerings respect the past dates toggle.

WackyWeaver:
![](https://github.trello.services/images/mini-trello-icon.png) [Datepicker allows past dates in multiple time offerings #1343](https://trello.com/c/5SdSXlm1/826-datepicker-allows-past-dates-in-multiple-time-offerings-1343)

PR #1346, opened 2024-10-02
Date picker 1343

doucoureb:
### **_Issue Description_**
********************************************************
Fixes issue #1343
Within the multiple time offerings modal, the time picker previously allowed users to create events in the past without toggling the "Allow start date to be in the past" option.

### **_Changes_**
**************************************************************
- Updated the time picker logic to enforce the "Allow start date to be in the past" toggle.
- Implemented a validation check that only permits event creation in the past when the toggle is enabled.
- Added a flash message when attempting to select a past date without enabling the toggle.

### **_Testing_**
*****************************************************************
- Verified that the time picker correctly restricts past date selections when the "Allow start date to be in the past" option is disabled.
- Conducted tests using the following steps:
- Checkout the branch datePicker_1343
- Run with flask run
- Open the modal and try selecting a past date without enabling the toggle.
- Enable the toggle and confirm that past dates are now selectable.

The issue has been fixed! The multiple time offerings modal now properly respects the past dates toggle.


WackyWeaver (dismissed):
Looks good to me!

app/static/js/createEvents.js
```diff
@@ -173,48 +184,91 @@ $('#multipleOfferingSave').on('click', function() {
...
+  $('#textNotifier').removeClass().addClass('alert alert-' + type).text(message).show();
+  $('#textNotifierPadding').addClass('pt-5');
+  setTimeout(function() {
+    $('#textNotifier').hide();
+    $('#textNotifierPadding').removeClass('pt-5');
+  }, 5000);
+}
 
 // Save the offerings from the modal to the hidden input field
 function saveOfferingsFromModal() {
```
bledsoef: Since you deleted the call to this function, it is never used. Can it be removed?

app/static/js/createEvents.js
```diff
@@ -173,48 +184,91 @@ $('#multipleOfferingSave').on('click', function() {
     }
   }
 
+  // Add past date validation
+  datePickerInputs.each(function(index, element) {
+    if (!allowPastStart && isDateInPast($(element).val())) {
+      $(element).addClass('border-red');
+      hasPastDates = true;
```
bledsoef: I like how this works but not sure how I feel about the variable name. Just because the variable `hasPastDates` is false even when there are actually past dates, but only when the allowPastStart is false. Maybe consider renaming to something like `hasInvalidDates`? Just a nit, though.

app/static/js/createEvents.js
```diff
@@ -173,48 +184,91 @@ $('#multipleOfferingSave').on('click', function() {
...
     $("#checkIsMultipleOffering").prop('checked', true);
-    // Remove the modal and overlay from the DOM
     $('#modalMultipleOffering').modal('hide');
-    $('.invalidFeedback').css('display', 'none');
-    $('#textNotifierPadding').removeClass('pt-5');
     msgFlash("Multiple time offering events saved!", "success");
   }
 });
-
+function showModalFlashMessage(message, type) {
```
bledsoef: While I actually think this is a good way to do it, is there a reason you didn't use our built in msgFlash function?

bledsoef (changes requested):
Generally looks good! Couple things here and there.

app/static/js/createEvents.js
doucoureb: Good suggestion, Fixed!

app/static/js/createEvents.js
doucoureb: When I use the built in msgFlash function, the flash message appears behind the multiple time offering window so the user won't be able to see it.

app/static/js/createEvents.js
doucoureb: I'm not sure I understand this

WackyWeaver: Since you deleted the call to this function, there is nowhere else in the code base that uses this function. When I checked out the branch, the function was even grayed out. Can the function be removed?

app/static/js/createEvents.js
WackyWeaver: It would be good to add a comment, so that next time, we understand why we needed the extra function.

WackyWeaver (changes requested):
Mostly good, just some small touch up!

app/static/js/createEvents.js
```diff
@@ -173,49 +183,64 @@ $('#multipleOfferingSave').on('click', function() {
...
-    // Remove the modal and overlay from the DOM
     $('#modalMultipleOffering').modal('hide');
-    $('.invalidFeedback').css('display', 'none');
-    $('#textNotifierPadding').removeClass('pt-5');
     msgFlash("Multiple time offering events saved!", "success");
   }
 });
 
+// Custom flash for modal: ensures message visibility when modal is open
+function showModalFlashMessage(message, type) {
```
bledsoef: I noticed we have a displayNotification function that has a very similar functionality and is used in other locations. I checkout the branch and tried it it out and it looks to work the same is this function. Could we potentially use that instead so we don't have to be doing message flashes in multiple ways?

bledsoef (dismissed):

WackyWeaver (approved):
Great Job, guys! Nice work on the PR!

PR #1349, opened 2024-10-20
Multiple Offerings and Recurring Events Refactor

### LINKED ISSUE #1335: Refactor Multiple Event Offering and Recurring Event logic
TBD

ojmakinde:
## Issue Description

Fixes #1335
- Most of the existing logic for our multiple offerings and recurring events are duplicates. We need to merge both functionalities into one, and change the workflow on the frontend.

## Changes

- Added a new integer column, seriesId, into the model.
- Refactored logic that used MultipleOffering and RecurringId to use the seriesId, merging logic where necessary.
- Reworked the UI on Create Event to accomodate the change, using a single toggle for both multiple offerings and recurring series, instead of our previous two toggles for multiple offerings and recurring events.
- Updated tests.

## Testing

- Run the test suite and ensure it is functional.
- Step through parts of the application and ensure that they are functional despite the many introduced changes.


thealphagurlux:
The UI for the recurring events looks a little bit awkward when you first click on it and the style of the button is not consistent with the UI with the rest of the site.

ss1

I suggest changing the text of the button to something like "add another offering" and have the pop-up look populate like this by default

ss2

One last thing the button only appears in the center of the window, but you can click the highlighted area and it still activates its function.

ss3

Karina-Agliullova:
There is an issue with the toggle for multiple offerings, specifically, with the weekly option. If on the multiple offerings modal, we select weekly toggle, then we cancel it so that modal closes, then we reopen the modal for multiple offerings, and click on the weekly toggle, it shows the unconstructed table with Generated Events, but it is empty because I did not yet create any event, so it seems awkward to have a table with column names but no information. The table shouldn't be generated at all if only I did not already create the weekly event with the information.
Screenshot (25)

Karina-Agliullova:
Another issue with the weekly toggle for multiple offerings is that it saves all the information that I add to it. So if I create an event with a weekly toggle, and then I cancel it, and then reopen the modal, it will still show the modal with all the information I input previously, which shouldn't be the case unless we save it

Karina-Agliullova:
Screenshot (27)
When I open the modal, and I do not input any information, and hit Save, it executes successfully even though it shouldn't. It creates an empty table with no information, just the column names. And that is the same way if I click on the toggle for weekly events in the modal, and I don't input any information, and hit Save, it says that Events created, even though it is empty

Karina-Agliullova:
Another problem is with the error message in the modal. If I open the modal, and I click on the add button to add events, and I do not input any information, and hit Save, it highlights everything in red, which is good, and leaves an error flash warning message. That is good. But when I cancel the modal and reopen it, there is a warning message left still even though it is not applicable to the new modal opening. 
Screenshot (29)

Karina-Agliullova:
Also, the spacing for the weekly option in multiple offerings seems too tight. It would be nice to give it some space between the boxes and texts

thealphagurlux:
When making a event occur more than once the name input into the pop-up window will be what the event is displayed as in the event list
ss1

ss2

ss3



When saving the recurring event the event name on the create event page should probably reflect whatever changes are made in the pop-up window

Karina-Agliullova:
When I create an event, and I add multiple time offering, but I don't input all the info, like I only input Event name, but no dates for multiple offerings, and it still saves it even though it displays the empty table, and I add all the other info outside of modal, like place and description, and then save an event, it throws an error
Screenshot (30)

Karina-Agliullova:
Something is going on with the datepicker in the modal. I selected the date for today but the time is later than what it is right now, and it is throwing me a warning, saying that I picked the date in the past
Screenshot (31)

ojmakinde:
> Something is going on with the datepicker in the modal. I selected the date for today but the time is later than what it is right now, and it is throwing me a warning, saying that I picked the date in the past Screenshot (31)

I just tested this in development and the issue exists in development as well. Will fix in this PR though. Thanks for catching that.

app/controllers/admin/routes.py
```diff
@@ -383,21 +381,22 @@ def cancelRoute(eventId):
...
     try:
-        events = session['lastDeletedEvent']
-        for eventId in events: 
+        eventIds = session['lastDeletedEvent'] #list of Ids of the events that got deleted
+        for eventId in eventIds: 
             Event.update({Event.deletionDate: None, Event.deletedBy: None}).where(Event.id == eventId).execute()
             event = Event.get_or_none(Event.id == eventId)
-        recurringEvents = list(Event.select().where((Event.recurringId==event.recurringId) & (Event.deletionDate == None)).order_by(Event.id))          
-        if event.recurringId is not None:
+        repeatingEvents = list(Event.select().where((Event.isRepeating) & (Event.deletionDate == None)).order_by(Event.id))
```
BrianRamsay: You don't use the series id, so this will get ALL repeating events and you'll rename them.

app/logic/events.py
```diff
@@ -66,34 +67,33 @@ def deleteEvent(eventId):
...
         event = Event.get_or_none(Event.id == eventId)
         if event:
-            if event.recurringId:
-                recurringId = event.recurringId
-            allRecurringEvents = list(Event.select(Event.id).where(Event.recurringId == recurringId).order_by(Event.startDate))
-            eventId = allRecurringEvents[0].id
+            if event.seriesId:
+                seriesId = event.seriesId
+            allSeriesEvents = list(Event.select(Event.id).where(Event.seriesId == seriesId).order_by(Event.startDate))
+            eventId = allSeriesEvents[0].id
```
BrianRamsay: This if statement is the same as it was before, but what happens if there is no seriesId? this line will fail. All of it should be in the if statement I think

app/logic/events.py
```diff
@@ -113,23 +113,24 @@ def attemptSaveMultipleOfferings(eventData, attachmentFiles = None):
...
     allSavesWereSuccessful = True
     
-    # Creates a shared multipleOfferingId for all offerings to have
-    multipleOfferingId = calculateNewMultipleOfferingId()
+    seriesId = calculateNewSeriesId()
 
     # Create separate event data inheriting from the original eventData
-    multipleOfferingData = eventData.get('multipleOfferingData')
+    seriesData = eventData.get('seriesData')
+    isRepeating = bool(eventData.get('isRepeating'))
```
BrianRamsay, line 121: should we be using `preprocessEventData`? It's already doing the cleanup of these fields, right?

database/oneOff/merge-series-table.sql
```diff
@@ -0,0 +1,31 @@
...
+    declare cur1 cursor for select distinct multipleOfferingId from event where multipleOfferingId is not null;
+    declare continue handler for sqlstate '02000' set done = 1;
+    update event set seriesId=recurringId where recurringId is not null;
+    set seriesIdvalue = coalesce((select max(seriesId) from event), 0) + 1;
+
+    update event set seriesId=recurringId, isRepeating=1 where recurringId is not null;
+    open cur1;
+
+    fetch cur1 into mofferingid;
+    while done = 0 DO
```
BrianRamsay, line 16: Does done ever change?

database/oneOff/merge-series-table.sql
```diff
@@ -0,0 +1,31 @@
+DELIMITER //
+
+create procedure handleSeries()
+begin
+    declare mofferingid int default 0;
+    declare seriesIdvalue int default 0;
+    declare done int default 0;
+    declare cur1 cursor for select distinct multipleOfferingId from event where multipleOfferingId is not null;
+    declare continue handler for sqlstate '02000' set done = 1;
+    update event set seriesId=recurringId where recurringId is not null;
```
BrianRamsay: This line can be removed and the one on line 13 moved up here

database/oneOff/merge-series-table.sql
```diff
@@ -0,0 +1,31 @@
...
+    set seriesIdvalue = coalesce((select max(seriesId) from event), 0) + 1;
+
+    update event set seriesId=recurringId, isRepeating=1 where recurringId is not null;
+    open cur1;
+
+    fetch cur1 into mofferingid;
+    while done = 0 DO
+        
+        select mofferingid;
```
BrianRamsay: What is this line doing? If it's just for output, let's make the output a little more verbose

BrianRamsay (changes requested):

app/controllers/admin/routes.py
ojmakinde: Ah, thanks for the catch.

database/oneOff/merge-series-table.sql
ojmakinde: Yeah, I only put it there to "print" that the id was in fact incrementing over iterations. It's not useful anymore, given that the function now works. I will remove it.

database/oneOff/merge-series-table.sql
ojmakinde: Yes, it does. On line 9, its value is set to 1 once the iteration runs out of values to loop over.

`declare continue handler for sqlstate '02000' set done = 1;`

app/logic/events.py
ojmakinde: Resolved. Also added error handling if event does not exist

database/oneOff/merge-series-table.sql
ojmakinde: Solved.

PR #1353, opened 2024-10-24
Add third option for event type

BrianRamsay:
CELTS needs a third option for event types. This PR fixes the event creation to allow this, but doesn't dig into how we are separating events in the Event List.


WackyWeaver (approved):
Everything looks good to me!

PR #1355, opened 2024-10-25
Interface graduation management 

### LINKED ISSUE #1315: Create interface to allow Admin to mark students as graduated
We recently created functionality to allow students to be marked as "graduated" however this is pretty meaningless to us right now because we are unable to actually set someone as graduated. So, we need to create an interface that admin can access and mark students as graduated. I would recommend something kind of like "minor management" that we currently have, but instead there is a list of all students (with the ability to search) where we can check off students as graduated.

**Considerations**
1. How can we make this scalable and prevent it from getting super cluttered as the amount of students who are graduated grow?

My suggestion might be to hide students that are graduated and only allow them to searchable, or also implement a toggle like for past events which will show them.

2. Should we add this to an existing page or create a new sidebar option with something like "Student Management" or "Graduation Management".

Unless we were able to just mark a student as graduated via their profile (which would be kind of inconvenient to do in mass. I would recommend creating a new page, or adding a modal in Settings under User Management.

3. Should we assume Graduated student cannot attend login and thus we should not have to add any additional logic to deactivate them when they are marked as graduated? How should we adjust their user profile while still leaving it visible?

Would definitely talk to Brian about this before picking it up.

thealphagurlux:
## Issue Description

Fixes #1315 

- The purpose of this issue is to create a page that allows celts admins to mark students as graduated in the data base.

- The page should include a way for the user to filter students by searching for individual users, by Bonner Cohort (and what academic year they are associated with), as well as students who are participating in the CCE minor (whether they have completed it or made any progress towards it. ).

## Changes
- Created a new html, js, logic, and controller file called graduationManagement.
- Created new functions for created setting the graduation status of a user and fetching all eligible users.
- Added a new field in the database `hasGraduated`
- Created a toggle that allows you to hide or show all graduated students.
- Stored all filter and toggle data in session state so that a page reload doesn't get rid of all your changes.

## Testing
- Checkout `interface-graduation-management`
- Head to the Graduation Management page in the Admin sidebar
- Boot up the page and make sure all students are seniors
- Mark students as graduated and refresh the page to make their state is permanent.
- Try to filter and toggle various buttons and make sure they properly filter the data as well as being properly stored in the session state. Make sure you can refresh and changes are maintained.


app/controllers/admin/graduationManagement.py
```diff
@@ -0,0 +1,72 @@
...
+    bonnercohorts = getBonnerCohorts()
+
+    return render_template('/admin/graduationManagement.html', 
+                           users = users, 
+                           bonnercohorts = bonnercohorts,
+                           CCEstudents = CCEusers)
+
+
+@admin_bp.route('/<username>/hasGraduated/', methods=['POST'])
+def hasGraduated(username):
```
WackyWeaver: The "has" at the beginning of the function was a bit confusing, because it is an attributing verb. I usually associate it with a check, like "isDead" or "hasDied". Maybe change this use a better helping verb like "setGraduationStatus" or something similar.

app/controllers/admin/graduationManagement.py
```diff
@@ -0,0 +1,72 @@
...
+            return "", 200
+        else:
+            return "", 500
+
+    except Exception as e:
+        print(e)
+        return "Error Updating Graduation Status", 500
+
+@admin_bp.route('/<username>/hasNotGraduated/', methods=['POST'])
+def hasNotGraduated(username):
```
WackyWeaver: same here.

app/controllers/admin/graduationManagement.py
```diff
@@ -0,0 +1,72 @@
...
+            return "", 200
+        else:
+            flash("Error!", "Failed to update graduation status.")
+            return "", 500
+    except Exception as e:
+        print(e)
+        return "Error Updating Graduation Status", 500
+
+@admin_bp.route("/gradStudentsxls/<filterType>", methods=['GET'])
+def gradsxls(filterType):
```
WackyWeaver: Maybe make this camelCase to match the rest of the functions.

app/templates/admin/graduationManagement.html
```diff
@@ -0,0 +1,77 @@
...
+    </button>
+    <div class="dropdown-menu">
+      <a class="dropdown-item" data-filter="all" >All</a>
+      <a class="dropdown-item" data-filter="bonner" >Bonner Students</a>
+      <a class="dropdown-item" data-filter="cce" data-cce = "{{CCEstudents}}" >CCE Minor</a>
+    </div>
+  </div>
+  <div class="btn-group" id = "bongroup">
+    <button type="button" class="btn btn-primary dropdown-toggle bonner-filter" id = "cohortFilter" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"  >
+      Bonner Cohorts
```
WackyWeaver: I believe this is the line, or it is near the line responsible for this issue. I thought it was a bit confusing that this made the "academic year" selection drop down default to "Bonner Cohort instead of the current year range. 
image

app/static/js/graduationManagement.js
```diff
@@ -0,0 +1,193 @@
...
+        type: "POST",
+        url: graduationURL,
+        success: function(response) {
+
+            if ($('.alert').length > 1 ){
+
+                $('.alert').alert('close'); 
+            };
+            console.log("Graduation status updated successfully!");
+            msgFlash("Graduation status updated successfully!", "success");
```
WackyWeaver: This makes the flash messages a bit overwhelming if a user was attempting to graduate a lot of students at once. Also, when hitting "Select All" button, the message flashes for every user on the page individually. 
image

app/static/js/sidebar.js
```diff
@@ -1,5 +1,7 @@
 $(document).ready(function() {
 
+  $('.alert').alert('close');// close excess flash messages from graduation managment
```
WackyWeaver: I was mentioning excess flash messages, and this might be the solution. I did not see it effect the flash notification I was getting.

database/tmp-backups/2024-11-15-backup.sql
WackyWeaver, line 1: Why is this empty file here?

WackyWeaver (dismissed):
Almost to the finish line! I just had a few immediate changes I thought would be necessary, and there were one or typos I saw. I think that the flash messages still need to be handled, but otherwise, I'm excited to see this merged!

thealphagurlux:
The above naming conventions and issues with the multiple flash messages should be in working order. The sql file has also been removed.

app/templates/admin/graduationManagement.html
```diff
@@ -0,0 +1,77 @@
...
+      <a class="dropdown-item" data-filter="cce" data-cce = "{{CCEstudents}}" >CCE Minor</a>
+    </div>
+  </div>
+  <div class="btn-group" id = "bongroup">
+    <button type="button" class="btn btn-primary dropdown-toggle bonner-filter" id = "cohortFilter" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"  >
+      Bonner Cohorts by Year
+    </button>
+    <div class="dropdown-menu">
+      {%for year in bonnercohorts%}
+        <a class="dropdown-item dropdown-item-new" data-filter = "bonner" data-cohort-year= "{{year}}" data-cohort-users = "{{bonnercohorts[year]}}" >{{year}}-{{year+1}}</a>
```
WackyWeaver: Since you are filtering for years with Bonner cohorts, I was running into issues if I wanted to look for a CCE minor student that wasn't a Bonner. The drop down created the same effect as that of the Bonner students filter.

app/templates/admin/graduationManagement.html
```diff
@@ -0,0 +1,77 @@
...
+{% block app_content %}
+<h1 class="text-center mb-5">Graduation Managment</h1>
+<div class="col-md-3 me-auto"></div>
+<div class="d-flex justify-content-between align-items-center mb-3">
+  <div class="btn-group">
+    <button type="button" class="btn btn-primary dropdown-toggle" id = "main-filter" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" >
+      All
+    </button>
+    <div class="dropdown-menu">
+      <a class="dropdown-item" data-filter="all" >All</a>
```
WackyWeaver: I'm a little curious of this feature. If I select a specific year with the other drop down, the "all" filter then would take priority. Would we want "all" to be separate from "all of a specific year"? 
Intuitively, I was assuming that the left drop down was filtering type of student and the center filter was for _**just**_ year.

app/templates/admin/graduationManagement.html
```diff
@@ -0,0 +1,77 @@
...
+        <tr>
+            <th>Name</th>
+            <th>Graduated</th>
+            <th>Class Level</th>
+        </tr>
+    </thead>
+  <tbody>
+    {% for user in users %}
+    <tr data-student-type="{% if user.isBonnerScholar %}bonner{% else %}all{% endif %}" data-username="{{user.username}}">
+      <td><input type="checkbox" class="graduated-checkbox" id = "gradCheck" data-username ="{{ user.username }}" {% if user.hasGraduated %}checked{% endif %}></td>
```
WackyWeaver: When I am flask running, I don't see the checkbox anymore. I was wondering if this idea was scratched.

WackyWeaver (commented):
Here's the situation. You are so very close, but there are still a few things that need tweaked. The drop downs for academic year and student type need to respect each other when filtering the table. Also, I still would likely recommend renaming the "CheckIfNotGraduated" to something a bit more clear. I think you've made some good progress, and I like he filtering options. All together, keep it up!

app/templates/admin/graduationManagement.html
thealphagurlux: I don't think this is a flask problem, as I am able to see them. Double check as the check boxes are on the far left.

app/templates/admin/graduationManagement.html
thealphagurlux: The same user does appear in the table because they are both a Bonner and a CCE minor participant. In any case I think should be a different issue not related to this PR due to the lack of clarification from CELTS on what exactly they'd like, although I think that is a good feature that can be added.

app/logic/graduationManagement.py
```diff
@@ -0,0 +1,99 @@
...
+from app.models.user import User
+from app.models.bonnerCohort import BonnerCohort
+from app.logic.minor import getMinorProgress
+
+def getGraduationManagementUsers():
+    """
+    Function to fetch all senior students along with their CCE Minor Progress and Bonner Status 
+    """
+
+    elibibleUsers = (User.select(User.username, User.hasGraduated, User.classLevel, User.firstName, User.lastName, BonnerCohort.year)
```
hoerstl: Some might think that `eligibleUsers` is how it should be but I prefer `elibibleUsers`

app/logic/graduationManagement.py
```diff
@@ -0,0 +1,99 @@
...
+    return graduationManagementUsers
+
+
+def setGraduatedStatus(username, status):
+    """
+    Update a students graduation status based on the parameter status.
+    """
+    gradStudent = User.get(User.username == username)
+    
+    gradStudent.hasGraduated = int(status)
```
hoerstl, line 39: Since hasGraduated is a boolean field, I think it would be more accurate to cast status to a boolean instead
`gradStudent.hasGraduated = bool(status)`

app/templates/admin/graduationManagement.html
```diff
@@ -0,0 +1,82 @@
...
+    <div class="col col-xl col-12 d-flex justify-content-xl-start justify-content-center">
+        <div id="categoryDropdown">
+          <button type="button" class="btn btn-primary dropdown-toggle me-1" id="mainFilter" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+            Filter
+          </button>
+          <div class="dropdown-menu">
+            <a class="dropdown-item main-dropdown-item" data-filter="all">All</a>
+            <a class="dropdown-item main-dropdown-item" data-filter="bonner">Bonner Students</a>
+            <a class="dropdown-item main-dropdown-item" data-filter="cce" data-cce="{{CCEstudents}}">CCE Minor</a>
+          </div>
```
hoerstl, line 30: This works but I feel it should be rewritten as a select attribute using options and values rather than bootstrap's dropdown classes. Feel free to disagree and dismiss this comment.

app/static/js/graduationManagement.js
```diff
@@ -0,0 +1,142 @@
...
+        handleBonnerFilterChange(cohortYear, buttonText)
+    });
+
+    $('.graduated-checkbox').change(function() {
+        let hasGraduated = $(this).is(':checked');
+        let username = $(this).data('username');
+
+        $.ajax({
+            type: "POST",
+            data: {status: hasGraduated ? 1 : 0},
```
hoerstl: Let's remove the ternary operator and just return `hasGraduated` directly

app/static/js/graduationManagement.js
```diff
@@ -0,0 +1,142 @@
...
+            url: `/${username}/setGraduationStatus`,
+            success: function(response) {
+                location.reload()
```
hoerstl: Let's not have this reload the whole page and instead just change the data or remove a row from the table instead

app/static/js/graduationManagement.js
```diff
@@ -0,0 +1,142 @@
...
+            },
+            error: function(status, error) {
+                console.error("Error updating graduation status:", error);
+                location.reload()
```
hoerstl: There's no need to reload here. If there was an error then the graduation status wasn't updated and therefore there will be no corresponding change in the table that we'd need to reload for.

app/static/js/graduationManagement.js
```diff
@@ -0,0 +1,142 @@
...
+            }
+        });
+    });
+
+    $('#showGraduatedToggle').click(function() {
+        let isToggled = $(this).is(':checked')
+        sessionStorage.setItem('showGraduatedToggleState', isToggled ? 1 : 0)
+        location.reload()
```
hoerstl: I like setting this as part of the session for when people might come back to this page later but I don't think it should reload the page. We should find a way to give all seniors a row and conditionally hide the rows for those who have graduated.

app/static/js/graduationManagement.js
```diff
@@ -0,0 +1,142 @@
...
+    })
+
+    function filterTable(dataFilter, condition) {
```
hoerstl: Consider renaming the parameters to be more precise:
`function filterTable(dataField, expectedValue) {`

app/static/js/graduationManagement.js
```diff
@@ -0,0 +1,142 @@
...
+
+        if (mainFilterState == "bonner") {
+            var bonnerButtonText = "All"
+            if (cohortFilterState != "all") {
+                cohortFilterState = Number(cohortFilterState) 
+                bonnerButtonText = `${cohortFilterState}-${cohortFilterState+1}`
+            }
+            handleBonnerFilterChange(cohortFilterState, bonnerButtonText)
+        }
+        $('#showGraduatedToggle').prop('checked', Number(showGraduatedToggleState));
```
hoerstl: ### Big Bug

This breaks the `show graduated students` slider.

This line needs to be at the start of this function (and combined with the line that sets showGraduatedToggleState), otherwise the datatable filters out all of the graduates before the slider is checked.

app/static/js/graduationManagement.js
```diff
@@ -0,0 +1,142 @@
...
+            sessionStorage.setItem('cohortFilterState', null)
+            
+            filterTable('cce-progress', "True")
+        }
+    }
+
+    function initializePage() {
+        var mainFilterState = sessionStorage.getItem('mainFilterState') || 'all';
+        var cohortFilterState = sessionStorage.getItem('cohortFilterState') || 'all';
+        var showGraduatedToggleState = sessionStorage.getItem('showGraduatedToggleState') || false;
```
hoerstl, line 149: We only need this variable in one place. When we bring the line where we use this variable up to here we should just delete this initialization.

app/static/js/graduationManagement.js
```diff
@@ -0,0 +1,142 @@
...
+
+        if (mainFilterState == 'cce') {
+            var buttonText = 'CCE Minor'
+        } else if (mainFilterState == 'bonner') {
+            var buttonText = 'Bonner Students'
+        } else {
+            buttonText = 'Filter'
+        }
+        handleMainFilterChange(mainFilterState, buttonText)
```
hoerstl: I feel as though these `mainFilterState` conditionals should be moved inside the `handleMainFilterChange` function

app/templates/admin/graduationManagement.html
```diff
@@ -0,0 +1,82 @@
...
+      <a href="/gradStudentsxls/all" id="exportFile" target="_blank">
+        <button class="btn btn-success" id="export">Export</button>
+      </a>
+
+    </div>
+    <div class="col-2 col-xl-1 col-12 text-sm-center stepsIndicator">
+      <span class="step"></span>
+      <span class="step"></span>
+      <span class="step"></span>
+    </div>
```
hoerstl, line 44: I can't find what these empty spans do in the javascript anywhere so I think these should be deleted.

app/templates/macros/graduationIconMacro.html
```diff
@@ -0,0 +1,7 @@
+{% macro graduationIcon(width="20px", height="20px") %}
+  <svg width="20px" height="20px" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="currentColor" class="size-6">
```
hoerstl: If the macro takes in parameters, we should use them on this line to set the width and height.

hoerstl (changes requested):
There are a couple of bugs that need fixing ~~~

app/templates/macros/graduationIconMacro.html
bledsoef: You right you right.

app/templates/admin/graduationManagement.html
bledsoef: Copy and pasted from somewhere else in the app but you are right they are unnecessary.

app/logic/graduationManagement.py
bledsoef: I think elibible is quite apt

app/logic/graduationManagement.py
bledsoef: Intuitively this would make sense, however, Peewee isn't the smartest and for some reason it needs to be an int.

app/templates/admin/graduationManagement.html
bledsoef: I would generally agree that we should be consistent, but I think this looks better.

app/static/js/graduationManagement.js
bledsoef: Because we need it to be an int when saved to the db, if we do not do this it will be passed as the string 'true' and the int conversion won't work.

app/static/js/graduationManagement.js
bledsoef: Let us discuss this in person my friend. I think there are many arguments to be made.

app/static/js/graduationManagement.js
bledsoef: I think the reload is mainly for the flash error message.

app/static/js/graduationManagement.js
bledsoef: For readability and consistency with the ones above I like it how it is.

app/static/js/graduationManagement.js
bledsoef: My b my b.

app/static/js/graduationManagement.js
bledsoef: I might disagree since the only other place that this is used we don't need the conditionals to determine what the buttonText should be.

app/logic/graduationManagement.py
bledsoef: Maybe add comment for clarity

hoerstl (approved):
Beautiful work!!

PR #1356, opened 2024-11-15
Fix Firefox Timepicker on Create Event

bledsoef:
## Issue Description

Fixes issue #1267 
- The create event start and end timepickers are not working on Firefox.

## Changes
- Imported flatpickr into the JS file and initialize it for non chrome browsers.
- Attach a new flatpickr object to each datepicker created when a multiple offerings row is added.

## Testing
- Checkout `development`
- Open up CELTS in Firefox
- Verify date selecting does not work
- Checkout `1267-timepicker-issue`
- Open up CELTS in Firefox and Chrome
- Make sure you can select dates for a single event on both Firefox and Chrome
- Make the event you are creating a multiple offering event
- Add multiple rows and make sure you can select dates on each of those new rows.


stevensonmichel:
The time formatting is not quite correct.
Screenshot 2024-11-20 160023

thealphagurlux (approved):

PR #1359, opened 2024-11-22
Fixed bug with deleting recurring events

### LINKED ISSUE #1350: Events List Recurring Events break when deleting one of the recurring events
When you create an event and make it recurring (e.g. 10), it displays all 10 weeks of that event in the events list. However, if we try to delete one of the recurring events, something in the ordering goes wrong, and the weeks are getting duplicated, and some are disappearing. Example is in the screenshot. I deleted Week 3 recurring event, and it duplicated it and removed the week 2 at first, and then I removed week 1, and it removed week 1 and got week 2 back, and never deleted week 3, which I initially deleted.
Screenshot (20)

bledsoef:
## Issue Description

Fixes #1350 
- When attempting to delete multiple recurring events from a series stuff breaks.

## Changes
- Made sure the only recurring events we were attempting to rename were those that were active and not deleted, as that was causing the problem.

## Testing
- Checkout development
- Reset database
- Create a recurring series with 10 events
- Delete the 5th
- Delete the 1st
- The order should get weird now
- Keep deleting the 1st and verifying weird functionality.
- Checkout `1350_recurring_events_bug`
- Reset database 
- Repeat steps 3-5 but verify it works now.
- Rejoice!

### DIFF
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -46,12 +46,15 @@ def deleteEvent(eventId):
     if event:
         if event.recurringId:
             recurringId = event.recurringId
-            recurringEvents = list(Event.select().where(Event.recurringId==recurringId).order_by(Event.id)) # orders for tests
+            recurringEvents = list(Event.select().where(Event.recurringId==recurringId, Event.deletionDate.is_null(True)).order_by(Event.id)) # orders for tests
             eventDeleted = False
-
+            print([i.name for i in recurringEvents])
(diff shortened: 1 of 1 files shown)


Karina-Agliullova (approved):
Looks good! Would be nice if UI was changed to make the delete modal look better, but it might be for a separate issue

Mbeweg (approved):

PR #1360, opened 2024-11-22
Adding fix to same day multiple time offering

### LINKED ISSUE #1357: Picking today's date counts as a "past start date" in multiple event offerings
If you try to save a multiple event offerings time slot with today's date, even if the time is later than when you are creating it will give you an error like below. I do not believe this is the expected functionality.

Screenshot 2024-11-21 at 2 20 37 PM

WackyWeaver:
## Issue Description

Fixes #1357
If you try to save a multiple event offerings time slot with today's date, even if the time is later than when you are creating it will give you an error like below. I do not believe this is the expected functionality.

## Changes
- I added the start time to the date object in the isDateInPast function 

## Testing
- `git checkout fixSameDayEvent`
- start to create a multiple time offering


bledsoef (approved):
LGTM

PR #1383, opened 2025-01-16
Clarified Form Validation for Instructors on the SLC Page

### LINKED ISSUE #1367: SLC Form should clarify when the instructors' list is empty
When we create an SLC form, on the second page, where we need to select the instructor, there should be more clarification added. As of now, if the instructor for a course is not selected, it won't let the user continue to the third page. However, to select the instuctor, the user needs to click on the names appearing when they start typing the name of the instructor. If they don't click on the option with the instructor's name, and type the full name by themselves, it appears that no instructor is selected, and it is not clear for the user why they can't proceed to the next page. There should be a text in the empty table telling the user that they need to select the instructor and the table is empty

hoerstl:
## Issue Description

Fixes #1367 

When we create an SLC form, on the second page where we need to select the instructor, there should be more clarification.

 As of now, if the instructor for a course is not selected, it won't let the user continue to the third page. However, to select the instuctor, the user needs to click on the names appearing when they start typing the name of the instructor. If they don't click on the option with the instructor's name, and type the full name by themselves, it appears an instructor is selected when, in reality, there is not and it is not clear for the user why they can't proceed to the next page. 

## Changes

I did a few things to enhance transparency:
- There's a message that appears when no instructors have been added to the page
image

- Attempting to move onto the next page now highlights the empty table instead of the search field
image

- The search bar now has a tooltip explaining that a click is required in order to add them to the instructor table
image
I also made it so that the first name in the searched list is autofocused by default and can be conveniently added to the table by pressing enter.

- The search box now empties itself when you stop focusing it. This avoids any misunderstanding that you added a course instructor without clicking on their search result.
image
*unfocus
image

## Testing

In order to test, navigate to the course proposals tab,
image

Create a new course proposal
image

Navigate to the second page (tab) of the form.
image

Finally, pretend you've never interacted with the search instructor field and make sure its intuitive what went wrong when you try to navigate to the next page without entering a course instructor and that its intuitive how to add an instructor to the list.
image

Two instructors in the test database you can use are `Finn Bledsoe` and `Brian Ramsay`

### DIFF
--- a/app/static/js/instructorTable.js
+++ b/app/static/js/instructorTable.js
@@ -31,6 +31,8 @@ export function createNewRow(selectedInstructor) {
   let editLink = newRow.find("td:eq(0) a")
   editLink.attr("id", `editButton-${username}`);
 
+  console.log("Added Row:")
+  console.log(username)
   editLink.attr("data-username", username)
   newRow.prop("hidden", false);
   lastRow.after(newRow);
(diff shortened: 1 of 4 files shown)


hoerstl:
[SLC Form should clarify when the instructors' list is empty](https://trello.com/c/ZflfLO7P/839-slc-form-should-clarify-when-the-instructors-list-is-empty)

bledsoef (approved):
LGTM, would appreciate some more PR description detail next time.

PR #1384, opened 2025-01-17
Addition of Undo Action When Deleting Background Check

### LINKED ISSUE #1366: Add an UNDO option when deleting background check
When a user wants to delete their background check, there should be an UNDO option once they click the delete button for them to undo that action. As of now, there is none, so when they click delete, there is no way to restore it.

stevensonmichel:
## Issue Description

Fixes #1366 
- Previously, when a user wanted to delete a background check, there was no undo process. That means that if the deletion process resulted from a mistake, there was no possibility of recovering that background check. 

## Changes

- Created Undo button for undoing deletion of background check.
- Modified Background Check model to reflect that change

## Testing
- After setting up environment and resetting the databases, run the application using `flask run`.
- Go to My Profile, and then open the section Background Check.
- Create some background checks if needed.
- Click on X to delete background check.
- Look at the behavior if undo is not clicked.
- Create other background checks. 
- Undo background check deletion by clicking on UNDO that appears near the deletion message at the top of the screen.


app/controllers/admin/routes.py
```diff
@@ -379,7 +380,19 @@ def cancelRoute(eventId):
...
         abort(403)
-    
+
+@admin_bp.route('/profile/undo', methods=['GET'])
+def undoBackgroundCheck():
+    try:
+        username = g.current_user
+        bgCheckId = session['lastDeletedBgCheck']
+        BackgroundCheck.update({BackgroundCheck.deletionDate: None, BackgroundCheck.deletedBy: None}).where(BackgroundCheck.id == bgCheckId).execute()
+        flash("Deletion successfully undone.", "success")
```
bledsoef: Maybe change the verbiage here to sound a little better or be more precise?

bledsoef (dismissed):
Check with @BrianRamsay if you need to update prod backup, otherwise LGTM!

bledsoef (dismissed):
LGTM!

bledsoef (approved):
LGTM

PR #1389, opened 2025-01-30
Fixed Show Past Events Toggle on Past Terms

### LINKED ISSUE #1372: Events List - View Past Events Toggle
When you are in the current term on the events list and click the toggle View Past Events to be on, then you go to previous terms and the toggle stays on. However, when you are in the current term and toggle View Past Events off, then there it automatically sets the toggle in the past terms off without the option of turning it on. In the past terms, the toggle should always be on.

bledsoef:
## Issue Description

Closes #1372 
- If you untoggle "Show Past Events" and then switch to a past term, it will maintain that change and you won't be able to see any events.

## Changes
- Added a conditional that checks if it is a current term and if it is not it will ignore whatever is stored in the session state, which was causing the problem in the first place.
- Renamed event_list.js to be eventList.js

## Testing
- Replicate issue on development
- Reset database, source setup.sh, etc.
- Disable "Show Past Events" on the Event List.
- Switch to a past term and make sure it is enabled again.
- Return to a current or future term and make sure it restores to the state it was before.
- Merge :)


MImran2002:
Just like what Finn said, the function of the past events and the toggle buttons are owrking well. The button also toggle and untoggle depending on whether the year selected is past or future.

MImran2002 (approved):
Approved

PR #1425, opened 2025-03-11
CCE Minor Withdrawal Functionality

### LINKED ISSUE #1397: Add withdrawal functionality to Summer Experiences and Other Engagements
In the new table view we are creating in a different issue, we need to add the option for students to be able to withdraw their proposals.

Image

Kafui123:
## Issue Description

Fixes #1397 
- Implemented the withdrawal functionality for the other experience and summer experience creations. 

## Changes
- Added a withdrawal function in javascript that sends an ajax request to the withdrawProposal function in the routes.py file.
- Created a new withdrawProposal function in the routes.py file to check to ensure that the information got deleted from both of the models. 
- Added a new removeProposal function in the minor.py file to delete the file and information stored in the CCEMinor and AttachmentUpload modals. 
- Created a test_removeProposal() function in the tests_minor.py file to see if all the information and files are deleted once the user selects the withdraw button for a specific experience. i.e summerExperience or otherExperience. 

## Testing
- Checkout 1397_WithdrawFunctionality
- Navigate to Minor Management in the Admin sidebar dropdown.
- Click on the user ayisie to navigate to their CCE Minor Profile.
- Switch to the Manage Proposals tab at the top.
- Create an Other Engagement and a Summer Experience with random data.
- Return to the Manage Proposals page and verify both of these proposals look something like this:
- Click on the Select Action button and click on Withdraw
- Check the attachment upload and cceminorproposal tables in the database to see if the information you entered has been deleted.


app/logic/fileHandler.py
```diff
@@ -39,8 +41,10 @@ def getFileFullPath(self, newfilename=''):
                 filePath = (os.path.join(self.path, newfilename))
         except AttributeError:
             pass
+            print('AttributeError')
         except FileExistsError:
             pass
+            print('AttributeError')
```
hoerstl: I'm not sure if you meant to leave these two print statements in here. I think I'm personally okay with it if you add a little more description to what failed and why but then you'll need to delete the pass statements before them.

You've got two options for resolving this:
1. Remove the print statements
2. Improve the print statements to be more descriptive and delete the pass statements

app/logic/minor.py
```diff
@@ -318,3 +322,12 @@ def removeSummerExperience(username):
     """
     term, summerExperienceToDelete = getSummerExperience(username)
     IndividualRequirement.delete().where(IndividualRequirement.username == username, IndividualRequirement.description == summerExperienceToDelete).execute()
+
+def removeProposal(proposalID) -> None:
+    """
+    Delete summerexperience or otherexperience of the CCEMinorProposal Table  for entry proposalType
```
hoerstl: I'm not sure how to read the end of this docstring `for entry proposalType`. I'm not sure if you meant to leave this in here.

app/logic/minor.py
```diff
@@ -318,3 +322,12 @@ def removeSummerExperience(username):
...
+    """
+    proposalAttachment = AttachmentUpload.get_or_none(proposal=proposalID)
+    proposalFileHandler=FileHandler(proposalId=proposalID)
```
hoerstl: Please add spaces around your `=` 🙏 for variable assignment.

The lack of space around keyword parameters is okay.

app/logic/minor.py
```diff
@@ -318,3 +322,16 @@ def removeSummerExperience(username):
     """
     term, summerExperienceToDelete = getSummerExperience(username)
     IndividualRequirement.delete().where(IndividualRequirement.username == username, IndividualRequirement.description == summerExperienceToDelete).execute()
+
+def removeProposal(proposalID) -> None:
+    """
+    Delete summerexperience or otherexperience of the CCEMinorProposal Table  for entry proposalType
+    """
+
+    proposalAttachment = AttachmentUpload.get_or_none(proposal=proposalID)
```
hoerstl, line 408: A few lines later, we're casting proposalID to an Integer which is good but implies that this function may not always accept an integer. If an integer is not passed, this get_or_none will have odd behavior. I think you should cast `proposalID` to an integer once near the start of the function.

app/logic/minor.py
```diff
@@ -318,3 +322,16 @@ def removeSummerExperience(username):
...
+    if proposalAttachment:
+        proposalFileHandler=FileHandler(proposalId=proposalID)
```
hoerstl: Spaces around the `=` please!

app/static/js/minorProfilePage.js
```diff
@@ -2,13 +2,45 @@ import { validateEmail } from "./emailValidation.mjs";
 
 $(document).ready(function() {
   $("#supervisorEmail").on('input', validateEmail);
+    $("#withdrawBtn").on("click", withdraw);
```
hoerstl: unindent this line by two spaces pls

app/static/js/minorProfilePage.js
```diff
@@ -2,13 +2,45 @@ import { validateEmail } from "./emailValidation.mjs";
...
+      let proposalAction = action.value;
+      // decides what to do based on selection
+    if (proposalAction == "Withdraw"){
+        $('#proposalID').val(proposalID);
+        $('#withdrawModal').modal('show');
+      }
+    }
+
+
+  function withdraw(){
```
hoerstl: I think renaming this to `withdrawProposal` would be more descriptive!

app/static/js/minorProfilePage.js
```diff
@@ -2,13 +2,45 @@ import { validateEmail } from "./emailValidation.mjs";
...
+      // uses hidden label to withdraw course
+      let proposalID = $("#proposalID").val();
+      let username = $("#username").val()
+      $.ajax({
+        url: `/cceMinor/withdraw/${username}/${proposalID}`,
+        type: "POST",
+        success: function(s){
+          location.reload();
+        },
```
hoerstl, line 34: I would love to see the page refresh and return focus to the same manage proposals tab instead of returning to the sustained community engagements tab. 

Consider doing something like adding some logic to read in the query string from a route: 
`selectedTab = request.args.get("tab", "sustainedCommunityEngagements")`
and then passing that value through the `render_template` function. It seems like there's already functionality to select the right tab within the `profile.html` page if you just pass that value to the page using the keyword name `activeTab` (see here: `class="nav-link {{ 'active' if activeTab == 'sustainedCommunityEngagements' or activeTab is not defined else '' }}"`)

Finally, on the line of this comment, instead of reloading, run `window.location.href = "/profile/ayisie/cceMinor?tab=manageProposals";` which will reload and pass a key value pair `{"tab": "manageProposals"}` inside of the `request.args` object.

app/static/js/minorProfilePage.js
```diff
@@ -2,13 +2,45 @@ import { validateEmail } from "./emailValidation.mjs";
...
+      let proposalID = $("#proposalID").val();
+      let username = $("#username").val()
+      $.ajax({
+        url: `/cceMinor/withdraw/${username}/${proposalID}`,
+        type: "POST",
+        success: function(s){
+          location.reload();
+        },  
+        error: function(request, status, error) {
+            console.log(status,error);
```
hoerstl: small one, space after `,` pls

`console.log(status, error);`

app/templates/base.html
```diff
@@ -46,7 +46,7 @@
         <div id="maincontent" class="main-content container">
           <div class="flash_container" id="flash_container">
             {% with messages = get_flashed_messages(with_categories=true) %}
-            <script> var flashMessages = {{messages | tojson}} </script>
+            <script> var flashMessages = {messages,tojson} </script>
```
hoerstl: !!! SUPER IMPORTANT !!!
This line change breaks all flash messages on the site. This syntax wasn't a dictionary to begin with, the double curly braces were to signal that we're doing something with Jinja (which vscode doesn't always understand). The pipe character here (`|`) is a special Jinja character which means to apply the Jinja builtin filter `tojson` to the object on the left of the pipe. This should be changed back to how it was. If you'd like more reading about how and why it works, check out the [Jinja docs](https://jinja.palletsprojects.com/en/stable/templates/#filters)

app/templates/minor/cceMinorProposalManagement.html
```diff
@@ -46,6 +45,7 @@
     <a class="btn btn-primary" href='/cceMinor/{{user.username}}/otherEngagement'>Request Other Engagement</a>
     <a class="btn btn-primary" href='/cceMinor/{{user.username}}/summerExperience'>Request Summer Experience</a>
     </div>
+    <input id="proposalID" type="hidden">
```
hoerstl: You should remove this line. The same input tag with the same id is defined on line 39.

app/templates/minor/cceMinorProposalManagement.html
```diff
@@ -63,7 +63,8 @@ <h5 class="modal-title" id="withdrawModalLabel">Withdraw Proposal</h5>
             <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Cancel</button>
             <button type="button" class="btn btn-primary" id="withdrawBtn">Withdraw</button>
         </div>
-        </div>
+        </div>]
```
hoerstl: I believe you should delete the `]` here. If I'm silly and don't understand why then please disregard

tests/code/test_minor.py
```diff
@@ -551,3 +556,78 @@ def test_createOtherEngagementRequest():
...
+                g.current_user = "glek"
+                removeProposal(testProposalId)
+
+            assert list(CCEMinorProposal.select().where(CCEMinorProposal.id == testProposalId)) == []
+
+            assert not AttachmentUpload.select().where(AttachmentUpload.proposal_id == testProposalId, AttachmentUpload.fileName == f"{testProposalId}.pdf").exists()
+            assert 0 == AttachmentUpload.select().where(AttachmentUpload.proposal_id == testProposalId, AttachmentUpload.fileName == f"{testProposalId}.pdf").count()
+
+        except:
+            os.remove(handledProposalFile.getFileFullPath('999.pdf'))
```
hoerstl: This is a big one that makes this test not work.

An assertion error is also a kind of error that pytest uses to check to see if a test failed or not. Since it's wrapped within the try-except block, if any of the assertions fail, they'll be caught by the exception. I think what you're looking for here is a try-finally block to make sure the proposal file gets deleted.

tests/code/test_serviceLearningCourses.py
```diff
@@ -68,7 +68,6 @@ def test_getInstructorCourses():
     currentFaculty = User.get_by_id("ramsayb2")
     currentFacultyCourses = courseDict[currentFaculty]
     assert 'Spanish Help' in currentFacultyCourses
-    assert 'Databases' in currentFacultyCourses
```
hoerstl: This is an important part of the test and Databases is in fact in currentFacultyCourses as it should be. Please return this to how it once was.

hoerstl (dismissed):
I could see a lot of good insights and changes that were made based on what seems like should be correct and I'm very happy with that! Always keep learning and check everything that doesn't make sense.

Overall, the functionality to withdraw an other/summer experience wasn't working for me even though I saw the pop-up and there are a few big bugs in this that ought to be fixed before merging.

This is a well made PR and I'm happy to see us using our minds and double checking the logic of what was there before we got there. Good work! (●'◡'●)

app/static/js/minorProfilePage.js
Kafui123: This functionality is in a different pr please.

app/templates/minor/cceMinorProposalManagement.html
```diff
@@ -63,7 +62,8 @@ <h5 class="modal-title" id="withdrawModalLabel">Withdraw Proposal</h5>
             <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Cancel</button>
```
MImran2002: Screenshot 2025-04-04 at 5 39 19 PM
I cannot find which line it is but if you look at the picture at the bottom of the modal there is a bracket which online appears when the modal for withdrawal pops up so I think it's an unnecessary code.

app/templates/minor/cceMinorProposalManagement.html
```diff
@@ -1,6 +1,4 @@
-
```
MImran2002, line 1: So there's a front-end bug because you created a modal and that means you can click outside and the modal can disappear or cancel out. What happen is that the modal disappear and the withdrawn is still chosen in the drop down menu. If you can turn that into an alert which it is like an alert warning user if they actually want to delete it you should consider making sure the user cannot exit out of the modal box even if they click out on the screen. 
useful links for bootstrap: https://getbootstrap.com/docs/5.3/components/modal/#static-backdrop

app/templates/minor/cceMinorProposalManagement.html
```diff
@@ -38,6 +36,7 @@
```
MImran2002, line 36: Another front-end issue is that it will be better if there are red asterisk to display it is required when filling out the form for both other engagement and summer experience

app/templates/minor/cceMinorProposalManagement.html
```diff
@@ -1,6 +1,4 @@
-
 <div class="tab-pane fade show {{ 'active' if activeTab == 'manageProposals' else '' }}" id="pills-manageProposals" role="tabpanel" aria-labelledby="manageProposals">
-
     <div class="table-responsive" id="proposalTable">
     <table class="table table-striped">
```
MImran2002, line 3: For the summer experience, I cannot see any years when i click the select year dropdown bar but the term in other engagement works
issue:
Screenshot 2025-04-04 at 5 55 47 PM

MImran2002 (dismissed):

stevensonmichel:
A New Issue idea is to ensure that after the reload we are still on the tab Manage Proposals

stevensonmichel:
It appears that there is a bug in the Select Year field when creating a Summer Experience. Can you all check that out?

stevensonmichel:
Another enhancement idea: if somebody is on the Create Summer Experience or Create Engagement page, there is no way to cancel the process.

stevensonmichel:
Possible enhancement: If I click on Withdraw, and then in the modal I click on Cancel, the dropdown still has Withdraw as the selected option. From my viewpoint, it should be reset to "Select".

app/templates/minor/cceMinorProposalManagement.html
bledsoef: This is expected and functionality consistent throughout our application.

bledsoef:
> Another enhancement idea: if somebody is on the Create Summer Experience or Create Engagement page, there is no way to cancel the process.

Done in a different PR

bledsoef (dismissed):
LGTM

stevensonmichel (dismissed):

bledsoef (approved):
LGTM

PR #1426, opened 2025-03-13
Program Training Display Fix

### LINKED ISSUE #1375: Evaluate the Program Training determination
Both the checkmarks on the profile page and on the Event View tab. There should not be duplicate items, and fulfillment by attendance should work correctly.

ojmakinde:
## Issue Description

Fixes #1375 
- If there are multiple instances of the same training where the user only has to attend one instance, all duplicate instances of the test will show up when viewing training requirements for any given user.

## Changes

- Updated logic to track and conditionally display trainings for any given student.
- Fixed bug in development where trainings were not displaying for students in the Manage Volunteer view.

Old look:
image

New look:
image

## Testing

- Create sample trainings and ensure that the view works correctly.

### DIFF
--- a/app/logic/participants.py
+++ b/app/logic/participants.py
@@ -157,8 +157,16 @@ def getParticipationStatusForTrainings(program, userList, term):
     # Create a dictionary binding usernames to tuples. The tuples consist of the training (event object) and whether or not they attended it (bool)
     userParticipationStatus = {}
     for user in userList:
-        for training, attendeeList in trainingData.items():
-            userParticipationStatus[user.username] = userParticipationStatus.get(user.username, []) + [(training, user.username in attendeeList)]
+        seenEvents = {}     # Tracks the most recent index where each training event was processed for each user
+        for index, (training, attendeeList) in enumerate(trainingData.items()):
+            userAttendedTraining = user.username in attendeeList
+            if training.name in seenEvents:
+                targetIndex = seenEvents[training.name]
+                if userAttendedTraining:
+                    userParticipationStatus[user.username][targetIndex][1] = True
+            else:
+                userParticipationStatus[user.username] = userParticipationStatus.get(user.username, []) + [(training, user.username in attendeeList)]
+                seenEvents[training.name] = index
 
     return userParticipationStatus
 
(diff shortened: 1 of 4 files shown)


app/logic/participants.py
```diff
@@ -136,13 +136,14 @@ def getParticipationStatusForTrainings(program, userList, term):
 
     :returns: trainings for program and if the user participated
     """
-    isRelevantAllVolunteer = (Event.isAllVolunteerTraining) & (Event.term.academicYear == term.academicYear)
-    isRelevantProgramTraining = (Event.program == program) & (Event.term == term) & (Event.isTraining)
+    print(program)
+    isRelevantTraining = ((Event.isAllVolunteerTraining | ((Event.isTraining) & (Event.program == program))) & 
+                              (Event.term.academicYear == term.academicYear))
```
hoerstl, line 141: I notice that you've changed the trainings you've selected here. This new select statement now includes program trainings that have occurred in a different semester provided that it's within the same academic Year. For example, a training in Fall 2021 is still selected when loading trainings for Spring 2022. 

Previously, on the deleted line 140, we were only selecting program trainings that occurred in the same term while all volunteer trainings were the only ones that we allowed to be valid across semesters.

I talked with Brian and he said it was, but just to make sure, is this the intended behavior?

hoerstl (dismissed):
Good work! This should work.

app/logic/participants.py
ojmakinde: Yes, the change is intended. After reaching out to Sarah Heggen, Steven confirmed that all program trainings are valid for a full academic year, so we decided to merge the logic.

bledsoef (approved):
Looks good to me for the most part, will merge but want to pick your brain on these things @ojmakinde and @WackyWeaver. I have one logistics comment. I created two events for the semester that had the same name but were ultimately different events with different locations, dates, descriptions, etc but are not multiple series. Can we ever foresee them creating two trainings that they want to be individually required? If so we need to make it not name dependent, but I can't really see this being much of a frequent occurence.

Also, following up on Lawrence's comment, you said All Volunteer Trainings should last all year, but this makes it so that any type of training does as well, is this also expected?

Nice job guys!

PR #1429, opened 2025-03-18
Simplified/Refactored getParticipationStatusForTrainings

hoerstl:
## Issue Description

This is a very small PR, just refactoring a single function for readability.

I noticed that the `getParticipationStatusForTrainings()` function inside of `/home/vscode/celts/app/logic/participants.py` (which I wrote about a year and a half ago), was far more complicated than it needed to be since I didn't know about better data structures.

## Changes

I refactored the getParticipationStatusForTrainings function which should now, hopefully, be more readable.

## Testing

- Run the tests script and ensure all tests pass. 
- Look at the code and ensure that its readable enough, please suggest any ways I could further simplify this code.

### DIFF
--- a/app/logic/participants.py
+++ b/app/logic/participants.py
@@ -11,6 +11,7 @@ from app.logic.users import isEligibleForProgram
 from app.logic.volunteers import getEventLengthInHours
 from app.logic.events import getEventRsvpCountsForTerm
 from app.logic.createLogs import createRsvpLog
+from collections import defaultdict
 
 def trainedParticipants(programID, targetTerm):
     """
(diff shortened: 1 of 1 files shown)


app/logic/participants.py
```diff
@@ -144,23 +145,29 @@ def getParticipationStatusForTrainings(program, userList, term):
...
+                trainingData[training].add(training.eventrsvp.user_id)
         except AttributeError:
-            trainingData[training] = trainingData.get(training, [])
-    # Create a dictionary binding usernames to tuples. The tuples consist of the training (event object) and whether or not they attended it (bool)
-    userParticipationStatus = {}
-    for user in userList:
-        for training, attendeeList in trainingData.items():
-            userParticipationStatus[user.username] = userParticipationStatus.get(user.username, []) + [(training, user.username in attendeeList)]
-
-    return userParticipationStatus
```
hoerstl, line 163: I know someone will rightfully ask about why I made this change from development to be longer and more complex so I'll describe here. This functionality returns an attended status for every training made available instead of merging the trainings attendance by their `event.name`.

PR #1437, opened 2025-04-03
Bonner checkmark profile

### LINKED ISSUE #1370: Bonner requirements checkmarks on the profile page
The checkmarks for Bonner requirements on the profile page should respect the frequency of the requirement. Something that is required once, would have one checkmark or X. Something that is required twice should have two icons, every term should show a checkmark for each term and an x for each term missed. As part of this we should add a ‘Twice’ frequency.

MImran2002:
## Bonner requirements checkmarks on the profile page 

Fixes   #1370

- The checkmarks for Bonner requirements on the profile page should respect the frequency of the requirement. Something that is required once, would have (1/1) or (0/1). Something that is required twice should have (0/2) or (1/2) or (2/2), every term should show the numbers of time they have done the certification against the total expected certification based on their class level. For example if they are Junior they will likely have (0-5/5) or (0-6/6).

## Changes
- Add options for both term and twice in Bonner Management under Brian as admin.
- Reproduce the certification requirement with counts (1/1) if they have done and (0/1) if they have not (This may vary based on frequency.

## Edge-cases to consider before testing:
- These edge case are because the numbers populated are based on two things: the current term and the student's rawClassLevel.
- If any of the cases disrupt these two things it will display no data or wrong data. 
- Cases to avoid:
  1. If you create a new event on top of an existing event that has the same term and the same frequency, only one of those events will appear.
  
  2. Another scenario occurs when the current term is Fall 2023, but an event is created in Spring 2024, and a student is marked as attended for that Spring 2024 event. In this case, the student is being assigned attendance for a future term, so it will not appear in the count.
  
  3. This logic also applies to the past. For example, if a student is a sophomore and the current term is Spring 2023, the list only populates terms from Spring 2023 back to their Freshman year (Spring 2023, Fall 2022, Spring 2022, Fall 2021).
  Any events created before or after these term ranges will not appear in the counts.

  4. Another case is where if the event is created during summer terms it will not count, as certification events do not happen during summer and if the current term is summer and you hover over the view it will show the admin the Fall term  instead of Summer term as the idea is to inform the admin of certification being needed for Fall term as there's no certification for the Summer term. 

# Testing
## Navigating to the branch
- Use git checkout bonnerCheckmarkProfile
- Pull the code with git pull
- Reset the database with to test and run tests/run_tests.sh
- Run the application with flask run
- Make your way to the navigate bar and click on admin drop down feature
- Click Bonner_Management and go to Requirements and click on the "Add Requirement" button then name it and choose every term as the Frequency
- Note:  Make sure that the term that you picked for the event when creating it is a term at which the cohort you picked will be attending school (not graduated at that time ) and not earlier than the current term. Otherwise, it won't show up.

## Creating and checking the certification
***All of the testing must make sure that the term the event is created in because the event must be created between the current term and the term they came to Berea College (you can see the current term in the setting > term management or when you create an event the term that is auto selected is the current term)***
- Go to Create Event and Select Bonner Scholars to create certification events
- Create an Event and make sure you choose the Bonner Requirement that you added a while ago.
- Select the invite cohorts that has Zach Neill.
- And click Create Event
- Go to manage volunteers where the RSVP and Waitlist checked Zach Neill as attended and saved
- Then go to the side bar and navigate to Zach Neill as a user
- Go to My Profile and click the Bonner Program dropdown
- You will see the tick if you have finished and a number of cross on the terms you have not
- Hover over the requirement that you created and will see the tooltips


WackyWeaver (commented):
There is a html logic issue on the tool tip when there have been no attended terms. Also, the tooltip never shifts when there are fewer terms available for underclassmen.

WackyWeaver (dismissed):
LTGM! Please add a comment for the formula in missedTerms and rename certs

WackyWeaver (dismissed):
I appreciate the comments added to the calculating logic section.  A little long compared to a lot in the rest of the code base, but everything else LGTM!

bakobagassas:
When I follow the steps in the PR, everything works up to where I click to save the event. It is failing to save the event.

bakobagassas:
Also, I have noticed that when I try to create an event, before even selecting the requirement that had "all term", I am unable to select a term in the event itself.

bakobagassas:
Also, something seems to be breaking. When I try to access event lists (where we see the event we created), the page is not accessible on your branch; it gives an error page. nb: I already reset the database multiple times.

MImran2002:
> Also, something seems to be breaking. When I try to access event lists (where we see the event we created), the page is not accessible on your branch; it gives an error page. nb: I already reset the database multiple times.

There are tests breaking in development:
tests/code/test_serviceLearningCourses.py ...F......
tests/code/test_spreadsheet.py F..FFFFFFFFFFF
tests/code/test_term.py ...
tests/code/test_transcripts.py .F..

And for the event I am able to select the term:
image
 and I am able to see an event list : 
image

I can definitely guide you with the set up when might you be free on Wednesday or Thursday.

bakobagassas:
I wonder why it is only showing 3 checkmarks. Wasn't it supposed to be every term (so 8 terms)? 
image

MImran2002:
> I wonder why it is only showing 3 checkmarks. Wasn't it supposed to be every term (so 8 terms)? image

Hello Sara, Thank you for the question. That's what I thought as well but after reading through the issues and code I remember why. These are the few questions you might potentially have reviewing this issue.

Why am I seeing only three checkmarks, shouldn't it be eight checkmark?
    The reason you see three checkmarks instead of eight is because the student is a sophomore and it is either a summer term or a spring term. So the student who is a Bonner scholar should only at that point have three checkmarks or ticks because it is populated based on the terms that they are Bonner scholars. Not the overall term this is to address multiple scenario, if a Bonner student decides to drop out of the program there won't be unnecessary populated terms, if they decide to graduate early, the populating of the checkmark will not show a total checkmark at all but only show the checkmarks when they are Bonner scholars and importantly the CELTS department doesn't care about future certification requirment their concern more is if the student have done the previous and present requirement. That is why the populating of checkmark is based on every term they stay in the program or college.

Why is Summer Term move forward to Fall term?
    The reason we have a condition where we manually assigned the current term if its summer term to fall term all the time is because if the celts admin is checking this during the summer term we don't want summer term to be considered towards the overall term which will make it from maximum 8 checkmark to 11 checkmark and no Bonner program student is present during summer so it is more reasonable to show that there is a cross for next term then the summer term to always inform the admin about it. 

I hope these questions might help and if you want to confirm the logic of it you can definitely refer to the test where the logic is tested in multiple scenario with multiple class levels and terms.

bakobagassas (dismissed):
I had questions about whether the checkmarks represent the number of terms from the past or the total number of terms the person needs to spend in Berea. It seems to be the number of terms the person has already spent in Berea including the present. If this is what the client requires then everything looks good!

app/logic/certification.py
```diff
@@ -1,14 +1,68 @@
...
+from flask import g
+from app.models.event import Event
+from app.models.term import Term
 from app.models.certification import Certification
 from app.models.certificationRequirement import CertificationRequirement
 from app.models.requirementMatch import RequirementMatch
 from app.models.eventParticipant import EventParticipant
+from app.models.user import User
+
+def termsAttended(certification=None, username=None):
```
BrianRamsay: We should handle if username is not provided in a different way. Currently, we'll get an error, as `len(attendance)` will break in some fashion, as attendance won't have been defined if there is no username.

We should probably check for a missing username and exit the function right at the top if it is not provided.

Since we are handling that case, we should also have a test case when the expected parameters are not given, and make sure that the function fails gracefully (or fails the way we would expect)

app/logic/certification.py
```diff
@@ -1,14 +1,68 @@
...
+
+    Logic:
+    - Each class level is expected to participate in 2 terms per year.
+    - If the user is currently in their final spring term (e.g., Spring of senior year), 
+      they are expected to have completed all terms: missedTerms = (level + 1) * 2.
+    - Otherwise, assume they’ve had one fewer term to attend: missedTerms = ((level + 1) * 2) - 1.
+    - If the user's classification is None, assume just 1 expected term.
+    - Subtract the number of terms the student has attended from the expected total to get the missed count.
+    '''
+    classLevel = ["Freshman", "Sophomore", "Junior", "Senior"]
```
BrianRamsay, line 40: We should validate our parameters here, so we have more control over how the function handles missing data. Right now, we'll just get a `NotFoundException` when we try to run the User query, if it is missing.

app/logic/certification.py
```diff
@@ -1,14 +1,68 @@
...
+    
+    # looking into a scenario where the current term is summer so that we can reassigned the current term variable to the next term     
+    if currentTerm.isSummer == True:
+        currentDescription = f'Fall {currentTerm.year}'
+        currentTerm = Term.select(Term).where(Term.description == currentDescription).get()
+    else:
+        currentDescription = currentTerm.description
+
+    for level in range(4):
+        user = User.select().where(User.username == username).get()
```
BrianRamsay: We don't need to get the user every time in the loop, we can get it before the loop. In this particular case, we can get the user at the top of the function to validate the `username` parameter.

app/logic/certification.py
```diff
@@ -1,14 +1,68 @@
...
+    - Otherwise, assume they’ve had one fewer term to attend: missedTerms = ((level + 1) * 2) - 1.
+    - If the user's classification is None, assume just 1 expected term.
+    - Subtract the number of terms the student has attended from the expected total to get the missed count.
+    '''
+    classLevel = ["Freshman", "Sophomore", "Junior", "Senior"]
+    currentTerm = g.current_term 
+    currentDescription = currentTerm.description
+    
+    # looking into a scenario where the current term is summer so that we can reassigned the current term variable to the next term     
+    if currentTerm.isSummer == True:
```
BrianRamsay: `isSummer` is already a boolean, so we don't need to compare it to `True`

app/logic/certification.py
```diff
@@ -1,14 +1,68 @@
...
+        currentDescription = f'Fall {currentTerm.year}'
+        currentTerm = Term.select(Term).where(Term.description == currentDescription).get()
+    else:
+        currentDescription = currentTerm.description
```
BrianRamsay: This case is already handled by the variable initialization on line 40

app/logic/certification.py
```diff
@@ -1,14 +1,68 @@
...
+
+    for level in range(4):
+        user = User.select().where(User.username == username).get()
+        if user.rawClassLevel == classLevel[level] and currentDescription == f'Spring {currentTerm.year}':
+            missedTerms = (level + 1) * 2
+        elif user.rawClassLevel == classLevel[level]:
+            missedTerms = ((level + 1) * 2) - 1
+        elif str(user.rawClassLevel) == "None":
+            missedTerms = 1
```
BrianRamsay: Let's use `totalTerms` for this value, because it's not really a missed term until the calculation happens below.

ojmakinde (changes requested):
I think this approach is a bit unintuitive and unclear.
image
The X's don't fit with the rest of the text and don't tell users very much. Additionally, for events with multiple requirements, this is going to be hard to interact with on mobile, and our users don't want to count the number of events but see it at a glance instead. 

An alternative approach that I recommend would be using an approach like, 
Requirement Name (Frequency Attended / Required Frequency) and using a [tooltip](https://getbootstrap.com/docs/4.0/components/tooltips/) to show the number of events attended. An example of this would be
image

Next to the (1/1), add the suggested tooltip and show information for the attended events and the semester, just like we do in the volunteer view. 
image

BrianRamsay (changes requested):

ojmakinde (dismissed):
PR looks good. Requested changes have been addressed and it looks much better than it did previously.

app/logic/certification.py
BrianRamsay: Doesn't appear to be resolved. If `username` is `None`, `attendance` will be undefined when we try to use it on line 21.

(username if conditional check has been removed-Imran)

app/logic/certification.py
```diff
@@ -1,14 +1,102 @@
...
+    Returns a list of Fall + Spring semesters a student has attended, like ["Fall 2023", "Spring 2024"]
+    """
+    user = User.select().where(User.username == username).get()
+    currentTerm = g.current_term
+    studentStanding = user.rawClassLevel
+    currentTermIsFall = currentTerm.description.startswith("Fall")
+    fallAndSpringTerms = []
+    
+    classLevelToYears = {"Freshman": 1, "Sophomore": 2, "Junior": 3, "Senior": 4}
+    yearsCompleted = classLevelToYears.get(studentStanding)
```
BrianRamsay: This will fail as soon as we are looking at a student who doesn't have their class level reported properly. We need to handle that case in some fashion.

```
mysql> select rawClassLevel, count(*) from user where isStudent group by rawClassLevel;
+---------------+----------+
| rawClassLevel | count(*) |
+---------------+----------+
| NULL          |     1854 |
| Junior        |      383 |
| Freshman      |      503 |
| Sophomore     |      347 |
| Senior        |      276 |
| Graduating    |       73 |
| Non-Degree    |       33 |
+---------------+----------+
```

BrianRamsay (changes requested):

app/logic/certification.py
BrianRamsay: (probably we should just count them as seniors, or whatever we need to do to give them the max number of terms)

app/logic/certification.py
```diff
@@ -1,14 +1,84 @@
...
+            for k in range(totalTermsCount):
+                if k % 2 == 0:  
+                    season = "Fall"
+                    year = startYear + (k // 2)
+                else:           
+                    season = "Spring"
+                    year = startYear + (k // 2) + 1
+                totalTerms.append(f"{season} {year}")
+            break
+        elif user.rawClassLevel == "NULL" or user.rawClassLevel == "Graduating" or user.rawClassLevel == "Non-Degree":
```
ojmakinde: Should non-degree folks have any number of semesters? Can you confirm with Brian that they count as students?

app/logic/certification.py
```diff
@@ -1,14 +1,84 @@
...
+    '''
+    Calculate how many certification-eligible terms a student has missed based on their class level
+    and attendance record.
+    '''
+    totalTerms = termsInTotal(username)   
+    attendedTerms = termsAttended(certification, username)
+    missedTerms = [term for term in totalTerms if term not in attendedTerms]
+    return missedTerms
+
+def termsInTotal(username):
```
ojmakinde, line 28: Can you add a docstring here? The function doesn't return all the terms in total. It only returns the Fall + Spring and I think it's important to make that distinction in the comment. Also, I think the previous approach to calculating the terms was cleaner, because this is a bit hard(er) to understand, unless you disagree

app/logic/certification.py
MImran2002: username conditional check has been removed.

app/logic/certification.py
MImran2002: have put an elif condition that checked these rawClassLevel and populated 8 potential terms for the user

ojmakinde (changes requested):

app/logic/certification.py
MImran2002: From what I remember Brian showed me that that table he mentioned is underneath student which is why I have to consider that and he told me becasue all NULL, graduating and Bon-degree should populate like seniors @BrianRamsay please correct me if I am wrong. But I will double check with him else well in person.

app/logic/certification.py
MImran2002: > Yes, the reason I created this termsintotal is because for me this is easier to find the actual start year then populating it from there.  Especially when i was creating test cases I realize it always skipped a term, the count may be right but it never considers the current so I created this function that will find the start year and from there populate the term. I will add a docstring to further clarify it.

app/logic/certification.py
```diff
@@ -36,6 +35,20 @@ def termsMissed(certification, username):
     return missedTerms
 
 def termsInTotal(username):
+    '''
+    This function calculates all the terms a student have enrolled based on their class level.
+    This is under the impression that as Bonnor Scholars they start their term in Fall and end in four year along with Summer not being counted.
+    Thus, if the current term is summer it is changed to Fall for display purpose.
+    The logic is as follows:
```
ojmakinde: I'm sorry, but this is far too long. All I expected was a line or two describing this function and returning the groups, not a description of the logic

ojmakinde (changes requested):

app/logic/certification.py
BrianRamsay: We can treat them as seniors, though there wouldn't be any non-degree Bonner students, so we don't need a specific case for it, really. We just need to not break.

app/logic/certification.py
MImran2002: I am sorry, I have reduced it to two lines of explanation. Please tell me if this will work.

app/logic/certification.py
MImran2002: I am going to resolved this as, the code is considering all the edge cases as seniors.

app/logic/certification.py
ojmakinde: Looks good! Thanks

ojmakinde (approved):

ojmakinde (changes requested):
I pulled from prod-backup and searched up Syakira's profile. Syakira is a bonner that has a rawClassLevel of null. Here's the SQL query:

```
select user.username, user.rawClassLevel from user inner join bonnercohort on user.username=bonnercohort.user_id where rawClassLevel is NULL and user.username LIKE '%wij%';
+----------+---------------+
| username | rawClassLevel |
+----------+---------------+
| wijayas  | NULL          |
+----------+---------------+
1 row in set (0.001 sec)
```

Here's her bonner profile page:

Image


Also, here's Chase Ackerman's profile:
image

RueHaile:
> I pulled from prod-backup and searched up Syakira's profile. Syakira is a bonner that has a rawClassLevel of null. Here's the SQL query:
> 
> ```
> select user.username, user.rawClassLevel from user inner join bonnercohort on user.username=bonnercohort.user_id where rawClassLevel is NULL and user.username LIKE '%wij%';
> +----------+---------------+
> | username | rawClassLevel |
> +----------+---------------+
> | wijayas  | NULL          |
> +----------+---------------+
> 1 row in set (0.001 sec)
> ```
> 
> Here's her bonner profile page:
> 
> Image
> Also, here's Chase Ackerman's profile: image

That is a great point. We handle the rawClassLevel as an SQL NULL, rather than the string we had to catch that edge case. Now it is also working for syakira and other rawClassLevel = NULL people. It populates the number of terms as 8, as they have attended 8 terms to graduate.

Demonstration:
image

ojmakinde:
@RueHaile, your image says "Bonner Orientation mine 0/8". There's only 1 Bonner Orientation event, if I remember correctly, and the "mine" is probably a typo. Additionally, I'm certain that Syakira did her Bonner Orientation and all the other required events, so there's a problem of a data mismatch happening somewhere in our data.

It's likely that the events were not tagged as requirement matches, and in that case, we may need to scan through all the events attended, and find the matches. An alternative could also be writing a script that does that traversal for every student and populates the database. @BrianRamsay

RueHaile:
> @RueHaile, your image says "Bonner Orientation mine 0/8". There's only 1 Bonner Orientation event, if I remember correctly, and the "mine" is probably a typo. Additionally, I'm certain that Syakira did her Bonner Orientation and all the other required events, so there's a problem of a data mismatch happening somewhere in our data.
> 
> It's likely that the events were not tagged as requirement matches, and in that case, we may need to scan through all the events attended and find the matches. An alternative could also be writing a script that does that traversal for every student and populates the database. @BrianRamsay

@ojmakinde, the picture that you see above with the "mine" is mine, as the name implies, I created it, and so that is why she didn't attend any of it. If you reset the data to prod-backup, you can see that there is a bonner orientation of frequency of one, and not eight. I added the picture to demonstrate that the field is getting populated. The Bonner orientation has not been attended by her and almost all others. I will look into the data mismatch, as it is a possibility. 
image

RueHaile:
@ojmakinde: It seems like a lack of data, rather than a bug/data mismatch/ as we went in while the database was set to development.
PROOF: 
image
and we got:- 
image

WHICH means that no data is saved for the event of booner orientation other than for event_id= 476 & 708 (Doesn't have any participants), where  event_id = 476 is found to have data in terms of rsvp but not as eventattendees despite the addition of "utsa: a volunteer, she would be not get populated mark as bonner oriantation happens once for every stident on their freshman year.

JohnCox2211:
It appears that the credit for new requirements is not being added to user profiles when a new requirement is created.

image

image

image


The mark was not initially showing up for the test, so I edited the event to be an orientation, which then appeared. When I switched it back to the test requirement, the orientation mark stayed, and there was still no mark for the test requirement. 

When I tried this again, there was no longer a mark by the orientation when I changed the requirement to test, and there was still no check next to the test requirement. I then added an additional Bonner event for All Bonner Meeting, and added Zach; the check for the requirement did not appear next to All Bonner Meeting on their profile either, so neither the test requirement nor the All Bonner Meeting requirement appears. 

image

MImran2002:
> It appears that the credit for new requirements is not being added to user profiles when a new requirement is created.
> 
> image image image
> The mark was not initially showing up for the test, so I edited the event to be an orientation, which then appeared. When I switched it back to the test requirement, the orientation mark stayed, and there was still no mark for the test requirement.
> 
> When I tried this again, there was no longer a mark by the orientation when I changed the requirement to test, and there was still no check next to the test requirement. I then added an additional Bonner event for All Bonner Meeting, and added Zach; the check for the requirement did not appear next to All Bonner Meeting on their profile either, so neither the test requirement nor the All Bonner Meeting requirement appears.
> 
> image

Hi John, I have throughly looked through what you said definitely it is something to think about, I looked at and found the bug but the bug depends what the current term is in test the current term is flexible and we can assigned because of this line of code which it does is find all terms where the student have attended while the total term populate it base on the current term. Lets say the event is created at Spring 2025 while the current term is Spring 2021 because of the line: attendedTerms = set(term for term in attendedTerms  if term in totalTerms) it will only accept terms that are with the four year range below Spring 2021. That is why the bug exist. We can definitely talk about this at work and hopefully the bug recreation i did is like yours and this is the issue.  If this is the case then we should look into if it is an actual bug or more of a administrative edge case they wouldn't happen in real life. We should definitely talk to @ojmakinde !

JohnCox2211:
> > It appears that the credit for new requirements is not being added to user profiles when a new requirement is created.
> > image image image
> > The mark was not initially showing up for the test, so I edited the event to be an orientation, which then appeared. When I switched it back to the test requirement, the orientation mark stayed, and there was still no mark for the test requirement.
> > When I tried this again, there was no longer a mark by the orientation when I changed the requirement to test, and there was still no check next to the test requirement. I then added an additional Bonner event for All Bonner Meeting, and added Zach; the check for the requirement did not appear next to All Bonner Meeting on their profile either, so neither the test requirement nor the All Bonner Meeting requirement appears.
> > image
> 
> Hi John, I have throughly looked through what you said definitely it is something to think about, I looked at and found the bug but the bug depends what the current term is in test the current term is flexible and we can assigned because of this line of code which it does is find all terms where the student have attended while the total term populate it base on the current term. Lets say the event is created at Spring 2025 while the current term is Spring 2021 because of the line: attendedTerms = set(term for term in attendedTerms if term in totalTerms) it will only accept terms that are with the four year range below Spring 2021. That is why the bug exist. We can definitely talk about this at work and hopefully the bug recreation i did is like yours and this is the issue. If this is the case then we should look into if it is an actual bug or more of a administrative edge case they wouldn't happen in real life. We should definitely talk to @ojmakinde !

When looking into this further, I noticed the same issue with the mark for an event not appearing next to the requirement on the student's page, even when the event was set to have happened and was attended before the current term. The current term is summer 2021, and I created the event for November 2020. I followed the rest of the steps previously done and still ended up with 0/3 on the created requirement and on All Bonner Meeting.

RueHaile:
> > It appears that the credit for new requirements is not being added to user profiles when a new requirement is created.
> > image image image
> > The mark was not initially showing up for the test, so I edited the event to be an orientation, which then appeared. When I switched it back to the test requirement, the orientation mark stayed, and there was still no mark for the test requirement.
> > When I tried this again, there was no longer a mark by the orientation when I changed the requirement to test, and there was still no check next to the test requirement. I then added an additional Bonner event for All Bonner Meeting, and added Zach; the check for the requirement did not appear next to All Bonner Meeting on their profile either, so neither the test requirement nor the All Bonner Meeting requirement appears.
> > image
> 
> Hi John, I have throughly looked through what you said definitely it is something to think about, I looked at and found the bug but the bug depends what the current term is in test the current term is flexible and we can assigned because of this line of code which it does is find all terms where the student have attended while the total term populate it base on the current term. Lets say the event is created at Spring 2025 while the current term is Spring 2021 because of the line: attendedTerms = set(term for term in attendedTerms if term in totalTerms) it will only accept terms that are with the four year range below Spring 2021. That is why the bug exist. We can definitely talk about this at work and hopefully the bug recreation i did is like yours and this is the issue. If this is the case then we should look into if it is an actual bug or more of a administrative edge case they wouldn't happen in real life. We should definitely talk to @ojmakinde !

Note: There are edge cases that we need to account for when testing, such as the fact that certain students may not be present for specific semesters. As @MImran2002  said, we need to find a term that the student has attended while also making sure that, based on the current term that is set on the manage tab, it is a viable option as a term .

Karina-Agliullova (dismissed):
PR looks good! Tested it thoroughly and the code looks clean. Good job!

PR #1438, opened 2025-04-10
resolved the variable error that happened with finn and steven's merg…

MImran2002:
This is a quick solve on a variable merge conflict that happened out of the fact that steven and finn were working on the same file. After chaning the variable name it got fixed.


WackyWeaver (approved):
Looks good!

PR #1495, opened 2025-06-10
Finsbranch

adjagbodjouf:
## Issue Description

Fixes 1463
- The checkbox was not delivering a successful change message after being checked and unchecked.

## Changes

- Added a change to line 12 of the UserProfile.js file, which sends a success message after the checkbox is marked.

## Testing

- Ran the app and checked and unchecked the box to see if a message was being sent.


sheggen (approved):
Nice work!

PR #1497, opened 2025-06-10
fixed footer in Bonner Management

Arohasina:
## Issue Description
The footer that we have at the bottom of the Bonner Management page seems to move to the middle of the page if we open more than one option on the accordion. The footer needs to be fixated at the bottom and not move

## Changes
- position changed to fixed and bottom set to 0

## Testing

- flask run and expanded all to see if the footer was moving to the middle. It stays at the bottom now.


app/templates/base.html
```diff
@@ -22,7 +22,7 @@
 {% block content %}
 <div>
   <div class="row g-0">
-    <div class="col-md-auto d-print-none d-none d-lg-block" id="sideBarContainer">
+    <div class="col-md-auto d-print-none d-none d-lg-block" id="sideBarContainer" style="z-index: 11;">
```
sheggen: Styling should be done in the CSS file. In this case, base.css file, under the ID `sideBarContainer`:

```
#sideBarContainer{
    z-index: 11;
}
```

sheggen:
This fix didn't work when all three accordions on Bonner Management are open. The two buttons at the bottom are not clickable because they are covered by the footer. 

The bottom of the page should get a `<div>` with a height the same size as the footer. This will force the bottom of the page's actual content to always be above that same sized footer. (I think this will work... give it a shot!).

sheggen (approved):
Nice work!

Issue #1459, opened 2025-04-17
Service Transcript Redesign


Karina-Agliullova:
The service transcript page should be redesigned as it looks bad. There is also confusion between earned hours, service hours, and total hours. It needs to be structured better, The page needs to be more intuitive and user-friendly.

Karina-Agliullova:
Completed in https://github.com/BCStudentSoftwareDevTeam/celts/pull/1502

PR #1502, opened 2025-06-11
Redesigned Service Transcript Webpage  issue #1459

vazquezsotoj:
## Issue Description

#1459
- The service transcript page should be redesigned as it looks bad. There is also confusion between earned hours, service hours, and total hours. It needs to be structured better. The page needs to be more intuitive and user-friendly.

## Changes

- Moved print button to the top right of the page
- Fixed the order of hours, along with moving them to the left
- Changed volunteering to the top center
- Changed font sizes for headers
- centered general layout

## Testing
- In order to see changes, log in as admin, create 2-3 events, and add a student as a volunteer, switch to the student's account, and review the service transcript under My Profile.
-or you can search a student (ex: Zach Neill) and then go to their profile and click on view service transcript

### DIFF
--- a/app/templates/main/serviceTranscript.html
+++ b/app/templates/main/serviceTranscript.html
@@ -34,15 +34,11 @@
         </div>
     {% endfor %}
   {% endif %}
-  <h5 class="card-title bg-light ps-2 pe-5 border-bottom ms-3">Hours</h5>     
   {% if totalHours["totalHours"] > 0%}
-    <div class="float-front ms-3 pt-1">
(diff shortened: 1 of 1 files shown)


Karina-Agliullova:
There are a couple of issues with your changes to the View Service Transcript page. There is no need for heading "Hours". I think differentiating between the different kinds of hours would mean coming up with new language of how to word all the hours so that it does not sound as confusing. Moving everything to the right might be a good idea if only you align it well on the page without any extra spaces. Discuss the design of where the hours should be or the language for the hours with Scott or Brian or me.
Screenshot (77)

Karina-Agliullova (approved):
Looks good!

PR #1510, opened 2025-06-13
Deleted the "allow start date to be in the past" toggle

### LINKED ISSUE #1448: Delete the Allow Start Date to be in Past
Instead of having Allow Start Date to be in the Past, let the user know it is in the past with a little note. Move the multiple offerings toggle to be next to the start date.

BesherKitaz:
## Issue Description

Fixes #1448 
Delete the toggle "Allow start date to be in the past" and allow the user to pick any date in the past or future, prompting a message if in the past. Also change the multiple offerings toggle to be next to "start date."

Screenshot 2025-06-13 094504

## Changes

- Deleted the "allow start date to be in the past" toggle on "Create Event Page"
- Now the user can select dates in the past in the calendar, always, without needing to switch the toggle
- The page will show a message in case the selected start date and time is in the past or/and the end date is in the past
- Changed the "This is a series of events." toggle location in the page 

Screenshot 2025-06-13 094538

## Testing

- Open the CELTS website as an admin or Student Staff and go to "Create Event"
- Under "Programs" Choose any of the list of programs
- Go to start date and pick a date in the past or today
- Pick the start time to be in the past, and try end time in the future or past
- try the program on Firefox since there was a problem there
- Try edge cases in time like 12pm, 12:30pm, etc...

### DIFF
--- a/app/static/js/createEvents.js
+++ b/app/static/js/createEvents.js
@@ -509,8 +509,8 @@ $(document).ready(function() {
       let startDate = new Date($("#repeatingEventsStartDate").val());
       let endDate = new Date($("#repeatingEventsEndDate").val());
       let startTime = $("#repeatingEventsStartTime").val();
-      let endTime = $("#repeatingEventsEndTime").val();      
-startDatePicker
+      let endTime = $("#repeatingEventsEndTime").val();
+
       if (navigator.userAgent.indexOf("Chrome") == -1) {
         startTime = format12to24HourTime(startTime)
         endTime = format12to24HourTime(endTime)
@@ -652,3 +652,7 @@ startDatePicker
 
   setCharacterLimit($("#inputCharacters"), "#remainingCharacters"); 
 });
+
+
+
+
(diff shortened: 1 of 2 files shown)


Karina-Agliullova:
Your changes break the modal that we have for multiple offerings. Find what in your code is breaking it and fix it. I checked it in development, there is no such a problem.
Screenshot (80)

BesherKitaz:
We fixed the problem of the problem of the Modal of the Multiple offerings. There was an extra div in the HTML that was breaking it and we removed it. It should work now after the new pushes.

Karina-Agliullova:
I found another problem. When I click on the event that is happening right now, it displays the phrase that you added, which is good. But if I try to make a weekly event with a modal, and I set it up to be in the past and save it, the phrase doesn't disappear and still stays on the page, which makes it now a false statement. 
Screenshot (83)

BesherKitaz:
I fixed the problem in the new push. I just made sure it the text is removed when the toggle is active (checked), and to add the text again when the toggle is unchecked.

Karina-Agliullova:
Another issue. When I select an event for any date, it lets me use the calendar, which is nice. But when I click on the modal, it is breaking when I save a series of evens, and then uncheck the toggle for multiple series of events, as the calendar disappears and doesn't reappear after the toggle is unchecked. On development it is not an issue, it works correctly, so something in your code is breaking the calendar

BesherKitaz:
I fixed it now. Please let me know if there are any other issues.

Karina-Agliullova (approved):
Looks good! Great work!

PR #1513, opened 2025-06-13
Multi offer modal should have a different validation

### LINKED ISSUE #1447: Multiple Offering Modal should have a different validation
Multiple Offering Modal looks bad when all the fields are left blank. All the fields are red, and it is different from the validation we usually use.

BesherKitaz:
## Issue Description

Fixes #1447 
Multiple Offering Modal (In create event page) looks bad when all the fields are left blank. All the fields are red, and it is different from the validation we usually use.

## Changes
- Removed all border-red classes from the JavaScript file. 
- When the the data is not valid, JS will not insert border-red classes into the html, which removes this kind of styling
- The Flash still happens when the data is not valid, which should be enough for the user to see

Screenshot 2025-06-13 131419

## Testing

- Go to the CELTS website as an admin
- Go to create event page
- Toggle "This is a series of events." 
- On the top left side, if "This series repeats weekly" is toggled, untoggle it
- Click "Add Occurrence" two or three times
- then click "save" with  invalid data (like empty fields)
- Watch for the flash to happen. Red Boxes or borders should NOT happen

### DIFF
--- a/app/static/css/createEvent.css
+++ b/app/static/css/createEvent.css
@@ -37,7 +37,7 @@
     text-align: center;
     position: fixed;
     z-index: 10;
-    animation: flash 6s forwards;
+    animation: flash 10s forwards;
 }
 
 @keyframes flash {
--- a/app/static/js/createEvents.js
+++ b/app/static/js/createEvents.js
@@ -102,7 +102,7 @@ function initializeFlatpickr(obj) {
   });
 }
 
-function createOfferingModalRow({eventName=null, eventDate=null, startTime=null, endTime=null, isDuplicate=false}={}){
+function createOfferingModalRow({eventName=null, eventDate=null, startTime=null, endTime=null}={}){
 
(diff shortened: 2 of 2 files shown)


BrianRamsay (changes requested):
I don't think it's better to remove all of the border changes, because now it's much harder to tell what is wrong if you are missing one event name out of 6 events (for example).

I think a compromise approach would be to remove the red border around each event, but leave the borders around the inputs themselves. Otherwise it is difficult to tell which field had a problem.

Karina-Agliullova:
I noticed that there are now red borders around time selections when other fields and inputs are not selected, even though the time is selected. On development, the times are not in the red border because they have a value, that's how it should be
Screenshot (86)

Karina-Agliullova:
Also, the flash message disappears pretty fast. It would be nice to keep it for longer, maybe 10 seconds

BrianRamsay (approved):

PR #1516, opened 2025-06-16
Dietary restrictions volunteer details 1450

### LINKED ISSUE #1450: Dietary Restriction On Volunteer Details
- Make it Dietary Restriction(s)
- Have the default value be from what they input if they don’t have one.

adjagbodjouf:
## Issue Description

Fixes #1450 
- If a user hasn't inputted a dietary restriction, it would show up as a blank field in the volunteer details page. We needed this field to indicate that they haven't specified a dietary restriction yet. 
- "Dietary restriction" needed to be changed to "dietary restriction(s)". 

## Changes

-Changed     dietRestriction = TextField(null=True) in minor.py file to dietRestriction = TextField(null=False) to change the conditional statement for a null input. Now the volunteer details page shows "Not specified" if the dietRestriction is null.
- All "Dietary restriction" has been changed to "dietary restriction(s)". 

## Testing

- Ran the app and went to the user profile to make sure the words are changed. 
- Ran app and went to volunteer details to see what was displayed for volunteers with no specified dietary restrictions.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -1,6 +1,5 @@
 from app.models import *
 
-
 class User(baseModel):
     username = CharField(primary_key=True)
     bnumber = CharField(unique=True)
@@ -80,4 +79,3 @@ class User(baseModel):
         # Looks to see who the Program Manager for a specific event is
         return self.isProgramManagerFor(event.program)
 
-    
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -16,7 +16,7 @@ class User(baseModel):
     isStaff = BooleanField(default=False)
     isCeltsAdmin = BooleanField(default=False)
     isCeltsStudentStaff = BooleanField(default=False)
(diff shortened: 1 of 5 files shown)


BrianRamsay (dismissed):
- This isn't exactly what we are needing. We need to see a distinction between a volunteer having no dietary restrictions, and an empty field because they have not filled it out. Currently, both cases will show "None specified", which is the same as having it empty for both cases.

- Your description mentions that you changed the field in the database to not allow nulls, but I don't see that file in the pull request.

- I know the ticket asked for it, but I don't think it's necessary to have "Dietary Restriction(s)". We can put it back the way it was, "Dietary Restrictions". The plural is fine in all cases.

BrianRamsay (approved):

PR #1525, opened 2025-06-17
Event Name Population in Main Page

### LINKED ISSUE #1443: Event Name field population when creating event
When creating a new event and making it recurring weekly, you are asked to give it a name in a modal where recurrence happens. It disables the field for event name when you close the modal, and leaves it blank. It should not be so. Populate the event field with the name in the modal to reflect the new changes.

BesherKitaz:
## Issue Description

Fixes #1443
When creating a new event and making it recurring weekly, the event name field in main page appears empty and disabled.  
It disables the field for event name when you close the modal, and leaves it blank. It should not be so. We need to populated with the entered name (or names) from the multiple offerings modal.

image

## Changes

- Now the entered event name in the series event modal appears on the main page
- The main page name is read only when entered through the modal
- Added a function that retrieves the event name(s) from the modal and puts them in the even name field in the main page
- When the event is weekly, the name is added normally
- However, when the event is not weekly multiple names are possible through the modal, so all are added in the event name in the main page, separated by commas

image

image

## Testing

- Go to CELTS Website as an Admin
- From the navigation panel on the left, click "Create Event"
- Toggle "This is a series of events" to open a modal
- Try entering several events with different names and dates where "This series repeats weekly" is not toggled
- Click "Save" and see the result
- Now try to edit the event  and open the modal again, then toggle "This series repeats weekly" and enter name, start and end dates
- Click "Save" and see the result


app/static/js/createEvents.js
```diff
@@ -242,10 +242,36 @@ $('#saveSeries').on('click', function() {
...
+    let offeringsText = ""
+    names.forEach(offering => {
+    offeringsText += offering + ", "
+    })
+    offeringsText = offeringsText.slice(0, -2)
+    $('#inputEventName').prop('placeholder', offeringsText)
+  }
+  else {
+    // if weekly, take the name of the first item (which is the same for all) and take the word 'week'
+    let offeringText = offerings[0].eventName.split(" ").slice(0, -2).join(" ").trimEnd()
```
BrianRamsay: Weren't we just going to take the value in the event box that they just typed? That would be much simpler than pulling it out of the series data.

app/static/js/createEvents.js
```diff
@@ -242,10 +242,36 @@ $('#saveSeries').on('click', function() {
...
+    offerings.forEach(offering => {
+      names.add(offering.eventName)
+      console.log(offering.eventName);
+    });
+    // Check if the event is weekly
+    let offeringsText = ""
+    names.forEach(offering => {
+    offeringsText += offering + ", "
+    })
+    offeringsText = offeringsText.slice(0, -2)
```
BrianRamsay: A more concise way to turn this into a comma-separated string would be `offeringsText = Array.from(offerings).join(", ")`

app/static/js/createEvents.js
```diff
@@ -242,10 +242,36 @@ $('#saveSeries').on('click', function() {
...
+// Populate the Event Name field in the main page with the entered repeating events
+function updateEventNameField() {
+  let offerings = JSON.parse($("#seriesData").val())
+  let isSeries = $("#checkIsRepeating").is(":checked")
+  if (!isSeries) {
+    // if not weeekly, add them to a set to remove duplicates, then put them in a string to populate the field
+    let names = new Set()
+    offerings.forEach(offering => {
+      names.add(offering.eventName)
+      console.log(offering.eventName);
```
BrianRamsay: remove debug output please

BrianRamsay (changes requested):

app/static/js/createEvents.js
BesherKitaz: Fixed in the new push that I made.

app/static/js/createEvents.js
BesherKitaz: Fixed in the new push that I made.

app/static/js/createEvents.js
BesherKitaz: Fixed in the new push that I made.

app/static/js/createEvents.js
```diff
@@ -62,12 +62,16 @@ function setViewForSingleOffering(){
...
+  $('#checkIsSeriesToggleContainer').addClass('form-group col-md-6')
+  $('#checkIsSeriesToggleContainer').removeClass('form-group col-md-12')
 }
 
 function setViewForSeries(){
   $(".startDatePicker").prop('required', false);
   $("#multipleOfferingTableDiv").removeClass('d-none');
   $('#eventTime, #eventDate').addClass('d-none');
+  $('#checkIsSeriesToggleContainer').removeClass('form-group col-md-6')
+  $('#checkIsSeriesToggleContainer').addClass('form-group col-md-12')
```
BrianRamsay: `form-group` is a class we want in both cases, so we don't need to reference it at all in the remove/add lines. We can just change the col classes

BrianRamsay (changes requested):

app/static/js/createEvents.js
BesherKitaz: I removed the form-group class in the new pushes.

BrianRamsay (approved):
Nice work!

PR #1526, opened 2025-06-17
Replaced the Javascript with anchor tag for the service transcript 

### LINKED ISSUE #1518: Refactor Service Transcript link
The service transcript button that was created in this PR https://github.com/BCStudentSoftwareDevTeam/celts/pull/1498 is using javascript and an onclick handler to change to the service transcript page. This is way overkill. 

The button should just be a link with the appropriate bootstrap styling to make it look like a button, and an href to the page.

PuskarC:
## Issue Description

Fixes #1518 

- The 'View Service Transcript' button in Participation History is using javascript and an onclick handler to change to the service transcript page. We need anchor <a> tag instead and get rid of the Javascript.

## Changes

- Removed the javascript onclick handler and all the JS code that was related to the button. 
- Used anchor tag and a button instead to redirect the page to service transcript.

## Testing

- Flask Run and then go to My Profile.
- Under My Profile, open Participation History 
- Then, click on the 'View Service Transcript' button to see the results. 

image

### DIFF
--- a/app/static/js/userProfile.js
+++ b/app/static/js/userProfile.js
@@ -25,7 +25,6 @@ $(document).ready(function(){
       })
   $("#actions").on("change", changeAction)
   $("#phoneInput").inputmask('(999)-999-9999');
-  $("#serviceTranscript").click(viewTranscript); 
   $(".notifyInput").click(function updateInterest(){
     var programID = $(this).data("programid");
     var username = $(this).data('username');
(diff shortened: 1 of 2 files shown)


BrianRamsay (changes requested):
Better, but we don't need a `<button>` at all. Check out the kiosk entry button/link that you just were working on for an example.

BrianRamsay (approved):

PR #1527, opened 2025-06-17
Date Picker For Banning Volunteer 

### LINKED ISSUE #1456: Date Picker For Banning Volunteer
- Doesn’t work on Firefox. Needs to be fixed
- Modal isn’t clear what program they are being banned from
- Make the header a question mark.
- If you first fill out the description and then click the date it won’t let you submit the ban. Needs to be fixed

GalinaP7:
## Issue Description

Fixes #1456
-The datepicker would not work in FireFox but works in Chrome.
-Modal was unclear about what program they were being banned from.
-The question mark was needed in the header.
-Be able to submit a ban if a note was added prior to picking the date.

## Changes
-The datepicker is now able to work in firefox.
-The modal is clear about what program they are being banned from
-The question mark is a header
-Let's you submit the ban if you fill out the description then click the date.
image

## Testing
-Go to Student Search on the side bar and search for Ebenezer Ayisi.
-Select it and it will direct you to their profile where you select to change the eligibility of the individual in a program. 
-Try leaving a note prior to inputting a date to see if you are able to ban the volunteer.


app/controllers/main/routes.py
```diff
@@ -160,6 +160,8 @@ def viewUsersProfile(username):
     """
     This function displays the information of a volunteer to the user
     """
+    isChrome = True if "Chrom" in request.user_agent.string else False
```
BrianRamsay: This line is the same as `isChrome = "Chrom" in request.user_agent.string`. The conditional doesn't add anything

app/static/js/userProfile.js
```diff
@@ -102,14 +102,15 @@ $(document).ready(function(){
   }
 
   // This function is to disable all the dates before current date in the ban modal End Date picker
+  // console.log($.browser);
```
BrianRamsay: remove the debug line

app/static/js/userProfile.js
```diff
@@ -102,14 +102,15 @@ $(document).ready(function(){
...
   $(function(){
     var banEndDatepicker = $("#banEndDatepicker");
     banEndDatepicker.datepicker({
-      changeYear: true,
+      changeYear: true,   
       changeMonth: true,
       minDate:+1,
-      dateFormat: "yy-mm-dd",
-    }).attr('readonly','readonly');
+      dateFormat: "mm-dd-yy",
```
BrianRamsay: We should use the same date format everywhere, with either`m/d/Y` or `Y-m-d` (these aren't exact format strings)

app/static/js/userProfile.js
```diff
@@ -145,12 +146,15 @@ $(document).ready(function(){
...
-
-  $("#banNoteTxtArea, #banEndDatepicker").on('input' , function (e) { //This is the if statement the placeholder in line 45 is for #PLCHLD1
+ $("#banNoteTxtArea, #banEndDatepicker").on('input change' , function (e) { //This is the if statement the placeholder in line 45 is for #PLCHLD1
     var enableButton = ($("#banNoteTxtArea").val() && $("#banEndDatepicker").val());
     $("#banButton").prop("disabled", !enableButton);
   });
 
+$("#banEndDatepicker").on('change', function () {
+  console.log("Datepicker changed:", $(this).val());
+});
```
BrianRamsay: Remove debug output.

app/templates/main/userProfile.html
```diff
@@ -603,7 +603,9 @@ <h6 align="left">Reason for ban:</h6>
             <p align="left" id="banNote">Why the student was banned</p>
           </div>
           <div id="banEndDate">
-            <h6 align="left" class="d-inline-flex" id="unbanVolunteerEndDate">End Date: &nbsp<input id="banEndDatepicker" type="date"></h6>
+            <h6 align="left" class="d-inline-flex" id="unbanVolunteerEndDate">End Date: &nbsp
+              <input id="banEndDatepicker" readonly='readonly'  type= {% if isChrome == True %} "text" placeholder="mm/dd/yyyy" {% else %} "text" placeholder="mm/dd/yyyy" {% endif %}>
```
BrianRamsay: The `== True` part is redundant. Also, what is the difference between the Chrome case and the Firefox case?

BrianRamsay (changes requested):
Banning doesn't appear to work. In Chrome, when I ban a volunteer they still show up as 'Eligible' after banning.

app/controllers/main/routes.py
GalinaP7: Removed conditional and the whole address to the readonly attribute as it did nothing to change the code or UI.

app/static/js/userProfile.js
GalinaP7: Switching the date format in the following line (111) affects the ability for the ban button to work. Switched the format in the HTML file to keep y-m-d this format.

BrianRamsay (dismissed):
We should make the placeholder match the date format. Right now they both use dashes, but the format is different.

BrianRamsay (approved):

PR #1529, opened 2025-06-17
Making an Accordion for the CCE Minor and Changing Buttons

Meatchema:
## Issue Description

- Create the yes and no buttons that would allow for students to be added and removed to the CEE minor list

## Changes
- We added the accordion for CEE minor on the user profile page
- Got rid of the info box
- Changed the checkbox to two buttons (yes/no) 

## Testing

- We tried adding and removing the users from the CEE interested Minor list by clicking the yes and no buttons 
- We checked if the username was sent via Investigator tool
yesno

### DIFF
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -242,7 +242,7 @@ def viewUsersProfile(username):
                                 managersList = managersList,
                                 participatedInLabor = getCeltsLaborHistory(volunteer),
                                 totalSustainedEngagements = totalSustainedEngagements,
-                                expressInterest = volunteer.minorInterest # FIXME
+                                expressInterest = volunteer.minorInterest 
                             )
     abort(403)
 
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -226,6 +226,7 @@ def viewUsersProfile(username):
         totalSustainedEngagements = getEngagementTotal(getCommunityEngagementByTerm(volunteer))
 
         return render_template ("/main/userProfile.html",
+                                username=username,
                                 programs = programs,
                                 programsInterested = programsInterested,
                                 upcomingEvents = upcomingEvents,
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -584,7 +584,7 @@ def indicateMinorInterest(username):
     if g.current_user.isCeltsAdmin or g.current_user.username == username:
         data = request.get_json()
         isAdding = data.get("isAdding", False)
-        
+
         toggleMinorInterest(username, isAdding)
 
     else:
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -242,6 +242,7 @@ def viewUsersProfile(username):
                                 managersList = managersList,
                                 participatedInLabor = getCeltsLaborHistory(volunteer),
                                 totalSustainedEngagements = totalSustainedEngagements,
+                                expressInterest = volunteer.minorInterest # FIXME
                             )
     abort(403)
 
--- a/app/static/css/userProfile.css
+++ b/app/static/css/userProfile.css
@@ -31,10 +31,6 @@ dd {
 .bonnerNotes dd:nth-of-type(odd) { background-color:#FFF; } 
 .bonnerNotes dt:nth-of-type(odd) { background-color:#FFF; }
 
-.form-check.form-switch {
+#togglebuttonuser {
     position: relative; left: 12px; width: 95%;
-}
-
-.success {
-    color:green
 }
--- a/app/static/css/userProfile.css
+++ b/app/static/css/userProfile.css
@@ -26,15 +26,32 @@ dd {
     font-size: 15px;
 }
 
-.minorInfoBox {
-    position: absolute;
-    right: 0px;
-    width: 400px;
-    padding: 5px;
-    border: 1px solid #6666bb;
-    background-color: #e8f6ff; 
-    top: 10px;
+#expressInterest {
+  background-color: #4CAF50;
+  color: white;
+  border: none;
+  padding: 8px 23px;
+  font-size: 12px;
+  font-weight: bold;
+  border-radius: 8px;
+  cursor: pointer;
+  transition: background-color 0.3s ease, transform 0.2s ease;
 }
+
+#expressNoInterest {
+  background-color: #e82222;
+  color: white;
+  border: none;
+  padding: 8px 23px;
+  font-size: 12px;
+  font-weight: bold;
+  border-radius: 8px;
+  cursor: pointer;
+  transition: background-color 0.3s ease, transform 0.2s ease;
+}
+
+
+
 
 .bonnerNotes dd:nth-of-type(even) { background-color:#f2f2f2; } 
 .bonnerNotes dt:nth-of-type(even) { background-color:#f2f2f2; } 
--- a/app/static/css/userProfile.css
+++ b/app/static/css/userProfile.css
@@ -29,4 +29,12 @@ dd {
 .bonnerNotes dd:nth-of-type(even) { background-color:#f2f2f2; } 
 .bonnerNotes dt:nth-of-type(even) { background-color:#f2f2f2; } 
 .bonnerNotes dd:nth-of-type(odd) { background-color:#FFF; } 
-.bonnerNotes dt:nth-of-type(odd) { background-color:#FFF; } 
+.bonnerNotes dt:nth-of-type(odd) { background-color:#FFF; }
+
+.form-check.form-switch {
+    position: relative; left: 12px; width: 95%;
+}
+
+.success {
+    color:green
+}
--- a/app/static/css/userProfile.css
+++ b/app/static/css/userProfile.css
@@ -33,9 +33,6 @@ dd {
   padding: 8px 23px;
   font-size: 12px;
   font-weight: bold;
-  border-radius: 8px;
-  cursor: pointer;
-  transition: background-color 0.3s ease, transform 0.2s ease;
 }
 
 #expressNoInterest {
@@ -45,13 +42,7 @@ dd {
   padding: 8px 23px;
   font-size: 12px;
   font-weight: bold;
-  border-radius: 8px;
-  cursor: pointer;
-  transition: background-color 0.3s ease, transform 0.2s ease;
 }
-
-
-
 
 .bonnerNotes dd:nth-of-type(even) { background-color:#f2f2f2; } 
 .bonnerNotes dt:nth-of-type(even) { background-color:#f2f2f2; } 
--- a/app/static/css/userProfile.css
+++ b/app/static/css/userProfile.css
@@ -26,24 +26,6 @@ dd {
     font-size: 15px;
 }
 
-#expressInterest {
-  background-color: #4CAF50;
-  color: white;
-  border: none;
-  padding: 8px 23px;
-  font-size: 12px;
-  font-weight: bold;
-}
-
-#expressNoInterest {
-  background-color: #e82222;
-  color: white;
-  border: none;
-  padding: 8px 23px;
-  font-size: 12px;
-  font-weight: bold;
-}
-
 .bonnerNotes dd:nth-of-type(even) { background-color:#f2f2f2; } 
 .bonnerNotes dt:nth-of-type(even) { background-color:#f2f2f2; } 
 .bonnerNotes dd:nth-of-type(odd) { background-color:#FFF; } 
--- a/app/static/js/minorAdminPage.js
+++ b/app/static/js/minorAdminPage.js
@@ -85,15 +85,38 @@ function emailAll(){
 }
 
 
+// function getInterestedStudents() {
+//   // get all the checkboxes and return a list of users who's
+//   // checkboxes are selected
+//   let checkboxesDisplayedInModal = $("#addInterestedStudentsModal input[type=checkbox]:checked")
+//   let interestedStudentsList = []
+//   checkboxesDisplayedInModal.each(function(index, checkbox){
+//     interestedStudentsList.push(checkbox["value"])
+//   })
+//   return interestedStudentsList
+// }
+
 function getInterestedStudents() {
   // get all the checkboxes and return a list of users who's
   // checkboxes are selected
-  let checkboxesDisplayedInModal = $("#addInterestedStudentsModal input[type=checkbox]:checked")
-  let interestedStudentsList = []
-  checkboxesDisplayedInModal.each(function(index, checkbox){
-    interestedStudentsList.push(checkbox["value"])
+  // if (#ExpressInterest and "click"){
+  //   let interestedStudentsList = []
+  // ButtonDisplayedInModal.each(function(index, button){
+  //   interestedStudentsList.push(button["click"])
+  // })
+  // return interestedStudentsList
+  // } else if ($("#ExpressInterest") and "click"){
+  //   return false
+  // }
+$("#ExpressInterest").on("click", function(event) {
+  // let checkboxesDisplayedInModal = $("#addInterestedStudentsModal input[type=button]:event.target")
+  
+  let interestedStudentsList = [];
+  checkboxesDisplayedInModal.each(function(button){
+    interestedStudentsList.push("#ExpressInterest")
   })
-  return interestedStudentsList
+  return interestedStudentsList 
+},
 }
 
 function updateInterestedStudents(){
--- a/app/static/js/minorAdminPage.js
+++ b/app/static/js/minorAdminPage.js
@@ -85,38 +85,15 @@ function emailAll(){
 }
 
 
-// function getInterestedStudents() {
-//   // get all the checkboxes and return a list of users who's
-//   // checkboxes are selected
-//   let checkboxesDisplayedInModal = $("#addInterestedStudentsModal input[type=checkbox]:checked")
-//   let interestedStudentsList = []
-//   checkboxesDisplayedInModal.each(function(index, checkbox){
-//     interestedStudentsList.push(checkbox["value"])
-//   })
-//   return interestedStudentsList
-// }
-
 function getInterestedStudents() {
   // get all the checkboxes and return a list of users who's
   // checkboxes are selected
-  // if (#ExpressInterest and "click"){
-  //   let interestedStudentsList = []
-  // ButtonDisplayedInModal.each(function(index, button){
-  //   interestedStudentsList.push(button["click"])
-  // })
-  // return interestedStudentsList
-  // } else if ($("#ExpressInterest") and "click"){
-  //   return false
-  // }
-$("#ExpressInterest").on("click", function(event) {
-  // let checkboxesDisplayedInModal = $("#addInterestedStudentsModal input[type=button]:event.target")
-  
-  let interestedStudentsList = [];
-  checkboxesDisplayedInModal.each(function(button){
-    interestedStudentsList.push("#ExpressInterest")
+  let checkboxesDisplayedInModal = $("#addInterestedStudentsModal input[type=checkbox]:checked")
+  let interestedStudentsList = []
+  checkboxesDisplayedInModal.each(function(index, checkbox){
+    interestedStudentsList.push(checkbox["value"])
   })
-  return interestedStudentsList 
-},
+  return interestedStudentsList
 }
 
 function updateInterestedStudents(){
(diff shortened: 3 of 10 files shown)


app/static/js/minorAdminPage.js
```diff
@@ -88,6 +88,7 @@ function emailAll(){
 function getInterestedStudents() {
   // get all the checkboxes and return a list of users who's
   // checkboxes are selected
+  $("#addInterestedStudentsModal").on
```
BrianRamsay: That doesn't seem like it is supposed to be here.

app/static/css/userProfile.css
```diff
@@ -26,16 +26,33 @@ dd {
...
+  background-color: #e82222;
+  color: white;
+  border: none;
+  padding: 8px 23px;
+  font-size: 12px;
+  font-weight: bold;
+  border-radius: 8px;
+  cursor: pointer;
+  transition: background-color 0.3s ease, transform 0.2s ease;
+}
```
BrianRamsay: These styles are way too specific. Our goal is to have no custom styles. We don't always get there, but a lot of these should already be defined by bootstrap classes. Look at how other buttons are styled (or not).

app/static/js/userProfile.js
```diff
@@ -17,6 +18,26 @@ $(document).ready(function(){
...
+        success: function(s) {
+          msgToast("Changes saved successfully!", "Your interest has been updated.")
+        },
+        error: function(request, status, error) {
+          console.log(status, error)
+          msgToast("Error!", "Failed to save changes!")
+        }
+    });
+  });
+
```
BrianRamsay: Once you've got it working, combine the ajax requests so that you aren't using repeated code for each button.

BrianRamsay (changes requested):

app/static/js/minorAdminPage.js
Meatchema: We have now fixed this issue.

app/static/css/userProfile.css
Meatchema: We fixed this issue by erasing these lines from the yes and no button within this file.

app/static/js/userProfile.js
Meatchema: This issue has now been combined and fix.

app/static/js/userProfile.js
```diff
@@ -1,15 +1,15 @@
 $(document).ready(function(){
-  $("#expressInterest").on("click", function() {
+  $(".interest-button").on("click", function() {
     let username = $(this).data('username')
-    let isAdding = true
+    let isAdding = $(this).data('choice') === true || $(this).data('choice') === "true";
```
Karina-Agliullova: Why do you check for both boolean and string to be true? Stay consistent with one data type

app/static/css/userProfile.css
BrianRamsay: These lines are still here in the pull request

app/templates/main/userProfile.html
```diff
@@ -78,6 +65,43 @@ <h1>{{volunteer.firstName}} {{volunteer.lastName}}</h1>
...
+        <div class="accordion-body">
+          <div class="row table-responsive">
+             <label for="expressInterest">Did you know we have a Community and Civic Engagement Minor? Click 
+             <a href="https://berea.smartcatalogiq.com/en/current/catalog/majors-and-minors/interdisciplinary-minor/community-and-civic-engagement-minor/" target="_blank">here</a>
+             for more information. 
+             </label>
+        
+            <p for="expressInterest">Are you interested in pursuing the CCE Minor?</p>
+
+            <p for="expressInterest"><b>{{volunteer.firstName}} {{volunteer.lastName}} is interested in the CCE Minor.</b></p>
```
BrianRamsay: You don't actually check if they are interested or not, you just say they are.

BrianRamsay (changes requested):

app/static/js/userProfile.js
Meatchema: We have fixed this issue by only using the boolean data type version of true.

app/templates/main/userProfile.html
Meatchema: We have erased this line asking if they are interested.

app/templates/main/userProfile.html
```diff
@@ -78,6 +65,35 @@ <h1>{{volunteer.firstName}} {{volunteer.lastName}}</h1>
...
+<!-- ################# Community & Civic Engagement Minor ################ -->
+  <div class="accordion-item">
+    <h3 class="accordion-header" id="headingZero">
+      <button class="accordion-button {{focus}}" type="button" data-bs-toggle="collapse" 
+      data-bs-target="#collapseZero" aria-expanded="true" aria-controls="collapseZero">Community & Civic Engagement Minor</button>
+    </h3>
+      <div id="collapseZero" class="accordion-collapse collapse {{show}} " aria-labelledby="headingZero" data-bs-parent="#userProfile">
+        <div class="accordion-body">
+          <div class="row table-responsive">
+             <label for="expressInterest">Did you know we have a Community and Civic Engagement Minor? Click
```
sheggen: Clicking this label triggers the msgToast. Convert the label to a `<p>`

app/templates/main/userProfile.html
```diff
@@ -14,20 +14,7 @@
...
 
 {% block app_content %}
-{% if volunteer.isStudent and not totalSustainedEngagements%}
-  <div class="float-end rounded d-none d-lg-block minorInfoBox">
-    <p for="expressInterest">Did you know we have a Community and Civic Engagement Minor? Click 
-      <a href="https://berea.smartcatalogiq.com/en/current/catalog/majors-and-minors/interdisciplinary-minor/community-and-civic-engagement-minor/" target="_blank">here</a>
-      for more information.
-    </p>
-    {% if g.current_user.isAdmin %}
-      <label for="expressInterest">{{volunteer.firstName}} {{volunteer.lastName}} is interested in the CCE Minor.</label>
```
sheggen, line 24: There are three states this could be in: 

1. The user/admin has not made any indication of interest. The language should be something like `Are you interested in the CCE minor?`
2. The user/admin has indicated interest. The language should be something like `Ebenezer has indicated interest in the CCE minor`
3. the user/admin has indicated NO interest. The language should be something like `Ebenezer has indicated they are not interested in the CCE minor`

Also think about which user is looking at which profile. If it's the admin looking at a student profile, it should say their name (e.g., Ebenezer ...). If it's a student looking at their own profile, it should say "You" or "I"

app/templates/main/userProfile.html
Meatchema: This text has been changed to a paragraph.

app/templates/main/userProfile.html
Meatchema: This issue has been resolved.

Karina-Agliullova:
You are using toast to acknowledge the user that a change was made after the toggle is clicked. The toast appears for a second and disappears immediately due to a page reload. Since you are reloading the page after clicking on the toggle, I would ask to use a flash message instead and make it be on the screen for longer than a second

Karina-Agliullova:
Align the toggle with the text right above it. It seems to be a bit more on the left, which after reloading, misplaces the toggle that crosses the border of the accordion

Karina-Agliullova:
I believe that Brian, as an admin and non-student, should not have a toggle asking him to indicate an interest at all. First, it doesn't record his response anyway. If I click on interest as a user, Brian Ramsay, who is not a student, the toggle would be clicked, and the toast message appears; however, the minor management will not record his response. Double-check with Scott or Brian whether we can remove the functionality of the toggle for admins who are non-students. I believe that it is redundant and is not used anyway

petrosyantsa:
Changes:
- we have modified toast message so that it can display the message depending if they say yes or no to the minor
- we have changed the access to the minor accordion to only allow students and student staff to see it
- we have aligned the toggle with the text above it

app/controllers/main/routes.py
```diff
@@ -242,6 +242,7 @@ def viewUsersProfile(username):
                                 managersList = managersList,
                                 participatedInLabor = getCeltsLaborHistory(volunteer),
                                 totalSustainedEngagements = totalSustainedEngagements,
+                                expressInterest = volunteer.minorInterest # FIXME
```
BrianRamsay: we don't want to leave a `#FIXME` in the code. Does something still need to be fixed?

app/templates/main/userProfile.html
```diff
@@ -78,6 +65,45 @@ <h1>{{volunteer.firstName}} {{volunteer.lastName}}</h1>
 </div>
 <div class="accordion" id="userProfile">
 
+<!-- ################# Community & Civic Engagement Minor ################ -->
+{% if g.current_user.isCeltsStudentStaff or g.current_user.isStudent %}
```
BrianRamsay: We don't need both booleans here, because anyone who is Student Staff is already a student as well.

BrianRamsay: Also, you are checking to see if the logged in user is a student before showing this, not if the volunteer is a student.

app/templates/main/userProfile.html
```diff
@@ -78,6 +65,45 @@ <h1>{{volunteer.firstName}} {{volunteer.lastName}}</h1>
...
+  <div class="accordion-item">
+    <h3 class="accordion-header" id="headingZero">
+       {% set focus = "open" if visibleAccordion == "MinorTable" or (not visibleAccordion and volunteer.username != g.current_user.username) else "collapsed" %}
```
BrianRamsay: I don't think we ever want this to default to open. Only if MinorTable is specified.

BrianRamsay (changes requested):

app/controllers/main/routes.py
petrosyantsa: No, we just forgot to erase it. Now its fixed

app/templates/main/userProfile.html
petrosyantsa: Fixed

app/templates/main/userProfile.html
petrosyantsa: fixed

app/static/css/userProfile.css
```diff
@@ -26,17 +26,15 @@ dd {
...
-.bonnerNotes dt:nth-of-type(odd) { background-color:#FFF; } 
+.bonnerNotes dt:nth-of-type(odd) { background-color:#FFF; }
+
+.form-check.form-switch {
+    position: relative; left: 12px; width: 95%;
+}
+
+.success {
+    color:green
+}
```
BrianRamsay: It's very unlikely that we need a custom class for this. bootstrap already has some, most likely: https://getbootstrap.com/docs/4.0/utilities/colors/

app/static/css/userProfile.css
```diff
@@ -26,17 +26,15 @@ dd {
...
-    top: 10px;
-}
-
 .bonnerNotes dd:nth-of-type(even) { background-color:#f2f2f2; } 
 .bonnerNotes dt:nth-of-type(even) { background-color:#f2f2f2; } 
 .bonnerNotes dd:nth-of-type(odd) { background-color:#FFF; } 
-.bonnerNotes dt:nth-of-type(odd) { background-color:#FFF; } 
+.bonnerNotes dt:nth-of-type(odd) { background-color:#FFF; }
+
+.form-check.form-switch {
```
BrianRamsay: This isn't a very good selector because it will catch any `form-check form-switch` on the page, and we are styling a very specific positioning issue. I would create a new class for this element and use that here.

app/templates/main/userProfile.html
```diff
@@ -78,6 +65,45 @@ <h1>{{volunteer.firstName}} {{volunteer.lastName}}</h1>
...
+      data-bs-target="#collapseZero" aria-expanded="true" aria-controls="collapseZero">Community & Civic Engagement Minor</button>
+    </h3>
+      <div id="collapseZero" class="accordion-collapse collapse {{show}} " aria-labelledby="headingZero" data-bs-parent="#userProfile">
+        <div class="accordion-body">
+          <div class="row table-responsive">
+             <p for="expressInterest">Did you know we have a Community and Civic Engagement Minor? Click 
+             <a href="https://berea.smartcatalogiq.com/en/current/catalog/majors-and-minors/interdisciplinary-minor/community-and-civic-engagement-minor/" target="_blank">here</a>
+             for more information. 
+             </p>
+              <div class="form-check form-switch"class=".form-check.form-switch">
```
BrianRamsay: that's an invalid class name.

app/static/js/userProfile.js
```diff
@@ -1,17 +1,5 @@
...
-        $("#diet").val("No dietary restrictions");
-
-    } else {
-        $("#dietContainer").show();
-    }
-  });
-
-
-  $("#expressInterest").on("click", function() {
+  $(".form-check-input").on("change", function() {
```
BrianRamsay: This is a pretty generic class, when you are really trying to target the change behavior of one specific element. I would probably add an id to it and reference it that way.

app/templates/main/userProfile.html
```diff
@@ -78,6 +65,45 @@ <h1>{{volunteer.firstName}} {{volunteer.lastName}}</h1>
...
+              <strong><label class="custom-control-label" id="interestIndicatedText" for="expressInterest">
+                {% if expressInterest is none %}
+                  Are you interested in the CCE Minor?
+                {% else %} 
+                  {{volunteer.firstName}} {{volunteer.lastName}} has indicated
```
BrianRamsay: Let's change this from first and last name to "You have". Because only the student sees this now.

app/static/js/userProfile.js
```diff
@@ -20,15 +8,20 @@ $(document).ready(function(){
...
         contentType: "application/json",
-        success: function(s) {
-          msgToast("Changes saved successfully!", "Your interest has been updated.")
+        success: function(response) {
+          if (isAdding) {
+              msgToast("Interest Indicated", "You have shown interest in this minor.");
+              $("#interestIndicatedText").text("You're interested in this minor.");
+             } else {
+              msgToast("Interest Removed", "You are no longer interested in this minor.");
+              $("#interestIndicatedText").text("You're not interested in this minor.");
```
BrianRamsay: The message text here needs a little work. They are a little bland, and don't acknowledge the user's choice. Something more like "Thanks for expressing interest in the Community and Civic Engagement minor!" and "You have been removed from communication about the CCE minor."

BrianRamsay (changes requested):
Getting there!

app/static/css/userProfile.css
petrosyantsa: Fixed

app/static/css/userProfile.css
petrosyantsa: we changed the 2nd class to the id and now it works

app/templates/main/userProfile.html
petrosyantsa: fixed

app/static/js/userProfile.js
petrosyantsa: Fixed

app/templates/main/userProfile.html
petrosyantsa: Fixed

app/static/js/userProfile.js
petrosyantsa: Changed

Karina-Agliullova:
In UI, when I click on the toggle it gives me the text of "Thanks for expressing interest ...", and if I refresh the page, the clicked toggle says: "You have indicated interest in CCE Minor". Stay consistent instead of changing the text from one text to another. No need to thank the user for clicking the toggle. Just keep that text after they check that they are interested: "You have indicated interest in CCE Minor"

Karina-Agliullova:
Your toast messages are bulky containing too many words. Come up with a better toast message that doesn't take up so much space or spread the words accordingly so that it doesn't seem so bulky

Karina-Agliullova:
When logged in as a student staff, Zach Neill, I have access to other people's CCE minor interest accordion, which should not be the case. So if I am logged in as Zach Neill, I go to student search and search for "Ebenezer Ayisi", it opens the student profile, and shows me that I can indicate the minor interest toggle for that student, but then it fails and shows me error message. The CCE minor accordion should be hidden for when student staff sees other student's profile page. 
Screenshot (96)

Karina-Agliullova:
You use different messages for the toggle text when the interest is off. The first is "You have indicated you are not interested in the CCE Minor." and if I click on the toggle and unclick, it shows me a different text: "You have removed your interest from CCE minor." Keep it consistent. Leave only the first version "You have indicated you are not interested in the CCE Minor.". The second text message is already used in the toast message

GalinaP7:
-On the admin acount (ramsayb2), you can still see CCE minor with two separate buttons for it yes/no, but even if it was needed, it should match the same format as the other buttons. Same issue for faculty user (heggens) AND student staff (neillz). Also, clicking yes/no simply states that interest has been updated but does not fix to let the user know whether they are no longer interested. It will always display that the user is interested in the CCE Minor (for example Scott Heggen is interested in the CCE Minor).
image

app/templates/main/userProfile.html
```diff
@@ -78,6 +65,47 @@ <h1>{{volunteer.firstName}} {{volunteer.lastName}}</h1>
...
+                {% endif %}
+              </label></strong>
+                <input class="form-check-input" data-username ="{{volunteer.username}}" 
+                type="checkbox" id="checkIsInterest" name="isInterested" {%if expressInterest %} checked {% endif %} />
+             </div>
+          </div>
+        </div>
+     </div>
+  </div>
+  {% else %}
```
Karina-Agliullova: What's {%else%} for in here? I don't think it is needed

app/templates/main/userProfile.html
```diff
@@ -78,6 +65,47 @@ <h1>{{volunteer.firstName}} {{volunteer.lastName}}</h1>
...
+                  in the CCE Minor.                                
+                {% endif %}
+              </label></strong>
+                <input class="form-check-input" data-username ="{{volunteer.username}}" 
+                type="checkbox" id="checkIsInterest" name="isInterested" {%if expressInterest %} checked {% endif %} />
+             </div>
+          </div>
+        </div>
+     </div>
+  </div>
```
Karina-Agliullova, line 104: I think you are missing one more closing div here to close the accordion class

app/templates/main/userProfile.html
BesherKitaz: This is now removed.

app/templates/main/userProfile.html
BesherKitaz: This has now been added after the two end if statements.

Karina-Agliullova (dismissed):
Looking good!

app/static/js/userProfile.js
```diff
@@ -1,34 +1,29 @@
...
-  });
-
-
-  $("#expressInterest").on("click", function() {
+  $("#checkIsInterest").on("change", function() {
     let username = $(this).data('username')
     let isAdding = $(this).is(':checked');
-    
+    let accept = "You have indicated interest in CCE Minor.";
+    let decline = "You have indicated you are not interested in the CCE Minor.";
```
BrianRamsay: We should declare these variables next to where we use them.

app/static/js/userProfile.js
```diff
@@ -1,34 +1,29 @@
...
-        success: function(s) {
-          msgToast("Changes saved successfully!", "Your interest has been updated.")
+        success: function(response) {
+          if (isAdding) {
+              msgToast(accept);
+              $("#interestIndicatedText").text(accept);
+          } else {
+              msgToast(decline);
+              $("#interestIndicatedText").text(decline);
+          }
```
BrianRamsay: I remember talking about this before, and having variables for the message was part of the solution. But we also can get rid of the repeated lines with something like: 
```
let msg = isAdding ? accept : decline;
msgToast(msg);
$("#interestIndicatedText").text(msg);
```

BrianRamsay: Also, the toast looks a little weird with all of the text in the header. The header should probably be something like "Success"

BrianRamsay (changes requested):

app/static/js/userProfile.js
BesherKitaz: Done, moved it to within the success part of the ajax request, where they are used.

app/static/js/userProfile.js
BesherKitaz: Done for both.

BrianRamsay (approved):

PR #1531, opened 2025-06-17
Event title for celts sponsored

### LINKED ISSUE #1520: Fix event title for CELTS-Sponsored events
The event sub-title for CELTS-Sponsored event creation currently says "A CELTS-Sponsored Event Event". The second "Event" should be removed for this program.

Image

Make sure you don't break it for other programs!

- The subtitle on the View/Edit Event page should be correct for all programs.
- The Create Event page should read "Create a CELTS-Sponsored Event" instead of "Create Event for CELTS-Sponsored Event"

GalinaP7:
## Issue Description

Fixes #1520
-The Event Subtitle Displays "A CELTS-Sponsored Event Event"
-Create Event Page Title is messy "Create Event for CELTS-Sponsored Event"

## Changes
-Event only is placed in the title once.
-Create Event Page Title now states "Create a CELTS-Sponsored Event"
image
image

## Testing
-Go to Create Event and press CELTS-Sponsored Event
-Add a new event then check view/edit event for the subtitle that should display only event once

### DIFF
--- a/app/templates/events/createEvent.html
+++ b/app/templates/events/createEvent.html
@@ -320,6 +320,7 @@
         {% if event and not (event.isCanceled or eventPast) %}
         <button type="button" class="btn btn-warning bordered-element m-2" data-bs-toggle="modal"
           data-bs-target="#cancelWarning">Cancel Event</button>
+
         {% elif event and (event.isCanceled) %}
         <button type="button" class="btn btn-success" data-bs-toggle="modal" data-bs-target="#renewWarning">Renew
           Event</button>
--- a/app/templates/events/createEvent.html
+++ b/app/templates/events/createEvent.html
@@ -7,7 +7,13 @@
     {% set page_title = 'Create a ' + eventData["program"].programName %}
 
   {% elif template.tag == 'single-program' %}
-    {%set page_title = 'Create a ' + eventData["program"].programName %}
+    {% if eventData["program"].programName == 'CELTS-Sponsored Event' %}
+      {%set page_title = 'Create a ' + eventData["program"].programName %}
+    {% elif eventData["program"].programName == 'Adopt-a-Grandparent' %}
+      {% set page_title = 'Create an ' + eventData["program"].programName + ' Event' %}
+    {% else %}
+      {%set page_title = 'Create a ' + eventData["program"].programName + ' Event' %}
+    {% endif %}
 
   {% elif eventData["program"].programName == 'CELTS-Sponsored Event' and template.tag == 'all-volunteer' %}
     {% set page_title = 'Create All Volunteer Training' %}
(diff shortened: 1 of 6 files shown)


BrianRamsay (changes requested):
This has all of the changes from your other PR.
The fix for this issue doesn't work, in any case, because it breaks for the other programs.

BrianRamsay (changes requested):
It still doesn't work for the regular program case. "A Berea Tutoring Event" now just reads "A Berea Tutoring", for example.

app/templates/events/createEvent.html
```diff
@@ -7,7 +7,13 @@
     {% set page_title = 'Create a ' + eventData["program"].programName %}
 
   {% elif template.tag == 'single-program' %}
-    {%set page_title = 'Create Event for ' + eventData["program"].programName %}
+    {% if eventData["program"].programName == 'CELTS-Sponsored Event' %}
+      {%set page_title = 'Create a ' + eventData["program"].programName %}
+    {% elif eventData["program"].programName == 'Adopt-a-Grandparent' %}
```
sheggen: Can we future-proof this to check if the program starts with A, E, I, O, or U, and use a/an appropriately?

app/templates/events/createEvent.html
```diff
@@ -7,7 +7,13 @@
...
+      {% set page_title = 'Create an ' + eventData["program"].programName + ' Event' %}
+    {% else %}
+      {%set page_title = 'Create a ' + eventData["program"].programName + ' Event' %}
+    {% endif %}
 
   {% elif eventData["program"].programName == 'CELTS-Sponsored Event' and template.tag == 'all-volunteer' %}
```
sheggen, line 15: This condition cannot be reached anymore, because your first condition will catch "CELTS-Sponsored Event"

app/templates/events/createEvent.html
BrianRamsay: Just like we do for the subtitle on the other page.

app/templates/events/createEvent.html
```diff
@@ -7,7 +7,13 @@
     {% set page_title = 'Create a ' + eventData["program"].programName %}
```
BrianRamsay: If CELTS-Sponsored Event no longer triggers this conditional (because it always has a template, maybe) it should be removed.

BrianRamsay (dismissed):

Karina-Agliullova:
Something in your code breaks the Manage Volunteers tab when I am in editing mode for an event I created. Investigate and solve the problem
Screenshot (89)

Karina-Agliullova:
Other tabs except View Event and Edit Event are broken too. Same error for all of them

app/templates/main/userProfile.html
```diff
@@ -597,7 +597,7 @@ <h5 class="modal-title">Add Notes</h5>
     <div class="modal-dialog">
       <div class="modal-content">
         <div class="modal-header">
-          <h5 class="modal-title-ban"></h5>
+          <h5 class="modal-title-ban">Ban Volunteer</h5>
```
Karina-Agliullova: Why is this fix here? Seems unrelated to an issue

Karina-Agliullova (approved):
Looks good!

PR #1533, opened 2025-06-19
Labor only option

### LINKED ISSUE #1521: Add a Labor Only option for events.
We need a new boolean option for events that indicates whether the event is Labor Only. This will involve a new database field, defaulting to False, and should be added to the optional section of the event creation/edit page.

BesherKitaz:
## Issue Description

Fixes #1521
We need to add a new toggle in the "Create Event Page" that makes the event "Labor Only" as a Boolean value. The toggle value should be sent to the database in a new field and we should be able to view it when editing the event or view it. 

## Changes
- Added a new toggle with a label of "This event is for labor only" under the optional section of Create/edit event page under the description
- Added a field in the database to capture the result of this toggle
- Changed the event controllers and logic so that they handle the new toggle (processEventData function and saving the event)
- Added a text in the event view page that says "This event is for labor students only." if the value of the field in the database is 1 (true) (if the newly added toggle is checked)
- If the toggle is unchecked, nothing new happens in the event view page

image

image

## Testing

- Go to CELTS website as an admin
- Click Create Event Option on your left-hand side panel
- Select any program, **except for** "All Volunteer Training"
- Fill out the requirement event information
- On the right-hand side in the bottom of the page, see under (Optional section) the new toggle
- Try checking and unchecking it to make sure it works
- Click "save" while the toggle is checked to save
- You will prompted with an edit edit page. The toggle should remain checked
- Try navigating to the view page as the image below shows:

image

- You should be able to see the new text there, as highlighted in the image below:

image

- Try navigating back to "Edit Event" using the same navigation bar you used to view the page
- Uncheck the toggle and click "save" to save
- Navigate back to "View Event." The text you saw earlier should be removed:

image

- Lastly, try creating a new event without checking the toggle. The text should not be there. And when you edit the event and check it, the text should appear

### DIFF
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -116,6 +116,8 @@ def createEvent(templateid, programid):
     if request.method == "POST":
         savedEvents = None
         eventData.update(request.form.copy())
+        print("Hello World")
+        print(eventData)
         eventData = preprocessEventData(eventData)
         if eventData.get('isSeries'):
             eventData['seriesData'] = json.loads(eventData['seriesData'])
--- a/app/controllers/admin/routes.py
+++ b/app/controllers/admin/routes.py
@@ -269,7 +269,7 @@ def eventDisplay(eventId):
         event = Event.get_by_id(eventId)
         invitedCohorts = list(EventCohort.select().where(
             EventCohort.event == event
-        ))
+        )) 
         invitedYears = [str(cohort.year) for cohort in invitedCohorts]
     except DoesNotExist as e:
         print(f"Unknown event: {eventId}")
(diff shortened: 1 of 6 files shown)


BrianRamsay (approved):
Nice work! Very thorough, finding the other places that we show the event details.

BrianRamsay (changes requested):
There is a test failing that we need to fix

BesherKitaz:
> There is a test failing that we need to fix

The tests are now working. The code was expecting a value for the toggle, but the test data did not have that since it is a new toggle, so I just added it to the test and associated the value "True" for test data.

app/controllers/admin/routes.py
```diff
@@ -117,6 +117,7 @@ def createEvent(templateid, programid):
         savedEvents = None
         eventData.update(request.form.copy())
         eventData = preprocessEventData(eventData)
+        print("Debug here", eventData)
```
BrianRamsay: No debug output, please

BrianRamsay (dismissed):

BrianRamsay (approved):

PR #1536, opened 2025-06-23
Event participant none hours

### LINKED ISSUE #1162: Remove the ability for an event participant to have None hours earned
**Low priority issue, leave for summer cohort**

The `EventParticipant` model having a nullable `hoursEarned` field is old functionality leftover from when `EventParticipant` and `EventRsvp` were the same table. There really shouldn't be a case when their hours earned is `None`, at the very least it should only be 0. We need to change the following:

- The `EventParticipant` model's `hoursEarned` field
- Test data where `hoursEarned` is set to 0
- Tests that test for `hoursEarned` to be None (at least one is in test_spreadsheet
- Make sure there isn't any place in logic where we are setting their `hoursEarned` values

GalinaP7:
## Issue Description

Fixes #1162
- hoursEarned in EventParticipant model should not have a nullable parameter
- There shouldn't be a case when hoursEarned is None (should be 0)

## Changes
- hoursEarned now has null set to false and a default of 0
- in test_spreadsheet added a program 4 that tests event4 where eventparticipant4 has 0 hours earned and is part of term2
- 
image
image
image

## Testing
- go to the test_spreadsheet.py file
- run the following in the terminal: pytest -v tests/code/test_spreadsheet.py
-ensure that the test for hoursEarned to be 0 passes and the test for hoursEarned None to fail

### DIFF
--- a/app/models/eventParticipant.py
+++ b/app/models/eventParticipant.py
@@ -5,7 +5,7 @@ from app.models.event import Event
 class EventParticipant(baseModel):
     user = ForeignKeyField(User)
     event = ForeignKeyField(Event, backref="participants")
-    hoursEarned = FloatField(null=True)
+    hoursEarned = FloatField(null = False, default = 0)
 
     # Add this property so that we can combine these objects with EventRsvp objects in one array
     @property
--- a/app/templates/events/manageVolunteers.html
+++ b/app/templates/events/manageVolunteers.html
@@ -117,11 +117,7 @@
                           type ="number"
                           min="0"
                           step="0.01"
-                            {% if participant.hoursEarned == None %}
-                              value=""
-                            {% else %}
                               value='{{participant.hoursEarned}}'
-                            {% endif %}
                         />
                       </td>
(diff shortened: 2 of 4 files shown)


app/models/eventParticipant.py
```diff
@@ -5,7 +5,7 @@
 class EventParticipant(baseModel):
     user = ForeignKeyField(User)
     event = ForeignKeyField(Event, backref="participants")
-    hoursEarned = FloatField(null=True)
+    hoursEarned = FloatField(null = False, default = 0)
```
sheggen, line 8: @BrianRamsay Will this leave problems elsewhere? Are we using `Null` checks elsewhere in code?

BrianRamsay: A good catch. There's a check in `app/templates/events/manageVolunteers.html` that is no longer relevant.

BrianRamsay (changes requested):

BrianRamsay (approved):

PR #1537, opened 2025-06-23
User management refactor

### LINKED ISSUE #1517: program info should be converted into an ajax request
- the adminProgramManagement in usermanagement.html should not be dumped directly into the front end and should be an ajax request

BesherKitaz:
## Issue Description

Fixes #1517
The adminProgramManagement data for program information in usermanagement.html is being dumped directly into the frontend using Jinja, for all programs, even those we don't request the details for. We need to refactor this into an AJAX request. When the page is loaded, the data will not be dumped. Instead, it will be only requested through AJAX after we request the program detailed through "edit details" button.

## Changes

- No UI or functionality changes for the user, only changes in how the request works 
- Removed the data attribute to the "Edit details" button that contained all program details, in file userManagement.html
- We only have the programid through this button
- The rest of the data is requested upon clicking the button through an AJAX request
- Edited the JavaScript userManagement.js to receive the data from an AJAX request instead of getting it from the html attribute to populate the form with the existing data
- Added a new API endpoint in Python (userManagement.py) that sends the program data to accommodate the AJAX request

## Testing

- Go to CELTS website as an Admin
- On the left-hand side navigation panel, click "admin" to view administrator options
- From the new appearing options, click "settings"
- Click the accordion button that says "Program Management"
- Click "Edit Details" for any program
- The input fields should be populated with old data
- Try changing this data, and save
- Try clicking "edit" again to see if the changes happened
- Try going to the main CELTS page and see if the changes also reflect there

image


app/controllers/admin/userManagement.py
```diff
@@ -76,6 +78,24 @@ def updateProgramInfo(programID):
             abort(500,'Error while updating program.')
     abort(403)
 
+
+@admin_bp.route('/admin/getProgramInfo/<programID>', methods = ['GET'])
+def getProgramInfo(programID):
+    if g.current_user.isCeltsAdmin:
```
BrianRamsay, line 84: Student staff can update program information for the programs they manage, so program managers need access to their programs.

app/controllers/admin/userManagement.py
```diff
@@ -76,6 +78,24 @@ def updateProgramInfo(programID):
...
+        targetProgram = Program.get_by_id(programID)
+        programInfo = model_to_dict(targetProgram, recurse=False)
+        try:
+            return jsonify([programInfo])
+        except DoesNotExist:
+            flash('item not found')
```
BrianRamsay: "item not found" is not a good error message to display to the user. Is this something that they can control?

BrianRamsay: Also, the `Does Not Exist` error is thrown by `get_by_id()`, isn't it? not `jsonify`

app/static/js/userManagement.js
```diff
@@ -77,10 +77,26 @@ $(document).ready(function(){
...
+    let programInfoData;
+    try {
+      programInfoData = await $.ajax({
+        method: 'GET',
+        url: "/admin/getProgramInfo/" + programid,      
+      })
+    }
+    catch (error) {
+      console.error("Failed to retrieve program info", error);
+      alert("Could not retrieve program info. Please try again later.")
```
BrianRamsay: We don't use `alert` for our error handling. Use toasts, or contextual messages.

app/static/js/userManagement.js
```diff
@@ -77,10 +77,26 @@ $(document).ready(function(){
...
+
+    // Get the program data using an AJAX request and the retrieve programid
+    let programInfoData;
+    try {
+      programInfoData = await $.ajax({
+        method: 'GET',
+        url: "/admin/getProgramInfo/" + programid,      
+      })
+    }
+    catch (error) {
```
BrianRamsay: Let's use the `$.ajax` handlers for this, rather than javascript try/catch.

app/static/js/userManagement.js
```diff
@@ -77,10 +77,26 @@ $(document).ready(function(){
...
     submitTerm();
   });
-  $('[data-bs-target="#adminProgramManagement"]').on('click', function() {
-    // Get the JSON data from the data-programinfo attribute
-    const programInfo = JSON.parse($(this).attr('data-programinfo'));
-    // Directly populate modal fields
+
+  $('[data-bs-target="#adminProgramManagement"]').on('click', async function() {
+    // Get the programid from the attribute data-programid
+    const programid = JSON.parse($(this).attr('data-programid'))
```
BrianRamsay: Why do we need to parse this as JSON?

BrianRamsay (changes requested):

app/static/js/userManagement.js
BesherKitaz: Removed it in new pushes.

app/static/js/userManagement.js
BesherKitaz: Done.

app/static/js/userManagement.js
BesherKitaz: Done.

app/controllers/admin/userManagement.py
BesherKitaz: Done, both.

app/controllers/admin/userManagement.py
BesherKitaz: I added access to student staff.

BesherKitaz:
I added (with Scott's help) some code to make sure the modal is toggled after the data is loaded. However, if the button is clicked twice in a row (before it opens), and then closed, it locks the screen and the user cannot do anything and they have to reload. I will address this tomorrow to fix it.

app/controllers/admin/userManagement.py
```diff
@@ -76,6 +78,25 @@ def updateProgramInfo(programID):
             abort(500,'Error while updating program.')
     abort(403)
 
+
+@admin_bp.route('/admin/getProgramInfo/<programID>', methods = ['GET'])
+def getProgramInfo(programID):
+    if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff:
```
BrianRamsay: There is an `isAdmin` property of user that handles both cases

app/static/js/userManagement.js
```diff
@@ -77,49 +77,82 @@ $(document).ready(function(){
...
-    $('#modalProgramImage').val('');
-    $('#modalProgramImageContainer').html('');
-
-    handleFileSelection('modalProgramImage', true);
-    // Update the form action URL dynamically
-    let updateForm = $('#updateProgramForm');
-    updateForm.attr('action', "/admin/updateProgramInfo/" + programInfo.programid);
+
+  $('.editDetails').on('click', function() {
+    // Get the programid from the attribute data-programid
```
BrianRamsay: These comments are just restating what the code below is doing. The code should speak for itself as much as possible.

app/static/js/userManagement.js
```diff
@@ -77,49 +77,82 @@ $(document).ready(function(){
...
+    const programid = $(this).attr('data-programid')
+    // Get button id so we can target it to disable it, since the spinner does not disable it automatically
+    const buttonId = $(this).attr('id')
+    // buttonTextId and loadingSpinnerId all end the same loop index variable, which can retrieve so we can target them
+    const buttonTextId = 'editDetailsButtonText' + $(this).attr('loop-index')
+    const loadingSpinnerId = 'editDetailsButtonSpinner' + $(this).attr('loop-index')
+    // Remove the regular text of the button and adding the spinner and the word "loading" and disable the button
+    $('#' + buttonTextId).addClass('d-none')
+    $('#' + loadingSpinnerId).removeClass('d-none')
```
BrianRamsay: Let's use `.show()` and `.hide()` for this

app/static/js/userManagement.js
```diff
@@ -77,49 +77,82 @@ $(document).ready(function(){
...
+          $("#bereaUrl").val(programInfo.bereaUrl);
+          $('#modalProgramImage').val('');
+          $('#modalProgramImageContainer').html('');
+
+          handleFileSelection('modalProgramImage', true);
+          // Update the form action URL dynamically
+          let updateForm = $('#updateProgramForm');
+          updateForm.attr('action', "/admin/updateProgramInfo/" + programid);
+          // Making sure the data loads before the modal opens, which avoids erros in case the frontend is disconnected from the back
+          // Scott Suggestion
```
BrianRamsay: We don't need to know this in the code.

app/static/js/userManagement.js
```diff
@@ -77,49 +77,82 @@ $(document).ready(function(){
...
+          modal.show();
+
+          // Remove the spinner and the word "loading" and adding back the regular text
+          $('#' + buttonTextId).removeClass('d-none')
+          $('#' + loadingSpinnerId).addClass('d-none')
+          $('#' + buttonId).prop( "disabled", false );
+        }, 
+        error: function () {          
+            console.error("Failed to retrieve program info", error);
+            msgToast("Could not retrieve program info.")
```
BrianRamsay: Messages to the users will ideally have some suggestion for them. Should they try again? Call someone?

app/static/js/userManagement.js
```diff
@@ -77,49 +77,82 @@ $(document).ready(function(){
...
+        }
     });
+  });
+
+  $(".editProgramManagersButton").on('click', function(){
+    $('#programPlaceholder').attr('data-programid', $(this).attr('data-programid'));
```
BrianRamsay: jquery provides a `.data()` method for interacting with data parameters. Where possible, let's use it instead of `.attr` (this isn't code you originally wrote, but it's relevant for that code, too)

BrianRamsay (changes requested):

app/controllers/admin/userManagement.py
BesherKitaz: Yes. We changed that to "g.current_user.isAdmin()"

However, we just want to note that it is actually a method. This is what we found in user.py in models.

app/static/js/userManagement.js
BesherKitaz: Tried to change it as we could and there are less comments now.

app/controllers/admin/userManagement.py
BesherKitaz: Sorry. You were right. It is a property. It is my first time seeing the decorator "@property". Thank you.

app/static/js/userManagement.js
BesherKitaz: Removed.

app/static/js/userManagement.js
BesherKitaz: Done. We removed the d-none class. However, we had to use plain CSS (display: none;) for the spinners, because we need it to be initially hidden and we cannot wait for the JavaScript to work to hide it. .show() and .hide() don't work with d-none, but it works with plain CSS.

app/static/js/userManagement.js
BesherKitaz: Will let them contact support.

BrianRamsay (approved):

PR #1542, opened 2025-06-25
1454 travel forms button

### LINKED ISSUE #1454: Add A Print Travel Forms Button To Volunteer Details
Add A Print Travel Forms Button To Volunteer Details

mathew2005:
## Issue Description

Fixes #1454
- Add a dropdown with the Print Travel Forms Option to Volunteer Details

## Changes
- Added a Dropdown that encompasses the Print Travel Form button with Print Volunteer List in the volunteer details page.
- "Prints Travel Form" option on the dropdown prints Travel forms for everyone in the volunteer's list, depending on the number of volunteers. 
- image
- Used User's Travel Form page to work for `eventTravelForm` controller, allowing us to reuse the `travelForm.html` for both. 
- Preview, when printed:  
[Welcome to CELTS Link.pdf](https://github.com/user-attachments/files/20981286/Welcome.to.CELTS.Link.pdf)

## Testing
1.  As an admin, go to Events List 
2. Create an event if you haven't already 
3.  Select an event
4. Navigate to Manage Volunteers (add volunteers if needed)
5. Go to Volunteer Details
6.  Now you should be able to click on the Print Dropdown -> Print Travel Form


BrianRamsay (changes requested):
What happened to the previous travel form printing? We shouldn't have two separate pages that do the same thing (but one loops). Either have one template and route that handles both cases, or have separate paths that re-use code so that we aren't repeating queries and html.

Karina-Agliullova:
I agree with Brian that we should reuse what we already have for printing travel forms and not duplicate html. Another issue for me is I think the button should say Print Travel Form (and Forms whenever there is more than one person in volunteers). Also, there is another Print button right next to yours and it makes it confusing. Make sure you differentiate what each button is doing exactly because as of now it doesn't seem intuitive

mathew2005:
> What happened to the previous travel form printing? We shouldn't have two separate pages that do the same thing (but one loops). Either have one template and route that handles both cases, or have separate paths that re-use code so that we aren't repeating queries and html.

We were able to change the format of the data being sent by the `travelForm` controller to a list so it can loop for both event travel form page and user's travel form page.

mathew2005:
> I agree with Brian that we should reuse what we already have for printing travel forms and not duplicate html. Another issue for me is I think the button should say Print Travel Form (and Forms whenever there is more than one person in volunteers). Also, there is another Print button right next to yours and it makes it confusing. Make sure you differentiate what each button is doing exactly because as of now it doesn't seem intuitive

We were able to use a drop down to resolve the issue in our latest commit.

app/templates/main/travelForm.html
```diff
@@ -146,13 +148,12 @@ <h2>Statement</h2>
     <br>
     <div class='underlined d-print-block'>{{userData.healthIssues}}</div>
 </div>
+{% endfor %}
+{% endif %}
 
+{% block script%}
+<script>
+    print()
```
Karina-Agliullova, line 154: Not sure what is going on here, but this is incorrect and not needed here

app/templates/base.html
```diff
@@ -73,8 +73,7 @@
           </div>
           <div class="spacer d-print-none my-0"></div>
         </div>
-      </div>
-
+        </div>
```
Karina-Agliullova: You added the wrong indentation and I don't think you needed to make any changes to this file

app/templates/main/travelForm.html
Shenderaz: The "print()" simply gives a print preview on the travel form so the user can print it.

app/templates/base.html
Shenderaz: Reverted base.html, mistakenly changed from its original state.

Karina-Agliullova (dismissed):
Looks good and works well!

app/templates/main/travelForm.html
```diff
@@ -5,8 +5,10 @@
 <link rel="stylesheet" href="{{url_for('static', filename = 'css/travelForm.css')  }}?u={{lastStaticUpdate}}">
 {% endblock %}
 
-{% block app_content %}
-<div class= "w-100 d-print-block">
+{% block content %}
+{% if userList %}
```
BrianRamsay: Is there a case where we aren't passing in userList? I'd say we can get rid of that conditional.

app/templates/events/volunteerDetails.html
```diff
@@ -127,7 +127,14 @@ <h3 style="text-align: center; width:180%;">
               <button class="btn btn-primary" id="tableCardToggle" nowrap>Card View</button>
             </div>
             <div class="col-auto align-items-end">
-              <button id="printVolunteerInfo" class="btn-primary btn">Print</button>
+               <select class="form-select" id="printSelect" align="right">
```
BrianRamsay: Let's make that `<select>` a bootstrap dropdown, so it matches the buttons styling: https://getbootstrap.com/docs/4.0/components/dropdowns/

app/static/js/volunteerDetails.js
```diff
@@ -1,4 +1,11 @@
 $(document).ready(function () {
+	var userdata = $('.volunteerInfoEntries');
+	var users = userdata.map(function(index, row) {
+      return $(row).data('user');
+    }).get();
```
BrianRamsay, line 5: let's use an arrow function here to get a nice, clean looking one-liner.

BrianRamsay (changes requested):
I like the code re-use!

app/templates/main/travelForm.html
mathew2005: That makes sense, the volunteer details tab already does the checking for us.

app/templates/events/volunteerDetails.html
mathew2005: Resolved in a newer commit.

app/static/js/volunteerDetails.js
mathew2005: Resolved in the newer commit.

app/controllers/main/routes.py
```diff
@@ -329,13 +329,48 @@ def travelForm(username):
...
+    try:
+        event = Event.get_by_id(eventID)
+    except DoesNotExist as e:
+        print(f"No event found for {eventID}", e)
+        abort(404)
+
+    if not (g.current_user.isCeltsAdmin):
+        abort(403)
+
+    if request.method == "POST" or request.form.getlist("username") != (None or []):
```
BrianRamsay: The second part of this expression doesn't handle the None case, because `(None or [])` is evaluated first, and always evaluates to `[]` because of short-circuit evaluation. So the expression is equivalent to `request.form.getlist("username") != []`. Do you need to be checking for None?

BrianRamsay (changes requested):
One minor thing left!

BrianRamsay (approved):

Issue #1385, opened 2025-01-27
Normalize flash message usage


bledsoef:
All over CELTS different pages use different kind of flash messages. We have toasts, flash messages from flask, and flash messages from Javascript. They all behave differently and look differently. It would be a good idea to attempt to normalize these the best we can.

bledsoef:
![](https://github.trello.services/images/mini-trello-icon.png) [Normalize flash message usage](https://trello.com/c/siaQWhwd/858-normalize-flash-message-usage)

Karina-Agliullova:
Completed in https://github.com/BCStudentSoftwareDevTeam/celts/pull/1543

PR #1543, opened 2025-06-26
Editing toasts flashes #1385

Jsmi8334:
## Issue Description

Issue: #1385
All over CELTS different pages use different kind of flash messages. We have toasts, flash messages from flask, and flash messages from Javascript. They all behave differently and look differently. It would be a good idea to attempt to normalize these the best we can.

## Changes

The logic we used for implementing flashes and toasts and custom validity tags is that we are using toasts when we are on a Modal and we need updates, success, deletion messages popped up next to the modal. If there is a required field a custom Validity tag will show up.  We are using flashes for page reloads and anything that does not use a modal so the user will see the flash in the center of the page. We are using Custom validity tags for all required fields throughout the page.

this is a toast: 
image

This is a flash: 
image

This is a custom validity tag:
image

- Create Event page we added custom validity tags to all required fields. On the same page we changed the order of the (Must select one) and the optional boxes so the user can easily recognize the required fields 
- On my profile we added success, deletion messages for adding background checks, as well as for the notes section
- On my profile we created flashes for the changing of the phone number 
- Fixed all toast to be flash messages and added custom validity tag to Status and Date in Description
- On the proposal For Service-Learning Course page we created custom validity tags for the required fields.
- On Settings --> program management page --> edit manager, we added msg toast for Manager deletion. 
- On my profile --> Description, we added the required sign onto the status and date of background check 
- We added a timer to the toast 3 sec (you can test this by creating a note under My Profile --> Note --> save)

## Testing

- User Profile: When opening Background Check the status and the date are required fields and they now show custom validity tags. If bg check is deleted then we get flash message.
image
- Create Event pg: 
- All of the required fields will show a custom validity tag if not filled out. (asteris- shows required field). There are three main forms a user can fill out are All Volunteer Training, Bonnar scholar, and the rest of the other avaible forms. 
image
- we added success and deletion messages for all of the specific pages in itself for the three main pages shown above. When a form is submitted or deleted or if there is an error. 
- we edited the series of events modal if the toggle is switched on:  Adding custom validity tags to the required fields. This functionality is also seen when on the modal and "this series repeats weekly is selected. 
- Under Admin settings: 
image
-  if the removed button is pressed then a success flash will show up for deletion
- On the same page and under program management if Edit details or Edit managers is selected a success flash will show. On the Edit Managers Modal, if a manager is removed a toast will pop up.
image
- On the CCE Minor Page 
image  if declare student button is pressed or you remove a student we added a Flash
image

### DIFF
--- a/app/static/js/base.js
+++ b/app/static/js/base.js
@@ -6,7 +6,17 @@ function flashMessageResponse(message) {
   return '';
 }
 
-function msgFlash(flashMessage, status, timeout=2500){
+function msgFlash(flashMessage, status, timeout=2500, afterReload=false) {
+    if (afterReload) {
+      // Save message to sessionStorage for next page load
+      sessionStorage.setItem('flashMessage', JSON.stringify({
+          message: flashMessage,
+          type: status,
+          timeout: timeout
+      }));
+      return;
+    }
+
     if (!["success", "warning", "info", "danger"].includes(status)) status = "danger";
     $("#flash_container").prepend(`
       <div class="alert alert-${status} alert-dismissible alert-success" role="alert">${flashMessage}
--- a/app/static/js/base.js
+++ b/app/static/js/base.js
@@ -6,6 +6,20 @@ function flashMessageResponse(message) {
   return '';
 }
 
+function enableLiveCustomValidityClearing(selectors) {
+  selectors.forEach(selector => {
+    $(selector).each(function () {
+      // Avoid rebinding listeners on already-bound elements
+      if (!$(this).data("has-clearing-listener")) {
+        $(this).on("input", function () {
+          this.setCustomValidity("");
+        });
+        $(this).data("has-clearing-listener", true); // flag it
+      }
+    });
+  });
+}
+
 function msgFlash(flashMessage, status, timeout=2500, afterReload=false) {
     if (afterReload) {
       // Save message to sessionStorage for next page load
--- a/app/static/js/base.js
+++ b/app/static/js/base.js
@@ -42,6 +42,10 @@ function msgToast(head, body){
   $("#toast-header").html(head)
   $("#toast-body").html(body)
   toastList[0].show()
+
+  setTimeout(() => {
+    toastList[0].hide();
+  }, 3000);
 }
 
 function setupPhoneNumber(editButtonId, phoneInput){
--- a/app/static/js/base.js
+++ b/app/static/js/base.js
@@ -6,7 +6,7 @@ function flashMessageResponse(message) {
   return '';
 }
 
-function msgFlash(flashMessage, status, timeout=null){
+function msgFlash(flashMessage, status, timeout=2500){
     if (!["success", "warning", "info", "danger"].includes(status)) status = "danger";
     $("#flash_container").prepend(`
       <div class="alert alert-${status} alert-dismissible alert-success" role="alert">${flashMessage}
@@ -114,7 +114,7 @@ function validatePhoneNumber(editButtonId, phoneInputId, username) {
             "phoneNumber":phoneInput.val()},
       success: function(s){
         $(phoneInputId).attr("data-value",phoneInput.val())
-        msgToast("Phone Number", "Successfully updated the phone number.")
+        msgFlash("Successfully updated the phone number.","success" )
       },
       error: function(request, status, error) {
         msgFlash("Phone number not updated.", "danger")
--- a/app/static/js/base.js
+++ b/app/static/js/base.js
@@ -34,7 +34,7 @@ $(document).ready(function() {
 
 });
 
-function msgToast(head, body){
+function msgToast(head, body, duration=3000){
   if ($("#liveToast").is(":visible") == true){
     $('#liveToast').removeClass("show")
     $('#liveToast').addClass("hide")
@@ -45,7 +45,7 @@ function msgToast(head, body){
 
   setTimeout(() => {
     toastList[0].hide();
-  }, 3000);
+  }, duration);
 }
 
 function setupPhoneNumber(editButtonId, phoneInput){
--- a/app/static/js/base.js
+++ b/app/static/js/base.js
@@ -17,6 +17,19 @@ function msgFlash(flashMessage, status, timeout=2500, afterReload=false) {
       return;
     }
 
+    if (!flashMessage || !status) {
+      const storedMessage = sessionStorage.getItem('flashMessage');
+      if (storedMessage) {
+        const messageData = JSON.parse(storedMessage);
+        flashMessage = messageData.message;
+        status = messageData.type;
+        timeout = messageData.timeout ?? 2500;
+        sessionStorage.removeItem('flashMessage');
+      } else {
+        return; // Nothing to show
+      }
+    }
+
     if (!["success", "warning", "info", "danger"].includes(status)) status = "danger";
     $("#flash_container").prepend(`
       <div class="alert alert-${status} alert-dismissible alert-success" role="alert">${flashMessage}
--- a/app/static/js/bonnerManagement.js
+++ b/app/static/js/bonnerManagement.js
@@ -76,9 +76,7 @@ $(document).ready(function(){
         } else {
             fileName = `Bonner Spreadsheet, ${Number(startingYear) - Number(noOfYears)} - ${startingYear}`;
         }
-        // const fileName = noOfYears === "all" 
-        //     ? "Bonner Spreadsheet, All Cohorts" 
-        //     : `Bonner Spreadsheet, ${Number(startingYear) - Number(noOfYears) + 1} - ${Number(startingYear) + 1}`
+
         $.ajax({
             url: url,
             method: "GET",
--- a/app/static/js/bonnerManagement.js
+++ b/app/static/js/bonnerManagement.js
@@ -183,10 +183,10 @@ function saveRequirements() {
                     row.data('id', id);
                 }
             });
-            msgToast("Bonner", "Updated Bonner Requirements");
+            msgFlash("Updated Bonner Requirements","success");
         },
         error: function(e) {
-            msgToast("Error", "Error Saving Requirements");
+            msgFlash("Error Saving Requirements","danger");
         }
     });
 }
@@ -249,6 +249,7 @@ function addRequirementsRowHandlers() {
         // Only remove if it isn't the last row
         if($("#requirements tbody tr").length > 1) {
             $(e.target.closest('tr')).fadeOut(function() { this.remove() });
+            msgFlash("Succsessful deletion of Bonner Requierment", "success")
         }
     });
 }
--- a/app/static/js/bonnerManagement.js
+++ b/app/static/js/bonnerManagement.js
@@ -76,6 +76,9 @@ $(document).ready(function(){
         } else {
             fileName = `Bonner Spreadsheet, ${Number(startingYear) - Number(noOfYears)} - ${startingYear}`;
         }
+        // const fileName = noOfYears === "all" 
+        //     ? "Bonner Spreadsheet, All Cohorts" 
+        //     : `Bonner Spreadsheet, ${Number(startingYear) - Number(noOfYears) + 1} - ${Number(startingYear) + 1}`
         $.ajax({
             url: url,
             method: "GET",
@@ -183,10 +186,10 @@ function saveRequirements() {
                     row.data('id', id);
                 }
             });
-            msgFlash("Updated Bonner Requirements","success");
+            msgToast("Bonner", "Updated Bonner Requirements");
         },
         error: function(e) {
-            msgFlash("Error Saving Requirements","danger");
+            msgToast("Error", "Error Saving Requirements");
         }
     });
 }
@@ -249,7 +252,6 @@ function addRequirementsRowHandlers() {
         // Only remove if it isn't the last row
         if($("#requirements tbody tr").length > 1) {
             $(e.target.closest('tr')).fadeOut(function() { this.remove() });
-            msgFlash("Succsessful deletion of Bonner Requierment", "success")
         }
     });
 }
--- a/app/static/js/createEvents.js
+++ b/app/static/js/createEvents.js
@@ -179,7 +179,7 @@ $('#saveSeries').on('click', function(e) {
   let startDateInput = $("#repeatingEventsStartDate");
   let endDateInput = $("#repeatingEventsEndDate");
   
-  let hasErrors = false; // Single flag to track all validation errors
+  let hasErrors = false; 
   
   // Validate repeating events fields first if it's a repeating event
   if (isRepeatingStatus) {
@@ -203,7 +203,6 @@ $('#saveSeries').on('click', function(e) {
     
   } else {
     // Validate individual event offerings for non-repeating events
-    
     // Check event name fields
     eventNameInputs.each((index, eventNameInput) => {
       if (eventNameInput.value.trim() === '') {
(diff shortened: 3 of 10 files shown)


app/static/js/base.js
```diff
@@ -114,7 +114,7 @@ function validatePhoneNumber(editButtonId, phoneInputId, username) {
             "phoneNumber":phoneInput.val()},
       success: function(s){
         $(phoneInputId).attr("data-value",phoneInput.val())
-        msgToast("Phone Number", "Successfully updated the phone number.")
+        msgFlash("Successfully updated the phone number.","success" )
```
BrianRamsay, line 158: Aren't we wanting to use toasts for ajax calls?

app/static/js/bonnerManagement.js
```diff
@@ -76,6 +76,9 @@ $(document).ready(function(){
         } else {
             fileName = `Bonner Spreadsheet, ${Number(startingYear) - Number(noOfYears)} - ${startingYear}`;
         }
+        // const fileName = noOfYears === "all" 
+        //     ? "Bonner Spreadsheet, All Cohorts" 
+        //     : `Bonner Spreadsheet, ${Number(startingYear) - Number(noOfYears) + 1} - ${Number(startingYear) + 1}`
```
BrianRamsay: What's up with this commented out section? It seems like it makes more sense than the if statement above it. Was it causing an error having fileName redefined?

BrianRamsay (changes requested):
There are a lot of debug comments in your javascript. Also, let's be consistent with using jquery to select elements. 

Particularly for the create events validation function, it would be helpful to have some comments that indicate what different sections of code are doing.

app/static/js/base.js
Jsmi8334: we thought it looked more normalized as a flash whenever the phone number is updated since we decided to use toast only for modals. 
but if you like the style better as a toast then we can switch it to a toast

Karina-Agliullova:
When I create or edit an event, there is a modal for selecting a series of events. I can create event on different dates and time with different names in that modal. As of now, it doesn't update me when I create or delete events on that modal. Would be nice to have a flsha message for both occasions (addition and deletion) once details in the modal are saved and page is reloaded to update the user that the change has been made

app/static/js/userProfile.js
```diff
@@ -265,16 +267,20 @@ $(document).ready(function(){
 
         if (bgStatus == '') {
           bgStatusInput.focus()
-          bgStatusInput.addClass("invalid");
-          window.setTimeout(() => bgStatusInput.removeClass("invalid"), 1000);
+           $('.form-select').each(function() {
+                bgStatusInput[0].setCustomValidity("Please enter a status");
```
Karina-Agliullova, line 266: The validity tag shows "Please enter status" even when the status is already filled. The validity tag should disappear when the person selects the status.

app/static/js/createEvents.js
```diff
@@ -384,6 +384,120 @@ function formatDate(originalDate) {
...
+      $(".series").each(function(){
+      if (!$(this).is(":visible") || $(this).is(":disabled")) return;
+            // Skip event type checkboxes from regular validation
+      let elementId = $(this).prop("id");
+      if (elementId === "checkIsTraining" || elementId === "checkServiceHours" || //checking if any of the toggles are toggled
+          elementId === "checkEngagement" || elementId === "checkBonners") {
+        return; // Skip these, they'll be validated separately
+      }
+
+      if ($(this).val() === ""){
```
Karina-Agliullova: The value should be trimmed if the person uses empty space - .trim()

app/static/js/createEvents.js
```diff
@@ -384,6 +384,120 @@ function formatDate(originalDate) {
...
+              document.getElementById("checkEngagement").setCustomValidity("Please select at least one of the event options.");
+              document.getElementById("checkEngagement").reportValidity();
+              allFieldFilled = false;
+            } else {
+              document.getElementById("checkEngagement").setCustomValidity("");
+          }
+      } 
+      else if (isAllVolunteer) {
+      $(".allV").each(function(){
+        if (!$(this).is(":visible") || $(this).is(":disabled")) return;
```
Karina-Agliullova: The validity tag should disappear when the value is actively being entered in the datepicker, the event name, the description, the location, and the toggles. It may also be on one of the lines below, possibly, reportValidity, line 454. For example, once they start entering the description, the tag should disappear.

app/static/js/createEvents.js
```diff
@@ -167,19 +167,27 @@ $('#saveSeries').on('click', function() {
   eventNameInputs.each((index, eventNameInput) => {
     if (eventNameInput.value.trim() === '') {
       isEmpty = true;
-      $(eventNameInput).addClass('border-red');
+      //$(eventNameInput).addClass('border-red');
```
Karina-Agliullova: Remove the comment lines that are not used in the code in this file

app/static/js/minorAdminPage.js
```diff
@@ -19,6 +19,7 @@ $(document).ready(function() {
           contentType: "application/json",
           success: function(response) {
             location.reload()
+            msgFlash("Candidate minor succsessfully removed", "success")
```
Karina-Agliullova: This message disappears very quickly after a page reload when i remove a student from the table. Make it stay for longer

Karina-Agliullova: At times, theflash message doesn't even appear on the screen when I remove the student

app/static/js/userProfile.js
```diff
@@ -265,16 +267,20 @@ $(document).ready(function(){
 
         if (bgStatus == '') {
           bgStatusInput.focus()
-          bgStatusInput.addClass("invalid");
-          window.setTimeout(() => bgStatusInput.removeClass("invalid"), 1000);
+           $('.form-select').each(function() {
```
bakobagassas, line 265: when we save either the date or the status in the background check (one of them being empty and the other one being completed) twice, then we come back to the row, there is a custom validity tag that remains stuck to the status box even after we select something from the status box until we click somewhere on the screen and then the custom validity tag finally disappears (though I don't know if that is actually a problem)

app/static/js/createEvents.js
```diff
@@ -148,103 +148,139 @@ function createOfferingModalRow({eventName=null, eventDate=null, startTime=null,
...
+      } else {
+        $(eventNameInput)[0].setCustomValidity("");
+      }
+    });
 
+    // Check date picker fields
+    datePickerInputs.each((index, datePickerInput) => {
+      if (datePickerInput.value.trim() === '') {
+        hasErrors = true;
+        $(datePickerInput)[0].setCustomValidity("Please enter an event date");
```
bakobagassas, line 221: When we try to save a fully empty form, there is a custom validity tag that appears for the "must select one" section, while the calendar displays at the same time. However, we should have the tag appearing at the calendar's level instead, since the calendar is the one opening up at first.

Karina-Agliullova:
I was testing creating all volunteer training, and i discovered that if I create series of events, and try to delete it in that modal, there is a toast message with Deletion info saying that "You have successfully deleted a serie of events." There is a typo, should be saying "series" instead of "serie." Very minor fix. I can't find that message in your code though, maybe that toast message was created before you picked up the issue, but would be nice if you could fix it

Karina-Agliullova:
When I add a note on User Profile and hit save button, it save and toast message is displayed for less than a second, it feels like. I can't even tell what is written on it. Make the toast message for Notes longer

Karina-Agliullova:
Can we add Validation Tags for text and datepicker for Ban/Unban Volunteer modal?
Screenshot (106)

edit: Brian told us we should leave it for now

Karina-Agliullova:
Can we add flash message when cancelling event? As of now, there is nothing showing up when I cancel event except that text under the title that stays there for as long as the event is cancelled.
Screenshot (107)

Karina-Agliullova:
Weird behavior of validity tags when creating course proposals. When I create a new course proposal and it asks me for course instructor, course name, etc. for some reason, if I click save and continue, it doesn't let me and shows me validity tag, which is good. However, that validity tage points at Course instructor, which is the first question asked in that form, but it puts my cursor for typing something into a Course name text box, so it seems like that validity tag is now associated with a different text box
Screenshot (108)

Karina-Agliullova:
Related to creating course proposals. There is a behavior similar to the one that you fixed on another page. It asks me with validity tag to input course isntructor, and when I do, it still shows me a validity tag for that specific textbox, even though I added one instructor already

app/static/js/minorAdminPage.js
```diff
@@ -2,29 +2,43 @@ import searchUser from './searchUser.js'
...
+        type: 'POST',
+        url: '/profile/' + username + '/indicateInterest',
+        data: JSON.stringify({ "isAdding": isAdding }),
+        contentType: "application/json",
+        success: function(response) {
+            // Store message in sessionStorage to survive page reload
+            sessionStorage.setItem('flashMessage', JSON.stringify({
+                message: "Candidate minor successfully removed",
+                type: "success"
+            }));
```
BrianRamsay: We should consolidate the session stuff in our `msgFlash` function, so that people who want to use flashes don't have to know how they are persisted. Maybe you could just call `msgFlash` like usual, but with a parameter that indicates it should be after the next page load, instead of immediately.

BrianRamsay (changes requested):

app/static/js/userProfile.js
```diff
@@ -216,8 +216,13 @@ $(document).ready(function(){
...
              "bonner": isBonner ? "yes" : "no"},
       success: function(response) {
+        msgToast("Successfully added a note", "", 2000)
         target = isBonner ? "bonner" : "notes"
-        reloadWithAccordion(target)
+        // Delay reload to let user see the message
+            setTimeout(function() {
+                reloadWithAccordion(target)
+                location.reload()
+            }, 2500)
```
BrianRamsay: Shouldn't this be using the new flash that shows up on the next page reload instead of a toast?

app/static/js/userProfile.js
```diff
@@ -239,7 +245,24 @@ $(document).ready(function(){
...
+      // Avoid rebinding listeners on already-bound elements
+      if (!$(this).data("has-clearing-listener")) {
+        $(this).on("input", function () {
+          this.setCustomValidity("");
+        });
+        $(this).data("has-clearing-listener", true); // flag it
+      }
+    });
+  });
+}
```
BrianRamsay: We create this function twice on different pages. We should probably have it in base.js and you can just pass in the selectors you want to work with.

app/static/js/userProfile.js
```diff
@@ -249,41 +272,48 @@ $(document).ready(function(){
...
+        $.ajax({
+          url: "/addBackgroundCheck",
+          type: "POST",
+          data: data,
+          success: function(s){
+            var date = new Date(data.bgDate + " 12:00").toLocaleDateString()
+            msgFlash(`Successfully added background check`, "success", 1300)
+            setTimeout(function() {
+            reloadWithAccordion("background")
+            }, 1500)
```
BrianRamsay: Same detail here, instead of delaying reload we could just show the flash after reload

app/templates/events/createEvent.html
```diff
@@ -75,14 +75,10 @@ <h1 id="pageTitle">{{page_title}}</h1>
   <div class="form-group col-md-6" id="eventDate">
     <label class="required form-label me-2" for="startDatePicker-{{pageLocation}}"><strong>Start Date</strong></label>
     <div class='input-group date startDate' id="startDate-{{pageLocation}}" data-page-location="{{pageLocation}}">
-      <input autocomplete="off" type='text' class="form-control datePicker startDatePicker readonly"
-        value="{{ eventData.startDate.strftime('%Y/%m/%d') if eventData.startDate and eventData.startDate.strftime else eventData.startDate}}"
-        name="startDate" placeholder="Pick a start date" id='startDatePicker-{{pageLocation}}'
+      <input autocomplete="off" type='date' class="form-control datePicker startDatePicker readonly main allV"
```
BrianRamsay, line 78: There are a few places with `all` and `allV` classes, but I don't see anywhere they are referenced in css or js. What do they do?

app/templates/events/createEvent.html
```diff
@@ -246,63 +241,58 @@ <h1 id="pageTitle">{{page_title}}</h1>
...
+          <label class="custom-control-label" for="checkFood"><strong>This event will provide food.</strong></label>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
+{% endif %}
+<div class="form-group mb-5">
+  <label class="mb-2" for='attachmentObject'><strong>Add Files</strong></label>
+  <input type="file" class="form-control" id="attachmentObject" name="attachmentObject" multiple />
```
BrianRamsay: Looks like we lost the list of file types it accepts (old line 304). Was that intentional?

app/templates/events/createEvent.html
```diff
@@ -466,22 +456,21 @@ <h5 class="modal-title">This is an event with multiple time offerings.</h5>
...
             </div>
-          </div>
-            
              <!-- Repeating Event End Date Picker -->
              <div class="form-group col-md-3">
               <label class="required form-label me-2 mb-0" for="repeatingEventsEndDate"><strong>Last Event Date</strong></label>
               <div class='input-group date'>
                 <input autocomplete="off" type='date'
-                  class="form-control repeatingEventsEndDate repeatingEventsField readonly"
+                  class="form-control repeatingEventsEndDate repeatingEventsField readonly seriesWeekly"
```
BrianRamsay, line 473: Is `seriesWeekly` used anywhere?

app/templates/events/createEvent.html
```diff
@@ -551,7 +540,7 @@ <h3 class="mt-1 d-flex justify-content-center"> Generated Events: </h3>
             <div class='input-group date multipleOfferingDate' id="multipleOfferingDate-{{pageLocation}}"
               data-page-location="{{pageLocation}}">
               <input autocomplete="off" type='date'
-                class="form-control datePicker multipleOfferingDatePicker readonly"
+                class="form-control datePicker multipleOfferingDatePicker readonly-series"
```
BrianRamsay: Is `readonly-series` used anywhere?

BrianRamsay (changes requested):

BrianRamsay:
As part of these changes we should write up a page in http://ssdt-documentation.berea.edu about our usage of flashes and toasts, and what our policy is, and a page about using custom validity, what we want, and how to implement it.

app/templates/events/createEvent.html
Arohasina: it is a way to classify the different required fields based on what type of template they are on
allv: start date and event name (on the modal)
all: location, description, event name 
we had to split them up so that only the field that are on the page are being checked and it would allow submission

app/templates/events/createEvent.html
Arohasina: we added it back

app/templates/events/createEvent.html
Arohasina: should be series

app/templates/events/createEvent.html
Arohasina: yes in this function:
function checkValidation() in createEvents.js line 452

Karina-Agliullova (approved):
Everything looks good!

Arohasina:
added Flashes, Toasts, and Custom Validity Tags explanation and usage to the documentation: http://ssdt-documentation.berea.edu/en/ssdt-program/FlashesToasts

BrianRamsay (approved):

PR #1546, opened 2025-06-30
Added a Location field for every single event listing

### LINKED ISSUE #1540: Adding a Location field for every single event listing
In the create event page, there is only one location field. This is convenient if there is only 1 event. However, when want to add different events through the modal, it is not convenient for multiple events with different locations. We need to add a Location field for every single event listing in the multi offer modal, instead of one location field in the main page.

PuskarC:
## Issue Description

Fixes #1540 
- The original create event page had only one location field, which worked for single events but was inconvenient for adding multiple events with different locations using the weekly offerings modal.

## Changes

- Added a new Location field inside the weekly event offerings modal (weekly_offerings_modal.html).
- Removed the single Location field from the main create event when the user clicks the 'series of events' toggle. 
- Updated JavaScript logic to handle the dynamic location field input and ensure it's tied to the correct event entry.
- Made it mandatory to enter an event name before toggling the multi-offering modal (to avoid blank entries).
- Adjusted modal logic to ensure each event listing now includes Name, Date, Time, and Location.
- Also added session number for every event added when we click on 'This is a series of events' and then click on Add Occurrence

## Testing

- Navigate to the CELTS Website and go to 'Create Event' from the dashboard.
- Select a Program and then check the event creation form
- Confirm the main Location field is no longer present when we click on the toggle for 'This is a series of events'
- Toggle weekly event offerings and add multiple weekly events.
- Each entry should now have its own Location field alongside Name, Date, and Time.
- Look for session number for every event added by clicking 'Add Occurrence'
-Try entering different locations for each event and save.
- After submission, verify that all event details, including locations, are stored correctly in the database.

***BEFORE***

image

In the modal when we click on 'Add Occurrence', we don't see a textbox for Event Location and then the Location textbox still appears outside the modal. 

image

There's no Event Location in this modal too when we click on 'This series repeats weekly'

image

There's no session name like session 1, session 2..... when we click on 'Add Occurrence' 

***AFTER***

image

image

image

image


bakobagassas:
When we save an occurrence without inputting a location, it still saves, although there is that red star that indicates that the location input is compulsory. The user should not be able to save when the location field is empty

bakobagassas:
When we save an event that repeats weekly for 1 month for instance, the location fields remain undefined even though we inputted a location while creating the weekly event. We should have that location appearing on each of the weekly events (or have the option to select different locations too, I guess?)

Karina-Agliullova:
The code is broken. As Sara mentioned above, first of all, it lets me save weekly event even without inputting location. But when I hit save, it outputs a broken table with nothing in it. And if I hit save that whole event, it throws a validation error
Screenshot (101)
Screenshot (102)

BesherKitaz:
The modal does not allow two events with the same date, name and time even though the location is different. For example, if there is an event that happens in two separate locations, but the other details are identical, they should be able to add it without a problem. We can see in the picture below that the location is different for each, yet still showing an error.
image

Also, when all details are identical, the validation message says "Event listing cannot have the same event name, date, and start time" and does not mention location. We should clarify to the user that the location is also part of the problem:
image

app/controllers/admin/routes.py
```diff
@@ -71,7 +71,7 @@ def switchUser():
 
 @admin_bp.route('/eventTemplates')
 def templateSelect():
-    if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff:
+    if g.current_user.isCeltsAdmin or g.current_user.isCeltsStudentStaff or g.current_user.isStudent:
```
BesherKitaz: Are you sure we should let students access this page to create new events? Just checking because I don't know if that's a new thing

app/static/js/createEvents.js
```diff
@@ -61,6 +64,8 @@ function calculateRepeatingEventFrequency(){
 function setViewForSingleOffering(){
   $(".startDatePicker").prop('required', true);
   $("#multipleOfferingTableDiv").addClass('d-none');
+  $("#eventLocation-main").removeClass('d-none');
```
BesherKitaz: I know it was initially my idea to use removeClass() and addClass(), but Brian later said that we should try to use .hide() and .show(). So I would recommend using them here if possible.

app/static/js/createEvents.js
```diff
@@ -69,6 +74,8 @@ function setViewForSingleOffering(){
 function setViewForSeries(){
   $(".startDatePicker").prop('required', false);
   $("#multipleOfferingTableDiv").removeClass('d-none');
+  $("#eventLocation-main").addClass('d-none');
+  $("#inputEventLocation-main").prop('required', false);
```
BesherKitaz: Same as comment above.

Jonimz:
We've addressed several of the comments on our PR. Regarding Sara's feedback, Aro and Juno are currently working on the flash messages, and Brian advised us to wait for their changes to be merged before proceeding further on that part. We've updated our PR, please let us know if there are any additional comments.

app/controllers/admin/routes.py
```diff
@@ -84,7 +84,7 @@ def templateSelect():
 
 @admin_bp.route('/eventTemplates/<templateid>/<programid>/create', methods=['GET','POST'])
 def createEvent(templateid, programid):
-    if not (g.current_user.isAdmin or g.current_user.isProgramManagerFor(programid)):
+    if not (g.current_user.isAdmin or g.current_user.isProgramManagerFor(programid) or g.current_user.isStudent):
```
Karina-Agliullova: As Besher mentioned above, the student user should not be able to create events

app/controllers/admin/routes.py
Jonimz: Thanks for pointing this out. We actually spoke with Dr. Heggen and confirmed that it should remain this way. The reasoning is that if it's the same session occurring at the same time and on the same date, but in different locations, while possible, it defeats the purpose of having a single, coherent event. At that point, users likely wouldn’t be going to the website to look up which location to attend, especially in a worst-case scenario.

Karina-Agliullova:
The same logic you implemented for multiple offerings should be applied for weekly events too. That means event name and location should be prefilled from what I input outside the modal
Screenshot (111)

app/static/js/createEvents.js
```diff
@@ -463,34 +490,47 @@ $(document).ready(function() {
...
 
-    if(!($('#inputEventName').val().trim() == '')){
-      //keeps main page event name for multiple event modal
-      $('#eventName').val($('#inputEventName').val());
+  
+let modalOpenedByEditButton = false;
+// #checkIsRepeating, #checkIsSeries are attributes for the toggle buttons on create event page
+
+$("#checkIsSeries, #edit_modal").click(function(event) {
+  eventSessionNum = 0;
```
Karina-Agliullova: Since you are assigning a variable in javascript, you should use "let" in front of eventSessionNum

app/static/js/createEvents.js
```diff
@@ -463,34 +490,47 @@ $(document).ready(function() {
...
-      $('#inputEventName').prop('readonly', true)
-      $('#inputEventName').val('')
-    } else {
-      setViewForSingleOffering()
-      $('#multipleOfferingTableDiv').addClass('d-none');
-      // Enable single event name field
-      $('#inputEventName').prop('readonly', false)
-      $('#inputEventName').prop('placeholder', 'Enter event name')
+    if ($('#inputEventLocation-main').val().trim() == '') {
+      $('#checkIsSeries').prop('checked', false)
```
Karina-Agliullova: In a lot of lines like this one, you are missing a semicolon. Go over all added lines in js and do some cleanup by adding semicolons and checking if there are correct indentations and with all closing curly braces and brackets

adjagbodjouf:
When an event is the last in its series, it would be nice to have a link from that instance to the nearest occurrence, the same way earlier instances have a link to later occurrences. Makes it easier to identify which events are related to one another. Also, having the session name of the other occurrence alongside the date would make the description much more helpful in knowing when and what the next event or previous occurrence is/was.

image
In this image, adding the session name of the next occurrence would be helpful. Different sessions could have different names, completely unrelated to the previous or future ones, so it would be helpful to have a name that users can identify when reading the description.

image
A link to the other occurrence, like the previous image, would be helpful. Like I said, as of right now, only earlier occurrences reference later occurrences, but later occurrences don't link back to their previous occurrence.

app/static/js/createEvents.js
PuskarC: Thanks for the suggestion! We updated the code to use .show() and .hide() instead of .addClass('d-none') / .removeClass('d-none') as recommended. However, we noticed that in some cases, especially with elements like #multipleOfferingTableDiv, replacing class toggling with .show()/.hide() caused the table to not appear and broke the expected functionality.

To resolve this, we’re now using a combination of .hide()/.show() and .addClass('d-none')/.removeClass('d-none') where necessary to ensure everything works as intended.

app/static/js/createEvents.js
PuskarC: Fixed!!

app/static/js/createEvents.js
PuskarC: Fixed the issue!!!

app/controllers/admin/routes.py
PuskarC: Fixed the Issue!!!

app/static/js/createEvents.js
PuskarC: Thanks for pointing that out! We've gone through all the recently added lines and cleaned up the code. Semicolons have been added where needed, and we double-checked the indentation, closing braces, and brackets for consistency. Let us know if anything else looks off!

app/static/js/createEvents.js
PuskarC: Also, applied the same prefill logic to weekly events

Karina-Agliullova:
I have created a name and location and then went to create a weekly event that would last 10 weeks. The event name and location were already prefilled, which is great. However, when I try to edit the modal, I go to change a name, click save button, and it doesn't show any difference. The name is still the one that was before, not the newly saved one. When I click edit again, and change the location along with the name, that is when it actually implements the change. Fix that behavior because we should be able to edit just the event name or just the location, without the need to change both. Such a behavior works well on the multi offerings without the weekly events

app/static/js/createEvents.js
```diff
@@ -33,43 +33,52 @@ function format12to24HourTime(timeStr) {
...
-                            endDate:$("#repeatingEventsEndDate").val()}
+function calculateRepeatingEventFrequency() {
+  var eventDatesAndName = {
+    name: $("#repeatingEventsNamePicker").val(),
+    isRepeating: true,
+    startDate: $("#repeatingEventsStartDate").val(),
+    endDate: $("#repeatingEventsEndDate").val(),
+    location: $("#repeatingEventsLocationPicker").val() || ''
+  }
+  console.log(eventDatesAndName);
```
BrianRamsay: No debug output please

app/static/js/createEvents.js
```diff
@@ -327,28 +349,36 @@ function verifyRepeatingFields(){
...
+      newOfferingModalRow.css('background-color', i % 2 ? '#f2f2f2' : '#fff');
+    }
+  });
 }
 
-function loadRepeatingOfferingToModal(offering){
+function loadRepeatingOfferingToModal(offering) {
   var seriesTable = $("#generatedEventsTable");
   var eventDate = new Date(offering.date || offering.eventDate).toLocaleDateString();
+  console.log(offering);
```
BrianRamsay: remove debug output

tests/code/test_events.py
```diff
@@ -299,13 +299,15 @@ def test_calculateRecurringEventFrequency():
 
     eventInfo = {'name': "testEvent",
                  'startDate': parser.parse("02/22/2023"),
-                 'endDate': parser.parse("03/11/2023")}
+                 'endDate': parser.parse("03/11/2023"),
+                 'location':("a big room")}
```
BrianRamsay: Is there a reason we are passing location as a list?

BrianRamsay (dismissed):

app/static/js/createEvents.js
Jonimz: fixed

app/static/js/createEvents.js
Jonimz: fixed

tests/code/test_events.py
Jonimz: I didn’t write that code, but I think it was meant to be a string. If we only need one location, it should just be "a big room." I just changed it.

app/static/js/createEvents.js
```diff
@@ -668,70 +719,74 @@ $("#cancelEvent").on('click', function (event) {
...
+
 
     if (startDateSelected < now && endDateSelected > now) {
       $("#pastDateWarningText").text("This event is currently in progress!")
     }
     else if (startDateSelected < now && endDateSelected < now) {
       $("#pastDateWarningText").text("This event is in the past!")
     }
-    else 
+    else
```
Karina-Agliullova: Open the curly bracket for else statement. Use right formatting

app/static/js/createEvents.js
```diff
@@ -594,38 +635,48 @@ $("#cancelEvent").on('click', function (event) {
...
       $("#repeatingEventsDiv").addClass('d-none');
       $("#multipleOfferingSlots").removeClass('d-none');
     }
   });
   
-  $("#repeatingEventsDiv").change(handleRepeatingEventsChange)
+  $("#repeatingEventsNamePicker, " + "#repeatingEventsLocationPicker").on("input", handleRepeatingEventsChange); 
+
+  $("#repeatingEventsStartDate, " +
+ "#repeatingEventsEndDate, " +
```
Karina-Agliullova: Fix indentation in this block

Jonimz: Done.

app/static/js/createEvents.js
Jonimz: done

Karina-Agliullova (dismissed):
Looks good

Karina-Agliullova (dismissed):
Reviewed this PR once more. Merged development into this pr. Everything seems good, tests passing, functionality working. It is ready for merge

BesherKitaz:
The Edit button is not fully working after we add multiple event listings. Let's avoid merging until this is fixed.

BesherKitaz:
All issues are fixed and everything seems to be good. From my side, this looks good to merge.

MImran2002:
I don't think this is a bug or should be a separate issue but because you are working on this, this can be worked on. First of all, it works the events are created successfully. 
But in an instance where the user create events in weird time sequence like Dec 27>Jan10>Jan6. The display after the recurring event is create should be in chronological order but the display is showing this way:
image
So potentially we could make it in order it will be great.

BesherKitaz:
I added the changes Imran suggested. The offerings will now show on the table in a sorted manner based on date and time, even if the user enters them otherwise.

MImran2002:
Sadly I think it is still broken: 
image

MImran2002 (changes requested):
Have left comment

PR #1547, opened 2025-06-30
Program manager 1545

### LINKED ISSUE #1545: Program Managers should be able to change program settings
Right now the Admin->Settings page is not available to program managers. Students who manage a program should be able to access the program settings for the programs they manage, but should not have access to other admin settings.

jjijon7000:
fixes #1545 
 
Issue Description:  the Admin->Settings page is not available to program managers. Students who manage a program should be able to access the program settings for the programs they manage, but should not have access to other admin settings.

Changes

- made a new method that determines if the user is a program manager or not
- added jinja conditions so that the html is dynamically changed depending if the user is program manager
- changed the permissions on the program manager page so that students that are program managers can edit the event / create

## Testing

*Before*
image

*After*
image

- run flask 
- sign in into user neillz and see the settings tag on the sidebar
- click the settings tab and see the admin page that only shows the programs that that user can manage
- ensure that the edit details buttons on the programs work


Karina-Agliullova:
On the sidebar, when I, as a student program manager, select the tab Settings, it should be highlighted just like any other tab on the sidebar that is selected. Keep it consistent

app/models/user.py
```diff
@@ -78,4 +78,9 @@ def isProgramManagerFor(self, program):
     def isProgramManagerForEvent(self, event):
         # Looks to see who the Program Manager for a specific event is
         return self.isProgramManagerFor(event.program)
+    
+    def isProgramManager(self):
+        from app.models.programManager import ProgramManager
```
Karina-Agliullova, line 85: all the imports should be on top of the files

app/templates/admin/userManagement.html
```diff
@@ -166,7 +167,40 @@ <h5>Programs</h5>
...
+          </div>
+          <div class="d-flex gap-2">
+            <button class="btn btn-success btn-sm editDetails" 
+                    role="button" 
+                    id="programManagementButton{{ loop.index }}"  
+                    loop-index="{{ loop.index }}"
+                    data-programid="{{program.id}}">
+                <span style="display: none;" id="editDetailsButtonSpinner{{ loop.index }}">
+                  <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>
+                  <span class="sr-only">Loading...</span>
```
Karina-Agliullova: No need for Loading and spinning. It seems inconsistent with the functionality in Program Management for the admin

app/models/user.py
jjijon7000: program manager seems to be imported always in the function that its being used in (ie. addProgramManager(), removeProgramManager(), isProgramManagerFor() fuctions)

Karina-Agliullova:
When I, as an admin, add a student, Ebenezer Ayisi, as a student program manager to the program People Who Care, and then I switch to be logged in as a student Ebenezer Ayisi, it keeps me in settings for program management; however, in the sidebar, the Settings is absent, so I, as a student program manager, Ebenezer Ayisi, cannot go back to view the program that I am managing. So ithe Settings display works for student staff but not for student, even though both a program managers

app/templates/admin/userManagement.html
jjijon7000: fixed

app/controllers/admin/userManagement.py
```diff
@@ -65,7 +65,7 @@ def deleteProgramFile():
 
 @admin_bp.route('/admin/updateProgramInfo/<programID>', methods=['POST'])
 def updateProgramInfo(programID):
-    if g.current_user.isCeltsAdmin:
+    if g.current_user.isCeltsAdmin or g.current_user.isProgramManager():
```
Karina-Agliullova: This logic would allow any program manager update any program, even if they are not program managers for other programs. That is mostly applicable for student program managers. Fix it with program id

app/controllers/admin/volunteers.py
```diff
@@ -217,7 +217,7 @@ def deleteBackgroundCheck():
 
 @admin_bp.route('/updateProgramManager', methods=["POST"])
 def updateProgramManager():
-    if g.current_user.isCeltsAdmin:
+    if g.current_user.isCeltsAdmin or g.current_user.isProgramManager():
```
Karina-Agliullova: This logic would let student program managers add or remove other program managers to programs. This should not be allowed for student program managers. This should be allowed only for admins

app/models/user.py
Karina-Agliullova: You will still be able to use it. Just move it to the top of the files. All the imports are stored there

app/templates/admin/userManagement.html
Karina-Agliullova: I don't see any changes. Did you commit and push?

app/models/user.py
jjijon7000: we tried to put it on top of the file, but it runs into an error. So we kept it there like the others.

app/templates/sidebar.html
```diff
@@ -89,6 +89,11 @@
       </li>
 
       {% endif %}
+      {% if g.current_user.isProgramManager() %}
+          <a href="/admin" class="nav-link text-white {{'active' if request.path =='/admin'}}" {{"aria-current='page'" if request.path =='/admin'}}>
```
Karina-Agliullova: Nesting html "aria-current='page'" inside the jinja tags {{}} seems weird. Try to put them inside the jinja conditionals but outside the jinja tags

petrosyantsa: Fixed

app/models/user.py
BrianRamsay: Because this file is imported directly by `app/__init__.py` you run into circular imports if you put this and other imports at the top of the file. They have to go in the functions so that they aren't executed when the file is loaded by app

app/controllers/admin/userManagement.py
```diff
@@ -65,7 +65,7 @@ def deleteProgramFile():
 
 @admin_bp.route('/admin/updateProgramInfo/<programID>', methods=['POST'])
 def updateProgramInfo(programID):
-    if g.current_user.isCeltsAdmin:
+    if g.current_user.isCeltsAdmin or g.current_user.isProgramManagerFor(Program.get_by_id(programID)):
```
BrianRamsay: You can just pass `programID` to the method, rather than doing a database query to get a `Program` object and passing that.

app/models/user.py
```diff
@@ -78,4 +78,9 @@ def isProgramManagerFor(self, program):
     def isProgramManagerForEvent(self, event):
         # Looks to see who the Program Manager for a specific event is
         return self.isProgramManagerFor(event.program)
+    
+    def isProgramManager(self):
```
BrianRamsay, line 84: Let's make this a property rather than a method, because in usage it is just like `isCeltsAdmin` or `isAdmin`

app/models/user.py
```diff
@@ -78,4 +78,9 @@ def isProgramManagerFor(self, program):
...
+        from app.models.programManager import ProgramManager
+        return ProgramManager.select().where(ProgramManager.user == self).exists()
```
BrianRamsay: We should cache this result like we do in `isProgramManagerFor`, because otherwise we run a query every time we call the method

app/templates/admin/userManagement.html
```diff
@@ -166,7 +167,39 @@ <h5>Programs</h5>
       </div>
     </div>
   </div>
-  
+{% endif %}
+{% if g.current_user.isProgramManager() and not g.current_user.isCeltsAdmin %}
```
BrianRamsay: Two things here. 

1. You shouldn't be able to edit all programs just because you're a program manager for one program.
2. We shouldn't duplicate all of this html between the admin vs the program manager cases. Look at using a macro to emit the markup that is shared.

app/templates/sidebar.html
```diff
@@ -89,6 +89,13 @@
       </li>
 
       {% endif %}
+      {% if g.current_user.isProgramManager() %}
+        <a href="/admin"
+           class="nav-link text-white {{'active' if request.path == '/admin'}}"
+           {% if request.path == '/admin' %}aria-current="page"{% endif %}>
+          Settings
```
BrianRamsay: Let's call this "Program Details", since that's what we call it in the page

app/controllers/admin/userManagement.py
```diff
@@ -100,20 +100,34 @@ def getProgramInfo(programID):
...
             .select(
                 Program,
                 fn.GROUP_CONCAT(fn.COALESCE(fn.CONCAT(User.firstName, ' ', User.lastName, '#', User.username), '')).alias('managers')
             )
-            .join(ProgramManager, JOIN.LEFT_OUTER, on=(Program.id == ProgramManager.program))
+            .join(ProgramManager, JOIN.INNER, on=(Program.id == ProgramManager.program))
             .join(User, JOIN.LEFT_OUTER, on=(ProgramManager.user == User.username))
+            .where(ProgramManager.user == g.current_user.username)
             .group_by(Program.id)
-    )
```
BrianRamsay, line 112: You can re-use this query instead of duplicating it in the different cases. The only real difference is the WHERE clause added to the second case (since the WHERE clause effectively turns the left join into an inner join automatically). So you can create the query, and if they are a program manager add the where clause to it.

BrianRamsay (changes requested):

app/templates/admin/userManagement.html
```diff
@@ -166,7 +167,39 @@ <h5>Programs</h5>
...
+                <span id="editDetailsButtonText{{ loop.index }}">Edit Details</span>
+            </button>
+          </div>
+        </div>
+        {% endfor %}
+      </div>
+    </div>
+  </div>
+</div>
+{% endif %}
```
BrianRamsay, line 190: We should create a jira macro that can output this chunk of html, since it's almost the same as the output for admins. a parameter can control whether it includes the edit managers button or not (and related styling).

BrianRamsay (changes requested):
Some nice changes! Just one more fix to go.

app/templates/admin/userManagement.html
jjijon7000: fixed based on the comment

BrianRamsay (approved):
Nice work!

PR #1551, opened 2025-07-01
New req

### LINKED ISSUE #1468: Enforce Required Frequency on Bonner Requirements
We should not be able to save a Bonner Requirement with an empty frequency

bakobagassas:
We should not be able to save a Bonner Requirement with an empty frequency. If we want it to be optional, lets make that clear. Issue with trying to modify the requirement section, along with Dr. Heggen, not liking the general functionality of how we save the rows and add them.
Fixes #1468 

Issue with selecting the requirement section in Bonner management, clearing up misunderstanding of allowing frequency to be empty, issues with the process of saving and creating rows
Changes

- Removed save changes button from bonnerManagement.HTML
- Added new save button to HTML file
- Created a saveRequirement function in bonnerManagement.js to handle both creating and updating rows
- Created an addRequirement function to handle the creation of a new row by setting placeholder IDs to new rows
- Created a remove button event handler to delete a specific row in the JavaScript file
- #reqAdd is now disabled once a new row is created in the JavaScript file
- created /deleteReq route to handle deleting rows, and modified /saveRequirements/ route to handle both new and updating rows
- Created a deleteRequirement function in certification.py to delete the row and return if successful or not
- Modified updatecertRequirements to handle creating new rows with placeholder IDs or updating previous rows and replacing old information
- Modified the testing file for certificate and removed tests we were not using.

Testing

(If while trying to see our changes, you notice that the buttons on the bonner management page are not working then it is probably an issue with the symlink. You can solve that on your side by removing sortable.min.js and then doing the command ln -s sortable-1.15.0.min.js sortable.min.js in the VSCode terminal. 
- Go into admin > Bonner Management > Requirements
- Click on add requirements
- Enter the information required
- Click the save button on that specific new row
- Toast message will appear on the top right if successful, along with the page will reload with the new row already there
- To test updating a row, change any information in a previous row and click save to update that row
- To test the removal of a row, click the remove button on any row, and the page will refresh with that row being gone.


app/static/js/bonnerManagement.js
```diff
@@ -87,46 +130,74 @@ $(document).ready(function(){
...
     var table = $("#requirements");
-    var newRow = table.find("tbody tr:last-child").clone()
-    newRow.data("id", "X");
+    var newRow = table.find("tbody tr:last-child").clone();
+    // newRow.data("id", "X");
+    newRow.attr('id','requirement_save-new');
+    $(newRow).attr("data-id", "save-new");
     newRow.find("input").val("");
-
     newRow.find("select.frequency-select option:first-child").attr('selected', true);
```
RueHaile, line 158: 1. The naming of the variable is confusing, as the parameters for the table are Necessity and Frequency, while in the code, Frequency is referred to as the first child in the drop-down of Necessity, which is "required". 
2. Line "  newRow.find("select. required-select option:last-child").attr('selected', true);" sets the Necessity to the last child by default, and the following line sets a default for the Frequency parameter. Which in return always enables the user to save without putting in the "Requirement Name".
3. Even if "We should not be able to save a Bonner Requirement with an empty frequency" has been solved, it has been so because of the default, which makes it way easier to make mistakes as the user and just save requirements without thinking about it. 
4. Remove the default and make it so that the user has to intentionally put in a value.

app/controllers/admin/routes.py
```diff
@@ -675,9 +675,28 @@ def saveRequirements(certid):
...
+        return jsonify({"success": True})
+    else:
+        return jsonify({"success": False, "message": "Failed to delete requirement."})
+    
+
+    
+
+
+
+
```
Karina-Agliullova: Too much space between 692 and 699. Remove it

app/logic/certification.py
```diff
@@ -79,28 +88,49 @@ def updateCertRequirements(certId, newRequirements):
         A list of CertificationRequirement objects corresponding to the given `newRequirements` list.
     """
     # check for missing ids to remove
-    saveIds = [requirementData['id'] for requirementData in newRequirements]
-    CertificationRequirement.delete().where(CertificationRequirement.certification_id == certId, CertificationRequirement.id.not_in(saveIds)).execute()
+    # saveIds = [int(requirementData['id']) for requirementData in newRequirements]
```
Karina-Agliullova: If you are not using this comment, remove it

app/logic/certification.py
```diff
@@ -79,28 +88,49 @@ def updateCertRequirements(certId, newRequirements):
...
+        actualRequirements = newRequirements['save-new']
         newRequirement.certification = certId
-        newRequirement.isRequired = bool(requirementData['required'])
-        newRequirement.frequency = requirementData['frequency']
-        newRequirement.name = requirementData['name']
-        newRequirement.order = order
+        newRequirement.isRequired = bool(actualRequirements.get('required', False))
+        newRequirement.frequency = actualRequirements['frequency']
+        newRequirement.name =actualRequirements['name']
+        # newRequirement.order = order
```
Karina-Agliullova: same here. If you don't need to use it, remove it

app/logic/certification.py
```diff
@@ -79,28 +88,49 @@ def updateCertRequirements(certId, newRequirements):
...
+    newRequirement = CertificationRequirement()
+    newRequirement.isRequired = bool(requirementData('required'))
+    newRequirement.frequency = requirementData['frequency']
+    newRequirement.name = requirementData['name']
+    # newRequirement.order = 0  #???
+    newRequirement.save()
+    return newRequirement.get_id()
+
+
+
```
Karina-Agliullova: Having just one line in between functions would be fine. Remove the rest. Do it all over the code that you changed for code cleanup

app/logic/certification.py
```diff
@@ -79,28 +88,49 @@ def updateCertRequirements(certId, newRequirements):
...
 
     # update existing and add new requirements
-    requirements = []
-    for order, requirementData in enumerate(newRequirements):
-        try:
-            newRequirement = CertificationRequirement.get_by_id(requirementData['id'])
-        except DoesNotExist:
-            newRequirement = CertificationRequirement()
 
+    if newRequirements.get("save-new",None) is not None:
```
Karina-Agliullova, line 89: From what I understand in the docstring, newRequirements should be a list of dictionaries, which you do not reflect on this line. It seems to me that you are trying to use it as a list only, not a list of dictionaries.

app/logic/certification.py
```diff
@@ -79,28 +88,49 @@ def updateCertRequirements(certId, newRequirements):
...
+        # newRequirement.order = order
+        certRequirement.save()
+        return True
+
+def saveRequirement(requirementData):
+    newRequirement = CertificationRequirement()
+    newRequirement.isRequired = bool(requirementData('required'))
+    newRequirement.frequency = requirementData['frequency']
+    newRequirement.name = requirementData['name']
+    # newRequirement.order = 0  #???
```
Karina-Agliullova: Remove unnecessary comments everywhere in your added code

app/logic/certification.py
```diff
@@ -79,28 +88,49 @@ def updateCertRequirements(certId, newRequirements):
...
-
-    return requirements 
-
+        return True
+    
+    else:
+        certKey = list(newRequirements.keys())[0]
+        try:
+            certRequirement = CertificationRequirement.get(CertificationRequirement.id == certKey)
+        except DoesNotExist:
```
Karina-Agliullova, line 97: In this case, maybe CertificationRequirement.DoesNotExist would be of more use?

app/static/js/bonnerManagement.js
```diff
@@ -41,16 +41,62 @@ function updateExportText(){
...
+        var row_data = {[el.data("id")]:
+                            {
+                                'id': row_el.data("id"),
+                                'name': row_el.find("input").val(),
+                                'required': row_el.find("select.required-select").val() == 'Required' ? true : false,
+                                'frequency': row_el.find("select.frequency-select").val()
+                            }}
+    }
+    else {
+        console.log("saveRequirement called");
```
Karina-Agliullova: Remove print statements from all over your code

app/logic/certification.py
```diff
@@ -79,28 +88,49 @@ def updateCertRequirements(certId, newRequirements):
...
+        certRequirement.isRequired = bool(requirement_info.get('required', False))
+        certRequirement.frequency = requirement_info['frequency']
+        certRequirement.name =requirement_info['name']
+        # newRequirement.order = order
+        certRequirement.save()
+        return True
+
+def saveRequirement(requirementData):
+    newRequirement = CertificationRequirement()
+    newRequirement.isRequired = bool(requirementData('required'))
```
Karina-Agliullova: Wrong use of bool function. Modify it so that it works for your "requirementData". Since it is not a function, you cannot use it like that. Check similar implementation in other lines of code you wrote, it is correct in other places

app/static/js/bonnerManagement.js
```diff
@@ -41,16 +41,62 @@ function updateExportText(){
...
+
     $("input[type=search]").each((i, inputElement) => addSearchCapabilities(inputElement));
     $(".removeBonner").on("click", function(){
         let year = $(this).data('year');
         cohortRequest(year, "remove", $(this).data("username"));
     });
 
+
     // Add requirements sorting
     // https://github.com/SortableJS/Sortable
```
Karina-Agliullova, line 98: unnecessary comments?

app/static/js/bonnerManagement.js
```diff
@@ -41,16 +41,62 @@ function updateExportText(){
...
+                        {
+                            'id': row_el.data("id"),
+                            'name': row_el.find("input").val(),
+                            'required': row_el.find("select.required-select").val() == 'Required' ? true : false,
+                            'frequency': row_el.find("select.frequency-select").val()
+                        }}
+}
+    $.ajax({
+        method: 'POST',
+        url: '/saveRequirements/1', // Bonner certification id hard-coded here
```
Karina-Agliullova: Why is this hardcoded? It should not be hard coded, so fix it

app/static/js/bonnerManagement.js
```diff
@@ -87,49 +133,50 @@ $(document).ready(function(){
...
 
 /* Add a new requirements row and focus it */
-function addRequirement() {
+function addRequirement() {                            
     var table = $("#requirements");
-    var newRow = table.find("tbody tr:last-child").clone()
-    newRow.data("id", "X");
+    var newRow = table.find("tbody tr:last-child").clone();
+    newRow.attr('id','requirement_save-new');
+    $(newRow).attr("data-id", "save-new");
```
Karina-Agliullova: Not sure why you are using $(newRow) as jQuery here. Seems like you can use just newRow.attr

tests/code/test_certification.py
```diff
@@ -44,66 +48,6 @@ def test_getCertRequirementsWithCompletion():
 
         transaction.rollback()
 
-@pytest.mark.integration
-def test_updateCertRequirements():
```
Karina-Agliullova, line 48: Why was this entire test removed?

bakobagassas: Because the function that was being tested has been changed and is not fruitful anymore

BrianRamsay: But the function still exists. It still needs a test. Also, there is no test of your new delete function

app/controllers/admin/routes.py
```diff
@@ -675,10 +675,20 @@ def saveRequirements(certid):
...
+@admin_bp.route("/deleteReq", methods =["POST"] )
+def deleteReq():
+    if not g.current_user.isCeltsAdmin:
+        abort(403)
+
+    deleteReq = deleteRequirement(request.get_json())
+    if deleteReq == True:
+        return jsonify({"success": True})
+    else:
+        return jsonify({"success": False, "message": "Failed to delete requirement."})
```
BrianRamsay, line 688: No `success: True` or `success: False` please. Use statuses for failure where possible

In addition, if they ask to delete something and it doesn't exist, who cares? They didn't want it in the first place. That should still count as succeeding.

app/logic/certification.py
```diff
@@ -73,33 +77,37 @@ def updateCertRequirements(certId, newRequirements):
...
+        except DoesNotExist:
+            abort(403)
+    
+        requirement_info = newRequirements[certKey]
+        certRequirement.certification = certId
+        certRequirement.isRequired = bool(requirement_info.get('required', False))
+        certRequirement.frequency = requirement_info['frequency']
+        certRequirement.name =requirement_info['name']
+        certRequirement.save()
+        return True
```
BrianRamsay: You have removed ordering, but we need to have the concept of order, so we know the order the requirements should be listed.

app/logic/certification.py
```diff
@@ -73,33 +77,37 @@ def updateCertRequirements(certId, newRequirements):
 
     Arguments:
         certId - The id of the certification whose requirements we are updating
-        newRequirements - a list of dictionaries. Each dictionary needs 'id', 'required', 'frequency', and 'name'.
+        newRequirements - a dictionary of dictionaries. Each dictionary needs 'id', 'required', 'frequency', and 'name'.
```
BrianRamsay: What are the keys for the new dictionary? You don't mention it. Do we really need a dictionary of dictionaries? Aren't we only updating one thing at a time now?

app/logic/certification.py
```diff
@@ -73,33 +77,37 @@ def updateCertRequirements(certId, newRequirements):
...
+            certRequirement = CertificationRequirement.get(CertificationRequirement.id == certKey)
+        except DoesNotExist:
+            abort(403)
+    
+        requirement_info = newRequirements[certKey]
+        certRequirement.certification = certId
+        certRequirement.isRequired = bool(requirement_info.get('required', False))
+        certRequirement.frequency = requirement_info['frequency']
+        certRequirement.name =requirement_info['name']
+        certRequirement.save()
```
BrianRamsay: We are duplicating the process of putting the information in an object and saving. We do mostly the same thing whether we are making a new one or updating an existing one, it should be the same code.

app/logic/certification.py
BrianRamsay: It looks like they are assuming it is a single dictionary, rather than a list of dictionaries. Which also doesn't match their comments.

app/logic/certification.py
```diff
@@ -64,7 +60,15 @@ def getCertRequirements(certification=None, username=None):
...
-def updateCertRequirements(certId, newRequirements):
+def deleteRequirement(rowID):
+    try:
+        req_id = CertificationRequirement.get_by_id(rowID)
+        req_id.delete_instance()
+        return True
+    except DoesNotExist:
+        return False
+
+def updateCertRequirements(newRequirements, certId=Certification.BONNER):
```
BrianRamsay, line 71: We shouldn't call it `newRequirements`, because we don't know if it is new or not. Existing updates happen here, too.

app/static/js/bonnerManagement.js
```diff
@@ -41,19 +40,61 @@ function updateExportText(){
...
+        $(".saveBtn").removeAttr("disabled");
+        let row_el = $("#requirement_" + el.data("id"));
+        var row_data = {[el.data("id")]:
+                            {
+                                'id': row_el.data("id"),
+                                'name': row_el.find("input").val(),
+                                'required': row_el.find("select.required-select").val() == 'Required' ? true : false,
+                                'frequency': row_el.find("select.frequency-select").val()
+                            }}
+    console.log(row_data);
```
BrianRamsay: No debug output, please

app/static/js/bonnerManagement.js
```diff
@@ -41,19 +40,61 @@ function updateExportText(){
...
+    else {
+        let rowid = parseInt(el.data("id"));
+        let row_el = $("#requirement_" + rowid);
+        var row_data = {[rowid]:
+                        {
+                            'id': row_el.data("id"),
+                            'name': row_el.find("input").val(),
+                            'required': row_el.find("select.required-select").val() == 'Required' ? true : false,
+                            'frequency': row_el.find("select.frequency-select").val()
+                        }}
```
BrianRamsay, line 66: This code is exactly the same as the code for a new requirement (almost)

app/static/js/bonnerManagement.js
```diff
@@ -41,19 +40,61 @@ function updateExportText(){
...
+}
+    $.ajax({
+        method: 'POST',
+        url: url,
+        contentType: 'application/json',
+        dataType: 'json',
+        data: JSON.stringify(row_data), 
+
+        success: function(ids) {
+            msgToast("Bonner", "Updated Bonner Requirements");
```
BrianRamsay: The header should be "Success", not "Bonner"

app/static/js/bonnerManagement.js
```diff
@@ -41,19 +40,61 @@ function updateExportText(){
...
+            $("#reqAdd").removeAttr("disabled");
+            // location.reload();
```
BrianRamsay: Because you don't reload, you end up with a bunch of duplicate  "save-new" related ids if you create another requirement. You either need to reset those ids or reload the page so that jinja does it for you.

app/static/js/bonnerManagement.js
```diff
@@ -41,19 +40,61 @@ function updateExportText(){
...
     $("input[type=search]").each((i, inputElement) => addSearchCapabilities(inputElement));
     $(".removeBonner").on("click", function(){
         let year = $(this).data('year');
         cohortRequest(year, "remove", $(this).data("username"));
     });
-
-    // Add requirements sorting
-    // https://github.com/SortableJS/Sortable
-    // https://sortablejs.github.io/Sortable/
     var requirementsObj = new Sortable($('#requirements tbody')[0], {
```
BrianRamsay: Why did we remove the reference to the source of the Sortable object?

app/static/js/bonnerManagement.js
```diff
@@ -87,50 +128,45 @@ $(document).ready(function(){
...
+    var newRow = table.find("tbody tr:last-child").clone();
+    newRow.attr('id','requirement_save-new');
+    newRow.attr("data-id", "save-new");
     newRow.find("input").val("");
-
     newRow.find("select.frequency-select option:first-child").attr('selected', true);
     newRow.find("select.required-select option:last-child").attr('selected', true);
-
+    newRow.find(".saveBtn").attr('id',"save-new");
+    newRow.find(".saveBtn").data('id',"save-new");
```
BrianRamsay, line 160: Why do we save this twice?

app/static/js/bonnerManagement.js
```diff
@@ -87,50 +128,45 @@ $(document).ready(function(){
...
+    let newSaveBtn = newRow.find(".saveBtn")[0];
+    newRow.find("select.frequency-select").attr("name", "frequency-new");
     table.append(newRow)
+    // newSaveBtn.addEventListener("click", enableSave());
```
BrianRamsay: If this isn't needed, we should get rid of it.

app/static/js/bonnerManagement.js
```diff
@@ -208,48 +203,61 @@ function addRequirementsRowHandlers() {
...
+        let rowid=parseInt(el.data("id"));
+        $.ajax({
+            method : "POST",
+            url: "/deleteReq",
+            contentType: 'application/json',
+            dataType: 'json',
+            data: JSON.stringify(rowid), 
+            success: function(response) {
+                msgToast("Bonner", "Removing Bonner Requirement");
+                location.reload();
```
BrianRamsay, line 226: Reloading is fine, but it would probably be cleaner to just delete the row from the table.

app/static/js/bonnerManagement.js
```diff
@@ -208,48 +203,61 @@ function addRequirementsRowHandlers() {
...
+            },
+            error: function(e) {
+                msgToast("Error", "Error Removing Requirement");
+                location.reload();
```
BrianRamsay: If you reload right after the toast you won't see the toast at all.

BrianRamsay (changes requested):

app/static/js/bonnerManagement.js
bakobagassas: When we remove the reload, the row gets automatically deleted, but it doesn't display until we manually reload. Should we still delete the reload or keep it?

ojmakinde:
Implementation in this PR has grown too complex from the initially simple issue. To solve this, simply prevent saving with an empty frequency

PR #1556, opened 2025-07-02
Save button on Create Event page 

### LINKED ISSUE #1444: Save button on Create Event page
Save button on the create event page can be found only by scrolling. Make it fixed to the bottom

Shenderaz:
## Issue Description

Fixes #1444 
- Updates the save button to fixed position to work on different screen sizes, where it was found only through scrolling. 

## Changes

-  Added a media query to change save button's position to fixed depending on the screen.
- Added `updatePosition` function that changes save button to absolute position when the save button reaches the footer, preventing overlap.

image

## Testing

- As an admin, go to "Create event". 
- Pick one of the programs.
- Check the save button

For further testing:
- Go to DevTools (Ctrl+Shift+I)
- Go to Toggle Device ToolBar or Emulation (Ctrl+Shift+M), to test out in different screen sizes.
- Check the save button.


jjijon7000 (commented):
looks good on different devices

jjijon7000 (dismissed):

app/static/js/createEvents.js
```diff
@@ -697,6 +697,24 @@ function handleTimeFormatting(timeArray){
   setCharacterLimit($("#inputCharacters"), "#remainingCharacters"); 
 });
 
+$(function () {
```
BrianRamsay: Why is this code wrapped in a function?

BrianRamsay (changes requested):
I like the fixed behavior (mostly), but you're not solving the problem when the content is bigger than the screen. Go add a list of weekly events on any screen size and the Save button gets bumped down below again.

mathew2005:
> I like the fixed behavior (mostly), but you're not solving the problem when the content is bigger than the screen. Go add a list of weekly events on any screen size and the Save button gets bumped down below again.

We were able to resolve it by changing the save button's parent div.

Jonimz:
I noticed there are two "Save" buttons on the page. Not sure if this is intentional. 
Screenshot 2025-07-22 at 9 36 06 AM

app/templates/events/createEvent.html
```diff
@@ -300,6 +300,7 @@ <h1 id="pageTitle">{{page_title}}</h1>
         {{ displayFiles(filepaths, 'Event Attachments', '/deleteEventFile', eventData.id)}}
       </div>
       {% endif %}
+        <input type="submit" class="btn btn-primary saveBtn" value="Save" style="width: 5rem;" />
```
Jonimz: I tested it out by removing this extra submit line since we’re already using #saveButton with JS. Tested it, and everything still works as expected. 
Screenshot 2025-07-22 at 10 11 48 AM
Screenshot 2025-07-22 at 10 11 59 AM

Jonimz: Fixed it

BrianRamsay:
Was still broken when the page was just too long, like when there was a very large event series. Fixed in https://github.com/BCStudentSoftwareDevTeam/celts/pull/1605

Issue #1565, opened 2025-07-08
Admin Reports page


Karina-Agliullova:
When we get volunteers report list, there should not be Bonner Scholar information. In addition, add indentations in the volunteers report

PR #1568, opened 2025-07-10
Admin report page 

Arohasina:
## Issue Description

#1565

-When we get the student transcript, there should not be Bonner Scholar information. In addition, remove indentations in the volunteer's report. 

## Changes

- Removed the Bonner Scholar information from the transcript
- We have now condensed the CSS file, reducing custom classes. We have changed the styling of the student service Transcript. 
- Added onto the slcourse section code to filter out Bonner course classes that may contribute to the hours within the summary table
- Removed Berea Logo
- Changed the color of the student's name to black in both regular and print version and changed the service transcript subtitle to grey.

- before:
image

-after
image

## Testing
- Search student --> enter student name --> then under participation history --> View Service Transcript--> Modified spacing and alignment, and modified the Bonner Scholar element of the Transcript. When looking at printing view, make sure to turn on Background settings. To retrieve the result seen in the picture above.


Karina-Agliullova:
Remove the symbol that I highlighted in red
Screenshot (112)

Karina-Agliullova:
The information about hours should be below everything else, below volunteer details and below the courses

app/templates/main/serviceTranscript.html
```diff
@@ -1,85 +1,157 @@
...
 
-    <div class="ms-3">
-      <div class="float-front pt-1">
-      <p class="fw-bold pb-3 d-inline fs-6 ">Service Hours:</p>
-      <p class="d-inline bg-light ps-2 pe-5 border-bottom fs-6" id="hourTotal">{{ totalHours["totalCourseHours"]|round(2) }}</p>
+  <!-- Service Learning Courses - Adjusted table styling -->
+  {% if slCourses.exists() %}
+    <div class="card mb-4 avoid-break" id ="service-learning-courses-card">
+      <div class="card-header bg-light">
+        <h5 class="mb-0">Service Learning Courses</h5>
```
Karina-Agliullova: Service Learning should be Service-Learning

Karina-Agliullova:
The bright blue color on Volunteer Hours seems to attract all the attention only to the hours. That should not be the case.

Karina-Agliullova:
Not sure if we want to keep the name of the user and Volunteering since in blue color. Might need to be in black

Jsmi8334:
> The information about hours should be below everything else, below volunteer details and below the courses

We figured the main point behind a service transcript is to highlight the hours the person has worked. This is why we moved it to the top section of the transcript. making it stand out to the person receiving the transcript.

app/static/css/serviceTranscript.css
```diff
@@ -14,9 +14,9 @@
     margin: 0.5in 0.1in 0.5in 0.1in;
   }
 
-  .footer-logo {
+  /* .footer-logo {
     width: 180px;
     max-width: 100%;
     height: auto;
-  }
+  } */
```
Karina-Agliullova: If we don't need this code, remove it fully. Seems like a redundant comment

Karina-Agliullova:
While it is true that you removed Bonner Scholars info from the service transcript, the amount of hours still reflect as if the bonner scholars events were counted as event hours and accumulated with other events hours that the student had. You would need to remove the Bonner Scholars event hours from the summary of the hours

Meatchema:
> While it is true that you removed Bonner Scholars info from the service transcript, the amount of hours still reflect as if the bonner scholars events were counted as event hours and accumulated with other events hours that the student had. You would need to remove the Bonner Scholars event hours from the summary of the hours

The hours are now removed from the summary of the hours.

app/templates/main/serviceTranscript.html
```diff
@@ -1,85 +1,159 @@
...
+              {% for item in allEventTranscript[program] %}
+                {% if item[1] is not none %}
+                  <div class="mb-2">
+                    <small class="text-muted">{{ item[0] }} - {{ item[1] | round(2) }} hour(s)</small>
+                  </div>
+                {% endif %}
+              {% endfor %}
+              {% if programTitle != "CELTS Sponsored Events" %}
+                <p class="text-muted small mb-2">{{ program.description }}</p>
+                <a class="d-print-none small" href="{{ program.url }}" target="_blank">More Information</a>
```
Karina-Agliullova, line 107: The link is broken. it leads to the "Page Not Found"

app/models/program.py
```diff
@@ -18,9 +18,17 @@ class Program(baseModel):
...
     def url(self):
+        if self.bereaUrl:
+            return self.bereaUrl
+        if self.instagramUrl:
+            return self.instagramUrl
+        if self.facebookUrl:
+            return self.facebookUrl
+    
+        if self.programName == "Hunger Initiatives":
+            return "https://www.berea.edu/centers/center-for-excellence-in-learning-through-service/hunger-initiatives"
```
Karina-Agliullova: There should be another way to route to the url that you want. As of now, you are hardcoding the url only for one specific program, which is something we should not do. See how routing to the url that we want is done on the landing page with all programs displayed

Karina-Agliullova:
Also, Brian told me that we want to get rid of the word "Event" and anything it is referencing. It is misleading and we should use "Volunteer" and "Volunteering" instead. Please change the language

app/templates/main/serviceTranscript.html
Meatchema: We have now fixed this issue by assigning the link of the website into the sql database.

Karina-Agliullova (approved):
Good work!

PR #1578, opened 2025-07-14
changed the name from student-led services to Volunteer programs

### LINKED ISSUE #1570: Rework Event List tabs
We are changing the tabs available on the event list page.

- 'Other Events' should be removed.
- 'Student-Led Services' should be renamed to 'Volunteer Opportunities', and should only show events with the `isService` flag set to true.
- We should add a Labor tab, only visible to CELTS labor students, with the events that are marked Labor Only
- The Training and Education and Engagement tabs should exclude events that are marked Labor Only

bakobagassas:
## Issue Description

Fixes #1570 
- Rename student-led services to Programs or Volunteer Programs

## Changes

- We changed the name displayed in the js file from student-led services Volunteer Programs

## Testing

- Log in as an admin 
- go to the events list page 
- the changes should be reflected on the first tab on the page

### DIFF
--- a/app/static/js/eventList.js
+++ b/app/static/js/eventList.js
@@ -94,7 +94,7 @@ function updateIndicatorCounts(isChecked){
       $("#viewPastEventsToggle").prop(toggleStatus, true);
 
       // use ternary operators to populate the tab with a number if there are events, and clear the count if there are none
-      studentLedEventsCount > 0 ? $("#studentLedEvents").html(`Student-Led Services (${studentLedEventsCount})`) : $("#studentLedEvents").html(`Student-Led Services`)
+      studentLedEventsCount > 0 ? $("#studentLedEvents").html(`Volunteer Programs (${studentLedEventsCount})`) : $("#studentLedEvents").html(`Volunteer Programs`)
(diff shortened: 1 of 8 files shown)


app/static/js/eventList.js
```diff
@@ -94,7 +94,7 @@ function updateIndicatorCounts(isChecked){
       $("#viewPastEventsToggle").prop(toggleStatus, true);
 
       // use ternary operators to populate the tab with a number if there are events, and clear the count if there are none
-      studentLedEventsCount > 0 ? $("#studentLedEvents").html(`Student-Led Services (${studentLedEventsCount})`) : $("#studentLedEvents").html(`Student-Led Services`)
+      studentLedEventsCount > 0 ? $("#studentLedEvents").html(`Volunteer Programs (${studentLedEventsCount})`) : $("#studentLedEvents").html(`Volunteer Programs`)
```
Karina-Agliullova, line 97: This change is not fully reflected. In html, it still is called Student Led services. So when I reload page, it first shows me "Student-Led Services" and only atfer a moment, it would show a new name. Fix that. Also, double check with Brian if we want to rename all the variables in all the files associated with student-led services. That variable is in a lot of places

BrianRamsay:
I've changed the original ticket to encompass all of the changes that are needed.

PR #1579, opened 2025-07-14
Added a new filter for the labor history

### LINKED ISSUE #1576: New labor filter in participation history
Now that we can record participation as labor, we need to update the participation history accordion on the user profile.

We should add an indication of whether the participation was from labor, and add the ability to filter the list to show and hide labor participation.

PuskarC:
## Issue Description

Fixes #1576 
- Add a filter in the Participation history that shows labor history 

## Changes

- Added a HTML line that uses anchor tag and has the button name "View Labor History" in userProfile.html file

## Testing

- Go to CELTS Website and then go to 'My Profile'
- Click on the participation history accordion
- Now, you can see a button that shows "Labor history" next to the View Service Transcript

BEFORE

image

AFTER

Screenshot 2025-07-14 145610 (1)

### DIFF
--- a/app/templates/main/userProfile.html
+++ b/app/templates/main/userProfile.html
@@ -195,6 +195,7 @@
           </table>
         </div>
         <a class="btn btn-primary btn-sm" href="{{volunteer.username}}/serviceTranscript" > View Service Transcript</a>
+        <a class="btn btn-primary btn-sm" href="{{volunteer.username}}/laborHistory" > View Labor History</a>
       </div>
     </div>
   </div>


Karina-Agliullova:
Labor history leads nowhere. Page not found is what it displays. Double check what labor history actually is. If it is the same as service transcript, then we don't need it because we have the button for service transcipt already

BrianRamsay:
I don't really understand this one. Mainly, because no page exists with a `laborHistory` route (that I see), so this link doesn't go anywhere.

I've updated the original issue with more detail of what we need. If you don't have enough information in the ticket you should ask for clarification. In any case, this issue is on hold until we have our database field in `eventparticipation` which is still in progress in another issue.

PR #1588, opened 2025-07-22
Labor tab

### LINKED ISSUE #1523: Add a Manage Labor to the event page
- On Edit event, in manage volunteers, we are going to implement two more columns: "labor" and "volunteer service" checkboxes which are mutually exclusive. There should be a count of hours for either one of the checkboxes checked or for just participation with no labor or service hours. 
- "Add volunteers" button should be renamed to "add participants"
- Add an X for deleting participants in the modal when you added one and want to remove it in the modal.
- When we calculate RSVP and waitlist, they should not be added as labor
- Add to "eventparticipant" model new columns in database: isLabor, isService
- When we calculate RSVP and waitlist, they should not be added as labor

adjagbodjouf:
## Issue Description

Fixes #1523 

## Changes

- Added a new Labor tab
- Added new fields to the participants table in the database to work for labor as well.
- Created new HTML, JS files, routes, and more for the labor tab.

image

## Testing

- In the menu of the app, select Event list
- Select any term (eg: spring 2021) 
- go to the training tap 
- Click on any event there 
- then click on the Manage Labor tap 
- There you will be able to see the new page we have added


app/logic/sharedLogic.py
BrianRamsay, line 1: We already have a file `events.py` for shared event-related logic

BrianRamsay (changes requested):
I don't see any `manageLabor.html` template

PR #1589, opened 2025-07-22
Working Code Checkpoint: Added html file and JS file

Meatchema:
## Issue Description

Did not have a number 

Fixes #add issue number

Dietary Restrictions seem to be broken. When I use the toggle "No restrictions," and then I save it, it shows me that there are no dietary restrictions in the text window. Then, if I unclick it and save it, the page is reloaded and still keeps the "No restrictions" toggle clicked. Fix the behavior of the toggle when we save information or save no information

## Changes

- We added wording to the placeholder in the HTML 
- Added a string value for  when the person has dietary restrictions

## Testing

- Select "my profile.": 
image

- Select the "dietary restrictions": 
image

- Then turn the toggle on and off. 
- When the toggle is off, feel free to input data
-Then save the changes to ensure that the changes have been saved. 
-Then turn the toggle on  to see if the other direction was saved 
image

### DIFF
--- a/app/static/js/userProfile.js
+++ b/app/static/js/userProfile.js
@@ -14,7 +14,7 @@ $(document).ready(function(){
         success: function(response) {
           let accept = "You have indicated interest in CCE Minor.";
           let decline = "You have indicated you are not interested in the CCE Minor.";
-        
+       
           let msg = isAdding ? accept : decline;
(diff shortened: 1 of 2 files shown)


Karina-Agliullova (approved):
Looks good!

PR #1604, opened 2025-08-01
Add a list of training options to create event page

### LINKED ISSUE #1564: Required Training list on Create Event page
When we create an event, we should have a list of required training from which the user could choose one or more to complete the training(s). It should be under the start time and end time

Arohasina:
## Issue Description

Fixes #[1564](https://github.com/BCStudentSoftwareDevTeam/celts/issues/1564)
-When we create an event, we should have a list of required training from which the user could choose one or more to complete the training(s). It should be under the start time and end time

## Changes

- Updated the event creation template to display a user-friendly list of required training options.
- Ensured all duplicate training entries are removed from the list.
- Enabled users to choose none, one, or multiple required trainings.
- Updated the form logic to correctly handle multiple selections.
image

## TO DO NEXT:
-Create a dedicated RequiredTraining table: move training options out of the current hardcoded or ad hoc structure into their own persistent model.
-Refactor event-training relationship: establish a relationship between events and required trainings using a foreign key or many-to-many association.
-Link training selections in the event views: display the associated required trainings in event detail and summary views.
image


fritzj2:
Shouldn't `eventView.html` also be updated to include the new additions to the event?
This should be reflected in the Program training tab, which currently ignores the input of these additional trainings.
Screenshot 2026-09-21 104629
Screenshot 2026-09-21 104833

fritzj2:
Also, this doesn't work with the Production backup as it involves adding new fields to the database without a way to transfer them. There should be lines in the `prod-backup.sql` that alter the tables to add the columns to the existing rows.

Kafui123:
> Shouldn't `eventView.html` also be updated to include the new additions to the event? This should be reflected in the Program training tab, which currently ignores the input of these additional trainings. Screenshot 2026-09-21 104629 Screenshot 2026-09-21 104833

addressed this issue.

PR #1610, opened 2025-08-27
Create event hot fix

### LINKED ISSUE #1609: Fix for recurring all volunteer training
When saving all volunteer training event, and later editing it to make it recurring, there is an issue, as when saving the recurring one, the event name gets erased

ojmakinde:
## Issue Description

Fixes #1609
- If a series is created off an already existing event, the initial event loses its name and the new events are not created.

- Previously, there was no logic to account for creating a series off an existing event. It becomes more of a design decision as to whether we want users to have the ability to create a series off an existing event. Do we also want them to be able to create a series off an existing event in a series? I don't think so.

Ideally, it'd be easiest if we removed the "Series event" toggle on an already created event. That way, we don't have to merge this logic and we keep it simple. If we decide to merge this, then we need to add more logic to prevent series from being created off events that are already in series.

## Testing

- It works


ojmakinde:
Brian and I ultimately decided to prevent users from creating series off of already created events. It keeps the logic simple and stops us from opening a web of messy skeletons.

PR #1622, opened 2025-09-04
Fixed issue with syncing hours on service transcript and spreasheet

### LINKED ISSUE #1617: Missing hours investigation
Can't find all hours for collinsc2 when generating report for AY 2024-2025. Investigate what generates them more hours than their service hours show.
Findings:
- The spreadsheet generation was not considering whether the event is canceled, or is a service, or is training. 
- The check in both the spreadsheet and transcript should check for whether the event is canceled, or is a service, or is training.

Karina-Agliullova:
## Issue Description

Fixes #1617 
- There are hours in the reports spreadsheet that exceed for some students compared to their hours on their service transcript

## Changes

- Now we are checking if events are not cancelled, and they are not trainings, and not deleted for all events that earn service hours in the transcript and spreadsheet for volunteers

## Testing

- In production db, check the user collinsc2, view hours at their service transcript
- Go to reports and download reports for AY 2024-2025, and check volunteer hours for collinsc2
- The hours in the reports spreadsheet and on the user's transcript should reflect the same number

### DIFF
--- a/app/logic/transcript.py
+++ b/app/logic/transcript.py
@@ -18,12 +18,16 @@ def getProgramTranscript(username):
     """
     # Add up hours earned in a term for each program they've participated in
 
+    print("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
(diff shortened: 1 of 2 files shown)


BrianRamsay:
Changes incorporated into https://github.com/BCStudentSoftwareDevTeam/celts/pull/1624

PR #1633, opened 2025-10-11
flash message

### LINKED ISSUE #1629: Add flash message on the term management
Make the flash message show up when you create a new term. As of now, it is not showing.

bakobagassas:
## Issue Description

Fixes #1629 
Make the flash message show up when you create a new term. As of now, it is not showing.

## Changes

- added a flash message line in the Python file from the controller. 
image

## Testing

- go to admin 
- then to settings 
- then the term management accordion 
- then add a new term
- You should now be able to see a message saying that it has been successfully added.

### DIFF
--- a/app/controllers/admin/userManagement.py
+++ b/app/controllers/admin/userManagement.py
@@ -118,7 +118,9 @@ def userManagement():
     
     currentAdmins = list(User.select().where(User.isCeltsAdmin))
     currentStudentStaff = list(User.select().where(User.isCeltsStudentStaff))
+    print(terms, "Here")
     if g.current_user.isCeltsAdmin or g.current_user.isProgramManager:
+        print(terms, "Slim shady")
         return render_template('admin/userManagement.html',
                                 terms = terms,
                                 programs = list(currentPrograms),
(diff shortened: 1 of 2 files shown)


app/controllers/admin/userManagement.py
```diff
@@ -118,7 +118,9 @@ def userManagement():
     
     currentAdmins = list(User.select().where(User.isCeltsAdmin))
     currentStudentStaff = list(User.select().where(User.isCeltsStudentStaff))
+    print(terms, "Here")
     if g.current_user.isCeltsAdmin or g.current_user.isProgramManager:
+        print(terms, "Slim shady")
```
ojmakinde: please remove

ojmakinde:
There's a strange bug with the page. If there are more than 9 terms on a single page, any additionally created terms will not be visible, and you'd have to switch to a more recent term, and then create another term to see the additional terms. 

For example, in the image below, I clicked the button to create an additional term. Although it was created, I could not see it, as the page caps at Fall 2028.

image

If I switch my current term to something more recent, like Spring 2028 and click on create new terms, then I can see all the created terms that were invisible to me. This is a bug your PR likely did not introduce, but I would like you to fix it in this PR, since the PR is closely related to term creation.

ojmakinde (changes requested):

BrianRamsay (approved):

BrianRamsay (approved):

PR #1647, opened 2025-12-18
Events header

### LINKED ISSUE #1636: Inconsistent Headers in Events List Tabs
When on the Events List page, some tabs will show a header as such with no upcoming events:

Image

Some will show like this:

Image

And the Student-Led-Services shows like this:

Image

This seems to be because there are past events in these categories, but in the second case I showed, there are no past events. Find a way to make these consistent with one another.

bakobagassas:
## Issue Description

Fixes #1636 

- When on the Events List page, some tabs do not display the same message when there are no upcoming events

## Changes

- Standardized display message format when there are no events 
- Added a new function for past events 
- Added a new logic for past and upcoming events notifications on the accordions 
- Corrected the number of events displayed on top of the tab. When the past toggle is not checked, we only see the number of upcoming events. However, when it is checked, we should see the number of upcoming AND past events. 

## Testing

- Go to events list 
- You will be able to see the different tabs with no upcoming events and the display. 
- You can create events in the tabs and play around to see what appears where.

### DIFF
--- a/app/static/js/eventList.js
+++ b/app/static/js/eventList.js
@@ -96,7 +96,11 @@ function updateIndicatorCounts(isChecked){
       
       $("#viewPastEventsToggle").prop(toggleStatus, true);
       
-      // use ternary operators to populate the tab with a number if there are events, and clear the count if there are none
+      // Update tab labels with event counts:
+      // - When toggle is ON, show total volunteer opportunities (upcoming + past)
+      // - When toggle is OFF, show upcoming volunteer opportunities only
+      // - For all tabs, show counts only if greater than zero; otherwise show the label without a count
+
       if (toggleStatus === "checked") {
         // Toggle ON: show total (upcoming + past)
         if (volunteerOpportunitiesCount > 0) {
--- a/app/static/js/eventList.js
+++ b/app/static/js/eventList.js
@@ -94,7 +94,7 @@ function updateIndicatorCounts(isChecked){
       const celtsLaborCount = Number(eventsCount.celtsLaborCount);
       const toggleStatus = eventsCount.toggleStatus;
       
-      $("#viewPastEventsToggle").prop(toggleStatus, true);
+      $("#viewPastEventsToggle").prop("checked", toggleStatus === "checked");
       
       // Update tab labels with event counts:
       // - When toggle is ON, show total volunteer opportunities (upcoming + past)
--- a/app/static/js/eventList.js
+++ b/app/static/js/eventList.js
@@ -88,7 +88,6 @@ function updateIndicatorCounts(isChecked){
     success: function(eventsCount) {
       const volunteerOpportunitiesCount = Number(eventsCount.volunteerOpportunitiesCount);
       const upcomingVolunteerCount = Number(eventsCount.countUpcomingVolunteerOpportunitiesCount);
-      const pastVolunteerCount = Number(eventsCount.countPastVolunteerOpportunitiesCount);
       const trainingEventsCount = Number(eventsCount.trainingEventsCount);
       const engagementEventsCount = Number(eventsCount.engagementEventsCount);
       const bonnerEventsCount = Number(eventsCount.bonnerEventsCount);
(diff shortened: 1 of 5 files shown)


MImran2002:
Hello here are some of my reviews as I looked into the issue John mentioned how there are three different ways of showing events are not there. They want standardization so when it comes to Volunteer Opportunities there is one bug that you should look into and one more standardization. 

As volunteer opportunities when it should shows this:
image

Is showing this instead:
image

The bug I see also is when the Volunteer Opportunities is showing (1) when there is no event coming, I think it is counting one because of the drop down opportunities instead of event:
image

app/static/js/eventList.js
```diff
@@ -87,6 +87,8 @@ function updateIndicatorCounts(isChecked){
     },
     success: function(eventsCount) {
       const volunteerOpportunitiesCount = Number(eventsCount.volunteerOpportunitiesCount);
+      const upcomingVolunteerCount = Number(eventsCount.countUpcomingVolunteerOpportunitiesCount);
+      const pastVolunteerCount = Number(eventsCount.countPastVolunteerOpportunitiesCount);
```
MImran2002: this variable is not used in the script

app/templates/events/eventList.html
```diff
@@ -175,29 +175,29 @@ <h1 class="text-center">Events List for {{selectedTerm.description}}</h1>
...
+              <button class="accordion-button {{ 'show' if programID == program.id else 'collapsed' }}"
+                      type="button"
+                      data-bs-toggle="collapse"
+                      data-bs-target="#accordion__body_{{ program }}_num_{{ loop.index }}"
+                      aria-expanded="true"
+                      aria-controls="accordion__body_{{ program }}">
+                {{ program.programName }}
+                {% set upcoming = countUpcomingVolunteerOpportunities.get(program.id, 0) %}
+                {% set past = countPastVolunteerOpportunities.get(program.id, 0) %}
+                  <span class="ms-auto fw-light fst-italic"> {{ upcoming }} upcoming event{% if upcoming != 1 %}s{% endif %} and {{ past }} past event{% if past != 1 %}s{% endif %} </span>
```
MImran2002: this shows the count regardless of when it is past or present i think you should show upcoming when the past toggle is unchecked and only show past counts each when the toggle is in the past.

MImran2002:
1. this variable is not in used:  const pastVolunteerCount :  
    Number(eventsCount.countPastVolunteerOpportunitiesCount);
3. This display I think will be better if it only shows the past count when the past toggle is checked and when unchecked can show the upcoming:  
    <span class="ms-auto fw-light fst-italic"> {{ upcoming }} upcoming event{% if upcoming != 1 %}s{% endif %} and {{ past }} past event{% if past != 1 %}s{% endif %} </span>
5. This opening div tag needs correct indentation: 
image

MImran2002:
this variable in test_getPastVolunteerOpportunitiesCount() has not been used at all in the test but the peewee is being inserted so I think you should remove the variables.:
        pastSameDayEvent = Event.create(
            name="Test same-day past AGP event",
            term=currentTestTerm,
            description="Same day but earlier time.",
            timeStart="04:00:00",
            timeEnd="04:30:00",
            location="Venus",
            isTraining=False,
            isService=True,
            startDate="2021-08-01",
            program=3
        )
        futureAgpEvent = Event.create(
            name="Test future AGP event",
            term=currentTestTerm,
            description="Future volunteer opportunity.",
            timeStart="06:00:00",
            timeEnd="07:00:00",
            location="Moon",
            isTraining=False,
            isService=True,
            startDate="2021-08-02",
            program=3
        )

MImran2002:
1. So in getPastVolunteerOpportunitiesCount( ) function, when you are extracting pastCount and you are comparing Event.startDate with currentTime there can be a potential type mismatch because Event.startDate's type is 2025-12-30 15:14:56.873933 <class 'datetime.datetime'> whereas currentTime is  2025-09-11 <class 'datetime.date'> I just want you to make sure to take care of the differences. So with other filtering as well ((Event.startDate == currentTime). Moreover, 16:00:00 <class 'datetime.time'> for Event.timeStart is also inconsistent with currentime type too. 

2. "// use ternary operators to populate the tab with a number if there are events, and clear the count if there are none"
this comment in line 99 of eventList.js is for the condition you have changed below so this comment should be updated

MImran2002 (changes requested):

app/static/js/eventList.js
bakobagassas: It is. Line 111 if (upcomingVolunteerCount > 0)

bakobagassas:
Solved all comments, ready for new review

MImran2002 (dismissed):
Looks good

Karina-Agliullova (approved):
Looked thoroughly through it, and it works well and looks good. Fixed minor issues, and now approving!

MImran2002 (approved):
After careful review, with check different scenario with the UI, I think this PR is good for merging.

PR #1664, opened 2026-01-20
1660 CCE minor students not showing

### LINKED ISSUE #1660: CCE Minor Declare students list not shown
Issue description: When the student underneath CCE Minor Candidates is declared (ie declare student button is clicked on the list). The student name should enter CCE Minor Progress table to start their minor progress. But this doesn't happen this needs top be fixed.

Navigating to the issue page: Admin > Minor Management

Kafui123:
## Issue Description

Fixes #1660 
- Students who were declared as CCE minors did not appear on the CCE Minor Progress table.

## Changes
- Created a new query function called getDeclaredMinorStudentsWithProgress() that starts from the user table and uses left outer joins to include declared minors when they have no IndividualRequirement records yet.
- Computes engagementCount, hasSummer, and hasCCEMinorProposal directly from the database.
- Added email to the returned declared-minor progress data to support existing email string generation logic.
- Modified the tests to use the getDeclaredMinorStudentsWithProgress function.  

## Testing
- Click on Admin and then Minor Management
- Add students to the interested section of cce minor candidates
- Declare the interested student 
- The interested student should appear under the cce minor progress table 

## The student has been declared
Screenshot 2026-01-20 153528

## Student shows up on the CCE Minor Progress Table
Screenshot 2026-01-20 153651


app/controllers/admin/minor.py
```diff
@@ -4,7 +4,7 @@
 
 from app.controllers.admin import admin_bp
 
-from app.logic.minor import getMinorInterest, getMinorProgress, toggleMinorInterest, getMinorSpreadsheet, getDeclaredMinorStudents
+from app.logic.minor import getDeclaredMinorStudentsWithProgress, getMinorInterest, getMinorProgress, toggleMinorInterest, getMinorSpreadsheet, getDeclaredMinorStudents
```
JohnCox2211: getDeclaredMinorStudents is an unused import
```suggestion
from app.logic.minor import getDeclaredMinorStudentsWithProgress, getMinorInterest, getMinorProgress, toggleMinorInterest, getMinorSpreadsheet, getDeclaredMinorStudents
```

app/logic/minor.py
```diff
@@ -182,6 +182,52 @@ def declareMinorInterest(username):
     except Exception as e:
         raise RuntimeError(f"Failed to declare interested student: {e}")
     
+def getDeclaredMinorStudentsWithProgress():
```
JohnCox2211: Add a docstring

tests/code/test_minor.py
```diff
@@ -672,15 +672,12 @@ def test_declareMinorInterest():
...
+def test_getDeclaredMinorStudentsWithProgress():
     
     with mainDB.atomic() as transaction:
         # Get all the declared students
-        declaredStudents = getDeclaredMinorStudents()
-        
-        assert declaredStudents == []
-        assert len(declaredStudents) == 0
-        
+        declaredStudents = getDeclaredMinorStudentsWithProgress()
```
JohnCox2211: You never use declaredStudents. Originally, it was used to make sure that declaredStudents starts off as empty. Maybe you can reimplement this?

tests/code/test_minor.py
```diff
@@ -526,10 +526,10 @@ def test_createSummerExperience(testUser, testTerm, testProposal):
...
         testProposal["term"] = testTerm
 
-        User.create(username="glek",
-                    firstName="kafui",
-                    lastName="gle",
-                    email="kaf@berea.edu",
+        User.create(username="nimelyj",
+                    firstName="joyce",
+                    lastName="nimely",
+                    email="joycenimelyberea.edu",
```
JohnCox2211: Missing @ symbol in the email here

JohnCox2211 (changes requested):

tests/code/test_minor.py
Kafui123: Thanks for catching this. I have reimplemented it.

database/test_data.py
```diff
@@ -1027,7 +1042,15 @@
 ]
 
 IndividualRequirement.insert_many(individualReqs).on_conflict_replace().execute()
+# This ensures every user with an individual requirement has declaredMinor set to True so that when our query runs it can find users with  a "declaredMinor = True"
+req_usernames = {r["username"] for r in individualReqs if r.get("username")}
+print("Users with individual requirements to update:", req_usernames)
```
JohnCox2211: remove print

tests/code/test_graduationManagement.py
```diff
@@ -114,7 +114,7 @@ def test_getGraduationManagementUsers():
         BonnerCohort.create(year=2025, user=testUser1)
         BonnerCohort.create(year=2024, user=testUser2)
 
-        sustainedEngagement = {"username": testUser3,
+        sustainedEngagement = {"username": testUser3.username,
```
JohnCox2211: revert this

JohnCox2211:
Comments from PR review have been resolved

JohnCox2211:
My last minor thing would be the remove from minor button. I think it would be better if remove from minor was the header of that part of the table and the button said 'remove,' which would also add consistency with other buttons, but it works fine as is as well.

image

JohnCox2211 (dismissed):
You have addressed all of the comments that I left and the your implementation to solve the original issue works well, good work.

ojmakinde:
On development, I added myself as interested and then marked as declared, it looked like this:
image

Then, I switched to your branch, and it looked like this:
image

Where are the other students, who were in the minor? I didn't reset the database, so they should show up. @Kafui123

app/logic/minor.py
```diff
@@ -182,6 +185,55 @@ def declareMinorInterest(username):
...
+    """
+    This function retrieves a list of students who have declared the CCE minor along with their engagement progress.
+    It returns a list of dictionaries containing student information and their engagement details and adds students who have no requirements but have declared the minor with 0 engagements.
```
ojmakinde, line 191: Also, this comment says that it returns a list essentially containing students who have no requirements but have declared the minor. Doesn't that mean they have no progress?

ojmakinde (changes requested):

app/controllers/admin/minor.py
```diff
@@ -25,10 +25,38 @@ def manageMinor():
...
+            **student,
+            'isDeclaredMinor': student['username'] in declaredUsernames
+        }
+        
+    for student in declaredStudentsList:
+        if student['username'] not in sustainedUsernames:
+            cceMinorStudents[student['username']] = {
+                **student,
+                'isDeclaredMinor': True
+            }
```
ojmakinde: Here, it would be more efficient to merge two sets instead of using two for loops. We are essentially trying to see if students are in declared or not. With that, we should be able to run this with one or two set/dictionary comprehensions to make it much cleaner, faster and shorter.

app/logic/minor.py
```diff
@@ -182,6 +185,55 @@ def declareMinorInterest(username):
...
+        .join(CertificationRequirement, JOIN.LEFT_OUTER, on=(IndividualRequirement.requirement_id == CertificationRequirement.id))
+        .switch(User)
+        .join(CCEMinorProposal, JOIN.LEFT_OUTER, on=(User.username == CCEMinorProposal.student))
+        .where(
+            (User.declaredMinor == True) &
+            (
+                (CertificationRequirement.certification_id == Certification.CCE) |
+                (CertificationRequirement.id.is_null(True))
+            )
+        )
```
ojmakinde: Lastly, I found an incredibly nifty bug here, and I'm going to explain what (I think) is going on:

Assume we have this hypothetical result table after all the join statements, but just before the where condition:

User A, individualreq=3, declaredMinor=True
User B, individualreq=2, declaredMinor=True
User B, individualreq=3, declaredMinor=True
User C, individualreq=2, declaredMinor=True

In this case, user A will not show up in our result table, when in reality, they should. This is because of how the where clause is constructed. (Here's the where clause)

```
.where(
            (User.declaredMinor == True) &
            (
                (CertificationRequirement.certification_id == Certification.CCE) |
                (CertificationRequirement.id.is_null(True))
            )
        )
```

For user A, declaredMinor is true. But, they don't have a certification requirement with a certification ID for CCE Minor (which is `id=2`) nor do they have a null certificationRequirement. Because of that, the result table does not have user A, even though they have been declared.

What do you think @BrianRamsay?

app/logic/minor.py
```diff
@@ -182,6 +185,55 @@ def declareMinorInterest(username):
...
+    It returns a list of dictionaries containing student information and their engagement details and adds students who have no requirements but have declared the minor with 0 engagements.
+    """
+    summerCase = Case(None, [(CCEMinorProposal.proposalType == "Summer Experience", 1)], 0)
+
+    q = (
+        User
+        .select(
+            User,
+            fn.COUNT(fn.DISTINCT(IndividualRequirement.id)).alias("rawEngagementCount"),
+            fn.COALESCE(fn.SUM(fn.DISTINCT(summerCase)), 0).alias("summerCount"),
```
ojmakinde: this returns a 1 for all summer engagements, because the summer case is

`summerCase = Case(None, [(CCEMinorProposal.proposalType == "Summer Experience", 1)], 0)`

Essentially, you are returning 1 if it is true. So if there are 5 summer experiences, we get [1, 1, 1, 1, 1], and the distinct merges it into a set of [1] because that's the only distinct one. Instead, we want a set of the different individual requirements, so that way, we get a correct distinct count.

Additionally, if there is a bug in how `summerCase` is calculated, then
` engagementCount = int(s.rawEngagementCount) - int(s.summerCount or 0)` 
will not work as intended

ojmakinde (changes requested):

app/logic/minor.py
ojmakinde: I fixed this. You can take a look at my commit to see my solution. You can mark as resolved if you find it satisfactory

ojmakinde:
> On development, I added myself as interested and then marked as declared, it looked like this: image
> 
> Then, I switched to your branch, and it looked like this: image
> 
> Where are the other students, who were in the minor? I didn't reset the database, so they should show up. @Kafui123

I took a look at this and it was fixed by improving the existing logic.

app/logic/minor.py
ojmakinde: I have implemented the correct solution. Test and mark as resolved when you find it satisfactory.

app/controllers/admin/minor.py
ojmakinde: Implemented. Please look over the solution and mark as resolved if satisfactory

app/logic/minor.py
ojmakinde: @Kafui123

Kafui123: Yes, that's correct, but once they have been declared for the minor, they need to show up on the table regardless of whether they have made progress or not.

ojmakinde (dismissed):
Changes have been addressed. Ready for merge.

JohnCox2211:
When someone has 0/4 for their sustained engagements, if you add a summer experience, the summer experience column shows completed, but the sustained engagements column shows -1/4 now. This does not happen if the person already has a sustained engagement completed. 

image

JohnCox2211 (changes requested):
address summer experience comment

app/logic/minor.py
```diff
@@ -235,8 +235,7 @@ def getDeclaredMinorStudents():
 
     result = []
     for s in q:
-     
-        engagementCount = int(s.allEngagementCount or 0) - int(s.summerEngagementCount or 0)
+        engagementCount = int(s.allEngagementCount or 0)
```
ojmakinde, line 241: this still returns an incorrect count. if a user has 2 summer engagements and 0 other engagements, this variable `engagementCount` will return 0. it should return 2.

Kafui123: Yes, that is how(I think) it is supposed to work because the "allengagementCount"  counts individual requirements  only  when a certification requirement exists, and summer engagements are not included in that anymore.

JohnCox2211 (approved):
Made all requested changes. Looks good to me

PR #1666, opened 2026-01-21
Fixed insurance check box issue.

### LINKED ISSUE #1655: Insurance information unchecked by default
When the user navigates to the Volunteer Details tab, the Insurance Information column is displayed by default. This should not occur, as insurance data is sensitive and should only be shown when explicitly requested by the user through the checklist.
Image

Jonimz:
Fixes #1655 

Issue Description
After creating an event and navigating to the Volunteer Details tab, users see a table of volunteers along with a checklist that controls which columns are displayed. Currently, the Insurance Information option is checked by default, causing the insurance column to appear automatically.

Changes
Updated the default state of the Insurance Information checkbox (insuranceSelect) to be unchecked on page load
Ensured the insurance column is hidden by default
Kept existing behavior for all other checklist options and table functionality

Screenshot 2026-01-21 at 3 36 16 PM

Screenshot 2026-01-21 at 3 45 53 PM

Image
Testing
Checked out the branch
git checkout insurance-info
git pull
Created or opened an event
Navigated to Volunteer Details
Verified the Insurance Information column is hidden by default
Verified checking the Insurance option displays the column
Verified unchecking it hides the column again

### DIFF
--- a/app/static/js/volunteerDetails.js
+++ b/app/static/js/volunteerDetails.js
@@ -78,7 +78,7 @@ $(document).ready(function () {
 			$(e).addClass(i % 2 ? 'custom-odd' : 'custom-even')
 		})
 	}
-
+    $("#insuranceSelect").prop("checked", false);
 	getCheckBoxes()
 	hideDuplicateVolunteers()
 	sortVolunteers()
--- a/app/static/js/volunteerDetails.js
+++ b/app/static/js/volunteerDetails.js
@@ -78,7 +78,6 @@ $(document).ready(function () {
 			$(e).addClass(i % 2 ? 'custom-odd' : 'custom-even')
 		})
 	}
-    $("#insuranceSelect").prop("checked", false);
 	getCheckBoxes()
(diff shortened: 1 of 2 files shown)


app/static/js/volunteerDetails.js
```diff
@@ -78,7 +78,7 @@ $(document).ready(function () {
 			$(e).addClass(i % 2 ? 'custom-odd' : 'custom-even')
 		})
 	}
-
+    $("#insuranceSelect").prop("checked", false);
```
Karina-Agliullova: This solution works, but I don't recommend using it. Instead, look at your HTML file, as it has the "checked" next to insurance info over there, and so your JS is basically overriding HTML, which is not ideal. Instead, just remove the "checked" from HTML, and it should produce the same result, and it won't override anything

Jonimz: Fixed, thank you.

Karina-Agliullova (approved):
This pr is looking good, issue was resolved!

PR #1668, opened 2026-01-23
Non enrolled and graduated students

### LINKED ISSUE #1649: Graduated and not-enrolled students don't show up in graduation management
Graduated students do not show up in the graduation management page. The only graduated students that show up are the seniors that we explicitly check to say that they graduated. Other students who have graduated before we go check, do not show up in the list at all. We should have a designation for them the same way we have one for senior, junior, etc. And use that designation to display graduated students. Same thing for students that left before they graduated.

### LINKED ISSUE #1657: Class level is missing for non-enrolled students
Graduated or non-enrolled students don't have any designation on their profile, like Senior, Junior, Sophomore, Freshmen students do.

bakobagassas:
## Issue Description

Fixes #1649  
Graduated students do not show up in the graduation management page. The only graduated students that show up are the seniors that we explicitly check to say that they graduated. Other students who have graduated before we go check, do not show up in the list at all. We should have a designation for them the same way we have one for senior, junior, etc. And use that designation to display graduated students. Same thing for students that left before they graduated.

Fixes #1657  
Graduated or non-enrolled students don't have any designation on their profile, like Senior, Junior, Sophomore, Freshmen students do.

## Changes

- New functions to handle alumni and enrollment 
- Modified the display on the graduation management page, now students who graduated recently appear with the Alumni class level. 
- Ticked students become alumni automatically 
- Students who dropped out or transferred appear as non enrolled 

the page then (marked students who graduated last semester as seniors): 

image

the page now:

image

The profile display for graduated students:

image

The profile display for non-enrolled students:

image

## Testing

-  database/reset_database.sh test  reset data base with test 
- tests/run_tests.sh -v  Run the tests 
- Use backup data database/reset_database.sh from-backup
- Then Flask run 
- Then go on the website 
- Click on admin
- Then go on the graduation management page 
- People marked as alumni, are people who recently graduated. When you untick the "include graduated students" toggle, they disappear from the list 
- When you click on Veronica's name, she becomes part of the alumni and disappear from the list unless you click on including graduated student. In case it is a mistake you can safely uncheck the box and she becomes a senior. 
- If another student is marked as alumni and they realize that it is a mistake, they can also untick the box and the person becomes a Senior instead.
- On the student profile search for coronad, she transferred last year and will appear as "Non'enrolled" on her profile.

### DIFF
--- a/app/static/js/graduationManagement.js
+++ b/app/static/js/graduationManagement.js
@@ -46,6 +46,7 @@ $(document).ready(function() {
     function getRowStatus(row) {
         return $(row).data('status');
     }
+    
     function filterTable(dataField, expectedValue) {
         gradStudentsTable.rows().every(function() {
             const row = this.node();
--- a/app/static/js/graduationManagement.js
+++ b/app/static/js/graduationManagement.js
@@ -43,22 +43,26 @@ $(document).ready(function() {
 
     })
 
+    function getRowStatus(row) {
+        return $(row).data('status');
+    }
     function filterTable(dataField, expectedValue) {
         gradStudentsTable.rows().every(function() {
-            var hasGraduated = $(this.node()).find('input[type="checkbox"]').is(':checked');
-            if (!showGraduatedStudents() && hasGraduated) {
-                $(this.node()).addClass('hidden')
-                return 
+            const row = this.node();
+            const status = getRowStatus(row);
+            if (!showGraduatedStudents() && status === 'alumni') {
+                $(row).addClass('hidden');
+                return;
             }
-            var data = $(this.node()).data(dataField); 
+            const data = $(row).data(dataField);
 
             if (data === expectedValue) {
-                $(this.node()).removeClass('hidden')
+                $(row).removeClass('hidden');
             } else {
-                $(this.node()).addClass('hidden')
+                $(row).addClass('hidden');
             }
         });
-        redrawTable()
+        redrawTable();
     }
 
     function handleBonnerFilterChange(cohortYear, buttonText) {
@@ -94,10 +98,10 @@ $(document).ready(function() {
 
             gradStudentsTable.search('').draw();
             gradStudentsTable.rows().every(function() {
-                var hasGraduated = $(this.node()).find('input[type="checkbox"]').is(':checked');
-                if (!showGraduatedStudents() && hasGraduated) {
-                    $(this.node()).addClass('hidden')
-                    return 
+                const status = getRowStatus(this.node());
+                if (!showGraduatedStudents() && status === 'alumni') {
+                    $(this.node()).addClass('hidden');
+                    return;
                 }
                 $(this.node()).removeClass('hidden');
             });
@@ -134,7 +138,14 @@ $(document).ready(function() {
             success: function(response) {
                 initializePage()
                 msgFlash(`Saved graduation status for ${username}.`, "success", 1000)
-                $(`#${username}ClassLevel`).html(hasGraduated ? "Graduated" : "Senior")
+                const row = $(`tr[data-username="${username}"]`);
+                if (hasGraduated) {
+                    row.data('status', 'alumni');
+                    $(`#${username}ClassLevel`).text("Alumni");
+                } else {
+                    row.data('status', 'enrolled');
+                    $(`#${username}ClassLevel`).text("Senior");
+                }
             },
             error: function(status, error) {
                 console.error("Error updating graduation status:", error);
--- a/app/static/js/graduationManagement.js
+++ b/app/static/js/graduationManagement.js
@@ -142,6 +142,10 @@ $(document).ready(function() {
                 if (hasGraduated) {
                     row.data('status', 'alumni');
                     $(`#${username}ClassLevel`).text("Alumni");
+                    if (!showGraduatedStudents()) {
+                        row.addClass('hidden');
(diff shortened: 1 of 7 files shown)


MImran2002:
Fix 1: `if status == 1` can just be `if status` as integer 1 and above is always True and integer 0 is always False

Fix 2: in app>models>user.py the `if self.hasGraduated or self.rawClassLevel == "Graduating": in processedClassLevel(self)` can use self.isAlumni as both has the same logic. 

Fix 3: When the `include graduating students` is not toggle and you click a student graduating checkbox their status turn alumni but they don't disappear, they should disappear as the filter is toggle:
image

Fix 4: Because you have changed and assigned in setGraduationStatus, there should be test case for the part where their rawClassLevel switch where there is a change in status right now there is test on hasGraduated.

Question & suggestion 1: In app>logic>graduationManagement>`setGraduatedStatus()` the if logic assigned the rawClassLevel as "Graduating" which means the someone who will be graduating will always be graduating and there won't be any code or function where the student is assigned "alumni". I am asking this because I know the "alumni" is returned in the `processedClassLevel` but this property method is not affecting the database at all. Which means this peewee command `.where((User.rawClassLevel == 'Senior') | (User.rawClassLevel == "Graduating") | (User.hasGraduated == True) | (User.rawClassLevel == "Alumni")))` where there is `User.rawClassLevel == Alumni` will never return Alumni. Because the Alumni is never added into the database and we can use the property method here too so if you are just going along without adding "Alumni" in the database the peewee filter should go. 

NitPicking 1: The two functions need space :) 
`function getRowStatus(row) {
        return $(row).data('status');
    }`
`function filterTable(dataField, expectedValue) {`

bakobagassas:
> Fix 1: 
>  
Done 
> Fix 2:
> 
 Done
> Fix 3: 
>  
Done
> Fix 4: Because you have changed and assigned in setGraduationStatus, there should be test case for the part where their rawClassLevel switch where there is a change in status right now there is test on hasGraduated.
>
Done
> Question & suggestion 1: In app>logic>graduationManagement>`setGraduatedStatus()` the if logic assigned the rawClassLevel as "Graduating" which means the someone who will be graduating will always be graduating and there won't be any code or function where the student is assigned "alumni". I am asking this because I know the "alumni" is returned in the `processedClassLevel` but this property method is not affecting the database at all. Which means this peewee command `.where((User.rawClassLevel == 'Senior') | (User.rawClassLevel == "Graduating") | (User.hasGraduated == True) | (User.rawClassLevel == "Alumni")))` where there is `User.rawClassLevel == Alumni` will never return Alumni. Because the Alumni is never added into the database and we can use the property method here too so if you are just going along without adding "Alumni" in the database the peewee filter should go.
> 
- So when a student is "Graduating" it means that the graduation term has passed but hasGraduated may not yet be set. Fall 2025 grads for example would be "Graduating" + hasGraduated=False, but once confirmed we have hasGraduated=True. But either way, the UI shows Alumni because either way, they are Alumni. processedClassLevel truly is just for the front-end use, we are intentionally not storing Alumni. 
- I fixed the pewee query by removing User.rawClassLevel == Alumni, that is never true 
> NitPicking 1

Done

MImran2002:
Fix 1: there are the three comments, I don't think all three will be necessary but if you want to add them you can add them in the docstring. 

Looks great, i couldn't find any other edge cases. @ojmakinde Do you want to check Sara's too. She mentioned that she wants you to that's why.

MImran2002:
Fix FInal: So  `"Graduating" means: graduation term has passed but hasGraduated may not yet be set` comment should be integrated into the docstring as docstring is to write about the purpose and input and return of a function. 

You have the comment or docstring in three different location: `def getGraduationManagementUsers()`, `def setGraduatedStatus(username, status)` and `def isAlumni(self)`. In all those three function and property the most appropriate location to add this information is in `setGraduatedStatus()` because this is where logically Graduating is first set so this function should be the only one with the information that tells why there is graduating.

MImran2002:
Job well done, we just need to wait for a second person to approve it.

MImran2002 (dismissed):
Job well done. Just need a second person to check.

app/static/js/graduationManagement.js
Arohasina, line 1: In app/models/user.py, the new isCurrentlyEnrolled property is defined as "isStudent" and "not isAlumni". However, the setGraduatedStatus function in app/logic/graduationManagement.py currently only updates hasGraduated and rawClassLevel.

If a student graduates, they are typically no longer considered an active 'student.' Should setGraduatedStatus also toggle the isStudent boolean to False? Keeping these in sync could ensure that any logic elsewhere in the app that filters by isStudent correctly excludes alumni.

Arohasina (dismissed):

app/static/js/graduationManagement.js
bakobagassas: So I think that in our model isStudent says if the user originated as a student account, not if they are currently enrolled. The student being actively enrolled is determined by isCurrentlyEnrolled from my understanding, which coming from isStudent and alumni status. So if they are not a student and they are not an alumni then they are not enrolled. Alumni are just graduated **students**, (graduated is just the status of the student, it is not their identity as a student). Toggling isStudent can be problematic as it can make alumni not be able to see events and all that.

Arohasina: I see. All good then.

PR #1675, opened 2026-02-16
Paging issue

### LINKED ISSUE #1656: Inconsistencies in Volunteer Details table
Arrange the flow of the list in volunteer details when RSVP is unchecked. Here, for instance, Joyce's details appear on the third page, as she is the only person listed. So logically, she should be appearing on the first page. Which means whenever these checkmarks are done the page should reconsolidate the data and repopulate them to fit the filter that is there like 50-100.

Image

bakobagassas:
## Paging issue

Fixes #1656 

- Arrange the flow of the list in volunteer details when RSVP is unchecked. Here, for instance, Joyce's details appear on the third page, as she is the only person listed. So logically, she should be appearing on the first page. Which means whenever these checkmarks are done the page should reconsolidate the data and repopulate them to fit the filter that is there like 50-100.

## Changes

- Changed mainly the JavaScript file 
- Now we are not anymore manually hiding table rows when checkboxes were unchecked but we use the DataTables custom filtering
- Removed Duplicate DataTable Initialization

Now joyce appears on the first page when RSVP unchecked
image

## Testing

- Go to events page
- Change to the fall 2025 term
- go to Bonner Scholars tab 
- Click on the only event there 
- Click on volunteer details
- You can check and uncheck the RSVP box

### DIFF
--- a/app/templates/events/volunteerDetails.html
+++ b/app/templates/events/volunteerDetails.html
@@ -59,15 +59,13 @@
 {% macro printParticipants(type, attended, rsvp, waitlist) %}
     {% set combinedParticipants = attended + rsvp + waitlist %}
     {% for p in combinedParticipants | unique %}
-      {% if p in rsvp %}
-          {{createTable(p, 'rsvp') if type == 'table' else createCard(p, 'rsvp') }}
-      {% endif %}
-      {% if p in attended%}
-          {{createTable(p, 'attended') if type == 'table' else createCard(p, 'attended') }}
-      {% endif %}
-      {% if p in waitlist %}
-          {{createTable(p, 'waitlist') if type == 'table' else createCard(p, 'waitlist') }}
-      {% endif %}
+        {% if p in attended %}
+            {{ createTable(p, 'attended') if type == 'table' else createCard(p, 'attended') }}
+        {% elif p in rsvp %}
+            {{ createTable(p, 'rsvp') if type == 'table' else createCard(p, 'rsvp') }}
+        {% elif p in waitlist %}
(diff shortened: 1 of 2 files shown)


app/templates/events/volunteerDetails.html
```diff
@@ -57,17 +57,20 @@ <h4 class="nameSelect" nowrap><b>{{ participant.user.firstName }} {{ participant
...
-      {% endif %}
-      {% if p in waitlist %}
-          {{createTable(p, 'waitlist') if type == 'table' else createCard(p, 'waitlist') }}
-      {% endif %}
+    {% for p in combinedParticipants %}
+        {% set username = p.user.username %}
+        {% if username not in seen %}
+            {% set _ = seen.append(username) %}
+            {% if p in attended %}
+                {{ createTable(p, 'attended') if type == 'table' else createCard(p, 'attended') }}
```
JohnCox2211: Maybe instead of using create table 3 times, create a new variable called 'status', for example, and set that variable to be 'attended', 'rsvp', and 'waitlist within the if statements. Then after the if statements create the table, ex:

```
{% set status = None %}
{% if p in attended %}
    {% set status = 'attended' %}
{% elif p in rsvp %}
    {% set status = 'rsvp' %}
{% elif p in waitlist%}
    {% set status = 'waitlist' %}
{% endif %}

{% if status %}
    {{ createTable(p, status) if type == 'table' else createCard(p, status) }}
{% endif %}
```

This would make general code maintenance in the future easier.

JohnCox2211 (dismissed):

ojmakinde (approved):
LGTM. Fixes the bug fairly elegantly with datatable sorting.

PR #1681, opened 2026-02-23
training message

### LINKED ISSUE #1432: Remove training info from event registration success message
When you RSVP for an event you get a flash message with a million different trainings that really shouldn't be necessary. Let's remove these.

bakobagassas:
## Issue Description

Fixes #1432 
- When you RSVP for an event you get a flash message with a million different trainings that really shouldn't be necessary. They need to be removed.

## Changes

- Changed the routes flash message and removed unused variables 

## Testing

- Create a new event that requires an RSVP (while you are logged in as admin)
- Then log in as a student 
- Go to events list 
- RSVP for that event, you should see the new flash message without the trainings.

### DIFF
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -493,9 +493,7 @@ def volunteerRegister():
     program = event.program
     user = g.current_user
 
-    isAdded = checkUserRsvp(user, event)
     isEligible = isEligibleForProgram(program, user)
-    listOfRequirements = unattendedRequiredEvents(program, user)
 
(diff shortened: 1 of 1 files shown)


Meatchema (commented):
It worked. I received the flash without all the additional trainings needed.

Meatchema (commented):
It worked. I received the flash without all the additional trainings needed.

JohnCox2211 (approved):
The flash works; the unnecessary training list is gone as well.

PR #1684, opened 2026-02-25
Labor attendance report 1626

### LINKED ISSUE #1626: Add Labor Attendance report
We should have a sheet in the Reports downloads that shows labor attendance for the academic year. The report should have a list of labor students and the number of labor meetings that they attended for each term in the academic year.

Arohasina:
## Issue Description

Fixes #1626
- We should have a sheet in the Reports downloads that shows labor attendance for the academic year. The report should have a list of labor students and the number of labor meetings that they attended for each term in the academic year.

## Changes

- Implemented `laborAttendanceByTerm(term)` function to query labor students and count their meeting attendance by term
- Added "Labor Attendance By Term". There are two sheets: one for Fall term and one for Spring term (for the academic year)
- added the test function `test_laborAttendanceByTerm` in `test_spreadsheet.py `
 
This is how the spreadsheet looks like: 
image

## Testing

- run `database/reset_database.sh test` and then `tests/run_tests.sh` or if you want to test the laborAttendanceByTerm function specifically, you can run `pytest tests/code/test_spreadsheet.py::test_laborAttendanceByTerm -v`
- after it passed, run `database/reset_database.sh from-backup` and then `flask run`
- Once you are in the Celts app, go to event list choose Fall 2024, and then choose a couple of events there, such as these events
image

- And for each, click on the event,  turn on the toggle "This event is for labor only" , and then click "save"
image

- And then do the same for Spring 2025. 

- Then go to admin --> report --> select academic year 2024 - 2025 --> download the spreadsheet, open it, and you should see a sheet called "Labor Attendance By Term" with the reporting, one for Fall and one for Spring term.

### DIFF
--- a/app/logic/volunteerSpreadsheet.py
+++ b/app/logic/volunteerSpreadsheet.py
@@ -226,28 +226,8 @@ def calculateRetentionRate(fallDict, springDict):
 
     return retentionDict
 
-def laborAttendanceByTerm(academicYear):
-#     """Get labor students and their meeting attendance count for each term"""
-#     base = getBaseQuery(academicYear)
-
-#     query = (base.select(
-#         fn.CONCAT(User.firstName, ' ', User.lastName).alias('fullName'), 
-#         User.bnumber, 
-#         fn.CONCAT(EventParticipant.user_id, '@berea.edu').alias('email'),
-#         Term.description, 
-#         fn.COUNT(EventParticipant.event_id).alias('meetingsAttended'), 
-#     )
-#     .where(Event.isLaborOnly == True)
-#     .group_by(EventParticipant.user_id, Term.description)
-#     .order_by(User.lastName, User.firstName, Term.description)
-#     )
-
-#     columns = ("Full Name", "B-Number", "Email", "Term", "Meetings Attended")
-#     results = list(query.tuples())
-#     return (columns, results)
-
-
-    query = (
+def laborAttendanceByTerm(academicYear): 
+    laborQuery = (
         CeltsLabor
         .select(
             fn.CONCAT(User.firstName, ' ', User.lastName).alias('fullName'),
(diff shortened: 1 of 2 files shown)


tests/code/test_spreadsheet.py
```diff
@@ -683,5 +685,15 @@ def test_getUniqueVolunteers(fixture_info):
         ("Test Tester", "testt@berea.edu", "B55555"),
     ])
 
+@pytest.mark.integration
+def test_laborAttendanceByTerm(fixture_info):
+    EventParticipant.create(event=fixture_info["event1"], user=fixture_info['user1'], hoursEarned=1)
```
MImran2002: I think one practice in testing would be to use : with mainDB.atomic() as transaction: and transaction.rollback() this helps with reducing the time to always create unique data and instead you can always create the same data insert it into the database and roll it back.

app/logic/volunteerSpreadsheet.py
```diff
@@ -225,6 +226,74 @@ def calculateRetentionRate(fallDict, springDict):
 
     return retentionDict
 
+def laborAttendanceByTerm(term):
+    laborQuery = ( #so that all Celts Labor students appear even if they didn't attend anything 
+        CeltsLabor
+        .select(
+            fn.CONCAT(User.firstName, ' ', User.lastName).alias('fullName'),
```
MImran2002: these fn.CONCAt can be turn into variables
for example: full_name = fn.CONCAT(User.firstName, ' ', User.lastName).alias('fullName')
so you should do that for email, meetingsattended too. So that you can use these variables in both labor query and nonlaborquery

app/logic/volunteerSpreadsheet.py
```diff
@@ -225,6 +226,74 @@ def calculateRetentionRate(fallDict, springDict):
...
+        .join(
+            EventParticipant,
+            JOIN.LEFT_OUTER,
+            on=(CeltsLabor.user == EventParticipant.user)
+        )
+        .join(
+            Event,
+            JOIN.LEFT_OUTER,
+            on=(
+                (EventParticipant.event == Event.id) &
```
MImran2002: these can also be turn into variables like valid event =               (EventParticipant.event == Event.id) &
                (Event.term == term) &
                (Event.isLaborOnly == True) &
                (Event.deletionDate.is_null()) &
                (Event.isCanceled == False) then it can be reused in laborquery and nonlaborquery

app/logic/volunteerSpreadsheet.py
```diff
@@ -225,6 +226,74 @@ def calculateRetentionRate(fallDict, springDict):
...
+                (Event.term == term) &
+                (Event.isLaborOnly == True) &
+                (Event.deletionDate.is_null()) &
+                (Event.isCanceled == False)
+            )
+        )
+        .where(
+            (CeltsLabor.term == term)
+        )
+        .group_by(CeltsLabor.user)
```
MImran2002: these can be made in singlie lines like 
.where((CeltsLabor.term == term))
.group_by(celtsLabor.user)

app/logic/volunteerSpreadsheet.py
```diff
@@ -225,6 +226,74 @@ def calculateRetentionRate(fallDict, springDict):
...
+            CeltsLabor,
+            JOIN.LEFT_OUTER,
+            on=(EventParticipant.user == CeltsLabor.user) &
+                (CeltsLabor.term == term) 
+        )
+        .where(CeltsLabor.user.is_null())
+        .group_by(EventParticipant.user)
+    )
+
+    query = laborQuery.union(nonLaborQuery).order_by(SQL('fullName'))
```
MImran2002, line 270: if the above laborQuery and nonlaborquery are only for them to be join I think there are other solutions where we can combine those laborquery and nonlaborquery into a single peewee command so we don't even need to do this again. where we are doing union

MImran2002 (changes requested):

ojmakinde (dismissed):
Looks good

app/logic/volunteerSpreadsheet.py
ojmakinde: @Arohasina good add.

Arohasina: fixed!

Arohasina:
I noticed the spring report didn't include zero attendance (for labor students). Brian suggested a roster membership rule for labor attendance reports. Include a student in the labor roster if either:
 1) the spreadsheet term exactly matches their CeltsLabor term, OR
 2) their CeltsLabor row is an academic-year assignment and belongs to the same academic year as the spreadsheet term.

Academic-year labor rows are stored once (typically on Fall), but we generate both Fall and Spring
sheets. Without this rule, Spring can miss labor students who attended 0 labor events, so those students do not appear with a zero count. So I added that fix.

MImran2002 (commented):
this is still under review. I am revisiting the peewee again during lunch.

MImran2002 (dismissed):
Looks good to me!

Meatchema (approved):
I have tested this PR and it works as intended. 

Image

MImran2002 (approved):
double check and all good was worried about choosing union over union_all but the data lost there won't happen as there won't be duplicates based on the where statements so I am going to approve it

PR #1685, opened 2026-02-26
Removed labor meetings from participation history

### LINKED ISSUE #1683: Remove "Labor Meetings" from participation history for CELTS students and track it separately
The screenshot here shows a student’s participation history, and this is a CELTS student, so it shows many labor meetings. If an event is checked as "For Labor students only", remove those events from the Participation history. We should track attendance for Labor meetings elsewhere, such as printing out a report for the semester's meetings to date

Image

Karina-Agliullova:
Fixes #1683 

There was a request to remove labor meeting events from the Participation history. And also, those labor meetings should be tracked in reports for each academic year per each term. The problem with removing it from participation history was inconsistency in naming of labor meetings and also filtering: sometimes they are named "CELTS labor meeting" or "Labor meeting(date)", so it was confusing to remove Labor only events as there are trainings that are also filtered the exact same way.

## Changes

- Added a new function to exclude anything labeled labor meeting
- Got rid of labor meetings from participation history, while preserving trainings

## Testing

- Go to the user profile from backup data, such as Kai Craig
- Go to participation history
- Check if there are no labor meetings
- Download the report for AY 2025-2026, and find the column Labor Meetings
- Check their name, Kai Craig, as an attendee (confirming that there is data with labor meetings under their name)

### DIFF
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -380,7 +380,7 @@ def getParticipatedEventsForUser(user):
                                .join(Program, JOIN.LEFT_OUTER).switch()
                                .join(EventParticipant)
                                .where(EventParticipant.user == user,
-                                      Event.isAllVolunteerTraining == False, Event.deletionDate == None, checkIfLaborMeeting)
+                                      Event.isAllVolunteerTraining == False, Event.deletionDate == None, ~checkIfLaborMeeting)
                                .order_by(Event.startDate, Event.name))
 
     allVolunteer = (Event.select(Event, "")
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -364,6 +364,16 @@ def getUpcomingEventsForUser(user, asOf=datetime.now(), program=None):
 
     return eventsList
 
+def excludeLaborMeeting():
+    eventName = fn.LOWER(Event.name)
+    eventDescription = fn.LOWER(Event.description)
+
+    return (
+        eventName.contains("labor meeting") |
+        eventName.contains("celts labor meeting") |
+        eventDescription.contains("labor meeting")
+    )
+
 def getParticipatedEventsForUser(user):
     """
         Get all the events a user has participated in.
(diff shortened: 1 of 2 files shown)


MImran2002 (dismissed):
I don't think I have any changes in mind, this is a quick fix and I have talked to Karina that a long-term fixed would possibly be an attribute but for now I think it is a great fix.

app/logic/events.py
```diff
@@ -373,17 +383,18 @@ def getParticipatedEventsForUser(user):
...
                                .where(EventParticipant.user == user,
-                                      Event.isAllVolunteerTraining == False, Event.deletionDate == None)
+                                      Event.isAllVolunteerTraining == False, Event.deletionDate == None, ~((Event.isLaborOnly == True) & excLaborMeeting), ~((Event.isTraining == True) & excLaborMeeting))
                                .order_by(Event.startDate, Event.name))
 
     allVolunteer = (Event.select(Event, "")
                          .join(EventParticipant)
                          .where(Event.isAllVolunteerTraining == True,
-                                EventParticipant.user == user))
+                                EventParticipant.user == user, ~((Event.isLaborOnly == True) & excLaborMeeting), ~((Event.isTraining == True) & excLaborMeeting)))
```
BrianRamsay: All Volunteer is always All Volunteer, without other qualifiers. If they screw up the data entry, that's on them.

app/logic/events.py
```diff
@@ -373,17 +383,18 @@ def getParticipatedEventsForUser(user):
         :return: A list of Event objects
     """
 
+    excLaborMeeting = excludeLaborMeeting()
     participatedEvents = (Event.select(Event, Program.programName)
                                .join(Program, JOIN.LEFT_OUTER).switch()
                                .join(EventParticipant)
                                .where(EventParticipant.user == user,
-                                      Event.isAllVolunteerTraining == False, Event.deletionDate == None)
+                                      Event.isAllVolunteerTraining == False, Event.deletionDate == None, ~((Event.isLaborOnly == True) & excLaborMeeting), ~((Event.isTraining == True) & excLaborMeeting))
```
BrianRamsay: We can simplify this by just excluding Labor Meetings. There aren't going to be other labor meetings that we want to put in this list.

app/logic/events.py
```diff
@@ -364,6 +364,16 @@ def getUpcomingEventsForUser(user, asOf=datetime.now(), program=None):
 
     return eventsList
 
+def excludeLaborMeeting():
```
BrianRamsay: This function makes a clause that determines if the event IS a labor meeting, but is called `excludeLaborMeeting`

app/logic/events.py
```diff
@@ -364,6 +364,16 @@ def getUpcomingEventsForUser(user, asOf=datetime.now(), program=None):
...
+    eventName = fn.LOWER(Event.name)
+    eventDescription = fn.LOWER(Event.description)
+
+    return (
+        eventName.contains("labor meeting") |
+        eventName.contains("celts labor meeting") |
```
BrianRamsay: redundant, because this would be caught be the first check.

app/logic/events.py
```diff
@@ -364,6 +364,16 @@ def getUpcomingEventsForUser(user, asOf=datetime.now(), program=None):
...
+        eventDescription.contains("labor meeting")
```
BrianRamsay: This could potentially catch too much. Let's not check event description.

BrianRamsay (changes requested):

PR #1687, opened 2026-03-09
Make Start Date appear

### LINKED ISSUE #1594: Make Start Date appear in the "This is a series of events" modal on Create Event page
When doing the usability study, in reference to task " Create an Event that Repeats Three Weeks in a Series", participants often faced a problem in which the start date created on the main page (after picking an Event type within "Create Event"), when they turned on the toggle ("This is a series of Events") the start date would disappear and didnt transfer to the series. The start date selected outside of the modal should transfer into the modal and should still be editable.

bakobagassas:
## Issue Description

Fixes #1594 
- When doing the usability study, in reference to task " Create an Event that Repeats Three Weeks in a Series", participants often faced a problem in which the start date created on the main page (after picking an Event type within "Create Event"), when they turned on the toggle ("This is a series of Events") the start date would disappear and didnt transfer to the series. The start date selected outside of the modal should transfer into the modal and should still be editable.

## Changes

- Only js file, made it so that the event is carried into the box, same as the name and location.
- Carried the date into the occurrence events 
- Carried the time too

## Testing

- Go to create an event 
- Select the program 
- Fill out name, location, and date 
- Click that it is a series of events 
- toggle the weekly repetition 
- You should see the name you previously inputted appear as the start date.

### DIFF
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -120,7 +120,7 @@ def attemptSaveMultipleOfferings(eventData, attachmentFiles = None):
 
     # Create separate event data for each event in the series, inheriting from the original eventData
 
-    seriesData = sorted(eventData.get('seriesData'), key=lambda x: datetime.strptime(f"{x['eventDate']} {x['startTime']}",'%m/%d/%Y %H:%M'))
+    seriesData = sorted(eventData.get('seriesData'), key=lambda x: datetime.strptime(f"{x['eventDate']} {x['startTime']}",'%m/%d/%Y %H:%M')) # sorts the events in the series by date and time so that the events are created in order and the naming convention of Week 1, Week 2, etc. is consistent with the order of the events.
     isRepeating = bool(eventData.get('isRepeating'))
     with mainDB.atomic() as transaction:
         for index, event in enumerate(seriesData):
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -120,7 +120,7 @@ def attemptSaveMultipleOfferings(eventData, attachmentFiles = None):
 
     # Create separate event data for each event in the series, inheriting from the original eventData
 
-    seriesData = sorted(eventData.get('seriesData'), key=lambda x: datetime.strptime(f"{x['eventDate']} {x['startTime']}",'%Y-%m-%d %H:%M'))
+    seriesData = sorted(eventData.get('seriesData'), key=lambda x: datetime.strptime(f"{x['eventDate']} {x['startTime']}",'%m/%d/%Y %H:%M'))
     isRepeating = bool(eventData.get('isRepeating'))
     with mainDB.atomic() as transaction:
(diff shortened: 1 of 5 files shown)


JohnCox2211 (changes requested):
The date appears in the multiple-offering modal when toggling 'This series repeats weekly'; however, it's not carried over when you press 'Add Occurrence', which I feel like it should. 

I think it would also be good to carry over the time from the initial form into the modal, since that would be a good quality-of-life change and very similar to the date appearing in the modal.

JohnCox2211 (changes requested):
- The start/end time is not carried over when you press the add occurrence button in the series of events modal 
- After saving recurring events in the modal, then exiting the modal to save the event overall, there is a value error: ValueError: time data '4/3/2026 15:00' does not match format '%Y-%m-%d %H:%M'

bakobagassas:
Ready for review.

app/logic/events.py
```diff
@@ -119,8 +119,9 @@ def attemptSaveMultipleOfferings(eventData, attachmentFiles = None):
     seriesId = calculateNewSeriesId()
 
     # Create separate event data for each event in the series, inheriting from the original eventData
-
-    seriesData = sorted(eventData.get('seriesData'), key=lambda x: datetime.strptime(f"{x['eventDate']} {x['startTime']}",'%Y-%m-%d %H:%M'))
+    print("seriesData received:", eventData.get('seriesData'))
```
MImran2002: I don't think this print statement is needed.

MImran2002 (changes requested):

tests/code/test_events.py
```diff
@@ -772,7 +772,7 @@ def test_deleteEvent():
 @pytest.mark.integration
 def test_upcomingEvents():
     with mainDB.atomic() as transaction:
-        testDate = datetime.strptime("2021-08-01 05:00","%Y-%m-%d %H:%M")
+        testDate = datetime.strptime("08/01/2021 05:00","%m/%d/%Y %H:%M")
```
MImran2002, line 775: I want to know why do you have to change this because if there is a reason there are other places in the folder like test_participants > test_getEventLengthInHours() that should be change too.

MImran2002 (changes requested):

tests/code/test_events.py
bakobagassas: I changed it because I was carrying the date from the main page to the occurrences and to the series. However, the two were formatted differently because they initially got their dates from different sources. Therefore, when I carried the date there, I encountered some mismatches because the format on the front-end was different from what the backend expected. Hence why I had to change some of the formats and make the tests follow that. All tests are passing.

MImran2002 (approved):

JohnCox2211 (approved):
The changes look good to me and the functionality works

PR #1694, opened 2026-04-07
Marking Volunteers as Attended from the RSVP List Wipes Current Volunteer Names Fix

### LINKED ISSUE #1692: Marking Volunteers as Attended from the RSVP List Wipes Current Volunteer Names
- **Fix**: When you have an event that has an RSVP and Waitlist, if you mark volunteers as attended and then go back to mark different volunteers, the current volunteers will disappear from the attended volunteer list.
- **Implement**: Make the 'Attended' checkbox match between the 'Attended Volunteers' and the 'RSVP and Waitlist' dropdowns. Ex: If John Cox is in the RSVP list and was marked as 'Attended,' the 'Attended checkbox should be checked in both dropdowns.

**Mark volunteers from the RSVP list as attended, and they will appear in the attended volunteers list (they should be in both lists still).**

Image
 

**Check different volunteers as attended from the RSVP list and save:**

Image

**Notice that previous volunteers are gone:**

Image

**When a volunteer in the above pictures is marked as 'Attended' in the 'Attended Volunteers' dropdown, 'Attended' is not also checked in the 'RSVP and Waitlist' dropdown as we want it to be.**

JohnCox2211:
Fixes issue #1692 

Changes:
When you mark volunteers as attended from the 'RSVP and Waitlist', they will now be added to the attended list and will still have their 'Attended' checkbox checked. Adding this change also fixed users disappearing from the 'Attended' table when different users are checked in the 'RSVP and Waitlist' table.

Testing:
Create Event:
- Create an event and make sure you toggle 'This event requires RSVP'
- The event should be upcoming so that you can RSVP volunteers to move them between tables
- Add RSVP volunteers
- When the event starts, you can move them between boxes with the 'Attended' toggle

Example:
image

image

You can freely add/remove volunteers from the 'Attended' table with the checkbox now without wiping the table, and the checkboxes are consistent once you hit save.

### DIFF
--- a/app/controllers/admin/volunteers.py
+++ b/app/controllers/admin/volunteers.py
@@ -115,7 +115,6 @@ def volunteerDetailsPage(eventID):
     
     waitlistUser = list(set([obj for obj in eventRsvpData if obj.rsvpWaitlist]))
     rsvpUser = list(set([obj for obj in eventRsvpData if not obj.rsvpWaitlist ]))
-
     return render_template("/events/volunteerDetails.html",
                             waitlistUser = waitlistUser,
                             attendedUser= eventParticipantData,
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -546,9 +546,11 @@ def serviceTranscript(username):
     slCourses = getSlCourseTranscript(username)
     totalHours = getTotalHours(username)
     allEventTranscript = getProgramTranscript(username)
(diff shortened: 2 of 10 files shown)


Kafui123:
When I rsvp'd some students and the event started, they automatically showed up in the attended table without me clicking on the attended table.

bakobagassas (approved):
The UI and code look good now! Good Job!

PR #1699, opened 2026-04-27
Participation type

### LINKED ISSUE #1631: Include participation type in participation history
In participation history accordion on the user's page, there should be a participation type: volunteer, labor, attendee

MImran2002:
## Issue Description

Fixes #1631
- In participation history accordion on the user's page, there should be a participation type: volunteer, labor, attendee

## Changes

- Added participation type table in Sidebar > Search Student > User profile > participation History drop down

 image

- Refactored the getParticipatedEventsForUser(user) function to also add the column of event participant type and if the event is for volunteer the student participation type is volunteer, if the event is labor only it is labor else it is attendees.

- Rationale: This implementation can only be accurate if the eventParticiapant table has a new column which is participant type and we would need a script to populate it. However, this implementation is to give an approximation before the full implementation. 

## Testing

- Switch to the participationType - `git checkout participationType`
- Pull the changes - `git pull` & `git fetch origin`
- Reset the database - `database/reset_database.sh from-backup`
- Run the application - `flask run`
- Go to Sidebar > Search Student > User profile > participation History drop down to see the changes

### DIFF
--- a/app/logic/volunteerSpreadsheet.py
+++ b/app/logic/volunteerSpreadsheet.py
@@ -206,6 +206,32 @@ def termParticipation(term):
 
     return dict(programParticipationDict)
 
+def graduatingSeniorsVolunteerHours(academicYear):
+    columns = ["Full Name", "Email", "B-Number", "Unique Volunteer Semesters", "Total Volunteer Hours"]
+
+    currentSeniors = (EventParticipant
+                    .select(EventParticipant.user_id)
+                    .join(User).switch(EventParticipant)
+                    .join(Event)
+                    .join(Term)
+                    .where(Term.academicYear == academicYear, User.rawClassLevel.in_(["Senior", "Graduating"]), Event.isService == True,
+                            Event.deletionDate == None, Event.isCanceled == False))
+
+    query = (EventParticipant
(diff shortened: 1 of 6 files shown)


bakobagassas (commented):
We need to have an option for Volunteer too as per the issue description. I created an event that earned volunteer hours however the person was marked as an attendee only. 

Image

bakobagassas (commented):
I think you also need to figure out how to label people who attend events that are marked as both "Labor only" and "Earns volunteer hours". Are those people attendees? are they Volunteers ? or is it just Labor?

MImran2002:
> I think you also need to figure out how to label people who attend events that are marked as both "Labor only" and "Earns volunteer hours". Are those people attendees? are they Volunteers ? or is it just Labor?

Solved it

app/logic/volunteerSpreadsheet.py
```diff
@@ -206,6 +206,32 @@ def termParticipation(term):
 
     return dict(programParticipationDict)
 
+def graduatingSeniorsVolunteerHours(academicYear):
```
bakobagassas, line 209: Maybe I'm missing something but why do we need the function for graduating seniors hours? I thought we were looking at the event participation types for this PR

PR #1702, opened 2026-04-28
solved the responsiveness issue

### LINKED ISSUE #1689: Event Included Information Checkbox Toggles Do Not Work
Upon going to a created event and clicking to the Volunteer Details tab, there are multiple checkboxes that are meant to toggle the information that is visible on the page. These checkboxes are not toggling the information shown as they should. When unchecked, information still appears anyway.

Image

bakobagassas:
## Issue Description

Fixes #1689 
- Upon going to a created event and clicking to the Volunteer Details tab, there are multiple checkboxes that are meant to toggle the information that is visible on the page. These checkboxes are not toggling the information shown as they should. When unchecked, information still appears anyway.

## Changes

- arranged the mispeled ID and made getcheckboxes hid and show table cells not just cards
- rearranged the order of the checkboxes to match the table columns 

## Testing

The checkboxes are now working!
image

### DIFF
--- a/app/static/js/volunteerDetails.js
+++ b/app/static/js/volunteerDetails.js
@@ -5,6 +5,15 @@ $(document).ready(function () {
     }).get();
 	users = new Set(users);
 	users = [... users]
+	    const columnMap = {
+        phoneSelect: 1,
+        emailSelect: 2,
+        statusSelect: 3,
+        dietRestrictionSelect: 4,
+        emergencyContactSelect: 5,
+        insuranceSelect: 6,
+    };
+
 	$("#tableCardToggle").on('click', function () {
 		$("#volunteerInformationCardToPrint").toggle()
 		$("#volunteerInformationTableToPrint_wrapper").toggle()
--- a/app/static/js/volunteerDetails.js
+++ b/app/static/js/volunteerDetails.js
@@ -56,20 +56,22 @@ $(document).ready(function () {
 		});
 	}
 	function getCheckBoxes() {
-		$(".displayCheckbox").each(function () {
-			let checkboxId = this.id;
-			if ($('#' + checkboxId).is(':checked')) {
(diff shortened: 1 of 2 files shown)


MImran2002:
UI testing fixes:
 This is the first page we don't uncheck anything:
image

This is the second page and even without and box uncheck we can see the insurance column being hidden but taking up space.
image

This is the second page and if we unchecked the second box we see the information is hidden but the columns spaces are still there: 
image

If we unchecked more the second page move to the first page again and would break the view too:
image

bakobagassas:
@MImran2002 Good catch! Ready for another review again

MImran2002 (changes requested):
Working perfectly, the only fix left is the responsiveness of the table. I wanted to see if it is necessary and it the development it is responsive: 
image

But in the branch it is not working. This is the bug: in the first populated table when you first get there the responsiveness is good then you deselect the checkmarks and the table loses the columns and retain the width but when you add the checkmark again the responsiveness break.  I think the columns are added again on top of the total width breaking the responsiveness aside. This could exist in the development too because in the development the checkmark itself is breaking so this responsiveness bug was hidden.  Aside from this the code working:
image

potential solution could be to wrapped the table with a fixed responsive div and allow the table to expand and contract by 100%, this could be a solution there can be more solutions too.

bakobagassas:
For some reason I am not able to reproduce this bug on my branch. Did you pull the changes?

MImran2002 (approved):
the table's responsiveness breaks when the checkmarks are toggle the table remove the columns but when the checkmark are ticked back the action broke the table's responsiveness. This isn't a bug created by Sara but it was responsive bug that already exist but was not discover because the checkmark was broken. The fixed now is to have a wrapper around the table to maintain its responsiveness while allowing the table to be responsive within the wrapper.

BrianRamsay (approved):

PR #1703, opened 2026-05-02
Participationtypes

### LINKED ISSUE #1631: Include participation type in participation history

MImran2002:
## Issue Description

Fixes #1631
- In participation history accordion on the user's page, there should be a participation type: volunteer, labor, attendee

## Changes

- Added participation type table in Sidebar > Search Student > User profile > participation History drop down

 image

- Refactored the getParticipatedEventsForUser(user) function to also add the column of event participant type and if the event is for volunteer the student participation type is volunteer, if the event is labor only it is labor else it is attendees.

- Added test case for getParticipatedEventsForUser function and checked the returned participation types.

- Rationale: This implementation can only be accurate if the eventParticiapant table has a new column which is participant type and we would need a script to populate it. However, this implementation is to give an approximation before the full implementation. 

## Testing

- Switch to the participationType - `git checkout participationType`
- Pull the changes - `git pull` & `git fetch origin`
- Reset the test database - `database/reset_database.sh test`
- Test the case - `tests/run_tests.sh`
- Then reset the database - `database/reset_database.sh from-backup`
- Run the application - `flask run`
- Go to Sidebar > Search Student > User profile > participation History drop down to see the changes

### DIFF
--- a/app/logic/events.py
+++ b/app/logic/events.py
@@ -1,5 +1,5 @@
 from flask import  url_for, g, session
-from peewee import DoesNotExist, fn, JOIN
+from peewee import DoesNotExist, fn, JOIN, Case, Value
 from dateutil import parser
 from datetime import timedelta, date, datetime
 from dateutil.relativedelta import relativedelta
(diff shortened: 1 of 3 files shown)


bakobagassas (approved):
The code and the UI all look good to me! I am just still wondering about the Labor & Volunteer tag, as you cannot get paid for something you are volunteering for, but I guess that would be another issue. Good Job!

RueHaile (commented):
I see that we are using alias here which makes  sense but it makes it merely a virtual column what do you think about making it a database column? As it seems like a good plus. It could be a new pr idea. 
All looks good!

PR #1722, opened 2026-06-16
scaling issues on buttons on course proposals is fixed

### LINKED ISSUE #1514: Resolve breakpoint between buttons on Course Proposals
Resolve the breakpoint when the buttons resize so that it fits the screen real estate when going to course proposals and opening an existing draft. Changes seen when resizing the web page horizontally.

munsakad:
##### Issue Description

Fixes #1514 
The navigation buttons on the course proposal page were split across two separate divs: some on the left and some on the right, causing them to scale and align inconsistently depending on screen size.

### Changes Made
Moved all action buttons (Next, Save & Exit, Exit, Submit and Approve) into a single left-aligned div.
Removed ms-2 margin classes from buttons and replaced with a gap-2 utility on the container for consistent spacing.
Removed the now-empty right-side button div.
### Testing
Navigate to the course proposal page (/serviceLearning/proposal/new).
Resize the browser window to different screen sizes and confirm all buttons stay grouped and scale correctly.
Verify the "Submit and Approve" button still only appears for CELTS admins.
Confirm Previous, Cancel, Save & Exit, Exit, Next, and Submit and Approve buttons all function as expected.


MImran2002 (changes requested):
fix 1: Make sure the steps are coherent as in this is 
the bigger version: image

the medium: 
image

the small: 
image

as you see the medium is this way because in the original when the page reaches that medium size the buttons goes around it that is why the step dots collapsed into triangle. When this new implementation is done you can see the step dots no longer needs to be triangle anymore because in the small screen you are able to achieve this without turning the step dots into triangle.

MImran2002 (changes requested):
fix 2: This is on the universal UI principle that we always want next to be on the right and previous to be on the left not both previous and next to be in the same category.

app/templates/serviceLearning/slcNewProposal.html
```diff
@@ -25,24 +25,24 @@
   <div class="tab text-break">
       {% include "serviceLearning/slcQuestionnaire.html" %}
   </div>
-  <div class="row align-items-end pt-5 pb-5">
-    <div class="col col-xl col-12 d-flex gap-2 justify-content-xl-start justify-content-center">
+  <div class="row align-items-center pt-5 pb-5">
+    <div class="col col-xl col-12 d-flex flex-wrap gap-2 justify-content-xl-start justify-content-center mb-2 mb-xl-0">
```
MImran2002: Image

i talked to Scott and he told me the buttons need to be closer as you shrink the page and when it is shrink smaller the buttons should only be on top of the three dots but not on top and bottom of the three dots.

MImran2002 (changes requested):

MImran2002 (changes requested):
There are insuranceinfo.html, emergencyContactInfo.html and docker-compos.yml being changed which are not related to issue description. Please remove them and if they are really required.

fritzj2:
The problem seems to be fixed by your changes; however, there are three extra files that don't address this issue. Remove them from the staged changes, and it should be good. app/templates/main/emergencyContactInfo.html, app/templates/main/insuranceInfo.html, and .devcontainer/docker-compose.yml seem not to be related to this issue.

NYABUTOA (dismissed):
This looks good. John and I have reviewed your work. The logic is solid, and placing the Next button on the right with the others on the left makes sense. It's more visible and keeps the page more organized.

johnlolonga19 (commented):
The course proposal page looks good as far as the buttons don't overlap and stay in the same line when scaling the page

MImran2002 (commented):
talked with Brian and Scott, implement the buttons as in if the button text is unnecessarily long reduce it without losing the semantic or tyry findings was to make the button more good looking study other forms like google form.

MImran2002 (changes requested):
Image

I think this is the last fix as you can see the buttons looks good awesome. but when the button stack with each other you can see the one with the left buttons are on top and the right ones are at the button in between them is extra space. With this i think everything looks good. I know this is going back and forth but this is software development and these PR processes would sometimes take days and weeks.

MImran2002 (changes requested):
Image
and the space between cancel and "save and exit" are not consistent with other button spacing on the left in other pages. This should be fixed too.

MImran2002 (dismissed):
Fixes: We generally don't create function unless it is truly logic and cannot be handle by html.
the hidebutton and showbutton are functions we won't need because there are build in functions for those. So I would not have those functions.

Fixes: The reason the spacing among the buttons are there is because eventhough for example $("#nextButton").hide(); hides the button it doesn't hide the div that wraps around the button.

A possible solution is to assign ids to those div that wrap around the buttons. Then use those ids to hide the div (if you hide the div you also hide the button nested in it) by adding a bootstrap class that is similar to display:none to hide it or remove the bootstrap class to make it appear. 

In this way you are not creating functions yet are using build in methods. 
Link to the resources for the classes and methods are here:
https://www.w3schools.com/jquery/html_addclass.asp
https://www.w3schools.com/jquery/html_removeclass.asp
https://getbootstrap.com/docs/4.0/utilities/display/

BrianRamsay (changes requested):
The first button is still not aligned properly.

Screenshot 2026-06-22 at 11 00 59 AM

app/static/css/slcNewProposal.css
```diff
@@ -26,4 +26,14 @@
...
 .step.finish {
   background-color: #0d6efd;
-}
\ No newline at end of file
+}
+
+@media (max-width: 767.98px) {
+  #proposalButtonControls {
+    --bs-gutter-y: 0.5rem;
+  }
```
BrianRamsay, line 34: Does this rule do anything? If I remove it I don't see a change

BrianRamsay (changes requested):
The first button is still not aligned properly.

Image

BrianRamsay (approved):

PR #1724, opened 2026-06-16
Admin tab intuitiveness

### LINKED ISSUE #1595: Admin Tab on the sidebar is unintuitive
During the usability study, for the task " Add a student as a CCE Minor Candidate", participants often weren't able to recognise that the "Admin" tab on the left sidebar of the webpage was a drop-down. Make it more intuitive for the user to understand it is a dropdown that they can use. One idea is to make it blue when it is hovered over so that it seems more clickable, but you can come up with another solution

nahom70:
## Issue Description

Fixes #1595 issue number
- The issue was that the admin tab was not intuitive and lacked clarity on whether it was a dropdown.

## Changes
* Added a hover on the admin tab to make it look different from the other tabs


MImran2002 (changes requested):
from design perspective I would want you to have the same indication with the course proposal blue color and length.

Image

MImran2002 (changes requested):
Image
as you can see here it is a bit of sticky situation because the proposal is blue because the main content page is rendering proposal which is valid and the admin drop down lights up and continues to lights up when hover and also when clicked: now this is something to consider the sidebar tabs only light up blue when the main content area renders the tab's information. that means admin has nothing to render it is just a dropdown that is why admin should light up blue when hover (which is what the issue description is) but should not stay blue when clicked

MImran2002 (approved):
looks good!

PR #1725, opened 2026-06-16
toggle name change

johnlolonga19:
## Issue Description

Fixes #add issue number
- We had a wording issue, and what we had to do was change the toggle wording so it's more direct.

## Changes

- changed the wording from "labor only" to "includes labor" so it's more straightforward to whoever is using the platform.

## Testing

- You can go to the templates drop-down
- go to events there, after go to createevent.html, and this way you can access the wording that we changed.


brightfietsop-ux:
Good job, that was a small yet necessary change

brightfietsop-ux (approved):

MImran2002 (approved):
looks good

PR #1727, opened 2026-06-16
exception handling in slcManagement.js

### LINKED ISSUE #1535: Handle the exception in the ajax request in the withdraw function slcManagement.js file
- currently the exception does not inform the user that an exception was thrown. 
- update the exception in the ajax request so that it informs the user rather than console.log

munsakad:
## Issue Description

Fixes #1535 
-The withdraw function in slcManagement.js was silently swallowing errors, so users had no way of knowing when a course withdrawal failed.
## Changes

Added a user-facing alert in the error handler of the withdraw ajax request so the user is notified when something goes wrong.
Fixed done to complete so the select reset actually runs after the request finishes.

## Testing
Open the Manage Service Learning page.
Select "Withdraw" on a course to trigger the withdraw modal.
Click the withdraw button while the server is unreachable or returning an error.
Confirm that an alert appears informing the user that the withdrawal failed instead of failing silently.


BhushanSah (dismissed):

ArtemKurasov (dismissed):
The warning seems to be displayed correctly

fritzj2 (changes requested):
The warning appears to display correctly; however, it is confusing to read. Change it to an if/else statement for clarity. Check msgFlash() instead of alert for consistency across other files.

zhytkovd (dismissed):
The msgFlash message shows under the pop-up window

fritzj2 (dismissed):
These changes make the code significantly easier to read and modify. The changes also allow the code to continue working as intended.

MImran2002 (changes requested):
The issue has been fixed in the backend so all you would need to do is removed the console.log as the issue described because the error messaging is catching from the backend already so if there is an error it will be highlighted from the backend already. The request.response.message you implement can only work if the return from withdrawCourse return and error with a message only then the logic you wrote in the code will work.

So for now I would suggest you to remove the console.log.

BhushanSah (dismissed):
I tested this locally, and it works as expected.

The change improves the withdraw error handling by showing a user-facing flash message instead of only logging the error in the console. I also like that it checks for request.responseJSON.message and falls back to a clear default message when the server does not return one.

Using complete here also makes sense, since the dropdown reset should occur after the request finishes, whether it succeeds or fails.

No issues from my testing. Approved.

app/static/js/slcManagement.js
```diff
@@ -140,10 +139,13 @@ function withdraw(){
       location.reload();
     },
     error: function(request, status, error) {
-        console.log(status,error);      
-        
+      let message = "An error occurred while withdrawing the course. Please try again.";
+      if (request.responseJSON && request.responseJSON.message) {
+        message = request.responseJSON.message;
+      }
+      msgFlash(message, "danger");
```
BrianRamsay: The changes we talked about for this file were either not added or not completed.

BrianRamsay (changes requested):

app/controllers/serviceLearning/routes.py
```diff
@@ -197,20 +197,22 @@ def updateInstructorPhone():
     instructorData = request.get_json()
     (User.update(phoneNumber=instructorData[1])
         .where(User.username == instructorData[0])).execute()
-    return "success"
+    return "success", 200
```
BrianRamsay: The 200 status is implied, and we don't add it for our other successful returns

app/static/js/slcManagement.js
```diff
@@ -125,7 +125,6 @@ function renew(){
         location = '/serviceLearning/editProposal/' + newID;
       },
       error: function(request, status, error) {
-          console.log(status,error);
```
BrianRamsay, line 128: We need to reload the page here as well, or else the flash messages from the backend aren't getting shown.

app/static/js/slcManagement.js
```diff
@@ -140,10 +139,14 @@ function withdraw(){
       location.reload();
     },
     error: function(request, status, error) {
-        console.log(status,error);      
-        
+      let message = "An error occurred while withdrawing the course. Please try again.";
+      if (request.responseJSON && request.responseJSON.message) {
+        message = request.responseJSON.message;
+      }
```
BrianRamsay: We are not passing JSON ever, and are catching all exceptions, so there won't ever be a message coming to use here.

app/static/js/slcManagement.js
```diff
@@ -140,10 +139,14 @@ function withdraw(){
...
+      msgFlash(message, "danger");
```
BrianRamsay: Let's not do a flash from javascript here. Reloading the page will allow the flashes from the backend to show, so we don't need another.

BrianRamsay (changes requested):

app/controllers/serviceLearning/routes.py
```diff
@@ -205,12 +205,14 @@ def withdrawCourse(courseID):
         if g.current_user.isAdmin or g.current_user.isFaculty:
             withdrawProposal(courseID)
             flash("Course successfully withdrawn", 'success')
+            return ""
         else:
             flash("Unauthorized to perform this action", 'warning')
     except Exception as e:
```
MImran2002, line 212: if we are adding server codes return "", 500 should be nested in the exception or else and internal server outside of the exception doesn't make sense for me if exception is catching the errors. What do you think @BrianRamsay ? Should the other catches also consider 200 and 403?

MImran2002 (changes requested):

MImran2002 (dismissed):
Have deleted the console.log, create the reload as Brian said and ensure http error code are part of the return

BrianRamsay (approved):

sheggen (approved):

Issue #1596, opened 2025-07-30
Back Button Implementation


Jsmi8334:
During the Usability study, participants had difficulty going back when they wanted to navigate the page. We should make "back" or "previous" buttons to make it easier to navigate the website without having to press the main tabs again. For example, when you go to the "create events" page and you click on an event type, you are not able to go back to the event type page. Another example is that if you search up a student on "student search" and click on the student's name, you are not able to press a back button to go back to the search bar.

Karina-Agliullova:
https://trello.com/c/BS6uKHhE/974-back-button-creation

PR #1735, opened 2026-06-18
Daniel back buttom

DanielRukwasha:
## Back Button Implementation
Fixes issue #1596
- The student search tab page and event tab did not have back button once inside the page.

## Changes

- We first create a back button at the end of the page next to the save button but replicate basic button creation but we added the javascript:window.history.back 
image
- we added a jinja line of code where we had a back button on the profile page to only appear  the admi and the volunteer are different, 

## Testing

-we checked through the website using inspect, and manupilate tabs.

### DIFF
--- a/app/templates/events/createEvent.html
+++ b/app/templates/events/createEvent.html
@@ -58,6 +58,7 @@
       {% else %}
         <div class="text-center">
           <h1 id="pageTitle">{{page_title}}</h1>
+      
         </div>
       {% endif %}
 
@@ -312,8 +313,11 @@
   </div>
   <div class="row justify-content-left mt-2">
     <div class="col-md-6 offset-md-3">
-      <div class="text-center mb-2">
+      <div class="w-100 m-auto">
         <!--Submission buttons-->
+            {% if not event %}
+        <a href="javascript:window.history.back();" class="btn btn-secondary" btn-sm m-2> Back</a>
+         {% endif %}
         {% if event %}
         <button type="button" class="btn btn-danger" data-bs-toggle="modal" data-bs-target="#deleteWarning">Delete
           Event</button>
@@ -326,7 +330,7 @@
         <button type="button" class="btn btn-success" data-bs-toggle="modal" data-bs-target="#renewWarning">Renew
           Event</button>
         {% endif %}
-        <input type="button" class="btn btn-primary" id="saveButton" value="Save" />
+            <input type="button" class="btn btn-primary" id="saveButton" value="Save" /> 
       </div>
     </div>
   </div>
--- a/app/templates/events/createEvent.html
+++ b/app/templates/events/createEvent.html
@@ -311,12 +311,12 @@
       {% endif %}
     </div>
   </div>
-  <div class="row justify-content-left mt-2">
+  <div class="row justify-content-center mt-2">
     <div class="col-md-6 offset-md-3">
       <div class="w-100 m-auto">
         <!--Submission buttons-->
             {% if not event %}
-        <a href="javascript:window.history.back();" class="btn btn-secondary" btn-sm m-2> Back</a>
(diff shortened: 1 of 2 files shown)


BrianRamsay (changes requested):
The buttons are no longer centered on the page at the bottom of the form.

DanielRukwasha:
> The buttons are no longer centered on the page at the bottom of the form.

Resolved Back Button

app/templates/events/createEvent.html
```diff
@@ -312,8 +313,11 @@ <h1 id="pageTitle">{{page_title}}</h1>
   </div>
   <div class="row justify-content-left mt-2">
     <div class="col-md-6 offset-md-3">
-      <div class="text-center mb-2">
+      <div class="w-100 m-auto">
         <!--Submission buttons-->
+            {% if not event %}
+        <a href="javascript:window.history.back();" class="btn btn-secondary" btn-sm m-2> Back</a>
```
MImran2002: I feel this violated the separation of concern that css should not be in HTML, javascript should not be in HTML. Instead because this back button we clearly knows where it will go we can use this href="/something" for example as we know the route.

app/templates/main/userProfile.html
```diff
@@ -663,6 +663,13 @@ <h6 align="left" class="d-inline-flex" id="unbanVolunteerEndDate">End Date: &nbs
 <!-- ################# Ban or Unban Modal ################ -->
 
 
+{% if volunteer != g.current_user %}
+<div class="text-center my-3">
+  <a href="javascript:window.history.back();" class="btn btn-secondary">Back</a>
```
MImran2002: same here too this violated the separation of concern

MImran2002 (changes requested):
branch: 
Image
still not centered

development:
Image

app/templates/events/createEvent.html
```diff
@@ -58,6 +58,7 @@
       {% else %}
         <div class="text-center">
           <h1 id="pageTitle">{{page_title}}</h1>
+
```
MImran2002, line 74: remove this space.

app/templates/events/createEvent.html
```diff
@@ -310,10 +311,13 @@ <h1 id="pageTitle">{{page_title}}</h1>
       {% endif %}
     </div>
   </div>
-  <div class="row justify-content-left mt-2">
+  <div class="row justify-content-center mt-2">
     <div class="col-md-6 offset-md-3">
-      <div class="text-center mb-2">
+      <div class="w-100 m-auto">
         <!--Submission buttons-->
+            {% if not event %}
```
MImran2002, line 333: This only show up when the event is create how about when it expired or the event is not expired but has not completed yet. In this cases I think there should be aback button because the issue description talks about this process in general not just specifically for the process of creating event.

MImran2002 (changes requested):

app/templates/events/createEvent.html
DanielRukwasha: Done

MImran2002:
Tell me when you guys are done and this issue as brian has commented means you also need to get his approval

MImran2002:
@nahom70 and @DanielRukwasha reach out to Brian so that you can get his approval and be able to merge this issue. if we don't proactively do this it becomes an outstanding PR.

fritzj2:
image

Fixed the centering on the `userProfile.html` page.

bakobagassas (changes requested):
We will also need the back button when we view an event that has already been created. Currently we only have the back button for events we are creating (good job on that by the way!) but we do not have the back button for events we are just viewing.

fritzj2:
Added the button into the view events page

BrianRamsay (approved):

Issue #1405, opened 2025-02-07
Add a CCE Minor Note feature


bledsoef:
Add a CCE Minor Note feature attached to the Notes section in the User profile

Karina-Agliullova:
https://trello.com/c/jrlmeR08/1012-add-a-cce-minor-note-feature

PR #1737, opened 2026-06-20
Add a CCE Minor Note feature attached to the Notes section in the User profile

munsakad:
## Issue Description

Fixes issue #1405 
- The user profile Notes section had no way to flag a note as related to the CCE Minor, even though the profile already tracks CCE Minor interest. This adds a "CCE Minor" note tag alongside the existing Bonner tag.

## Changes

- Added a "Related to CCE Minor" toggle to the Add Note modal, shown only when the user has a CCE Minor interest (volunteer.minorInterest).
-Display a "CCE Minor" prefix in the note's Visible To column, mirroring the existing Bonner prefix convention.
-Pass the flag through the Edit button via a data-cceminor attribute so the modal can pre-fill it.
-All changes are confined to app/templates/main/userProfile.html.

## Testing

Open a profile for a user with CCE Minor interest and click Add Note, confirm that the "Related to CCE Minor" toggle appears.
For a user without CCE Minor interest, confirm the toggle is hidden.
Verify a note flagged as CCE Minor shows the "CCE Minor" prefix in the Visible To column.
Confirm that the Bonner toggle and existing note behavior are unchanged.


DanielRukwasha:
Overall Impression

  The feature makes sense and the approach of mirroring Bonner tag behavior is a smart, consistent design decision.
  However, 

  The template references row.isCCEMinorNote and submits a cceMinor form field, but there are no backend changes in this
 My question will be 

  - Does the Note model already have an isCCEMinorNote field? If not, this will silently fail (Jinja will render it as
  False/empty with no error).
  - Is there a route handler that reads the cceMinor POST parameter and saves it?
e

DanielRukwasha (commented):
Overall Impression

  The feature makes sense and the approach of mirroring Bonner tag behavior is a smart, consistent design decision.
  However, 

  The template references row.isCCEMinorNote and submits a cceMinor form field, but there are no backend changes in this


  - Does the Note model already have an isCCEMinorNote field? If not, this will silently fail (Jinja will render it as
  False/empty with no error).
  - Is there a route handler that reads the cceMinor POST parameter and saves it?

MImran2002 (changes requested):
These are the changes or the bugs that I encounter:
Fix 1: I think this issue alone should not be solved with a HTML. This is broader because this could potentially include javascript and python. (Hint: look at how bonner notes work)

Fix 2: Because the code only holds html I am not seeing any cce minor flags from my side: 
(These three notes are created by toggling cceminor)

Image

Fix 3: When I toggle the CCEminor and save the note, just like the bonner if i were to create a bonner note and go back and look i will see the model has bonner note toggle
bonner example:

Image

cceminor: 

Image

if its cceminor the toggle should be toggled to show it is working. 

Fix 4: In an instance where both the bonner and cce minor are toggle the bonner notes accordion seems to overwrite the cceminor notes and the note only appear in bonner this scenario should be considered as in should the note that have both toggles need to appear on both or one of them.

MImran2002:
@munsakad and @brightfietsop-ux after the department issue you guys worked on, while waiting for that issue to be reviewed and merged. Worked on this as we don't want this to be dragged into the weeks ahead.

brightfietsop-ux:
## What i discovered before coding
- Curently when  a note was added without a toggle, the note gets saved.
- When the note is added with toggle related to CCE minor, the note gets saved, when you try to edit, the toggle doesn’t display that it was toggled in the first place
- when the note is added by toggling, related to bonner, it does not even gets saved in the first place, which is the same thing when both gets toggled.
 
## Summary

This PR adds support for marking profile notes as CCE Minor notes alongside the existing Bonner note option.

## Changes

- Added `isCCEMinorNote` to the `ProfileNote` model.
- Added a CCE Minor toggle to the Add/Edit Note modal.
- Updated the profile note routes and logic to save and edit the CCE Minor value.
- Updated the JavaScript to restore the Bonner and CCE Minor toggles when editing a note.
- Updated the Notes table to display the correct note visibility.
- Updated the database reset process so databases restored from `prod-backup.sql` include the new `isCCEMinorNote` column.
- Preserved the existing behavior where Bonner notes are visible to Bonner Scholars.

## Testing

Tested the following note combinations:

- Regular profile note
- Bonner note
- CCE Minor note
- Note marked as both Bonner and CCE Minor
- Editing an existing note and restoring its toggle values
- Deleting a note
- Resetting the database using:
  - `database/reset_database.sh test`
  - `database/reset_database.sh from-backup`

Verified in MySQL that:

- `isBonnerNote` saves as `0` or `1`
- `isCCEMinorNote` saves as `0` or `1`
- Note content, creator, visibility, and creation date are saved correctly

app/controllers/main/routes.py
```diff
@@ -378,21 +378,64 @@ def eventTravelForm(eventID):
...
-    This function adds a note to the user's profile.
-    """
     postData = request.form
+
+    noteTextbox = postData.get("noteTextbox", "").strip()
+
+    if not noteTextbox:
+        return "Note cannot be empty", 400
+
+    addProfileNote(
```
brightfietsop-ux: This does not handle a failed note insert. For example, if the user doesn't exist (e.g., postData["username"] isn't a real user), this function fails, and the route fails. Handle the failure using try: except:. 

The logic function should `raise` and the route function should try/except.

app/controllers/main/routes.py
```diff
@@ -378,21 +378,64 @@ def eventTravelForm(eventID):
...
+
+    return "success"
+    
+@main_bp.route("/<username>/editNote", methods=["POST"])
+def editProfileNote(username):
+    profileNoteID = request.form.get("id")
+    noteTextbox = request.form.get("noteTextbox", "").strip()
+    visibility = request.form.get("visibility", 1)
+    bonner = request.form.get("bonner") == "yes"
+    cceMinor = request.form.get("cceMinor") == "yes"
```
brightfietsop-ux: Both this and the one above can be a single function that reforms the request data. use (optional?) input parameters to handle the slight differences.

app/logic/users.py
```diff
@@ -124,19 +124,56 @@ def getUserBGCheckHistory(username):
...
     return bgHistory
 
-def addProfileNote(visibility, bonner, noteTextbox, username):
+def addProfileNote(
+    visibility,
+    bonner,
+    cceMinor,
+    noteTextbox,
+    username,
+):
```
brightfietsop-ux: make a single line since it is a function definition (not a dictionary or some other datatype)

app/static/js/userProfile.js
```diff
@@ -157,136 +157,234 @@ $(document).ready(function(){
...
-        $.ajax({
-            method: "POST",
-            url: "/" + username + "/deleteNote",
-            data: { "id": noteid }
-        })
+    const requestData = {
+        username: username,
+        visibility: visibility,
+        noteTextbox: noteTextbox,
+        bonner: isBonner ? "yes" : "no",
```
brightfietsop-ux: Is this ever used as "yes" or "no" or only as a boolean (which it is used as a boolean in the route).

app/static/js/userProfile.js
```diff
@@ -157,136 +157,234 @@ $(document).ready(function(){
...
-    });
-  });
 
-  $(".deleteNoteButton").click(function() {
-    $("#confirmDeleteNote").data('username', $(this).data('username'))
-    $("#confirmDeleteNote").data('noteid', $(this).data('noteid'))
-    $("#deleteNoteWarning").modal("show")
-  
-  });
+    saveButton.prop("disabled", true);
```
brightfietsop-ux, line 255: probably doesn't do anything

database/reset_database.sh
```diff
@@ -39,10 +39,12 @@ mysql -u root -proot --execute="CREATE DATABASE IF NOT EXISTS \`celts\`; CREATE
 rm -rf migrations
 rm -rf migrations.json
 
-echo -n "Creating database objects"
 if [ $BACKUP -eq 1 ]; then
     echo " from backup"
     mysql -u root -proot celts < prod-backup.sql
+    mysql -u root -proot -NBe "SELECT COUNT(*) FROM information_schema.COLUMNS WHERE TABLE_SCHEMA='celts' AND TABLE_NAME='profilenote' AND COLUMN_NAME='isCCEMinorNote';" |
+        grep -q '^1$' ||
+        mysql -u root -proot celts -e "ALTER TABLE profilenote ADD COLUMN isCCEMinorNote TINYINT(1) NOT NULL DEFAULT 0 AFTER isBonnerNote;"
```
brightfietsop-ux: Remove this from the PR, should not be here. For local testing only .

app/static/js/userProfile.js
brightfietsop-ux: The checkbox values are booleans in JavaScript. They are serialized as
"yes" or "no" because the request is sent as form data, and Flask's
request.form returns strings. The route converts those strings back to
booleans before passing them to the logic layer. This also follows the
existing Bonner note behavior.

app/static/js/userProfile.js
brightfietsop-ux: The line disables the Save button while the AJAX request is running, helping prevent duplicate note submissions. The error callback re-enables it:

app/controllers/main/routes.py
```diff
@@ -378,21 +378,66 @@ def eventTravelForm(eventID):
                            userList = userList,
                            )
 
-@main_bp.route('/profile/addNote', methods=['POST'])
+def getProfileNoteData(include_username=False, include_id=False):
```
MImran2002: if this is a function this should be in logic and should be tested

MImran2002: we use camelCase

app/controllers/main/routes.py
```diff
@@ -378,21 +378,66 @@ def eventTravelForm(eventID):
...
     try:
-        note = addProfileNote(postData["visibility"], postData["bonner"] == "yes", postData["noteTextbox"], postData["username"])
-        flash("Successfully added profile note", "success")
-        return redirect(url_for("main.viewUsersProfile", username=postData["username"]))
-    except Exception as e:
-        print("Error adding note", e)
-        flash("Failed to add profile note", "danger")
+        noteData = getProfileNoteData(include_username=True)
+        addProfileNote(**noteData)
+
```
MImran2002: make sure the flash message appears

app/logic/users.py
```diff
@@ -124,19 +124,38 @@ def getUserBGCheckHistory(username):
...
-                            createdOn = datetime.datetime.now(),
-                            noteContent = noteTextbox,
-                            noteType = "profile")
-    createProfileNote = ProfileNote.create(user = User.get(User.username == username),
-                                           note = noteForDb,
-                                           isBonnerNote = bonner,
-                                           viewTier = visibility)
-    return createProfileNote
+    noteForDb = Note.create( createdBy=g.current_user,  createdOn=datetime.datetime.now(),  noteContent=noteTextbox, noteType="profile", )
+    profileNote = ProfileNote.create( user=user, note=noteForDb, isBonnerNote=bonner,  isCCEMinorNote=cceMinor, viewTier=int(visibility),  )
```
MImran2002: changing the datatype this is great but make sure this happens outside the peewee command

MImran2002 (changes requested):

MImran2002:
Also your is only for the test data if you want to use the production data before after ./database/run_database.sh from-backup , you should log into mysql and enter ALTER TABLE profilenote
ADD COLUMN isCCEMinorNote BOOLEAN NOT NULL DEFAULT FALSE; then flask run to use on production data

tests/code/test_users.py
```diff
@@ -136,22 +136,35 @@ def test_removeUserInterestt():
         assert result == True
 
         transaction.rollback()
-
+@pytest.mark.integration
+def test_getProfileNoteData():
+    formData = { "visibility": "3","bonner": "yes", "cceMinor": "no",  "noteTextbox": "  Test profile note  ",  "username": "ramsayb2",  "id": "12", }
+    noteData = getProfileNoteData(  formData, includeUsername=True, includeId=True, )
+    assert noteData == {"visibility": 3,"bonner": True,   "cceMinor": False,"noteTextbox": "Test profile note", "username": "ramsayb2", "profileNoteID": "12",  }
+    invalidData = [
```
MImran2002, line 185: this is good error dummy test except you need to do assert to see if the function handles those error

app/controllers/main/routes.py
```diff
@@ -378,21 +378,46 @@ def eventTravelForm(eventID):
...
         flash("Failed to add profile note", "danger")
         return "Failed to add profile note", 500
+    return "success"
 
+@main_bp.route("/<username>/editNote", methods=["POST"])
+def editProfileNote(username):
+    try:
+        noteData = getProfileNoteData( request.form,includeId=True, )
+        profileNote = ProfileNote.get_by_id( noteData["profileNoteID"] )
+        if profileNote.user.username != username:
```
MImran2002: combine them together with or

MImran2002 (changes requested):

app/controllers/main/routes.py
```diff
@@ -378,21 +378,46 @@ def eventTravelForm(eventID):
...
+            abort(403)
+        if ( profileNote.note.createdBy != g.current_user and not g.current_user.isCeltsAdmin ):
+            abort(403)
+        updateProfileNote(**noteData)
+        flash("Successfully updated profile note", "success")
+    except ValueError as error:
```
MImran2002: remove those two errors value error and profile note as for getprofilenote it handles most error and we only care about the overall error here. this also means the same as well on addNote

MImran2002 (changes requested):

MImran2002 (dismissed):
approve with the condition the value and another error message is remove

MImran2002 (dismissed):

MImran2002 (changes requested):
Image eventhough the user is CCEminor we still don't see the the cceminor option

MImran2002 (commented):
it need to be updated from base branch

app/controllers/main/routes.py
```diff
@@ -39,7 +39,7 @@
 from app.logic.landingPage import getManagerProgramDict, getActiveEventTab
 from app.logic.minor import toggleMinorInterest, declareMinorInterest, getCommunityEngagementByTerm, getEngagementTotal
 from app.logic.participants import hasGoneToTraining, unattendedRequiredEvents, getParticipationStatusForTrainings, checkUserRsvp, addPersonToEvent
-from app.logic.users import addUserInterest, isBannedFromEvent, removeUserInterest, banUser, unbanUser, isEligibleForProgram, getUserBGCheckHistory, addProfileNote, deleteProfileNote, updateDietInfo, trainedParticipants
+from app.logic.users import  addUserInterest, removeUserInterest, banUser, unbanUser, isEligibleForProgram, getUserBGCheckHistory,addProfileNote, deleteProfileNote, updateDietInfo, updateProfileNote, getProfileNoteData, isBannedFromEvent, trainedParticipants
```
MImran2002: from app.logic.users import * would be better here at this point where the app.logic.users have 16 functions and 13 are imported here

MImran2002 (changes requested):

MImran2002 (approved):
Looks good, made sure the prod-backup works too, found a bug and has a new issue created on it. The cross bonner and cce minor not is also tested.

MImran2002 (approved):

Issue #1739, opened 2026-06-26
User will be required to read the handbook to continue participating in CELTS


MImran2002:
CELTS department wants to make sure students have read their handbook. Those who have read their handbook put their signature underneath the user profile page as an indication that they have finished. This will be done annually. if they don't they wont be able to earn neither service learning hours nor labor hours.

Design Consideration:
The signature should be in UserProfile page.
Should be happening annually.
Ways to either remind students on finishing the requirement.
Ways to prevent further activities on students who didn't follow the requirement.

(will have more update on July 1)

PR #1756, opened 2026-07-10
Built signature page and updated profile to show signature status

sheggen:
## Issue Description

Fixes issue #1739

## Changes
**Signatures:**
- Adds a signature field to each user's profile. The signature field is only accessible to:
  - the user whose profile you are viewing (i.e., no one can sign on someone else's behalf).
  - the user if they are attending or have attended All volunteers training (for non-labor students)
  - the user if they are attending or have attended All CELTS training (for labor students)
  - the CELTS staff have uploaded the appropriate handbook for each case above
- Only one signature is needed for an academic year
- Program Managers can view signature statuses in Rosters (see below)

**Rosters**
- Program Managers and admins can view rosters for participants in their program:
  - All interested volunteers (checked the box in their profile or in the extravaganza page)
  - All active volunteers (have earned service hours in their program)
  - All previous year's volunteers (earned service hours last AY)
- Rosters can be exported to Excel

**Volunteer Extravaganza**
- Added a page for the volunteer Extravaganza (/extravaganza). 
- Added a QR and link to the extravaganza in the "Create Event" templates.

**Term Management**
- CELTS admins can now upload:
  - All Volunteers Handbook (for all volunteers)
  - All CELTS Handbook (for labor)
  - The Background check form (which students will need to submit to Sheila in paper format, due to SSN and other private info we do not want to store)
- All three files above are stored for an AY, depending on the selected term (i.e., If the form is uploaded in Summer 2026, it applies to Fall 25, Spring 26, and Summer 26)

**Event Templates**
- Added two new event templates: All CELTS training and Weekly Labor Meeting

**Event Flags**
- Events now have a more robust way of ensuring we don't have combinations that are not allowed. 
- On production, the `fix_event_flags.py` script should be run to fix old events. This will require some manual attention, as the fix is only applied to events I have confirmed with CELTS. There may be more that need fixed in the production data. 

## Testing

**Signature Workflow**

- Reset your database (test) and run the app.
- Set the user to an admin or Program Manager
- View your profile. You should _not_ be able to sign anything yet.
- Set the current term, then upload a handbook (all volunteer or celts labor).
- Create a new All Volunteers/All CELTS labor event in the future. RSVP a user. Check that the user cannot sign the handbook yet (as the event hasn't happened yet)
- Move the event into the present or past. Mark the user as "attended". Try signing again (they should be able to sign now).
- After the handbook is signed, test that it persists for the current academic year, and then resets during the following AY.

**Rosters**
- Pick a program to test against. Mark a user as "interested" in the program. Check the roster as a PM and as an admin. 
- Add an event to the program. Add the user to the event. Check the roster (specifically, the "Engaged Volunteers" section).
- Add an event to the program in the previous academic year, and add a user to that event. Check the roster (specifically, the "Last Year's Volunteers" section). 
- Export the rosters and confirm the data matches on the spreadsheet. 

**Event Templates**
- Create events using each template. In particular, test recurring events for labor meetings.

## Remaining Tasks
- Prevent a user from participating in an event _that earns service hours_ if they have not completed the four requirements (signed the handbook, attended AVT, attended PST, and have submitted their background check form).
- How do we handle RSVPs vs. attended in the rosters?
- New functions need tests written
- How does /extravaganza look on a mobile, since that is how most will use it on that day?
- In the "Engaged Volunteer" and "Last Year Volunteer" rosters, add columns for # events they attended, as well as if they were trained (training matter for the case were a student did all the trainings but could not participate in the actual events).


app/logic/volunteerSpreadsheet.py
sheggen, line 1: `makeDataXls` is critically broken. It works for the new rosters but breaks for all Reports.

BrianRamsay (approved):

Issue #1743, opened 2026-07-02
Add Education & Training section to the Service Transcript


MImran2002:
Description: The Service Transcript should organize the non-service participation. So we should add a separate section titled Education & Training. Then underneath this category we will display attendance for educational and engagement events that the participants attended but receive 0 service hours.

PR #1760, opened 2026-07-16
Separate zero-hour events into Education and Training section

BhushanSah:
## Issue Description

Fixes issue #1743 
The Service Transcript was showing zero-hour participation under Volunteer Hour Details. These events are not service-hour records, so they should be separated from volunteer/service hours.

## Changes

- Added a new section titled **Education & Training**.
- Moved zero-hour educational and engagement event attendance into the new section.
- Kept zero-hour events separate from Volunteer Hour Details.
- Continued excluding Bonner Scholars from the displayed transcript sections.
- Updated Volunteer Hour Details so only events with earned service hours are shown there.

image


MImran2002 (approved):
looks good

PR #1765, opened 2026-07-16
solved issue service hours not being displayed in participation history

conwelld:
## 

Fixes issue #<1742>
- there was an issue in celts where service hours were not being displayed in participation history, making it streneous for the user to find their hours and they had to go to the service transcript

## Changes

- added peewee query to find the hours for the specified event, by ascending order 
- 
image
image

## Testing

- reset the databse and ran it with no issues
-

### DIFF
--- a/app/controllers/main/routes.py
+++ b/app/controllers/main/routes.py
@@ -233,7 +233,10 @@ def viewUsersProfile(username):
         managersProgramDict = getManagerProgramDict(g.current_user)
         managersList = [id[1] for id in managersProgramDict.items()]
         totalSustainedEngagements = getEngagementTotal(getCommunityEngagementByTerm(volunteer))
-        eventparticipant = list(EventParticipant.select(EventParticipant.hoursEarned).join(Event).where(EventParticipant.user == volunteer, EventParticipant.event == Event.id,).order_by(Event.id.asc()))
+        eventParticipant = list(EventParticipant.select(EventParticipant.hoursEarned)
+                                                .join(Event)
+                                                .where(EventParticipant.user == volunteer, EventParticipant.event == Event.id,)
+                                                .order_by(Event.id.asc()))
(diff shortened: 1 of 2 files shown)


app/controllers/main/routes.py
```diff
@@ -233,6 +233,7 @@ def viewUsersProfile(username):
         managersProgramDict = getManagerProgramDict(g.current_user)
         managersList = [id[1] for id in managersProgramDict.items()]
         totalSustainedEngagements = getEngagementTotal(getCommunityEngagementByTerm(volunteer))
+        eventparticipant = list(EventParticipant.select(EventParticipant.hoursEarned).join(Event).where(EventParticipant.user == volunteer, EventParticipant.event == Event.id,).order_by(Event.id.asc()))
```
BetterRocket: Looking at other long .select() queries, like on line 52, 204, 209, 334, and 363, it seems like it is preferable for the .join(), .where() and other such clauses to be on new lines for readability.

fritzj2: `eventparticipant` Variable is not in camel case

fritzj2 (changes requested):

BetterRocket (changes requested):

MImran2002 (changes requested):
swap the event dat column with service hours, and n/a would be better than a "-".

MImran2002 (approved):
looks good

fritzj2 (changes requested):
The number of hours in the service hours column does not match the number of volunteer hours earned on the service transcript for certain students.
Fritzj2 has 73.75 hours tracked on the table, while the service transcript shows he has only 72.
The all-volunteer training should not earn toward service hours and thus shouldn't be included in the table.

BrianRamsay (approved):

Issue #1747, opened 2026-07-02
Allow additional academic terms in Manage Proposal


MImran2002:
Description: Students may submit proposals in their <student>'s CCE Minor Profile > Manage Proposal. There the terms shown are limited CELTS request that there should be more options. 

<img width="1415" height="226" alt="Image" src="https://github.com/user-attachments/assets/7cc8b9a5-19f5-4f61-9018-e374a752e897" />

PR #1767, opened 2026-07-17
Fixing the List of Summer Terms on the Request Summer Proposal page

ArtemKurasov:
## Issue Description

Fixes issue #1747 
- If you request a summer proposal on a CCE Minor Profile page, you don't have enough Summer terms to choose from 

## Changes

- There are now more Summer Terms options for the CCE Minor Proposal Form
- The options reflect the academic year of a student (for example, if they are a freshman, they will 4 options, but if they are not, they will have less).

## Testing

1. Go to Admin > Minor Management
2. Choose different students on the CCE Minor Progress list/table
3. Click on the Request Summer Proposal button
4. Click on the Summer Year dropdown
5. Make sure that the displayed years match the Student's academic year


brightfietsop-ux:
Fixes issue ##1747

## Summary

Updated the CCE Minor Summer Experience form so students see summer years based on their estimated enrollment period.

## What Changed

- Added `selectAllSummerTerms()` to estimate a student’s admission year using:
  - Current class level
  - Current academic term
  - Graduation status
- Limited the dropdown to summers within the student’s estimated enrollment window.
- Included the current summer when the student is still enrolled.
- Prevented graduated students from receiving summer options.
- Updated dropdown options to use explicit `Term` IDs.
- Removed unused or broken JavaScript references from the form.
- Added Summer 2018–2020 to the demo/test data using non-conflicting IDs.

For example, a Senior in Summer 2021 sees Summer 2018–2021, while a Sophomore sees Summer 2020–2022.

## Testing

Added tests covering:

- Freshman, Sophomore, and Senior enrollment windows
- Current-summer behavior
- Graduated students
- Missing or unrecognized class levels
- Exclusion of summers beyond the estimated graduation window

app/controllers/minor/routes.py
```diff
@@ -117,13 +117,16 @@ def createSummerExperienceRequest(username):
         createSummerExperience(username, request.form)
         flash("Proposal successfully created.", "success")
         return redirect(url_for('minor.viewCceMinor', username=username, tab="manageProposals"))
+
+    student   = User.get_by_id(username)
+    year_name = User.rawClassLevel
```
MImran2002: This should be classLevel instead of year_name

app/controllers/minor/routes.py
```diff
@@ -117,13 +117,16 @@ def createSummerExperienceRequest(username):
         createSummerExperience(username, request.form)
         flash("Proposal successfully created.", "success")
         return redirect(url_for('minor.viewCceMinor', username=username, tab="manageProposals"))
+
+    student   = User.get_by_id(username)
```
MImran2002: incase username is not student I know this is not the case above it should be as cceMinor is reserve for students only so we need to ensure isStudent is user here

app/logic/utils.py
```diff
@@ -36,6 +36,43 @@ def selectSurroundingTerms(currentTerm, prevTerms=2, summerOnly=False):
 
     return surroundingTerms
 
+def selectAllSummerTerms(currentTerm, student):
+    """
+    Select the summer terms during which a CCE Minor student could be enrolled.
+
+    The user record does not store an admission date, so the admission academic
```
MImran2002: docstrings should give a reason for input and what returns and a single sentence of what it should do.

app/logic/utils.py
```diff
@@ -36,6 +36,43 @@ def selectSurroundingTerms(currentTerm, prevTerms=2, summerOnly=False):
...
+    year is inferred from the student's current class level. The range never
+    extends beyond the student's inferred final school year; students marked as
+    ``Graduating`` are treated as fifth-year/fall-graduating students.
+    """
+    classYears = {
+        "Freshman": 1,
+        "Sophomore": 2,
+        "Junior": 3,
+        "Senior": 4,
+        "Graduating": 5,
```
MImran2002: graduating is troublesome classLevel because when you run this command: SELECT * FROM user AS u WHERE u.rawClassLevel = 'Graduating' AND u.isGraduated = TRUE; you will see those two attributes can occur at the same time.

app/templates/minor/summerExperience.html
```diff
@@ -5,8 +5,6 @@
   {{ super() }}
   <script type="module" src="/static/js/minorProfilePage.js"></script>
   <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery.inputmask/3.3.4/jquery.inputmask.bundle.min.js"></script>
-  <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-select/1.5.4/bootstrap-select.js"></script>
```
MImran2002: why are these remove?

tests/code/test_utils.py
```diff
@@ -78,3 +81,48 @@ def test_isFutureTerm():
...
+@pytest.mark.parametrize(
+    "class_level, current_description, current_year, expected_years",
+    [
+        ("Freshman", "Fall 2090", 2090, [2091, 2092, 2093]),
+        ("Sophomore", "Summer 2091", 2091, [2090, 2091, 2092]),
+        ("Senior", "Fall 2090", 2090, [2088, 2089, 2090]),
+    ],
+)
+def test_selectAllSummerTerms_uses_estimated_enrollment_window(
+        class_level, current_description, current_year, expected_years):
```
MImran2002: make sure to test graduating too as you have special logic for graduating when populating for the student.

MImran2002 (changes requested):

app/logic/utils.py
```diff
@@ -36,6 +36,43 @@ def selectSurroundingTerms(currentTerm, prevTerms=2, summerOnly=False):
...
+
+    classYear = classYears[student.rawClassLevel]
+    # A Spring/Summer term belongs to the academic year that began the prior fall.
+    academicYearStart = (currentTerm.year if currentTerm.description.startswith("Fall")
+                         else currentTerm.year - 1)
+    inferredAdmissionYear = academicYearStart - (classYear - 1)
+
+    firstSummer = inferredAdmissionYear + 1
+    lastSummer = inferredAdmissionYear + max(4, classYear) - 1
+    if currentTerm.description.startswith("Summer"):
```
MImran2002: currentTerm.isSummer should and can be use here as we have a field for this.

app/logic/utils.py
```diff
@@ -36,6 +36,43 @@ def selectSurroundingTerms(currentTerm, prevTerms=2, summerOnly=False):
...
+    }
+
+    if student.hasGraduated or student.rawClassLevel not in classYears:
+        return []
+
+    classYear = classYears[student.rawClassLevel]
+    # A Spring/Summer term belongs to the academic year that began the prior fall.
+    academicYearStart = (currentTerm.year if currentTerm.description.startswith("Fall")
+                         else currentTerm.year - 1)
+    inferredAdmissionYear = academicYearStart - (classYear - 1)
```
MImran2002: what we can do here is a way to shrink these lines 63,65,66 to two lines in two of these forms:

eg 1: 
```firstSummer = (academicYearStart - (classYear - 1)) + 1```
```lastSummer = (firstSummer - 1) + max(4, classYear) - 1```

eg2: 
```inferredAdmissionYear = academicYearStart - (classYear - 1)

firstSummer, lastSummer = inferredAdmissionYear + 1, inferredAdmissionYear + max(4, classYear) - 1```

File: app/templates/minor/summerExperience.html [hunk 9]
```diff
@@ -47,7 +45,7 @@ <h4>Proposal for Community-Engaged Summer Experience</h4>
                             </option>  
                         {% endif %}
                         {% for term in selectableTerms %}
-                            <option value="{{ term }}" 
+                            <option value="{{ term.id }}"
```
MImran2002: Image changing it from term to term.id doesn't really matter as i tried doing {{ term.description }}, {{term}}, {{term.id}} and term and term.id are the same.

tests/code/test_utils.py
```diff
@@ -1,6 +1,9 @@
 import pytest
+from types import SimpleNamespace
```
MImran2002: we do not use SimpleNamespace unless we are trying to let's say fake a request in the test so that we can avoid create a whole class for it. For instance:    mockRequestProposalObject = SimpleNamespace(
        form=defaultProposal,
        files=SimpleNamespace(
            getlist=lambda key: [],
            get=lambda key: None
        ), in test_nimor.py file. Here your simplenamespace is just use to convert traditional dictionary access into attribute access. but, if my understanding is wrong and there is a rationale do tell me that can change my comment.

tests/code/test_utils.py
```diff
@@ -78,3 +81,48 @@ def test_isFutureTerm():
...
+    with mainDB.atomic() as transaction:
+        for year in range(2087, 2096):
+            Term.create(description=f"Summer {year}",
+                        year=year,
+                        academicYear=f"{year - 1}-{year}",
+                        isSummer=True,
+                        isCurrentTerm=False,
+                        termOrder=f"{year}-2")
+
+        student = SimpleNamespace(rawClassLevel=class_level,
```
MImran2002: From the above comment regarding simplenamespace convert those so that it reflects the dictionary usage.

MImran2002 (changes requested):

app/logic/utils.py
ArtemKurasov: Can we simplify it even further? 

```
firstSummer = academicYearStart - classYear + 2
lastSummer = firstSummer + max(4, classYear) - 2
```

tests/code/test_utils.py
ArtemKurasov: I converted them into Term and User objects instead

app/logic/utils.py
```diff
@@ -36,6 +36,42 @@ def selectSurroundingTerms(currentTerm, prevTerms=2, summerOnly=False):
...
+
+    if currentTerm.isSummer:
+        lastSummer = max(lastSummer, currentTerm.year)
+
+    return (Term.select()
+                .where(Term.isSummer,
+                       Term.year >= firstSummer,
+                       Term.year <= lastSummer)
+                .order_by(Term.termOrder))
+
```
ArtemKurasov: It seems like the current implementation of `SelectAllSummerTerms()` returns all the summers the student will have or had over the course of their study at Berea. Nonetheless, this function is not supposed to return any future years at all! 

Weirdly enough, it seemingly doesn't do that on the website itself... I think this is because the program _**selects**_ all the summer terms that are between the first and the last terms. In my opinion, that can potentially be a problem because what if the database has future years?

ArtemKurasov: Also, by the current logic, freshmen can propose a summer experience before they have even been to college, which is not how it is meant to work, I think.

PR #1769, opened 2026-07-17
Fix volunteer spreadsheet tests for term and senior results

BhushanSah:
## Issue Description

The spreadsheet CI tests were failing because the expected test output no longer matched the current spreadsheet behavior.

`getUniqueVolunteers()` now returns a `Term` column, but the test still expected only `Full Name`, `Email`, and `B-Number`.

`graduatingSeniorsVolunteerHours()` can also return existing senior records from the database, so the test should not expect the entire result list to be empty or only contain the test-created users.

## Changes

- Updated `test_getUniqueVolunteers` to include the `Term` column in the expected output.
- Added expected term values to each unique volunteer row.
- Updated `test_graduatingSeniorsVolunteerHours` to check for specific test users instead of assuming the full result list is empty.
- Adjusted Bob Builder’s expected semester count and total hours after adding another unique service semester.
- Added checks to make sure non-senior students are still excluded.
- Added checks to make sure non-service events do not increase the service semester count.

## Testing

- Ran:

```bash
pytest tests/code/test_spreadsheet.py -q

### DIFF
--- a/tests/code/test_spreadsheet.py
+++ b/tests/code/test_spreadsheet.py
@@ -642,15 +642,15 @@ def test_getAllTermData(fixture_info):
 @pytest.mark.integration
 def test_getUniqueVolunteers(fixture_info):
     columns, rows = getUniqueVolunteers("2023-2024-test")
-    assert columns == ["Full Name", "Email", "B-Number"]
+    assert columns == ["Full Name", "Email", "B-Number", "Term"]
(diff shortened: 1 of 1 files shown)


brightfietsop-ux (approved):
looks good, I approve.

MImran2002 (approved):
approved this fixed next time please look into the logic and see why the test is broken. so that we can know if its working or not.

PR #1771, opened 2026-07-27
Celts labor api

MImran2002:
## Issue Description

Fixes issue #https://github.com/BCStudentSoftwareDevTeam/lsf/issues/553
- It appears that the feature that imports CELTS labor positions from LSF is not working. For example, user ramireza2 had a 2024-2025 position with CELTS.
- Normal students can access CCE Minor proposal page
- Student search not searching prefix searches like "Nyan Lin Zaw" or "Emma Mae Grooms"

## Changes

- The API from LSF works it is sending correctly what happened is within getCeltsLaborHistory if the student is working the same position title when the for loops run through the query it keeps replacing the same key within the dictionary with the same position even though the term changes. 
- This is fixed by first sorting through the terms as we want to catch the edge case of  the student having a different position in spring we can now show a separate fall and spring labor position and if they only have fall labor it is assumed as AY 2025-2026. 
- Then the second loop is we use the id as a key and a tuple of term description and position to make sure it doesn't have the same error as before.
- The terms are also in ascending order by termorder.
- For the search there is an ambiguity problem of what could be counted as first name and last name it can be "Emma Mae" and "Grooms" or the name can also be "Emma" and "Mae Grooms" so we fixed that ambiguity problem along with prefix problem in case we type "Emma M" it will still show as "Emma Mae Grooms" as an option.
- For the access problem every student can access it before now only the User themselves, User.isAdmin and User.isStudentStaff can do so.
- I also found out Neillz is no longer a Student Staff so I swithc it with another person who is a Student Staff and not a Celts Admin to fully find it capability.
- Fixed the test case for the getCeltsLaborHistory function with an added test case for a labor change in spring and its expected outcome.

Images:
Before: 

image

After:

image

Search Bar:

image

## Testing

- Check out the branch `git checkout CELTSLaborInput` and `git pull`
- Reset the Database to `database/reset_database.sh test` and run the test `tests/run_tests.sh`
- Then switch back to `database/reset_database.sh from-backup`

### Connecting LSF and CELTS
- Go to LSF and run the program make sure to flask run it and get the ip address of it for example : http://127.0.0.1:8989 paste that address into lsfURL in getCeltsLaborFromLsf. For me I commented out LsfUrl in the code and use this: lsfUrl = "http://127.0.0.1:8989/api/org/2084"

Code example:

`# lsfUrl = f"{app.config['lsf_url'].strip('/')}/api/org/2084"`
`lsfUrl = "http://127.0.0.1:8989/api/org/2084"`

- Then I go to LSF and in the lsf > app > config > secret_config.yaml I pasted the valid request ip as '127.0.0.1'. 

### Testing the connection:

-From your CELTS run python3 app/scripts/import_labor_from_lsf.py this will pull the labor data from LSF. Now any approved form in LSF of the student will be pulled and it should display on CELTS. For the edge case like having new spring labor go ahead and create and approve the labor status form. Then run the import_labor_from_lsf.py script. It will pull the updated one and you will see the changes.

### Testing the search bar
-Navigate to sidebar > student search  and search either of these names "Emma M" or " Nyan Li" it will show those names. 

### Testing CCEMinor portal
-Navigate to sidebar > student search  and search Cole Collins and go to his CCE Minor page. Copy the URL : http://127.0.0.1:8080/profile/collinsc2/cceMinor then switch to student ayisie and paste it. 
- It should deny for ayisie while display for celtsadmin and studentstaff.

### DIFF
--- a/app/logic/searchUsers.py
+++ b/app/logic/searchUsers.py
@@ -8,14 +8,13 @@ def searchUsers(query, category=None):
     '''
     # add wildcards to each piece of the query
     splitSearch = query.strip().split()
-    firstName = splitSearch[0] + "%"
-    lastName = " ".join(splitSearch[1:]) +"%"
-
-    if len(splitSearch) == 1: # search for query in first OR last name
-        searchWhere = (User.firstName ** firstName | User.lastName ** firstName | User.username ** splitSearch)
-    else:                     # search for first AND last name
-        searchWhere = (User.firstName ** firstName & User.lastName ** lastName)
-
+    fullSearch = " ".join(splitSearch) + "%"
+    searchWhere = (User.firstName ** fullSearch | User.lastName ** fullSearch | User.username ** fullSearch)
+    for splitIndex in range(1, len(splitSearch)):
+        firstName = " ".join(splitSearch[:splitIndex]) + "%"
+        lastName = " ".join(splitSearch[splitIndex:]) + "%"
+        searchWhere |= (User.firstName ** firstName & User.lastName ** lastName)
+    
     if category == "instructor":
         userWhere = (User.isFaculty | User.isStaff)
     elif category == "admin":
--- a/app/templates/sidebar.html
+++ b/app/templates/sidebar.html
@@ -105,7 +105,7 @@
           <select name="newuser" class="form-select" style="margin-bottom: 10px" onchange="this.form.submit()">>
             <option {{"selected" if g.current_user.username == config.default_user }} value="{{config.default_user}}">Default User: {{config.default_user}}</option>
             <option {{"selected" if g.current_user.username == "ramsayb2"}} value="ramsayb2">Admin: ramsayb2</option>
-            <option {{"selected" if g.current_user.username == "neillz"}} value="neillz">Student Staff: neillz</option>
+            <option {{"selected" if g.current_user.username == "velazquezdiaza"}} value="velazquezdiaza">Student Staff: velazquezdiaza</option>
             <option {{"selected" if g.current_user.username == "ayisie"}} value="ayisie">Student: ayisie</option>
             <option {{"selected" if g.current_user.username == "heggens"}} value="heggens">Faculty: heggens</option>
           </select>
--- a/app/controllers/minor/routes.py
+++ b/app/controllers/minor/routes.py
@@ -29,6 +29,9 @@ def viewCceMinor(username):
     """
     Load minor management page with community engagements and summer experience
     """
+    if not (g.current_user.isAdmin or g.current_user.username == username or g.current_user.isCeltsStudentStaff):
+        return abort(403)
+
     sustainedEngagementByTerm = getCommunityEngagementByTerm(username)
 
     activeTab = request.args.get("tab", "sustainedCommunityEngagements")
--- a/app/logic/celtsLabor.py
+++ b/app/logic/celtsLabor.py
@@ -125,7 +125,7 @@ def getCeltsLaborHistory(volunteer):
                                               Term.isSummer)
                                       .join(Term, on=(CeltsLabor.term == Term.id))
                                       .where(CeltsLabor.user == volunteer)
-                                      .order_by(Term.id.asc()))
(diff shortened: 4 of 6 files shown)


app/templates/sidebar.html
```diff
@@ -105,7 +105,7 @@ <h6>Current User: {{g.current_user.username}}</h6>
           <select name="newuser" class="form-select" style="margin-bottom: 10px" onchange="this.form.submit()">>
             <option {{"selected" if g.current_user.username == config.default_user }} value="{{config.default_user}}">Default User: {{config.default_user}}</option>
             <option {{"selected" if g.current_user.username == "ramsayb2"}} value="ramsayb2">Admin: ramsayb2</option>
-            <option {{"selected" if g.current_user.username == "neillz"}} value="neillz">Student Staff: neillz</option>
+            <option {{"selected" if g.current_user.username == "velazquezdiaza"}} value="velazquezdiaza">Student Staff: velazquezdiaza</option>
```
sheggen: Does this work with our test data?

sheggen: Pick a different user, as this is exposing this user's labor and service records to our team.

MImran2002: Will definitely work on it and the config for api as well.

sheggen: While we are here, we will be changing the language from "Student Staff" to "Program Manager" for this role. Later, we will add another role of "Operations Team" which has slightly elevated privileges from a PM.

app/templates/sidebar.html
```diff
@@ -105,7 +105,9 @@ <h6>Current User: {{g.current_user.username}}</h6>
           <select name="newuser" class="form-select" style="margin-bottom: 10px" onchange="this.form.submit()">>
             <option {{"selected" if g.current_user.username == config.default_user }} value="{{config.default_user}}">Default User: {{config.default_user}}</option>
             <option {{"selected" if g.current_user.username == "ramsayb2"}} value="ramsayb2">Admin: ramsayb2</option>
-            <option {{"selected" if g.current_user.username == "neillz"}} value="neillz">Student Staff: neillz</option>
+            <option {{"selected" if g.current_user.username == "neillz"}} value="neillz">Program Manager: neillz</option>
+            <!--Scott Heggen as a placeholder for Operations Team until we have proper operations team member and their consent-->
+            <option {{"selected" if g.current_user.username == "heggens"}} value="heggens">Operations Team: heggens</option>
```
fritzj2: This value doesn't default to an Operations Manager. It just defaults to faculty Heggen once selected.

Image

app/logic/searchUsers.py
```diff
@@ -8,14 +9,29 @@ def searchUsers(query, category=None):
...
+    searchWhere = (User.firstName ** fullSearch | User.lastName ** fullSearch | User.username ** fullSearch)
+    for splitIndex in range(1, len(splitSearch)):
+        firstName = " ".join(splitSearch[:splitIndex]) + "%"
+        lastName = " ".join(splitSearch[splitIndex:]) + "%"
 
-    if len(splitSearch) == 1: # search for query in first OR last name
-        searchWhere = (User.firstName ** firstName | User.lastName ** firstName | User.username ** splitSearch)
-    else:                     # search for first AND last name
-        searchWhere = (User.firstName ** firstName & User.lastName ** lastName)
+        searchWhere |= (
```
fritzj2: Not sure if this is intentional, but the search seems _very_ lenient, almost to the point of making it hard to find someone. It does make it easier to find someone with a complicated name setup, such as 2 last names; however, it seems like it doesn't really care for order or the number of names at all.

Image

Image

MImran2002: now it will do that if i type Lin it will search all lin but as soon as I type zaw as well it will reduce the choices

app/templates/sidebar.html
MImran2002: remove it will add it in another issue as this enhancement is for development and we would want it for the celts to get other features.

BrianRamsay (approved):

